PR #2, opened 2016-12-08
Update AmqpSend.java

enuysa-msft:
Cloud 2 Device client hanging and doesn't throw any error when the connection string is incorrect
https://github.com/Azure/azure-iot-sdks/issues/938

### DIFF
--- a/service/iothub-service-sdk/src/main/java/com/microsoft/azure/iot/service/transport/amqps/AmqpSend.java
+++ b/service/iothub-service-sdk/src/main/java/com/microsoft/azure/iot/service/transport/amqps/AmqpSend.java
@@ -111,9 +111,14 @@ public class AmqpSend extends BaseHandler
             // Codes_SRS_SERVICE_SDK_JAVA_AMQPSEND_12_007: [The function shall initialize the Proton reactor object]
             this.reactor = Proton.reactor(this);
             // Codes_SRS_SERVICE_SDK_JAVA_AMQPSEND_12_008: [The function shall start the Proton reactor object]
-            this.reactor.run();
+            try
+            {
+              this.reactor.run();
+            }
+            catch (Exception e) {
+              System.err.println("Error "+ e.getMessage()); 
+            }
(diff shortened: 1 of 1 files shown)


tameraw:
@enuysa-msft - Since this is a code change, can you please submit it against the develop branch? Thanks.

enuysa-msft:
@tameraw - Done. I have submitted it to the develop branch

service/iothub-service-sdk/src/main/java/com/microsoft/azure/iot/service/transport/amqps/AmqpSend.java
```diff
@@ -111,9 +111,14 @@ public void send(String deviceId, Message message) throws IOException
             // Codes_SRS_SERVICE_SDK_JAVA_AMQPSEND_12_007: [The function shall initialize the Proton reactor object]
             this.reactor = Proton.reactor(this);
             // Codes_SRS_SERVICE_SDK_JAVA_AMQPSEND_12_008: [The function shall start the Proton reactor object]
-            this.reactor.run();
+            try
+            {
+              this.reactor.run();
+            }
+            catch (Exception e) {
```
prmathur-microsoft: I think it would be nice to throw exception to higher layers and notify then to catch it here.

tameraw:
@enuysa-msft - This still appears to be against master. Can you please resubmit against develop and address feedback from @prmathur-microsoft?
Thanks!

tameraw:
@enuysa-msft - Please address comments and resubmit against develp branch.

enuysa-msft:
I have resubmitted the change at developer branch but I will check and try to throw exception to higher layers

enuysa-msft:
@prmathur-microsoft  - I have checked  throw exception to higher layers but the higher layer is 
 org.apache.qpid.proton.reactor reactor class which does not allow me throw exception so I found that as a best solution
@tameraw  - I have resubmitted the change at developer branch

PR #33, opened 2017-02-02
requirement doc and unit test case changes

azmigproject:


device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/CustomLoggerTest.java
```diff
@@ -0,0 +1,92 @@
...
+	{
+        CustomLogger logger = new CustomLogger(this.getClass());
+        String methodName = logger.getMethodName();
+        assertThat("getExecutingMethodName", is(equalTo(methodName)));
+	}
+	// Tests_SRS_CUSTOMERLOGGER_25_002: [The function shall mock CustomLogger for unit testing.]
+	private void mockLogger() {
+        new MockUp<CustomLogger>() {
+            @Mock
+            public void LogInfo(String msg, Object... params) {
```
prmathur-microsoft: How about making sure that log info actually prints log at info level? Your test cases are only making sure that calls actually call the respective function. Use invocation or some other way to actually go through the real function calls

prmathur-microsoft: Same applies to other level logging too.

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/CustomLoggerTest.java
```diff
@@ -0,0 +1,254 @@
...
+        new Verifications()
+        {
+            {
+                mockLogger.error(any);
+                times = 1; 
+            }
+        };
+    }
+
+    // Tests_SRS_CUSTOMERLOGGER_25_002: [The function shall mock CustomLogger for unit testing.]
```
prmathur-microsoft: Do you still need this ?

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/CustomLoggerTest.java
```diff
@@ -0,0 +1,254 @@
...
+{
+    // Tests_SRS_CUSTOMERLOGGER_25_001: [The constructor shall create the logger instance.]
+    @Test
+    public void contructorCreatesLogger()
+    {
+        CustomLogger logger = new CustomLogger(this.getClass());
+        assertThat(logger, notNullValue());
+    }
+
+    // Tests_SRS_CUSTOMERLOGGER_25_002: [The function shall log message for all levels.]
```
prmathur-microsoft: You can remove this as this is individually tested.

PR #34, opened 2017-02-09
fix memory leak in AmqpsIothubConnection

hiro4kuribaya4:
If proton failed sending, release dlv object. Otherwise
 release it when received a disposition frame from proton.


prmathur-microsoft:
Added tests, edited indentation and checked in @a3da58c8a14b7c2ff372fe70aadc1b22de797f2d

PR #36, opened 2017-02-13
Update AmqpSend.java

enuysa-msft:
Fix for Cloud 2 Device client hanging and doesn't throw any error when the connection string is incorrect on Developer branch
https://github.com/Azure/azure-iot-sdk-java/issues/11#issuecomment-272250187

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSend.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSend.java
@@ -112,6 +112,14 @@ public class AmqpSend extends BaseHandler
             this.reactor = Proton.reactor(this);
             // Codes_SRS_SERVICE_SDK_JAVA_AMQPSEND_12_008: [The function shall start the Proton reactor object]
             this.reactor.run();
+            try
+            {
+              this.reactor.run();
+            }
+            catch (Exception e)
+            {
+              System.err.println("Error "+ e.getMessage());
+            } 
             this.reactor.free();
 
         }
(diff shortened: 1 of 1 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSend.java
```diff
@@ -112,6 +112,14 @@ public void send(String deviceId, Message message) throws IOException
             this.reactor = Proton.reactor(this);
             // Codes_SRS_SERVICE_SDK_JAVA_AMQPSEND_12_008: [The function shall start the Proton reactor object]
             this.reactor.run();
+            try
```
prmathur-microsoft: this.reactor.run(); is already on line #114. Why is it triggered again on #117 ?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSend.java
```diff
@@ -112,6 +112,14 @@ public void send(String deviceId, Message message) throws IOException
...
+            {
+              this.reactor.run();
+            }
+            catch (Exception e)
+            {
+              System.err.println("Error "+ e.getMessage());
```
prmathur-microsoft: We are using Log4j for logging Please use the same instead of System.err

enuysa-msft: I have checked and I didnt see any CustomLogger.java under services(com.microsoft.azure.sdk.iot.service) package which uses Log4j.
It is only exist under the com.microsoft.azure.sdk.iot.device package. 
Do you want me to create a Logger class?

enuysa-msft: @prmathur-microsoft 
this.reactor.run(); is removed from the line #114

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSend.java
enuysa-msft: this.reactor.run(); is removed from the line #114

PR #38, opened 2017-02-16
Update AmqpSend.java

enuysa-msft:
Fix for Cloud 2 Device client hanging and doesn't throw any error when the connection string is incorrect 

https://github.com/Azure/azure-iot-sdk-java/issues/11

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSend.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSend.java
@@ -108,20 +108,20 @@ public class AmqpSend extends BaseHandler
       {
         if  (amqpSendHandler != null)
         {
-            // Codes_SRS_SERVICE_SDK_JAVA_AMQPSEND_12_006: [The function shall create a binary message with the given content]
-            amqpSendHandler.createProtonMessage(deviceId, message);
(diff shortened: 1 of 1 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSend.java
```diff
@@ -112,8 +112,14 @@ public void send(String deviceId, Message message) throws IOException
             amqpSendHandler.createProtonMessage(deviceId, message);
             // Codes_SRS_SERVICE_SDK_JAVA_AMQPSEND_12_007: [The function shall initialize the Proton reactor object]
             this.reactor = Proton.reactor(this);
-            // Codes_SRS_SERVICE_SDK_JAVA_AMQPSEND_12_008: [The function shall start the Proton reactor object]
-            this.reactor.run();
+            // Codes_SRS_SERVICE_SDK_JAVA_AMQPSEND_12_008: [The function shall start the Proton reactor object]
```
prmathur-microsoft: Need to move the try around this whole block not just run.

enuysa-msft: @prmathur-microsoft:  Moved  the try around this whole block

prmathur-microsoft:
Your PR is checked in @commit 8c017a8382ce040eaa1524291c0abcbc6e41550d after squashing.

PR #39, opened 2017-02-20
fix issue that message properties are not sent out when using MQTT

erich-wang:
the Message's properties are sent out when using AMQP/HTTPS but MQTT

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttMessaging.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttMessaging.java
@@ -4,6 +4,7 @@
 package com.microsoft.azure.sdk.iot.device.transport.mqtt;
 
 import com.microsoft.azure.sdk.iot.device.Message;
+import com.microsoft.azure.sdk.iot.device.MessageProperty;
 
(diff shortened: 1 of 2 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttMessaging.java
```diff
@@ -167,10 +168,34 @@ public void send(Message message) throws IOException
             throw new IOException("Message cannot be null");
         }
 
+        MessageProperty[] messageProperties = message.getProperties();
+        String messagePublishTopic;
```
prmathur-microsoft: Thank you for your change. This looks good and is much needed. Can you please also add a test case around this in device\iot-device-client\src\test\java\tests\unit\com\microsoft\azure\sdk\iot\device\transport\mqtt\MqttMessagingTest.java ?

erich-wang: test case has been added, please review, thanks.

prmathur-microsoft:
Added a new requirement, squashed all the commits and checked in @f9e2af997932722357fa15b8eba13e3bbc2e7d76 commit. Thank you for your contribution.

PR #44, opened 2017-03-10
Don't write full stack trace to system console on exceptions while trying to send messages to IoT Hub

JMayrbaeurl:
# Checklist
- [ ] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down). No modification necessary

# Description of the problem
Reported by Partner Bosch Rexroth: If there are exceptions on trying to send messages to IoT Hub with class IotHubSendTask, full stack traces will be written to the console. This causes problems on devices with few memory and disk space (when forwarding the console output to log files)

# Description of the solution
Introduced new boolean class property 'verboseErrorMessaging' that defaults to false and is used in catch clause to determine if full stack traces should be written to the system console. This means that the default behavior of the class has been changed to NOT write full stack traces on exceptions.
iot-hub-exceptions


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubSendTask.java
```diff
@@ -28,12 +43,20 @@ public void run()
...
-            for (StackTraceElement el : e.getStackTrace())
+            
+            // Only output full stack trace on verbose Error messaging configuration
+            if (this.verboseErrorMessaging) 
             {
-                System.out.println(el);
+	            for (StackTraceElement el : e.getStackTrace())
+	            {
+	            	// JMayrbaeurl 2017-03-10: Should be replaced with real logging setup
+	                System.out.println(el);
```
prmathur-microsoft: I think we have logger use in this module and would recommend using logger here instead of println. So that interested users can debug it using log.
And if we go the route of logger you woudn't really need verboseErrMessaging anymore. By default you can always write fatal error message in this case.

prmathur-microsoft:
Thanks for your PR. Can you please address my comment and then we pull it in.

device/iot-device-client/pom.xml
```diff
@@ -82,6 +83,11 @@
             <artifactId>commons-codec</artifactId>
             <version>1.10</version>
         </dependency>
+        <!-- Logging support -->
+        <dependency>
+   			<groupId>org.slf4j</groupId>
+ 			<artifactId>slf4j-api</artifactId>
+    	</dependency>
```
prmathur-microsoft: Thank you for proving a sample. But for now we will investigate this option and add support to better logging in future. As of now if we let this change in then only part of the SDK will be supporting this method of logging and rest will still be dependent on log4j. So I will hold off on this change.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubReceiveTask.java
```diff
@@ -3,36 +3,49 @@
...
  */
 public final class IotHubReceiveTask implements Runnable
 {
     protected final IotHubTransport transport;
-
+    /**
+     * Private logger for class
+     */
+    final Logger logger = LoggerFactory.getLogger(IotHubReceiveTask.class);
+
```
prmathur-microsoft: Can this be changed to the logger (CustomLogger) in use currently until we address proper logging method.

prmathur-microsoft:
I'm very much interested in the original change that you mentioned is required by the customer related to printing stack trace and moving it to logger. If you can use custom logger at the moment until we address better logging across the SDK, I can pull this change in.

device/iot-device-client/pom.xml
JMayrbaeurl: This is no breaking change and I really don't like to make changes that introduce an old error once again. Transitioning to logging facades is no enhancement, but a bug fix and has to be done asap. It's probably better to change the CustomLogger or even better completely remove it/exchange it by using slf4j logging facades. This can be done in an hour! If you don't accept the PR this way, I will close it and will open an according issue for it.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubReceiveTask.java
JMayrbaeurl: No. Can't be changed to CustomLogger. See comment above.

device/iot-device-client/pom.xml
prmathur-microsoft: Sounds good. If you are willing to provide a PR then please do so. I will be happy to review it. 
Just to let you know we expect the following from the PR-
1. Please remove any occurrences of log4J from requirement docs (https://github.com/Azure/azure-iot-sdk-java/tree/master/device/iot-device-client/devdoc/requirement_docs/com/microsoft/azure/iothub), unit tests(https://github.com/Azure/azure-iot-sdk-java/tree/master/device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device) and code. 
2. Provide an implementation, requirements and unit tests for new logging method for the entire SDK and replace the occurrence of log4j logger with your approach. 
3. Update github documents (https://github.com/Azure/azure-iot-sdk-java/tree/master/device/iot-device-samples#logging) and our send-receive sample (https://github.com/Azure/azure-iot-sdk-java/tree/master/device/iot-device-samples/send-receive-sample) with proper instructions for logging.

If you wish to it can be a separate PR as well.

JMayrbaeurl: I have done this in a separate PR #57 . I tried to follow the three steps described above as good as possible.

JMayrbaeurl:
Since this is a critical bug fix for our partner I have changed back the logger implementation to use the CustomLogger class instead of slf4j (still believe that this is the only way to do logging).

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubReceiveTask.java
```diff
@@ -3,36 +3,48 @@
...
+    private final CustomLogger logger = new CustomLogger(this.getClass());
+    		
     public IotHubReceiveTask(IotHubTransport transport)
     {
         // Codes_SRS_IOTHUBRECEIVETASK_11_001: [The constructor shall save the transport.]
         this.transport = transport;
+        
+        if (transport == null)
+        	throw new IllegalArgumentException("Parameter 'transport' must not be null");
+
```
prmathur-microsoft: Move this check to line 22

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubSendTask.java
```diff
@@ -11,30 +13,43 @@
...
+     */
+    private final CustomLogger logger = new CustomLogger(this.getClass());
+    
     public IotHubSendTask(IotHubTransport transport)
     {
         // Codes_SRS_IOTHUBSENDTASK_11_001: [The constructor shall save the transport.]
         this.transport = transport;
+        
+        if (transport == null)
+        	throw new IllegalArgumentException("Parameter 'transport' must not be null");
```
prmathur-microsoft: Move this check to line 23

prmathur-microsoft:
This PR did not have the following : 

1. Did not address CR
2. Went back and forth between log4j and slf4j without addressing the issue at hand which was blocking the customer.
3. Author was unwilling to address any CR changes given in the past, hence I have addressed it myself to unblock the customer and taken it forward.
4. PR was not split well between what was needed vs what is good to have.

Addressed all the above, squashed all the commits, pushed changes to unblock customer @commit d9cebc3b5760b5b33a013c3ee8a762584eb99500

JMayrbaeurl:
Concerning @prmathur-microsoft  latest comments:

1. **Did not address CR**: First commit @commit fc925b35cb94ed3761eba24710252b2a38305f54 in PR provided 18 days ago completely addressed CR and didn't introduce any kind of other changes.
2. **Went back and forth between log4j and slf4j without addressing the issue at hand which was blocking the customer**: Introducing logging was proposed by @prmathur-microsoft and changed several times to fulfill non-standard logging approach required by reviewer.
3. **Author was unwilling to address any CR changes given in the past, hence I have addressed it myself to unblock the customer and taken it forward**: Latest version of PR was done with CustomLogger usage according to CR of @prmathur-microsoft . No unblocking of the customer so far, since released versions of the SDKs still have this issue.
4. **PR was not split well between what was needed vs what is good to have:** : First version of PR, provided 18 days ago, simply solved the customer's problem. Discussions about additional changes was started by @prmathur-microsoft

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubReceiveTask.java
JMayrbaeurl: @prmathur-microsoft : With current setup constructors of IotHubSendTask and IotHubReceiveTask will write a wrong log error message when parameter 'transport' is not null. If it's null, no log message will be written, since the code is no longer reachable

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubSendTask.java
JMayrbaeurl: @prmathur-microsoft : With current setup constructors of IotHubSendTask and IotHubReceiveTask will write a wrong log error message when parameter 'transport' is not null. If it's null, no log message will be written, since the code is no longer reachable

Issue #50, opened 2017-03-15
Don't use log4j directly for logging - Replace with logging facade slf4j


JMayrbaeurl:
- **OS and version used:** Windows 10 

- **Java runtime used:** 1.8 

- **SDK version used:** 1.0.21 

# Description of the issue:
It's best practice to not use a logging implementation like log4j directly, but use a logging facade instead. Then users of the SDK are not forced to use log4j and go with their own choice. Popular logging facade libraries are Apache Commons logging and slf4j. Since only slf4j is still maintained (new versions for actual JRE runtimes available) I would recommend to use slf4j (see https://www.slf4j.org/ )

# Code sample exhibiting the issue:

# Console log of the issue:

JMayrbaeurl:
See last pull request #44 as an example for using this.

prmathur-microsoft:
This issue and #51 will be addressed together once we investigate better logging options.

jasmineymlo:
This is a low priority feature that we are not going to get to.  Please feel free to reopen and submit a PR if interested.

PR #57, opened 2017-03-17
Don't use log4j directly for logging - Replace with logging facade slf4j 

JMayrbaeurl:
# Checklist

#50
# Description of the problem
New logging approach using logging facades with slf4j framework #50 

# Description of the solution
As discussed in PR #44

### DIFF
--- a/device/iot-device-client/pom.xml
+++ b/device/iot-device-client/pom.xml
@@ -88,14 +88,38 @@
    			<groupId>org.slf4j</groupId>
  			<artifactId>slf4j-api</artifactId>
     	</dependency>
-<!--         <dependency>
+        <dependency>
             <groupId>log4j</groupId>
             <artifactId>log4j</artifactId>
             <version>1.2.17</version>
+            <scope>runtime</scope>
         </dependency>
- -->    </dependencies>
+    </dependencies>   
     <build>
         <plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<configuration>
+					<excludes>
+						<exclude>**/CustomLogger.java</exclude>
+					</excludes>
+				</configuration>
+				<executions>
+					<execution>
+						<id>default-testCompile</id>
+						<phase>test-compile</phase>
+						<configuration>
+							<testExcludes>
+								<exclude>**/CustomLoggerTest.java</exclude>
+							</testExcludes>
+						</configuration>
+						<goals>
+							<goal>testCompile</goal>
+						</goals>
+					</execution>
+				</executions>
+			</plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
@@ -106,6 +130,9 @@
                     <argLine>
                         -javaagent:${settings.localRepository}/org/jmockit/jmockit/1.22/jmockit-1.22.jar
                     </argLine>
+                    <excludes>
+                    	<exclude>**/CustomerLoggerTest.java</exclude>
+                    </excludes>
                 </configuration>
             </plugin>
             <plugin>
--- a/device/iot-device-client/pom.xml
+++ b/device/iot-device-client/pom.xml
@@ -1,5 +1,6 @@
 <!-- Copyright (c) Microsoft. All rights reserved. --><!-- Licensed under the MIT license. See LICENSE file in the project root for full license information. -->
-<project>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <artifactId>iot-device-client</artifactId>
     <name>IoT Hub Java Device Client</name>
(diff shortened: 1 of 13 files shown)


device/iot-device-client/pom.xml
```diff
@@ -82,14 +83,43 @@
...
+						<exclude>**/CustomLogger.java</exclude>
+					</excludes>
+				</configuration>
+				<executions>
+					<execution>
+						<id>default-testCompile</id>
+						<phase>test-compile</phase>
+						<configuration>
+							<testExcludes>
+								<exclude>**/CustomLoggerTest.java</exclude>
```
prmathur-microsoft: Can you remove customlogger from the repo itself ? And have it included in the commit ?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -92,7 +95,7 @@
     protected ScheduledExecutorService taskScheduler;
     protected IotHubClientState state;
     protected long RECEIVE_PERIOD_MILLIS;
-    protected CustomLogger logger;
+    protected final Logger logger = LoggerFactory.getLogger(DeviceClient.class);
```
prmathur-microsoft: Why not use this.class where ever possible ? we would prefer that.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -26,6 +26,9 @@
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
```
prmathur-microsoft: Is there any better way of doing this then importing it in every class file ? That was the primary reason for having customlogger.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -160,8 +163,9 @@ else if (attr.startsWith(SHARED_ACCESS_TOKEN_ATTRIBUTE))
         }
 
         initIotHubClient(hostname, deviceId, sharedAccessKey, sharedAccessToken, protocol);
-        this.logger = new CustomLogger(this.getClass());
-        logger.LogInfo("DeviceClient object is created successfully, method name is %s ", logger.getMethodName());
+        
+        logger.info("DeviceClient object is created successfully, method name is {} ", 
+        		Thread.currentThread().getStackTrace()[2].getMethodName());
```
prmathur-microsoft: Please avoid hard coding values. Use private static final instead. 
Again another reason to isolate as a method this in another class and lessen the code footprint.

device/iot-device-client/pom.xml
```diff
@@ -82,14 +83,43 @@
...
+        <!-- Logging support -->
+        <dependency>
+   			<groupId>org.slf4j</groupId>
+ 			<artifactId>slf4j-api</artifactId>
+    	</dependency>
         <dependency>
             <groupId>log4j</groupId>
             <artifactId>log4j</artifactId>
             <version>1.2.17</version>
+            <scope>runtime</scope>
```
prmathur-microsoft: Do you still need this here ?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
JMayrbaeurl: What's the difference between importing CustomLogger or slf4j classes in each file? slf4j library is highly mature and quite a lot of professionell java project are doing it that way. Using custom logging classes is not best practice.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
JMayrbaeurl: this.class is not specified and having different loggers and logging configurations for classes in the same hierarchy wont work with this.class

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
JMayrbaeurl: This was just copied from the original code. BTW. It's wrong, too, because the wrong method name gets returned. See Issue #58 . It should completely removed or exchanged by putting the method name directly into the log string, since calling the stack trace is very very slow!

device/iot-device-client/pom.xml
JMayrbaeurl: Yes. Otherwise the test classes won't work. They need a logger implementation. But dependency to log4j should not be populated to users of the SDK.

JMayrbaeurl:
Basic goal of this PR is to provide a logging implementation as it's known as best practice using logging facades. And another goal is to get rid of the CustomLogger class, since it's only introducing errors and additional dependencies.

prmathur-microsoft:
You should stick to primary goal of this PR i.e to add slf4j and remove log4j. 
My ask is to wrap slf4j around customlogger instead of wrapping log4j around it. There are several reasons for this ask 
1. Other developers working on this SDK will not have breaking changes.
2. Changes will be minimal and logging code will not change all over the SDK. But changes will only reflect in class file i.e customlogger.java
3. Customlogger already has all the developer needs for this checkin to go through our gate i.e requirements, tests etc.

JMayrbaeurl:
Still don't think that using CustomLogger is a good idea. Any new user to the SDK, when he is an experienced Java developer, will expect to find a logging approach as shown with this PR. Using the CustomLogger class all over the source code adds unnecessary dependency on code that can potentially break with each commit. That's not the case for using a fixed tested version of slf4j. I'm convinced that's better to have a breaking change for few developers than confuse a lot of professional devs with unusual logging and introducing additional risk for the stability of the SDK. I'll close the PR now, since I don't think we'll find a logging approach that works for both us. Sorry!

JMayrbaeurl:
As already mentioned in my last comment, I will close the PR now (reason see above). BTW. Trying to help you increase the quality of the SDK is no easy task.

PR #68, opened 2017-03-22
Eliminate dependency from DeviceTwin to DeviceClient

michaeljqzq:
# Checklist

# Description of the problem

I'm currently doing some dev work on Azure IoT device, and related to device SDK.
In current SDK, DeviceClient and DeviceTwin class has two-way dependency. Since dependency from DeviceTwin to DeviceClient is not necessary, it is better to let DeviceTwin depends on Transport.

# Description of the solution
 
Instead of maintain a property "deviceClient" in DeviceTwin, a property "transport" in DeviceTwin is better. And the only thing deviceClient related to is sending event, which can be replaced by transport.addMessage. I already tested in local environment for all twin operations. However, some test cases may need to be updated.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
@@ -329,7 +329,7 @@ public final class DeviceClient implements Closeable
             /*
             **Codes_SRS_DEVICECLIENT_25_011: [**The function shall create a new instance of class Device Twin and request all twin properties by calling getDeviceTwin**]**
              */
-            deviceTwin = new DeviceTwin(this, this.config, deviceTwinStatusCallback, deviceTwinStatusCallbackContext,
+            deviceTwin = new DeviceTwin(this.transport, this.config, deviceTwinStatusCallback, deviceTwinStatusCallbackContext,
                                         genericPropertyCallBack, genericPropertyCallBackContext);
             deviceTwin.getDeviceTwin();
         }
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceTwin/DeviceTwin.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceTwin/DeviceTwin.java
@@ -5,6 +5,7 @@ package com.microsoft.azure.sdk.iot.device.DeviceTwin;
 
 import com.microsoft.azure.sdk.iot.device.*;
 import com.microsoft.azure.sdk.iot.deps.serializer.*;
+import com.microsoft.azure.sdk.iot.device.transport.IotHubTransport;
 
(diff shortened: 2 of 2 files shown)


prmathur-microsoft:
May I ask why is this required by you ? What was the problem using Device client ?

JMayrbaeurl:
Copied from http://softwareengineering.stackexchange.com/questions/11856/whats-wrong-with-circular-references : There are a great many things wrong with circular references:

- Circular class references create high coupling; both classes must be recompiled every time either of them is changed.
- Circular assembly references prevent static linking, because B depends on A but A cannot be assembled until B is complete.
- Circular object references can crash naïve recursive algorithms (such as serializers, visitors and pretty-printers) with stack overflows. The more advanced algorithms will have cycle detection and will merely fail with a more descriptive exception/error message.
- Circular object references also make dependency injection impossible, significantly reducing the testability of your system.
- Objects with a very large number of circular references are often God Objects. Even if they are not, they have a tendency to lead to Spaghetti Code.
- Circular entity references (especially in databases, but also in domain models) prevent the use of non-nullability constraints, which may eventually lead to data corruption or at least inconsistency.
- Circular references in general are simply confusing and drastically increase the cognitive load when attempting to understand how a program functions.

Please, think of the children; avoid circular references whenever you can.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -329,7 +329,7 @@ public void startDeviceTwin(IotHubEventCallback deviceTwinStatusCallback, Object
             /*
             **Codes_SRS_DEVICECLIENT_25_011: [**The function shall create a new instance of class Device Twin and request all twin properties by calling getDeviceTwin**]**
              */
-            deviceTwin = new DeviceTwin(this, this.config, deviceTwinStatusCallback, deviceTwinStatusCallbackContext,
+            deviceTwin = new DeviceTwin(this.transport, this.config, deviceTwinStatusCallback, deviceTwinStatusCallbackContext,
                                         genericPropertyCallBack, genericPropertyCallBackContext);
```
prmathur-microsoft: How do you plan to initialize this transport ? It is most certainly dependent on Device client. If not so you will have to recreate logic in device client to initialize transport

prmathur-microsoft:
We cannot accept this change as is because 

1. Breaks existing unit tests
2. No additional UT
2. It is not complete change (can't compile as is)
3. removes additional functionality of using executors to provide async mechanism in sending messages. 

If the above changes can be addressed we can take this PR forward.

Also please make sure to sign the CLA if interested in publishing these changes.

olivierbloch:
Hi @michaeljqzq, @JMayrbaeurl 
We are definitively planning to take the proposed changes in, but we cannot take this PR as is for the above-mentioned reasons.
I'll go ahead and close this PR for now. You should see the changes published in one of the next couple releases.
Thanks a lot for the contribution.

Issue #48, opened 2017-03-14
Endless loop in com.microsoft.azure.sdk.iot.device.DeviceClient.close() when remaining messages can't be sent


JMayrbaeurl:
- **OS and version used:** Windows 10 

- **Java runtime used:** 1.7 

- **SDK version used:** 1.0.21 

# Description of the issue:
Calling DeviceClient.close() will end up in an endless loop (see code below -> while (!this.transport.isEmpty()) ) if there are still messages in the queue that haven't been sent to IoT Hub so far. E.g. in case of a missing Internet connection, this will lead to an endless loop.

# Code sample exhibiting the issue:
    public void close() throws IOException
    {
        // Codes_SRS_DEVICECLIENT_11_031: [If the client is already closed, the function shall do nothing.]
        if (this.state == IotHubClientState.CLOSED)
        {
            return;
        }

        // Codes_SRS_DEVICECLIENT_11_010: [The function shall finish all ongoing tasks.]
        // Codes_SRS_DEVICECLIENT_11_011: [The function shall cancel all recurring tasks.]
        while (!this.transport.isEmpty())
        {

        }
        this.taskScheduler.shutdown();
        // Codes_SRS_DEVICECLIENT_11_037: [The function shall close the transport.]
        this.transport.close();
        this.state = IotHubClientState.CLOSED;
    }

# Console log of the issue:

prmathur-microsoft:
Our SDK doesn't yet have reconnection logic in place. It is a PBI in our backlog and we will address it soon. In the meantime, I'll look for a workaround.

Draczech:
While I don't need the reconnection logic exactly, I've experienced some similar problems with sending messages and connection faults. E.G. JMS specification states that Connection close tries waiting for pending receives. It can happen that during loss of connection those pending messages prevent the connection from being closed if they for don't time out for some reason.

It would be nice if the DeviceClient.close() returned List of com.microsoft.azure.iothub.Message instances which could not be sent or so. (Something in style of [ExecutorService.shutdownNow()](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#shutdownNow--)) Of course there might be some better solution. But being able to terminate all open connections manually by closing the DeviceClient and being able to tell what messages are yet waiting to be sent would be really nice feature. Because on my device I am able to detect any connection loss better than some messaging implementation like eclipse.paho or qpid.proton-j.

anhashia:
[...]

PR #92, opened 2017-05-02
Avoid endless loop on closing device client when remaining messages can't be sent

JMayrbaeurl:
# Checklist
  - [x This pull-request is submitted against the `master` branch. 

Solution for critical issue described in issue #48 

# Description of the problem
Calling DeviceClient.close() will end up in an endless loop (see code below -> while (!this.transport.isEmpty()) ) if there are still messages in the queue that haven't been sent to IoT Hub so far. E.g. in case of a missing Internet connection, this will lead to an endless loop.

# Description of the solution
Introduced concurrent task that tries to send all remaining messages, but will return anyway after a certain configurable time out (currently 30 seconds)


anhashia:
fixed in https://github.com/Azure/azure-iot-sdk-java/commit/9079bf185d4420bf58e14601036b89d642c1a94c

prmathur-microsoft:
As this issue is checked in closing this as per our discussion. Please let me know if you have any further issues.

PR #98, opened 2017-06-01
Added support for registering for callbacks when connection state cha…

richardonorato:
# Checklist

None
# Description of the problem
Could not get any status on the state of the connection between the client and the IoT Hub.  We have logic that needs to know what the status of the connection is.  If the connection fails to connect then we needed a way to be informed of that.  This would allow us to track the number of failures (maybe the Key is invalid and the connection will never establish without a new key) and react to the failure accordingly.

# Description of the solution
We looked at trying to expose the connection state, but found that that was not a good approach since the state did not appear to be updated in all cases and also required our code to constantly poll for the current state.  Then we looked at using a callback to report connection up and connection down events.  This approach allowed our code to just register a callback with the device and react to the changes as they occur.  This was the approach that we ended up going with.


device/iot-device-client/devdoc/requirement_docs/com/microsoft/azure/iothub/deviceclient_requirements.md
```diff
@@ -25,6 +25,8 @@ public final class DeviceClient
     public void sendReportedProperties(Set<Property> reportedProperties) throws IOException;  
 
     public void subscribeToDeviceMethod(DeviceMethodCallback deviceMethodCallback, Object deviceMethodCallbackContext, IotHubEventCallback deviceMethodStatusCallback, Object deviceMethodStatusCallbackContext) throws IOException  
+    
+    public void registerConnectionStateCallback(IotHubConnectionStateCallback callback);
```
prmathur-microsoft: Can you please add context to the call back ?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubConnectionStateCallback.java
```diff
@@ -0,0 +1,21 @@
...
+ * Developers are expected to create an implementation of this interface,
+ * and the transport will call {@link IotHubConnectionStateCallback#connectionUp()}
+ * when connection is established with the IoT Hub and call
+ * {@link IotHubConnectionStateCallback#connectionDown()} when connection is lost.
+ */
+public interface IotHubConnectionStateCallback {
+    /**
+     * Method executed when the connection with the IoTHub is lost.
+     */
+    void connectionUp();
```
prmathur-microsoft: To keep this aligned across various SDK's we have, It would be nice to replace this with enum connectionStatus {
CONNECTION_SUCCESS, CONNECTION_DROP } and provide the enum as status as opposed to having multiple methods. This way we can easily extend this enum to multiple states associated with connection (for eg - in future we can add retry in progress, retry time out etc etc)

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsTransport.java
```diff
@@ -55,6 +55,9 @@
     /** Messages whose callbacks that are waiting to be invoked. */
     private final Queue<IotHubCallbackPacket> callbackList = new LinkedBlockingDeque<>();
 
+    /** Connection state change callbacks */
+    private final List<IotHubConnectionStateCallback> stateCallbackList = new ArrayList<>();
+
```
prmathur-microsoft: Why would a list of callbacks be needed ? We would only need one per SDK. May be just having a member variable storing callback and context would suffice here. ArrayList won't be needed (unless you had a scenario in mind that I cannot see right now. Please provide details if so.)

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsTransport.java
```diff
@@ -439,6 +463,23 @@ public boolean isEmpty()
...
+    public void registerConnectionStateCallback(IotHubConnectionStateCallback callback) {
+        // Codes_SRS_AMQPSTRANSPORT_99_003: [The registerConnectionStateCallback shall register the connection state callback.]
+        this.stateCallbackList.add(callback);
+    }
+
+    /**
+     * Helper method for testing
+     */
+     public int totalConnectionStateCallbacks() {
+         return this.stateCallbackList.size();
```
prmathur-microsoft: perhaps you can get rid of this one as per comment above then.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTransport.java
```diff
@@ -343,4 +343,15 @@ public boolean isEmpty()
...
+
+    /**
+     * Registers a callback to be executed whenever the mqtt connection is lost or established.
+     * 
+     * @param callback the callback to be called.
+     */
+    public void registerConnectionStateCallback(IotHubConnectionStateCallback callback) {
+        //TODO: Add in registration logic if needed.  The transport would also need to be updated to
+        //      receive notifications from the MQTT Connection class when the connection is lost or
+        //      established.
```
prmathur-microsoft: If you wish to leave this as is then please throw UnSupportedOperation here. If not please implement the logic.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -725,4 +726,14 @@ public void setOption(String optionName, Object value)
 
     }
 
+    /**
+     * Registers a callback to be executed whenever the connection to the device is lost or established.
+     * 
+     * @param callback the callback to be called.
+     */
+    public void registerConnectionStateCallback(IotHubConnectionStateCallback callback) {
```
prmathur-microsoft: should null value of call back be accepted ? perhaps throw IllegalArgument Exception in such cases.

device/iot-device-client/devdoc/requirement_docs/com/microsoft/azure/iothub/deviceclient_requirements.md
richardonorato: I can add a Object stateStatusCallbackContext to the method

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubConnectionStateCallback.java
richardonorato: I will add the enum and also change the callback up to only have an execute method that takes in the enum and a context object.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsTransport.java
richardonorato: For my particular project we would only ever have a single callback registered.  I have no issues changes this to be a member variable and not a List.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsTransport.java
richardonorato: Yes, I will remove this method since it will no longer be needed.  I will also update the spec as well.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTransport.java
richardonorato: I will have this method throw an UnSupportedOperation exception.  I am not really sure what would really be needed for this to work correctly anyway.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
richardonorato: We should not accept a null value.  I will add a check for it and have it throw an exception if null is detected.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsTransport.java
```diff
@@ -324,6 +324,17 @@ public boolean isEmpty()
...
 
     /**
+     * Registers a callback to be executed whenever the https connection is lost or established.
+     * 
+     * @param callback the callback to be called.
+     * @param callbackContext a context to be passed to the callback. Can be
+     * {@code null} if no callback is provided.
+     */
+    public void registerConnectionStateCallback(IotHubConnectionStateCallback callback, Object callbackContext) {
+        //HTTPS is stateless so this should be a no-op
```
prmathur-microsoft: Can you throw IllegalStateException here ?

richardonorato: Yes.  We probably should not allow someone to register a callback on this type of connection.

richardonorato:
I added the throwing of UnsupportedOperationException for the registerConnectionStateCallback on HttpsTransport.

prmathur-microsoft:
merged this @5b1b00aff7034d754a3fabd4c40d5fba060e3e6c. Thank for contribution.

PR #108, opened 2017-06-25
Update HttpsIotHubConnection.java

Soltman:
You should not use static lock. Send and receive operations will become chained despite how many connections to different hosts you will open.
This is very strong bottleneck to stuck with this SDK.

 

# Checklist

# Static lock at HTTPS connection causes bottleneck in communication with different endpoints.

# Object-level lock is used


prmathur-microsoft:
What is the scenario you have in mind ? Are you trying to connect to multiple Hubs from the same process using HTTP ?

Soltman:
The task is to send 100 events fast enough.
Real scenario I've used is 2 hubs, 100 connections, one per event. Network latency is around 400ms. This amount takes around one minute to be sent.
The reason is: connection has static lock, single per java instance. No matter where this connection is pointed, when some thread is in sending routine, no others can come in and send of receive. Even creation of connection is synchronized by static lock, you cannot initialize connections from multiple threads.

prmathur-microsoft:
rebased your code to latest and checked in this PR  @bbae08c59076b4281224f136a893a96f087e1f73

Issue #159, opened 2017-10-10
JobResult did not use non-parameter constructor of DeviceTwinDevice when device id is null


zhang-hua:
- **OS and version used:** <VERSION> Windows 10
- **Java runtime used:** <VERSION> JDK 1.8.0_141
- **SDK version used:** <VERSION>  Java Service Client (com.microsoft.azure.sdk.iot:iot-service-client:1.9.24)
- **IDE**:  IntelliJ IDEA

# Description of the issue:

There is non-parameter constructor DeviceTwinDevice() which should be used when device id is null. It happened at:
https://github.com/Azure/azure-iot-sdk-java/blob/master/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobResult.java#L158
Need to check twinParser.getDeviceId(). If null, DeviceTwinDevice() should be used here.

related issue: https://github.com/Azure/azure-iot-sdk-java/issues/143

# Code sample exhibiting the issue:

# Console log of the issue:

timtay-microsoft:
The fix for this issue is now released. See the release below for versions:

https://github.com/Azure/azure-iot-sdk-java/releases/tag/2017-10-23

I will close this issue for now. Feel free to re-open if the release did not fix your issues.

PR #162, opened 2017-10-12
Fix issue #159 [Use non-parameter constructor of DeviceTwinDevice if device id is null]

zhang-hua:
When scheduling a twin job, the job result contains empty device id which
will fail the constructor of DeviceTwinDevice(String deviceId). The
non-parameter constructor should be used in this case.

 

# Checklist

https://github.com/Azure/azure-iot-sdk-java/issues/159

# Description of the problem

When scheduling a twin job, the job result contains null or empty device id which
will fail the constructor of DeviceTwinDevice(String deviceId). The
non-parameter constructor should be used in this case.
# Description of the solution

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobResult.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobResult.java
@@ -155,7 +155,8 @@ public class JobResult
         TwinParser twinParser = jobsResponseParser.getUpdateTwin();
         if(twinParser != null)
         {
-            this.updateTwin = new DeviceTwinDevice(twinParser.getDeviceId());
+            this.updateTwin = twinParser.getDeviceId() == null || twinParser.getDeviceId().isEmpty() ?
+                new DeviceTwinDevice() : new DeviceTwinDevice(twinParser.getDeviceId());
             this.updateTwin.setETag(twinParser.getETag());
             try
             {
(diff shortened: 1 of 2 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobResult.java
```diff
@@ -155,7 +155,8 @@
         TwinParser twinParser = jobsResponseParser.getUpdateTwin();
         if(twinParser != null)
         {
-            this.updateTwin = new DeviceTwinDevice(twinParser.getDeviceId());
+            this.updateTwin = twinParser.getDeviceId() == null || twinParser.getDeviceId().isEmpty() ?
+                new DeviceTwinDevice() : new DeviceTwinDevice(twinParser.getDeviceId());
```
prmathur-microsoft: Please add a test case for empty device ID in the corresponding test.

prmathur-microsoft: Test suite for this is located https://github.com/Azure/azure-iot-sdk-java/blob/master/service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/jobs/JobResultTest.java

zhang-hua: @prmathur-microsoft , a test case of using nullable deviceId has been added in the commit. please help review it. Thanks!

prmathur-microsoft:
This PR has been merged in master. Thanks for contributing!

PR #190, opened 2018-01-08
Split Twin from Serializer in deps

mamokarz:
Split Twin from Serializer in deps


mamokarz:
I'll not push it to master for now, so, I'll resend this PR over the developer branch RedesignTwin.

PR #197, opened 2018-01-23
Remove unused (and unavailable) InactiveGroupException import statement that breaks the build.

JonathanGiles:
A `import sun.rmi.server.InactiveGroupException` statement is added to DeviceTwin but the class is not used. This is fortunate as under JDK 9 this fails to compile. This PR removes this import and restores the ability to build this project under JDK 9.


JonathanGiles:
This import statement was added as part of the following commit: https://github.com/Azure/azure-iot-sdk-java/commit/b60a1215c091f33f00ee292ebae65b12cc0dfdd3

JonathanGiles:
What's the process of getting this merged - do you wait until a build completes or something similar? Thanks!

prmathur-microsoft:
Yes. We run builds sequentially on our internal CI system which is gated and runs several tests. And only if the tests pass, your check in gets through. Currently I have kicked a run for your PR, waiting for it to pass.

PR #199, opened 2018-02-01
docs: Add commit message guidelines

prmathur-microsoft:
# Checklist
- [X] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-java/blob/master/.github/CONTRIBUTING.md).
  - [X] This pull-request is submitted against the `master` branch. 

# Description of the problem

# Description of the solution

### DIFF
--- a/.github/CODING_GUIDELINES.md
+++ b/.github/CODING_GUIDELINES.md
@@ -8,6 +8,7 @@ Following is a definition of a minimum set of rulers that defines the Java SDK c
 - [Comment Style](#Comment-Style)
 - [Naming Conventions](#Naming-Conventions)
 - [JavaDoc Guidelines](#JavaDoc-Guidelines)
+- [Commit Guidelines](#Commit-Guidelines)
 
 # Programming Guidelines
 * Switch statements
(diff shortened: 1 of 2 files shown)


timtay-microsoft:
I don't love the idea of splitting commits by package. Generally, when we complete a feature that spans across 1 or more packages, it would never make sense to revert the changes to just one of those packages. Doing so would just break the other package.

What is the argument for splitting commits by package?

mamokarz:
Agree with timtay comment.

timtay-microsoft:
From what I can tell, the reason commits are being split by package in these guidelines is so that we have an easy way of checking which packages were updated by each commit. Instead of having one Scope per commit, we should just allow multiple Scopes per commit. That way, we are still defining which packages are affected in a clear way, but we won't have to separate commits by package.

.github/CODING_GUIDELINES.md
```diff
@@ -252,3 +253,109 @@ public class ThisIsASampleClassTest()
...
+
+**Footer**
+
+The footer should contain any information about Breaking Changes and is also the place to reference GitHub issues that this commit Closes.
+
+Breaking Changes should start with the word BREAKING CHANGE: with a space or two newlines. The rest of the commit message is then used for this.
+
+***Example commit messages***
+
+Good commit messages looks like below :
```
tameraw: nitpic: "look" (and below)

.github/CODING_GUIDELINES.md
```diff
@@ -252,3 +253,109 @@ public class ThisIsASampleClassTest()
...
+The subject contains succinct description of the change:
+
+* use the imperative, present tense: "change" not "changed" nor "changes"
+* no dot (.) at the end
+
+**Body**
+
+Just as in the subject, use the imperative, present tense: 
+
+"change" not "changed" nor "changes". The body should include the motivation for the change and contrast this with previous behavior.
```
tameraw: Do we need to state that? Can't we point to a link with that common practice?

.github/CODING_GUIDELINES.md
```diff
@@ -252,3 +253,109 @@ public class ThisIsASampleClassTest()
...
+
+    Github issue (fix#123)
+* feat(iot-service): Add continuation token to Query
+* docs: Update readme to reflect provisiong client
+
+Bad commit messages looks like below :
+
+* fix(iot-dev): small fix
+
+    I was trying to reconnect and network dropped. Fixing such random failures
```
tameraw: >Fixing [](start = 51, length = 6)

Shouldn't be "fix" if I follow the rule above for imperative?

.github/CODING_GUIDELINES.md
prmathur-microsoft: done

.github/CODING_GUIDELINES.md
prmathur-microsoft: This is to show how a bad commit message looks like. Hence "Fixing" is used on purpose.

.github/CODING_GUIDELINES.md
prmathur-microsoft: Will do that

prmathur-microsoft:
As per the conversation changed the following : 
1.Optionally split commits per packages. Just do the needful.
2. just call scope "samples"
3. addressed further comments

Issue #257, opened 2018-06-02
IotHubTransport generates tons of useless logs


a18250:
- **OS and version used:** Android 7.1.1 

- **Java runtime used:** Android ART Java 8 

- **SDK version used:** 1.11.1 

# Description of the issue:
I have log4j log level set to DEBUG so to get more useful logs for you guys. However, the IotHubTransport.handleMessage() prints useless log every 10ms. This makes the log file size grows quickly and results the whole log file to be useless. I'd think the method should only log when it really sees an incoming message.

# Code sample exhibiting the issue:
N/A

# Console log of the issue:
Piece of log samples are as below:

DEBUG   2018-06-01 23:25:42,856 [azure-iot-sdk-IotHubReceiveTask] Get the callback function for the received message, method name is getMethodName 
DEBUG   2018-06-01 23:25:42,868 [azure-iot-sdk-IotHubReceiveTask] Get the callback function for the received message, method name is getMethodName 
DEBUG   2018-06-01 23:25:42,874 [azure-iot-sdk-IotHubReceiveTask] Get the callback function for the received message, method name is getMethodName 
DEBUG   2018-06-01 23:25:42,884 [azure-iot-sdk-IotHubReceiveTask] Get the callback function for the received message, method name is getMethodName 
DEBUG   2018-06-01 23:25:42,894 [azure-iot-sdk-IotHubReceiveTask] Get the callback function for the received message, method name is getMethodName 
DEBUG   2018-06-01 23:25:42,904 [azure-iot-sdk-IotHubReceiveTask] Get the callback function for the received message, method name is getMethodName 
DEBUG   2018-06-01 23:25:42,914 [azure-iot-sdk-IotHubReceiveTask] Get the callback function for the received message, method name is getMethodName 
DEBUG   2018-06-01 23:25:42,924 [azure-iot-sdk-IotHubReceiveTask] Get the callback function for the received message, method name is getMethodName 
DEBUG   2018-06-01 23:25:42,934 [azure-iot-sdk-IotHubReceiveTask] Get the callback function for the received message, method name is getMethodName 
DEBUG   2018-06-01 23:25:42,944 [azure-iot-sdk-IotHubReceiveTask] Get the callback function for the received message, method name is getMethodName 
DEBUG   2018-06-01 23:25:42,954 [azure-iot-sdk-IotHubReceiveTask] Get the callback function for the received message, method name is getMethodName 

# Need Support?
[...]

PR #259, opened 2018-06-07
Fix issue 257: remove unnecessary logging in IotHubTransport

a18250:
# Checklist

https://github.com/Azure/azure-iot-sdk-java/issues/257

# Description of the problem
The useless log is printed every 10ms as long as loglevel is set to DEBUG.

# Description of the solution
Since the received message has already been logged in onMessageReceived, the log in handleMessage is redundant and can be removed.
Another INFO log is added into addReceivedMessagesOverHttpToReceivedQueue
so all received messages are equally logged in transport level.

Note, the change in IotHubTransportTest.java is not for covering added code. But if without this change, the added logging in addReceivedMessagesOverHttpToReceivedQueue() will result in jmockit to complain missing invocation of HttpsIotHubConnection.receiveMessage(). Not sure why, but since the case is only to ensure addReceivedMessagesOverHttpToReceivedQueue() is not called, the added expectation is for mocking that method only.


timtay-microsoft:
Looks good to me. I'll get this merged in for you

Issue #296, opened 2018-08-02
validateStringUTF8 issue in 


gabiSRC:
- **OS and version used:** macOs Sierra 10.12.6

- **Java runtime used:** Java 1.8.0_10

- **SDK version used:** iot-deps 0.5.3

# Description of the issue:
When trying to get Twin devices, I get the exception `java.lang.IllegalArgumentException: parameter contains non UTF-8 character`

# Code sample exhibiting the issue:
```
QueryCollection twinQuery = deviceTwinClient.queryTwinCollection("SELECT * FROM devices");

        while (deviceTwinClient.hasNext(twinQuery)) {
            try {
                QueryCollectionResponse<DeviceTwinDevice> response = deviceTwinClient.next(twinQuery);
                response.getCollection().forEach(twin -> devicesId.add(twin.getDeviceId()));
            } catch (IllegalArgumentException e) {
                LOGGER.warn("Error while getting twin device", e);
            }
        }
```

I checked to method and I thing the comparison is wrong: `if(str.getBytes("UTF-8").length != str.length()) ` whereas it should be `if(str.getBytes("UTF-8").length != str..getBytes().length) `

Maybe I am wrong so if so, please could you help me solve this?

I already checked my devices data and there is no non-UTF-8 characters.

Thank you.

timtay-microsoft:
It looks like you are correct about this. If you'd like to provide a PR for this, I'd be happy to merge it in for you.

JMayrbaeurl:
I don't think that 

`if (str.getBytes("UTF-8").length != str.getBytes().length)
`

is the correct way of checking for the existence of valid characters in the device ID, since `str.getBytes()` will always use the default encoding that's probably different on the different machines. According to the documentation a device ID is a case-sensitive string (up to 128 characters long) of ASCII 7-bit alphanumeric characters plus certain special characters: - . + % _ # * ? ! ( ) , = @ $ '. This means only 1 byte per character and that's what the original check in the source code is doing. Right? Btw. On my machine the provided source code snippet works correctly. What device IDs have you registered on your IoT Hub?

timtay-microsoft:
I'm pretty curious about the device ids involved here, too. @gabiSRC  Can you provide a list of your device ids?

gabiSRC:
@JMayrbaeurl @timtay-microsoft here are my IDs:
[...]

PR #297, opened 2018-08-05
Device twins can now contain double byte UTF-8 characters - fix for bug#296

JMayrbaeurl:
# Checklist

Fix for issue #296 

# Description of the problem
Device twins with valid UTF-8 characters, that used more than 1 byte like '§', caused an IllegalArgumentException, since the check in ParserUtility was wrong

# Description of the solution
Removed the call to the wrong validation in ParserUtilitiy and added check against empty string. Only problem left is that property and tag keys could now contain invalid characters. But this should not never happen, since this is already checked by the IoT Hub service.


timtay-microsoft:
Looks good, we'll get this merged in shortly. Thanks for the pr!

PR #301, opened 2018-08-08
Reactivated e2e tests for import/export of devices

JMayrbaeurl:
# Checklist

# Description of the problem
e2e test for importing and exporting devices didn't compile and was ignored due to service problems

# Description of the solution
Set Java compiler to version 1.8 and removed Ignore attribute from unit test

### DIFF
--- a/iot-e2e-tests/jvm/pom.xml
+++ b/iot-e2e-tests/jvm/pom.xml
@@ -2,7 +2,7 @@
   ~  Copyright (c) Microsoft. All rights reserved.
   ~  Licensed under the MIT license. See LICENSE file in the project root for full license information.
   -->
-<project>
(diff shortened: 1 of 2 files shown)


timtay-microsoft:
These tests are currently disabled due to a known issue on the service side where sometimes, scheduled jobs block indefinitely. Once the service folks have a fix for this issue, we will re-enable these tests.

JMayrbaeurl:
OK. You said sometimes. Because I've run the e2e tests several times and it always worked as expected.

JMayrbaeurl:
@timtay-microsoft Import/Export e2e IT test is again not automatically run in test phases. I've excluded it over the Maven pom.xml, but the code still gets compiled. This is probably the better approach to temporarily disable this test (till the bug on the server side gets fixed).

PR #312, opened 2018-08-20
Fix build failures under JDK 9

JonathanGiles:
A few test classes reference a Module class, but this results in compilation issues under JDK 9+, because java.lang.Module is implicitly imported, but in the test classes I fix in this PR, the Module class that is intended to be used is another, IoT-specific class.


timtay-microsoft:
This looks fine, I'll go ahead and get it merged in.

JonathanGiles:
Thanks.

PR #320, opened 2018-08-27
Upgrade Maven pom dependencies to support JDK 11

JonathanGiles:
Building the SDK under JDK 11 fails due to the outdated versions of plugins specified in the Maven pom file. This helps get this repo closer to supporting JDK 11.

### DIFF
--- a/deps/pom.xml
+++ b/deps/pom.xml
@@ -118,7 +118,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
-                <version>2.10.3</version>
+                <version>${maven.javadoc.version}</version>
                 <executions>
                     <execution>
                         <id>attach-javadocs</id>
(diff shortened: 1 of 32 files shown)


JMayrbaeurl:
Question for @timtay-microsoft : What is the preferred way to provide PRs? All changes with one commit or should we always squash it into one single commit?

timtay-microsoft:
The [coding guidelines](https://github.com/Azure/azure-iot-sdk-java/blob/master/.github/CODING_GUIDELINES.md) for this repo insist that you squash all of your commits into one so as to not clutter up the master branch's history more than necessary. Especially when all of these changes fall under a similar purpose. @JonathanGiles can you please squash all of your commits? I'd be happy to merge this in once you do.

timtay-microsoft:
Also, why are you adding new dependencies on javax.xml.bind? I don't see any code changes that necessitate adding it

JonathanGiles:
Happy to squash. The jaxb-api dependency (not to be confused with an impl) is to enable support beyond JDK 8 - when the jaxb API was removed from the JDK in JDK 9, the recommended approach became to bring in the jaxb-api dependency to enable continued compilation.

Issue #338, opened 2018-09-14
GetTwin Method Misapplies 5 nesting depth limit for module and device twin


sam-findler:
- **OS and version used:** Ubuntu 14.04

- **Java runtime used:** jdk 8u171

- **SDK version used:** 1.14.0

# Description of the issue:
Acccording to https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-device-twins#tags-and-properties-format , the nesting level inside of the reported_properties, desired_properties and tags is supposed to be 5, starting from the inside of the map.  However, the code in the client (https://github.com/Azure/azure-iot-sdk-java/blob/master/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinCollection.java) indicates that the nesting level is being counted from the top level.  We have not seen this issue in the c/python client, only in the java client.

# Code sample exhibiting the issue:

` 
//creating the property
private static Property createAlertProperty(int dataIndex) {
    Map<String, Object> alertProperties = new HashMap<>();
    Map<String, Object> sourceProperties = new HashMap<>();
    sourceProperties.put(TwinFields.OK.value(), 1);
    Map<String, Object> data = new HashMap<>();
    data.put("data_index", dataIndex);
    sourceProperties.put(TwinFields.DATA.value(), data);
    alertProperties.put("some_source_1", sourceProperties);
    alertProperties.put(TwinFields.PRIORITY.value(), 1);

    return new Property("some_alert_1", alertProperties);
  }
/*
This produces the reported property:
"some_alert_1":{  
            "some_source_1":{  
               "data":{  
                  "data_index":0
               },
               "ok":1
            },
            "priority":1
 }
Which is only a nesting level of 3 deep
*/
`
`
//sending the reported property
Property property = createAlertProperty(i);
System.out.println("Sending Property" + property.getKey());
moduleClient.sendReportedProperties(new HashSet<>(Arrays.asList(property)));
`
`
//trying to get the remote device twin state from the service client

DeviceTwin service = DeviceTwin
        .createFromConnectionString(iotHubTestController.getIotHubConnectionString());
DeviceTwinDevice device = new DeviceTwinDevice(moduleController.getModule().getDeviceId(),
        moduleController.getModule().getId());
//this is when the exception occurs
service.getTwin(device);
`

# Console log of the issue:
`
java.lang.IllegalArgumentException: Map exceed maximum of 5 levels

[...]

PR #339, opened 2018-09-14
fix nesting depth bug

sam-findler:
# Checklist
  - [NA] I edited the corresponding document in the `devdoc` folder and added or modified requirements.

https://github.com/Azure/azure-iot-sdk-java/issues/338

# Description of the problem
The nesting depth in TwinCollection.java didn't account for the full structure of a device/module twin, specifically, it didn't account for the following structure existing outside of the properties sent by a device:
```
{
    "properties":
        {
            "reported": { ... },
            "desired":{...},
            ...
        }
}
```

# Description of the solution
Increased the constant specifying nesting depth to 8 from 6 to account for the two layers that weren't considered.

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinCollection.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinCollection.java
@@ -95,8 +95,9 @@ import java.util.Map;
  */
 public class TwinCollection extends HashMap<String, Object>
 {
-    // By definition, Twin maps cannot contain more than 5 levels.
-    private static final int MAX_TWIN_LEVEL = 6;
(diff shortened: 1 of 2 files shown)


timtay-microsoft:
This PR looks good to me. I'll get this merged in shortly

sam-findler:
You're right that tags are different. I forgot that they are not nested in the "properties" section.  However, there is still 1 more layer than is accounted for, e.g. "tags:" { "first_layer": ... }, meaning that max depth for tags should be 7 rather than 6.

timtay-microsoft:
I've checked in a more comprehensive fix for this issue into master now, and we are releasing a new version of our SDK within the day. Because of that, I'm going to close this PR for now.

PR #343, opened 2018-09-27
(refactor samples)Update sample code for a downstream device that connects to an IoT Edge device

mrohera:
# Checklist
- [ x] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-java/blob/master/.github/CONTRIBUTING.md).
  - [x ] This pull-request is submitted against the `master` branch. 

# Description of the problem

Make it easier for users to connect downstream devices to an Azure IoT Edge device,

# Description of the solution
 
Updating the sample to highlights how a developer should put together a connection string for the Edge and install the relevant CA certificate required to authenticate the Azure IoT Edge server certificate.


device/iot-device-samples/send-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendEvent.java
```diff
@@ -68,14 +68,14 @@ else if (status == IotHubConnectionStatus.CONNECTED)
...
      *
      * @param args
-     * args[0] = IoT Hub connection string
-     * args[1] = number of requests to send
+     * args[0] = IoT Hub or Edge Hub connection string
+     * args[1] = number of messages to send
      * args[2] = protocol (optional, one of 'mqtt' or 'amqps' or 'https' or 'amqps_ws')
-     * args[3] = path to certificate to enable one-way authentication over ssl for amqps (optional, default shall be used if unspecified).
+     * args[3] = path to certificate to enable one-way authentication over ssl for amqps (optional, when connecting to IoT Hub, required if connecting to an Edge device using a non public root CA certificate).
      */
```
timtay-microsoft: "Optional when connecting to IotHub" is a little misleading. We shouldn't encourage users to provide special certificates just to connect to IotHub since the default certs work just fine.

how about "Not necessary when connecting directly to Iot Hub, but required if connecting to an Edge device using a non public root CA certificate"

mrohera: Will fix.

PR #345, opened 2018-09-27
Update java-devbox-setup.md

JonathanGiles:
Updating the maven versions.


jasmineymlo:
We have fixed it in the internal release scripts and the versions should be updated automatically going forward.  Thanks.

PR #347, opened 2018-10-04
Allow configurations that don't specify a schemaVersion

elvinm:
SchemaVersion is not a required property in configuration and deployments created by Azure don't specify them by default. Therefore it should not be a required property. This PR removes those checks.

 

# Checklist

# Description of the problem
* Create a configuration in azure portal.
** Note that the json manifest shows no schemaVersion at the top level.
* Calling RegistryManager.getConfigurations and it fails to retrieve the deployment due to the schemaVersion check in ConfigurationParser.

# Description of the solution
Removal of the schemaVersion check to allow it to be missing and adding accompanying unit tests.


timtay-microsoft:
This looks good to me, we'll try to get this merged in soon. We are currently hitting some CI issues, so it may take a few days.

PR #363, opened 2018-11-05
Stabilizing unit tests and some E2E tests

timtay-microsoft:
# Description of the problem
TransportClientIT will always fail on any retry attempts because it maintains the connection strings of deleted devices and tries to use them on the 2nd/3rd attempt. Because of that, the test will always fail with "Device not found" because the connection string belongs to a deleted device

JobClientIT and ExportImportIT suffer from when their spawned jobs take a while to run. Currently, our tests are expecting the jobs to finish too quickly, so this PR will up those time limits. Also, ExportImportIT needed to handle cases when the service throttles its job creation.

Unit tests are failing due to a known jmockit bug where running tests in parallel can cause what are essentially random failures. Context is [here](https://github.com/jmockit/jmockit1/issues/263). Since we run tests in parallel and do see this issue, it makes sense to add retry to our unit tests since we cannot count on jmockit to be correct 100% of the time.

### DIFF
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/serviceclient/ExportImportCommon.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/serviceclient/ExportImportCommon.java
@@ -31,9 +31,9 @@ import static junit.framework.TestCase.fail;
 
 public class ExportImportCommon
 {
-    private static final long IMPORT_EXPORT_TEST_TIMEOUT = 3 * 60 * 1000; //3 minutes
-    private static final long IMPORT_JOB_TIMEOUT = 3 * 60 * 1000; //3 minute
-    private static final long EXPORT_JOB_TIMEOUT = 3 * 60 * 1000; //3 minute
+    private static final long IMPORT_EXPORT_TEST_TIMEOUT = 8 * 60 * 1000; //3 minutes
+    private static final long IMPORT_JOB_TIMEOUT = 6 * 60 * 1000; //3 minute
+    private static final long EXPORT_JOB_TIMEOUT = 6 * 60 * 1000; //3 minute
 
     protected static String iotHubConnectionString = "";
     protected static String storageAccountConnectionString = "";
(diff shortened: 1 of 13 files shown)


jasmineymlo:
>             Thread.sleep(MAXIMUM_TIME_TO_WAIT_FOR_IOTHUB); // wait 10 seconds

nit: remove this comment, we are waiting for 0.2 sec increments.

---
Refers to: iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/serviceclient/JobClientCommon.java:435 in f52bb74. [](commit_id = f52bb743d182e2c8ae10651a79e283b1e3435fc0, deletion_comment = False)

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/serviceclient/ExportImportCommon.java
```diff
@@ -30,9 +31,9 @@
 
 public class ExportImportCommon
 {
-    private static final long IMPORT_EXPORT_TEST_TIMEOUT = 3 * 60 * 1000; //3 minutes
-    private static final long IMPORT_JOB_TIMEOUT = 1 * 60 * 1000; //1 minute
-    private static final long EXPORT_JOB_TIMEOUT = 1 * 60 * 1000; //1 minute
+    private static final long IMPORT_EXPORT_TEST_TIMEOUT = 8 * 60 * 1000; //3 minutes
```
jasmineymlo: >3 [](start = 76, length = 1)

I think it's safe to delete the command....it's pretty easy to understand these math here

iot-e2e-tests/jvm/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/serviceclient/ExportImportIT.java
```diff
@@ -10,11 +10,13 @@
 import com.microsoft.azure.sdk.iot.common.serviceclient.ExportImportCommon;
 import com.microsoft.azure.storage.StorageException;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 
 import java.io.IOException;
 import java.net.URISyntaxException;
 import java.security.InvalidKeyException;
 
+@Ignore
```
jasmineymlo: Are we adding the Ignore?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/serviceclient/ExportImportCommon.java
timtay-microsoft: Sure, I'll get rid of the comments

PR #384, opened 2018-11-29
feat(iot-dev): Enable fileupload with SelfSigned device client

### LINKED ISSUE #371: Support File Upload for X.509 authenticated devices
- **OS and version used:** Windows 10 (and Android sdk 27) 

- **Java runtime used:** 1.8.0_131 

- **SDK version used:** 1.14.1 

# Description of the issue:
When using a client that is authenticated using X.509 authentication the file upload feature does not work. As this is the primary use case for our project, it is very inconvenient that this is not supported.  We cannot use the SDK without file upload support and we must have X.509 authentication.

The documentation on the readme suggests that this is in progress. It also indicates that this will use HTTPS only. We are using MQTT for all other calls, will it simply always use HTTPS for file upload even if the protocol is specified as MQTT or is it all MQTT or all HTTPS?

Is there an ETA for file upload with X509 authentication planned?

Update 09-NOV-2018 06:03PM CST - I removed the following lines from DeviceClient.java that generate the exception and the file upload was successful with a device using X.509 authentication.... I believe this interaction uses https as the transport but at least it does not prevent it from uploading a file. What is the reason for generating this exception since it "appears" to work if I remove this check (at least the file ends up in blob storage)  :
`if (this.config.getAuthenticationType() == DeviceClientConfig.AuthType.X509_CERTIFICATE)
        {
            throw new UnsupportedOperationException("File Upload does not support x509 authentication");
        }`

# Code sample exhibiting the issue:
client.uploadToBlobAsync(fileName, new FileInputStream(file), file.length(), new FileUploadStatusCallBack(), null);

# Console log of the issue:
Exception in thread "main" java.lang.UnsupportedOperationException: File Upload does not support x509 authentication
	at com.microsoft.azure.sdk.iot.device.DeviceClient.uploadToBlobAsync(DeviceClient.java:429)
	at samples.com.microsoft.azure.sdk.iot.ProvisioningX509Sample.main(ProvisioningX509Sample.java:131)

# Need Support?
* Have a feature request for SDKs? Please post it on [User Voice](https://feedback.azure.com/forums/321918-azure-iot) to help us prioritize.
* Have a technical question? Ask on [Stack Overflow](https://stackoverflow.com/questions/tagged/azure-iot-hub) with tag “azure-iot-hub”
* Need Support? Every customer with an active Azure subscription has access to support with guaranteed response time.  Consider submitting a ticket and get assistance from Microsoft support team
* Found a bug? Please help us fix it by thoroughly documenting it and filing an issue on GitHub (C, Java, .NET, Node.js, Python).

timtay-microsoft:
Adding e2e test, too

fixes #371 

# Description of the problem
Currently, file upload code path throws unsupported operation exception when using self signed or ca signed certs as the authentication mechanism of your device client. While the service does not allow CA signed certs for file upload, it does allow self signed certs.

# Description of the solution
Removed the unsupported operation exception, and added a note about the gap with CA signed certs. Also added end to end tests to cover this scenario

### DIFF
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothubservices/FileUploadTests.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothubservices/FileUploadTests.java
@@ -13,12 +13,10 @@ import com.microsoft.azure.sdk.iot.device.IotHubClientProtocol;
 import com.microsoft.azure.sdk.iot.device.IotHubEventCallback;
 import com.microsoft.azure.sdk.iot.device.IotHubStatusCode;
 import com.microsoft.azure.sdk.iot.service.*;
+import com.microsoft.azure.sdk.iot.service.auth.AuthenticationType;
 import com.microsoft.azure.sdk.iot.service.exceptions.IotHubException;
 import com.microsoft.azure.sdk.iot.service.exceptions.IotHubNotFoundException;
-import org.junit.After;
-import org.junit.AfterClass;
-import org.junit.Before;
-import org.junit.Test;
+import org.junit.*;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
@@ -66,9 +64,14 @@ public class FileUploadTests extends MethodNameLoggingIntegrationTest
     private static DeviceClient deviceClient;
     private static ServiceClient serviceClient;
     private static Device scDevice;
+    private static Device scDevicex509;
     private static FileUploadState[] fileUploadState;
     private static MessageState[] messageStates;
     private static FileUploadNotificationReceiver fileUploadNotificationReceiver;
+
+    private static String publicKeyCertificate;
+    private static String privateKeyCertificate;
+    private static String x509Thumbprint;
 
     enum STATUS
     {
@@ -127,11 +130,15 @@ public class FileUploadTests extends MethodNameLoggingIntegrationTest
         }
     }
 
-    public static void setUp() throws IOException
+    public static void setUp(String publicK, String privateK, String thumbprint) throws IOException
     {
         registryManager = RegistryManager.createFromConnectionString(iotHubConnectionString);
         serviceClient = ServiceClient.createFromConnectionString(iotHubConnectionString, IotHubServiceClientProtocol.AMQPS);
 
+        publicKeyCertificate = publicK;
+        privateKeyCertificate = privateK;
+        x509Thumbprint = thumbprint;
+        
         fileUploadNotificationReceiver = serviceClient.getFileUploadNotificationReceiver();
         assertNotNull(fileUploadNotificationReceiver);
     }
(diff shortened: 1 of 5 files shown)


iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothubservices/FileUploadTests.java
```diff
@@ -578,7 +643,185 @@ public void run()
...
+        }
+
+        tearDownDeviceClient();
+    }
+
+    @Test (timeout = MAX_MILLISECS_TIMEOUT_KILL_TEST)
+    public void uploadToBlobAsyncSingleFileOnAMQPSelfSigned() throws URISyntaxException, IOException, InterruptedException, ExecutionException, TimeoutException
+    {
+        // arrange
+        setUpX509DeviceClient(IotHubClientProtocol.AMQPS);
```
jasmineymlo: need case for WS?

jasmineymlo: WS?

jasmineymlo: found it up there....nevermind

---
In reply to: [237619247](https://github.com/Azure/azure-iot-sdk-java/pull/384#discussion_r237619247) [](ancestors = 237619247)

jangajdorus:
Hello gentlemen, 
I have a project where we would need to upload files from devices, authorized with CA signed certificates (provisioned via DPS). Do you have some plans to introduce support for async file uploads for devices using CA signed X.509 certiticates ?

cima:
There are two parts of this problem. We have tried manually usinwg wget to request the upload link:

`wget --verbose --no-check-certificate --certificate=dopey.crt --private-key=dopey.pem --header='Content-type:application/json' --post-data='{"blobName":"pizzafile.txt"}' -O pizzafile.json  https://daedalus-iot-CDO-1.azure-devices.net/devices/dopey/files?api-version=2018-04-01`

while the resulting file contained:

`{"correlationId":"MjAxOTAXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX3ZlcjIuMA==","hostName":"daedalusstorage1.blob.core.windows.net","containerName":"gateway-logger-testing","blobName":"dopey/pizzafile.txt","sasToken":"?sv=2017-04-17&sr=b&sig=6BI75T%2FXXXXXXXXXXXXXXXXXXXiQgVahn2Ngp3p4%3D&se=2019-06-11T09%3A35%3A03Z&sp=rw"}`

then the URL for upload was composed, some file was uploaded sideways and it succeeded. Which concludes there are no restrictions on Azure side to use CA based X.509 atestation and file upload at the same time. The problem is in Java which wrongly implements the second part (upload to storage account blob). We have worked around this by using IoT hub client just to obtain the SAS token (by overloading one class and tap this information), path, etc. Implemented upload on our own.
It'd be definitely possible to find the weak point in code as it's opensource but our timeline is not in favor to opensource contribution. I hope there'll be some time in august for this.

PR #351, opened 2018-10-23
Add reporting of capabilities in DeviceTwin based on TwinState

elvinm:
# Checklist
- [x] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-java/blob/master/.github/CONTRIBUTING.md).
- [x] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).
- If this is a modification that impacts the behavior of a public API
  - [ ] I edited the corresponding document in the `devdoc` folder and added or modified requirements.
- I submitted this PR against the correct branch: 
  - [x] This pull-request is submitted against the `master` branch. 

# Reference/Link to the issue solved with this PR (if any)

# Description of the problem

Querying for devices does not properly show the device capabilities to reflect whether or not the device is an IoT Edge device.

# Description of the solution
 
The set of capabilities is already passed down through the twin state. It just needs to be set on the DeviceTwinDevice.

PR #393, opened 2018-12-07
feat(iot-service): Add reporting of capabilities in DeviceTwin based …

timtay-microsoft:
…on TwinState

 

# Description of the problem
Querying for devices does not properly show the device capabilities to reflect whether or not the device is an IoT Edge device.

# Description of the solution
The set of capabilities is already passed down through the twin state. It just needs to be set on the DeviceTwinDevice.

(rebased version of [pr 351](https://github.com/Azure/azure-iot-sdk-java/pull/351))

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
@@ -493,6 +493,7 @@ public class DeviceTwin
         deviceTwinDevice.setTags(twinState.getTags());
         deviceTwinDevice.setDesiredProperties(twinState.getDesiredProperty());
         deviceTwinDevice.setReportedProperties(twinState.getReportedProperty());
+        deviceTwinDevice.setCapabilities(twinState.getCapabilities());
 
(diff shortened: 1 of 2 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
```diff
@@ -493,6 +493,7 @@ private DeviceTwinDevice jsonToDeviceTwinDevice(String json) throws IOException
         deviceTwinDevice.setTags(twinState.getTags());
         deviceTwinDevice.setDesiredProperties(twinState.getDesiredProperty());
         deviceTwinDevice.setReportedProperties(twinState.getReportedProperty());
+        deviceTwinDevice.setCapabilities(twinState.getCapabilities());
```
prmathur-microsoft: Is this feature released from service side?

timtay-microsoft: Yes

prmathur-microsoft: Is this released for the code flow involving all the API's that call jsonToDeviceTwinDevice that exist here?

PR #410, opened 2019-01-11
refactor(e2e): Adding correlation logging details to e2e test failures

timtay-microsoft:
# Description of the problem
The e2e tests in this codebase don't always have output that correlates the test device id, hub name, etc. to it's failure logs. This PR adds in these details

# Description of the solution
All device/module-centric tests now include correlation details in all of their assertions/exceptions. Tests with just a service client don't have a big need for these correlation details as the hubs are static, and the timestamp is a part of the VSTS build already

I also added timeouts to all e2e tests that were missing a timeout

### DIFF
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/CorrelationDetailsLoggingAssert.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/CorrelationDetailsLoggingAssert.java
@@ -1,0 +1,260 @@
+/*
+ *  Copyright (c) Microsoft. All rights reserved.
+ *  Licensed under the MIT license. See LICENSE file in the project root for full license information.
+ */
+
+package com.microsoft.azure.sdk.iot.common.helpers;
+
+import com.microsoft.azure.sdk.iot.device.InternalClient;
+import org.junit.ComparisonFailure;
+
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Date;
+
+public class CorrelationDetailsLoggingAssert
+{
+    String hostname;
+    Collection<String> deviceIds;
+    String protocol;
+    Collection<String> moduleIds;
+
+    public CorrelationDetailsLoggingAssert(String hostname, String deviceId, String protocol, String moduleId)
+    {
+        this.hostname = hostname;
+        this.deviceIds = new ArrayList<>();
+        this.deviceIds.add(deviceId);
+        this.protocol = protocol;
+
+        if (moduleId != null && !moduleId.isEmpty())
+        {
+            this.moduleIds = new ArrayList<>();
+            this.moduleIds.add(moduleId);
+        }
+    }
+
+    public CorrelationDetailsLoggingAssert(String hostname, Collection<String> deviceIds, String protocol, Collection<String> moduleIds)
+    {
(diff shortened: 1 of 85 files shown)


iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/CorrelationDetailsLoggingAssert.java
```diff
@@ -0,0 +1,260 @@
...
+                moduleIds);
+    }
+
+    public void assertTrue(String message, boolean condition)
+    {
+        if (!condition)
+        {
+            fail(message);
+        }
+    }
```
prmathur-microsoft: why can't we use libraries to acheive this ? Why do we need to write this method?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/CorrelationDetailsLoggingAssert.java
```diff
@@ -0,0 +1,260 @@
...
+        return BuildExceptionMessage(
+                baseMessage,
+                client.getConfig().getDeviceId(),
+                client.getConfig().getProtocol().toString(),
+                client.getConfig().getIotHubHostname(),
+                client.getConfig().getModuleId());
+    }
+
+    public static String BuildExceptionMessage(String baseMessage, Collection<InternalClient> clients)
+    {
```
prmathur-microsoft: camelCase?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/CorrelationDetailsLoggingAssert.java
timtay-microsoft: These are instance methods that make use of the instance variables provided in the constructor. This assertTrue calls into fail() which builds a better failure message than the regular Assert.fail() would provide on its own

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/CorrelationDetailsLoggingAssert.java
timtay-microsoft: will update, good catch

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothubservices/TransportClientTests.java
```diff
@@ -644,7 +644,16 @@ public void invokeMethodAMQPSWSInvokeParallelSucceed() throws Exception
     @Test
     public void testTwin() throws IOException, InterruptedException, IotHubException, URISyntaxException
     {
-        TransportClient transportClient = setUpTwin();
+        TransportClient transportClient = null;
```
davilu: I would suggest to move this into `setUpTwin` function instead of swallow it with `fail`.
```
void setUpTwin() throws ...Exception {
   try {
   .....
   } catch(...Exception e) {
      log.error("Encountered exception during setUpTwin.");
      throw e;
   }
}
```

timtay-microsoft:
Closing this PR since I opened a different PR for the same feature + more

PR #414, opened 2019-01-23
E2E correlation details logging + E2E fixes

timtay-microsoft:
# Description of the problem
e2e tests do not currently log out any device id/hub name correlation details when they fail. This pr adds that functionality

Also, desired properties tests and transport client tests had issues with reruns, so this pr fixes those issues.

This pr also has a few small fixes in misc areas

For an outline of each commit, and why we are adding it see below:

 **Correlation logging details**: We need to log the relevant device id(s) and hub name for any failure

**Adjusting Timeouts**: Removed test timeouts at the individual test level, now all tests inherit a 15 minute timeout which is only for preventing stalled builds. Ideally, these 15 minute timeouts are never hit

**TransportClient test fixes**: Some state was maintained between test reruns since the arraylist was never cleared. Instead, we'll now use arrays so they can't grow over time like they used to

**DesiredProperties test fixes**: Similar to transport client. Some twin state was maintained from previous test runs.

**SendMessageErrInj**: The test class had a test that was hardcoded to use a device client, even when the ClientType given to it was ModuleClient. Also, migrating to using ClientType the enum instead of a string when passing Device vs Module to all test classes.

**Ignore Error Injection Auth/Throttle/Quota**: Disabling due to service side issue

### DIFF
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IntegrationTest.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IntegrationTest.java
@@ -10,7 +10,7 @@ import org.junit.rules.TestRule;
 import org.junit.rules.TestWatcher;
 import org.junit.runner.Description;
 
-public class MethodNameLoggingIntegrationTest
+public class IntegrationTest
 {
     @Rule
     public TestRule watcher = new TestWatcher()
@@ -18,6 +18,11 @@ public class MethodNameLoggingIntegrationTest
         protected void starting(Description description)
         {
             System.out.println("Starting test: " + description.getMethodName());
+        }
+
+        protected void finished(Description description)
+        {
+            System.out.println("Finished test: " + description.getMethodName());
         }
     };
 }
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IntegrationTest.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IntegrationTest.java
@@ -8,6 +8,7 @@ package com.microsoft.azure.sdk.iot.common.helpers;
 import org.junit.Rule;
 import org.junit.rules.TestRule;
 import org.junit.rules.TestWatcher;
+import org.junit.rules.Timeout;
 import org.junit.runner.Description;
 
 public class IntegrationTest
@@ -25,4 +26,10 @@ public class IntegrationTest
             System.out.println("Finished test: " + description.getMethodName());
         }
     };
+
+    public static final int E2E_TEST_TIMEOUT_MS = 15 * 60 * 1000;
+
+    //This timeout applies to all individual tests in classes that inherit from this class
+    @Rule
+    public Timeout timeout = new Timeout(E2E_TEST_TIMEOUT_MS);
 }
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IotHubServicesCommon.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IotHubServicesCommon.java
@@ -448,17 +448,4 @@ public class IotHubServicesCommon
 
         return false;
     }
-
-    public static String actualStatusUpdatesToString(List<Pair<IotHubConnectionStatus, Throwable>> actualStatusUpdates)
-    {
-        String statusString = "";
-        for (int i = 0; i < actualStatusUpdates.size(); i++)
-        {
-            IotHubConnectionStatus status = actualStatusUpdates.get(i).getKey();
-            Throwable cause = actualStatusUpdates.get(i).getValue();
-            statusString += i + "th status update: " + status + ((cause == null) ? "" : " with throwable " + Tools.getStackTraceFromThrowable(cause) + "\n");
-        }
-
-        return statusString;
-    }
 }
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IotHubServicesCommon.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IotHubServicesCommon.java
@@ -6,16 +6,17 @@
 package com.microsoft.azure.sdk.iot.common.helpers;
 
 import com.microsoft.azure.sdk.iot.device.*;
+import com.microsoft.azure.sdk.iot.device.DeviceTwin.Pair;
 import com.microsoft.azure.sdk.iot.device.transport.IotHubConnectionStatus;
 import com.microsoft.azure.sdk.iot.service.auth.AuthenticationType;
 import org.junit.Assert;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 
-import static junit.framework.TestCase.assertTrue;
-import static junit.framework.TestCase.fail;
+import static com.microsoft.azure.sdk.iot.common.helpers.CorrelationDetailsLoggingAssert.buildExceptionMessage;
 
 /*
  * This class contains common code for Junit and Android test cases
@@ -36,14 +37,14 @@ public class IotHubServicesCommon
                                     final long RETRY_MILLISECONDS,
                                     final long SEND_TIMEOUT_MILLISECONDS,
                                     long interMessageDelay,
-                                    List<IotHubConnectionStatus> statusUpdates) throws IOException, InterruptedException
+                                    List<Pair<IotHubConnectionStatus, Throwable>> statusUpdates) throws IOException, InterruptedException
     {
         try
         {
             openClientWithRetry(client);
             if (statusUpdates != null)
             {
-                confirmOpenStabilized(statusUpdates, 120000);
+                confirmOpenStabilized(statusUpdates, 120000, client);
             }
 
             for (int i = 0; i < messagesToSend.size(); ++i)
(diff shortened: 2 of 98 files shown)


iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothubservices/TransportClientTests.java
```diff
@@ -403,43 +403,43 @@ public void run()
...
+                verifyNotification(fileUploadNotification, fileUploadState[i], clientArrayList.get(indexJ));
+                Assert.assertTrue(buildExceptionMessage("File upload callback was not triggered for file upload attempt " + i, clientArrayList.get(indexJ)), fileUploadState[i].isCallBackTriggered);
+                Assert.assertEquals(buildExceptionMessage("File upload status was not successful on attempt " + i + ", status was: " + fileUploadState[i].fileUploadStatus, clientArrayList.get(indexJ)), fileUploadState[i].fileUploadStatus, SUCCESS);
+                Assert.assertEquals(buildExceptionMessage("Message status was not successful on attempt " + i + ", status was: " + messageStates[i].messageStatus, clientArrayList.get(indexJ)), messageStates[i].messageStatus, SUCCESS);
             }
         }
 
         executor.shutdown();
-        if (!executor.awaitTermination(10000, TimeUnit.MILLISECONDS))
+        if (!executor.awaitTermination(5 * 60 * 1000, TimeUnit.MILLISECONDS)) //5 minutes
```
jasmineymlo: >60 [](start = 43, length = 2)

declare this as psf?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothubservices/TransportClientTests.java
```diff
@@ -722,43 +724,43 @@ public void run()
...
                     }
-                    assertEquals(devicesUnderTest.get(finalI).deviceTwinStatus, SUCCESS);
+                    Assert.assertEquals(buildExceptionMessage("Expected SUCCESS but was " + devicesUnderTest[finalI].deviceTwinStatus, devicesUnderTest[finalI].deviceClient), SUCCESS, devicesUnderTest[finalI].deviceTwinStatus);
                 }
             });
         }
         Thread.sleep(MAXIMUM_TIME_TO_WAIT_FOR_IOTHUB_TWIN_OPERATION);
         executor.shutdown();
-        if (!executor.awaitTermination(10000, TimeUnit.MILLISECONDS))
+        if (!executor.awaitTermination(4 * 60 * 1000, TimeUnit.MILLISECONDS)) //4 minutes
```
jasmineymlo: use same value as above

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IntegrationTest.java
```diff
@@ -8,9 +8,10 @@
 import org.junit.Rule;
 import org.junit.rules.TestRule;
 import org.junit.rules.TestWatcher;
+import org.junit.rules.Timeout;
 import org.junit.runner.Description;
 
-public class MethodNameLoggingIntegrationTest
+public class IntegrationTest
```
jasmineymlo: nit: xxxxTest suggests test implementation.  But this class is used for test logging, maybe e2eTestLogging??

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/setup/DeviceTwinCommon.java
```diff
@@ -41,12 +42,13 @@
...
  * but any child class should.
  */
-public class DeviceTwinCommon extends MethodNameLoggingIntegrationTest
+public class DeviceTwinCommon extends IntegrationTest
 {
     // Max time to wait to see it on Hub
     protected static final long PERIODIC_WAIT_TIME_FOR_VERIFICATION = 100; // 0.1 sec
     protected static final long MAX_WAIT_TIME_FOR_VERIFICATION = 180000; // 180 sec
     protected static final long DELAY_BETWEEN_OPERATIONS = 200; // 0.2 sec
+    public static final long MULTITHREADED_WAIT_TIMEOUT_MS  = 5 * 60 * 1000; // 5 minutes
```
jasmineymlo: eventually, I think we should consolidate all timeout into one central place for reusing...

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IntegrationTest.java
timtay-microsoft: This class represents our most abstract concept of an integration test, and the rules it must follow. I'm not sure e2eTestLogging covers that, especially since we now have a test timeout rule defined in this class. I'd rather keep this as IntegrationTest, or maybe change it to AbstractIntegrationTest, or something similar. For now, I think IntegrationTest works well.

davilu: 1. `abstract`
2. Name as `AbstractWatchedTest`
3. use logger
```java
protected final Logger log = LoggerFactory.getLogger(getClass());
    @Rule
    public TestRule watcher = new TestWatcher() {
        @Override
        protected void starting(Description description) {
            log.info("Test[{}] started.", description.getMethodName());
        }
        @Override
        protected void failed(Throwable e, Description description) {
            log.error("Test[{}] failed.", description.getMethodName());
        }
        @Override
        protected void succeeded(Description description) {
            log.info("Test[{}] was successful.", description.getMethodName());
        }
    };
```

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IntegrationTest.java
```diff
@@ -19,5 +20,16 @@ protected void starting(Description description)
...
+        {
+            System.out.println("Finished test: " + description.getMethodName());
+        }
     };
+
+    public static final int E2E_TEST_TIMEOUT_MS = 15 * 60 * 1000;
+
+    //This timeout applies to all individual tests in classes that inherit from this class
+    @Rule
+    public Timeout timeout = new Timeout(E2E_TEST_TIMEOUT_MS);
```
davilu: `public Timeout timeout = new Timeout(15, TimeUnit.MINUTES);`

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IotHubServicesCommon.java
```diff
@@ -36,14 +37,14 @@ public static void sendMessages(InternalClient client,
                                     final long RETRY_MILLISECONDS,
                                     final long SEND_TIMEOUT_MILLISECONDS,
                                     long interMessageDelay,
-                                    List<IotHubConnectionStatus> statusUpdates) throws IOException, InterruptedException
+                                    List<Pair<IotHubConnectionStatus, Throwable>> statusUpdates) throws IOException, InterruptedException
```
davilu: I didn't see `Throwable` is used.

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/Tools.java
```diff
@@ -78,4 +84,136 @@ public static void removeDevicesAndModules(RegistryManager registryManager, Coll
...
+
+    public static Device addDeviceWithRetry(RegistryManager registryManager, Device device) throws IotHubException, IOException, InterruptedException
+    {
+        long startTime = System.currentTimeMillis();
+        Device ret = null;
+        while (System.currentTimeMillis() - startTime < RETRY_TIMEOUT_ON_NETWORK_FAILURE)
+        {
+            try
+            {
+                ret = registryManager.addDevice(device);
```
davilu: Is `addDevice` has timeout?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/Tools.java
```diff
@@ -78,4 +84,136 @@ public static void removeDevicesAndModules(RegistryManager registryManager, Coll
...
+
+    public static Module addModuleWithRetry(RegistryManager registryManager, Module module) throws IotHubException, IOException, InterruptedException
+    {
+        long startTime = System.currentTimeMillis();
+        Module ret = null;
+        while (System.currentTimeMillis() - startTime < RETRY_TIMEOUT_ON_NETWORK_FAILURE)
+        {
+            try
+            {
+                ret = registryManager.addModule(module);
```
davilu: The same as above

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/setup/DeviceMethodCommon.java
```diff
@@ -62,11 +62,10 @@
     //How many milliseconds between retry
     protected static final Integer RETRY_MILLISECONDS = 100;
 
-    private List<IotHubConnectionStatus> actualStatusUpdates;
+    private List<Pair<IotHubConnectionStatus, Throwable>> actualStatusUpdates;
```
davilu: Is `Throwable` used?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/setup/SendMessagesCommon.java
```diff
@@ -200,6 +200,20 @@ public SendMessagesTestInstance(InternalClient client, IotHubClientProtocol prot
             this.publicKeyCert = publicKeyCert;
             this.privateKey = privateKey;
             this.x509Thumbprint = x509Thumbprint;
+
+            String deviceId = "";
+            String moduleId = "";
+            if (identity instanceof Module)
```
davilu:
```java
deviceId = identity.getDeviceId(); 
if (identity instanceof Module) 
{
   moduleId = ((Module) identity).getId(); 
}
```

Issue #334, opened 2018-09-12
Android client can't connect if many messages in the queue


vernazza:
- **OS and version used:** <VERSION> Android

- **Java runtime used:** <VERSION> Android SDK 28

- **SDK version used:** <VERSION> com.microsoft.azure.sdk.iot:iot-device-client:1.14.0

# Description of the issue:
When in the Iot message queue there are many messages in the cloud to device message queue (i.e. at least 12), the iothub client can't connect.
The `DeviceClient.open()` function throws an exception, after very few mseconds

# Code sample exhibiting the issue:
>                deviceClient = new DeviceClient(connectionString, IotHubClientProtocol.MQTT);
>                deviceClient.setMessageCallback(this, this);
>                deviceClient.registerConnectionStatusChangeCallback(this, this);
>                deviceClient.open();

# Console log of the issue:
An exception is thrown. The exception is
`java.io.IOException: Could not open the connection`

The cause field of the exception is:
`com.microsoft.azure.sdk.iot.device.exceptions.ProtocolException: Unable to subscribe to topic :devices/DEVICEID/messages/devicebound/#`

And the cause field in the cause field is
`Timed out waiting for a response from the server (32000)`

The exception is thrown very quickly, so it doesn't seem to match the "timeout" reason

timtay-microsoft:
Are you seeing this behavior consistently? And what happens if you use AMQPS?

vernazza:
I did some extra tests and:
- yes, the behaviour is consistent
- the problem appears when there are 12 or more messages. It doesn't matter the size of the message. 11 messages are ok, 12 or more aren't
- the problem is happening with MQTT and MQTT_WS; while using AMQPS or AMQPS_WS the library works fine

jasmineymlo:
@vernazza, sorry for the delay.  I am able to repro this bug and it's not only on Android.  Investigating and will keep you posted.  Thanks.

timwieman:
+1. Seeing this on Linux Ubuntu 18.04

timwieman:
This is the stack trace for the exception.  Note that when stepping through in the debugger in VS Code, the connect exception is *not* hit. 
```
java.io.IOException: Could not open the connection
    at com.microsoft.azure.sdk.iot.device.DeviceIO.open (DeviceIO.java:164)
    at com.microsoft.azure.sdk.iot.device.InternalClient.open (InternalClient.java:130)
[...]

PR #418, opened 2019-01-25
fix(iot-deps,iot-device-client): introduced separate locks for status change and incoming/publishing message

davilu:
# Checklist

https://github.com/Azure/azure-iot-sdk-java/issues/334

# Description of the problem

[Device Client] Device Client connect timeout with MQTT if there are 10~ messages stack on IoT Hub.
# Description of the solution
 
It was cased by sharing the same lock in different threads when subscribe is called. Separating locks helped to avoid deadlock during this process. Next step we should look closer to the sync locks and avoid hold lock during I/O operation.

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/mqtt/MqttConnection.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/mqtt/MqttConnection.java
@@ -5,8 +5,8 @@
 
 package com.microsoft.azure.sdk.iot.deps.transport.mqtt;
 
-import com.microsoft.azure.sdk.iot.deps.util.ObjectLock;
-import org.eclipse.paho.client.mqttv3.*;
+import org.eclipse.paho.client.mqttv3.IMqttDeliveryToken;
+import org.eclipse.paho.client.mqttv3.IMqttToken;
 import org.eclipse.paho.client.mqttv3.MqttAsyncClient;
 import org.eclipse.paho.client.mqttv3.MqttCallback;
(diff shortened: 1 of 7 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/mqtt/MqttConnection.java
```diff
@@ -34,11 +34,6 @@
 
     private MqttListener mqttListener;
 
-    private final ObjectLock openLock = new ObjectLock();
-
-    // paho mqtt only supports 10 messages in flight at the same time
-    static final int MAX_IN_FLIGHT_COUNT = 10;
```
timtay-microsoft: >MAX_IN_FLIGHT_COUNT [](start = 21, length = 19)

This value is still being used in Mqtt.java

jasmineymlo:
>                 synchronized (this.publishLock)

do we still need it here  since we have it at a level up?

---
Refers to: device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java:225 in 20c6d0e. [](commit_id = 20c6d0e2f0247b518e28c8013004a6870d548662, deletion_comment = False)

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/mqtt/MqttConnection.java
davilu: No, it's not. It is `MqttConnection` in device project not deps.

PR #420, opened 2019-01-25
refactor(iot-dev): Simplify IotHubCertificateManager

timtay-microsoft:
IotHubCertificate manager has a lot of unnecessary code around reading certificate files into strings before reading them into the certificate factory. This PR simplifies the class and cuts down on some unnecessary unit tests


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubCertificateManager.java
```diff
@@ -105,83 +112,49 @@
...
-            //Codes_SRS_IOTHUBCERTIFICATEMANAGER_25_007: [**This method shall throw IllegalArgumentException if parameter is null.**]**
-            throw new IllegalArgumentException("Cert path cannot be null");
+            throw new IllegalArgumentException("Certificate cannot be null or empty");
         }
 
-        //Codes_SRS_IOTHUBCERTIFICATEMANAGER_25_008: [*This method shall save the cert provided by the user a s valid cert to be used to communicate with IotHub.**]**
-        this.validCert = cert;
+        //Codes_SRS_IOTHUBCERTIFICATEMANAGER_34_005: [**This function shall read the certificates from the provided
+        // string and save them into a certificate collection.**]**
+        try (InputStream reader = new ByteArrayInputStream(cert.getBytes()))
```
davilu: Please rename `reader` to `inputStream`

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubCertificateManager.java
```diff
@@ -105,83 +112,49 @@
...
-            }
-
-            if (cert.length() == 0)
-            {
-                //Codes_SRS_IOTHUBCERTIFICATEMANAGER_25_004: [**This method shall throw IllegalArgumentException if certificate contents were empty.**]**
-                throw new IllegalArgumentException("Cert cannot be Empty");
-            }
-
-            this.validCert = cert.toString();
+            certificates = certificateFactory.generateCertificates(fis);
```
davilu: Check `certificates` to not be empty will still be required.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubCertificateManager.java
timtay-microsoft: sure thing

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubCertificateManager.java
timtay-microsoft: sure thing

timtay-microsoft:
I have addressed all of the comments above now

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubCertificateManager.java
```diff
@@ -105,83 +110,60 @@
...
+        }
 
-        if (cert == null || cert.length() == 0)
+        //Codes_SRS_IOTHUBCERTIFICATEMANAGER_34_005: [**This function shall read the certificates from the provided
+        // string and save them into a certificate collection.**]**
+        try (InputStream inputStream = new ByteArrayInputStream(cert.getBytes()))
         {
-            //Codes_SRS_IOTHUBCERTIFICATEMANAGER_25_007: [**This method shall throw IllegalArgumentException if parameter is null.**]**
-            throw new IllegalArgumentException("Cert path cannot be null");
+            certificates = certificateFactory.generateCertificates(inputStream);
```
davilu: One side effect here is if `cert` is empty, certificates will still be replaced and the instance is polluted. It might be alright since if someone want to customize certificates, they won't care about default value any more. IMPO, since there is exception thrown, I'd prefer certificates not be changed, validate and then set.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubCertificateManager.java
```diff
@@ -105,83 +110,60 @@
...
-                cert.append("\r\n");
-                readLineCount++;
-            }
-
-            if (cert.length() == 0)
-            {
-                //Codes_SRS_IOTHUBCERTIFICATEMANAGER_25_004: [**This method shall throw IllegalArgumentException if certificate contents were empty.**]**
-                throw new IllegalArgumentException("Cert cannot be Empty");
-            }
+            certificates = certificateFactory.generateCertificates(fis);
```
davilu: The same as below.

PR #444, opened 2019-02-19
E2E tests - add extra string operation on method handler

BertKleewein:
Our tests are passing strings as method parameters, and these get passed differently depending on where the method is invokes.  From the service, they are not json-encoded.  From the (node) module client, they are.  The short-term fix is to add an extra encoding.  The longer-term fix is to either sort this out or to stop using string parameters.

Question for the team: some editor (I assume IntelliJ) reformatted a bunch of my catch blocks.  Did it change them to the right form or is this needless churn?


timtay-microsoft:
Closing this PR because its commit was merged in from a different PR

Issue #395, opened 2018-12-08
Callback for device method over MQTT gets stuck on network connection loss scenarios


anhashia:
- **OS and version used:** <VERSION> 
Windows 10

- **Java runtime used:** <VERSION> 

java version "1.8.0_111"
Java(TM) SE Runtime Environment (build 1.8.0_111-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.111-b14, mixed mode)

- **SDK version used:** <VERSION> 

Java IotHub Device Client (com.microsoft.azure.sdk.iot:iot-device-client:1.13.1)
# Description of the issue:

If device response for MQTT response on method invocation is not processed because of loss of network connection, that call seems to remain pending and is never processed

# Code sample exhibiting the issue:

```````
private static int method_command(Object command) 
    {
        System.out.println("invoking command on this device");
        
        try {
            Thread.sleep(1000 *60);
            return METHOD_SUCCESS;
        }
        catch (Exception e)
        {
            return 400;
        }
        // Insert code to invoke command here
        
    }
`````````````````
Loss of network connection at   Thread.sleep(1000 *60) for some time.

# Console log of the issue:

````````````````````````
IoT Hub responded to device method operation with status OK_EMPTY
invoking command on this device
IoT Hub responded to device method operation with status OK_EMPTY

CONNECTION STATUS UPDATE: DISCONNECTED_RETRYING
CONNECTION STATUS REASON: NO_NETWORK
CONNECTION STATUS THROWABLE: Mqtt connection lost

com.microsoft.azure.sdk.iot.device.exceptions.ProtocolException: Mqtt connection lost
	at com.microsoft.azure.sdk.iot.device.transport.mqtt.exceptions.PahoExceptionTranslator.convertToMqttException(PahoExceptionTranslator.java:73)
	at com.microsoft.azure.sdk.iot.device.transport.mqtt.Mqtt.connectionLost(Mqtt.java:340)
	at org.eclipse.paho.client.mqttv3.internal.CommsCallback.connectionLost(CommsCallback.java:292)
	at org.eclipse.paho.client.mqttv3.internal.ClientComms.shutdownConnection(ClientComms.java:423)
	at org.eclipse.paho.client.mqttv3.internal.CommsReceiver.run(CommsReceiver.java:181)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #445, opened 2019-02-20
fix(iot-dev): reuse IotHubConnection objects on reconnection #395

jasmineymlo:


jasmineymlo:
Preview for the fix and e2e tests addition if any will be added in the next iteration.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -164,6 +166,21 @@ public void addDeviceOperationSession(DeviceClientConfig deviceClientConfig) thr
         }
     }
 
+    public void reset(ScheduledExecutorService scheduledExecutorService) throws TransportException
```
timtay-microsoft: >reset [](start = 16, length = 5)

Could this logic be moved to the open call instead? That way, we only have to close, then open the xxxiothubconection objects

timtay-microsoft: if we can, we would also need to move some of the constructor logic into the open call, too

---
In reply to: [258633143](https://github.com/Azure/azure-iot-sdk-java/pull/445#discussion_r258633143) [](ancestors = 258633143)

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -112,7 +109,7 @@ public AmqpsIotHubConnection(DeviceClientConfig config, ScheduledExecutorService
             throw new IllegalArgumentException("hubName cannot be null or empty.");
         }
 
-        this.scheduledExecutorService = scheduledExecutorService;
+        //this.scheduledExecutorService = scheduledExecutorService;
```
timtay-microsoft: can this be removed?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -194,8 +169,20 @@ public void reset(ScheduledExecutorService scheduledExecutorService) throws Tran
...
+        reconnectionScheduled = false;
+        connectionId = UUID.randomUUID().toString();
+
+        this.scheduledExecutorService = scheduledExecutorService;
+
+        this.closeLatch = new CountDownLatch(1);
+        this.openLatch = new CountDownLatch(1);
+        this.savedException = null;
+
+        this.state = IotHubConnectionStatus.DISCONNECTED;
```
timtay-microsoft: Shouldn't this method check the state of the connection object before setting it to DISCONNECTED? What if it was already connected here?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
jasmineymlo: fixed

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
jasmineymlo: fixed

Issue #447, opened 2019-02-24
Downstream device connects to IoTHub instead of IoT Edge gateway


wickste:
I am trying to connect a downstream device with the Azure IoT Java device SDK to my IoT Edge gateway. I am using this guide: https://docs.microsoft.com/en-us/azure/iot-edge/how-to-connect-downstream-device#use-certificates-with-azure-iot-sdks  

However, specifying the GatewayHostName in the connection string seems to be ignored by the Java SDK. The device continues to connect to IoTHub directly, not to the Edge device. No error message on the device or in the edgeHub logs. I know I can reach the Edge device because when I put the gateway's DNS name as the Hostname in the connection string I see an error in the edgeHub logs. Also on the same machine it works correctly when using .NET and C-SDK.

- **OS and version used:** <VERSION> 
Windows 10 1809 (build 17763.316) Enterprise

- **Java runtime used:** <VERSION> 
openjdk 11.0.2 2019-01-15 LTS
OpenJDK Runtime Environment 19.1-(Zulu-11.29+3-win_x64)-Microsoft-Azure-restricted (build 11.0.2+7-LTS)
OpenJDK 64-Bit Server VM 19.1-(Zulu-11.29+3-win_x64)-Microsoft-Azure-restricted (build 11.0.2+7-LTS, mixed mode)

- **SDK version used:** <VERSION> 
1.13.30

# Description of the issue:
I am starting with the Java quickstart for sending telemetry to IoTHub. Then I modify the connection string by appending ";GatewayHostName=<gateway_dns_name>". This seems to be ignored by the Java SDK, it continues to connect to IoTHub instead of edgeHub. No error on the device or the edge.

# Code sample exhibiting the issue:
Repros with the quickstart sample code from the SDK

# Console log of the issue:

There is no error. The bug is that it connects to IoTHub instead of edgeHub.

# Need Support?
* Have a feature request for SDKs? Please post it on [User Voice](https://feedback.azure.com/forums/321918-azure-iot) to help us prioritize.
* Have a technical question? Ask on [Stack Overflow](https://stackoverflow.com/questions/tagged/azure-iot-hub) with tag “azure-iot-hub”
* Need Support? Every customer with an active Azure subscription has access to support with guaranteed response time.  Consider submitting a ticket and get assistance from Microsoft support team
* Found a bug? Please help us fix it by thoroughly documenting it and filing an issue on GitHub (C, Java, .NET, Node.js, Python).

[...]

PR #451, opened 2019-02-26
fix(iot-dev): Make "GatewayHostName" connection string attribute case…

timtay-microsoft:
… insensitive

allow attribute to be written as "GatewayHostname", "Gatewayhostname", etc.

issue #447


timtay-microsoft:
Updated this PR to extend to all connection string attributes including "HostName", "ModuleId", etc.

timtay-microsoft:
This feature was merged in through a separate pr

PR #460, opened 2019-03-07
fix(iot-service): Fix IotHubConnectionStringBuilder to be more thread…

timtay-microsoft:
… safe

keyValueMap may be cleared from another thread if two threads both call into parse(...) around the same time. This gives connection string objects that are missing important key-value pairs. It is safer to declare this map locally instead of globally


timtay-microsoft:
This feature was merged in through a separate pr

PR #462, opened 2019-03-08
fix(iot-dev): fix issue where mqtt can deadlock if too many messages …

timtay-microsoft:
…are sent at once

In the event that the device client sends a lot of telemetry at once, the mqtt publish lock will be grabbed by the sending thread and won't allow the receiving thread to ever propagate the ack up to IotHubTransport. The sending thread never lets go of the publish lock as it waits until the getPendingDeliveryTokens() call lowers, but it can't lower until the ack thread finishes. 

This fix reduces the scope of where the publishLock covers. Now, it only makes the actual sending and receiving of acks atomic. This still protects the shared unacknowledgedSentMessages object from race conditions.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -106,12 +106,15 @@ public void onMessageSent(Message message, Throwable e)
...
 
         // remove from in progress queue and add to callback queue
-        IotHubTransportPacket packet;
+        IotHubTransportPacket packet = null;
         synchronized (this.inProgressMessagesLock)
         {
-            //Codes_SRS_IOTHUBTRANSPORT_34_004: [This function shall retrieve a packet from the inProgressPackets
-            // queue with the message id from the provided message if there is one.]
-            packet = inProgressPackets.remove(message.getMessageId());
+            if (message != null)
```
jasmineymlo: >null [](start = 27, length = 4)

add else statement on line 107 instead of checkout message != null in multiple places

timtay-microsoft: sure thing

timtay-microsoft:
Adding the paho upgrade on top of this PR, since it is MQTT related

timtay-microsoft:
Nevermind, the paho upgrade seems to cause a regression in mqtt connectivity. I took it out from this PR

PR #470, opened 2019-03-22
Create mqtt perf test sample code to be reviewed

timtay-microsoft:
Not actually removing the old sample, this pr is only for reviewing how to best test perf numbers over mqtt

Do not merge

### DIFF
--- a/device/iot-device-samples/send-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendEvent.java
+++ b/device/iot-device-samples/send-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendEvent.java
@@ -8,31 +8,13 @@ import com.microsoft.azure.sdk.iot.device.transport.IotHubConnectionStatus;
 
 import java.io.IOException;
 import java.net.URISyntaxException;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 
 
 /** Sends a number of event messages to an IoT Hub. */
 public class SendEvent
 {
-    private  static final int D2C_MESSAGE_TIMEOUT = 2000; // 2 seconds
-    private  static List failedMessageListOnClose = new ArrayList(); // List of messages that failed on close
-
-    protected static class EventCallback implements IotHubEventCallback
-    {
-        public void execute(IotHubStatusCode status, Object context)
-        {
-            Message msg = (Message) context;
-
-            System.out.println("IoT Hub responded to message "+ msg.getMessageId()  + " with status " + status.name());
-
-            if (status==IotHubStatusCode.MESSAGE_CANCELLED_ONCLOSE)
-            {
-                failedMessageListOnClose.add(msg.getMessageId());
-            }
-        }
-    }
-
     protected static class IotHubConnectionStatusChangeCallbackLogger implements IotHubConnectionStatusChangeCallback
     {
         @Override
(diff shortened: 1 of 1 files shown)


device/iot-device-samples/send-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendEvent.java
```diff
@@ -10,233 +10,57 @@
...
-        // Set your token expiry time limit here
-        long time = 2400;
-        client.setOption("SetSASTokenExpiryTime", time);
-        System.out.println("Updated token expiry time to " + time);
-
-        client.registerConnectionStatusChangeCallback(new IotHubConnectionStatusChangeCallbackLogger(), new Object());
+        final long messagesToSend = 1000000;
+        String connString = "<device connection string for a device in your B3 or S3 iot hub>";
+        DeviceClient client = new DeviceClient(connString, IotHubClientProtocol.MQTT);
+        long clientSendInterval = 10; //Lower number here spawns send threads more frequently, can send more quickly. By default, value is 10
```
prmathur-microsoft: If it is default then why are we tweaking it?

device/iot-device-samples/send-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendEvent.java
```diff
@@ -10,233 +10,57 @@
...
-                EventCallback callback = new EventCallback();
-                client.sendEventAsync(msg, callback, msg);
-            }
-
-            catch (Exception e)
-            {
-                e.printStackTrace(); // Trace the exception
-            }
+            Message message = new Message("a");
+            EventCallback callback = new EventCallback();
```
prmathur-microsoft: Lets use a standard byte count here as oppose to a char eg say a msg for 64Bytes, 1KBytes and 1MBytes and note differences

prmathur-microsoft: And teh right way to measure time here would be to pass the start time in callback and note the stop time when ack is received. If not you are accounting for creating of the message as well. Even if that does't cause network IO, it may change the metrics.

device/iot-device-samples/send-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendEvent.java
timtay-microsoft: We aren't, but it remains in the sample so that we can change it easily if we want to test its impact

device/iot-device-samples/send-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendEvent.java
timtay-microsoft: Sure thing, good idea

timtay-microsoft: I can move the message creation outside of the timed loop, instead. I'll just have an array of messages created for the timed loop to pull from. 

I don't follow putting the start time in the callback. We are measuring the time it takes to send thousands of messages, not the time it takes to send each message

device/iot-device-samples/send-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendEvent.java
```diff
@@ -10,233 +10,79 @@
...
-        System.out.println("Opened connection to IoT Hub.");
-        System.out.println("Sending the following event messages:");
-
-        String deviceId = "MyJavaDevice";
-        double temperature = 0.0;
-        double humidity = 0.0;
-
-        for (int i = 0; i < numRequests; ++i)
+        final long startTime = System.currentTimeMillis();
+        for (int sentMessageCount = 0; sentMessageCount < numberOfMessagesToSend; sentMessageCount++)
```
prmathur-microsoft: Also this is sequential sending. We need to send it in parallel to execute the right threshholds of iothub

device/iot-device-samples/send-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendEvent.java
```diff
@@ -10,233 +10,79 @@
...
+        countDownLatch.await(90, TimeUnit.MINUTES);
 
-        // Wait for IoT Hub to respond.
-        try
-        {
-            Thread.sleep(D2C_MESSAGE_TIMEOUT);
-        }
+        final long stopTime = System.currentTimeMillis();
+        double secondsTaken = ((stopTime - startTime) / 1000.0);
+        System.out.println("Time taken: " + secondsTaken + " seconds");
```
prmathur-microsoft: I still see time being calculated from start to stop. So we are still accounting for extra processing. Can we use the timestamps exactly at the time of message being sent and ack received to get the right perf numbers?

prmathur-microsoft: Like I mentioned earlier, use the callback and stamp it when you get the ack. you should use context object to store this time for that message. Once you gather all the start and stop times for each message being sent you can average it by number of messages sent. This way of calculating time will be more accurate.

device/iot-device-samples/send-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendEvent.java
prmathur-microsoft: i.e change this to use executors

timtay-microsoft: done

device/iot-device-samples/send-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendEvent.java
timtay-microsoft: sure, I can try that

timtay-microsoft:
Closing this PR as it was never meant to be merged in

PR #484, opened 2019-04-09
fix(iot-dev): Fix issue where amqps connection did not refresh token …

timtay-microsoft:
…appropriately

cbs renewal task would wake up before the sas token should be ready to be proactively renewed. Now, when the cbs renewal thread wakes up, it forces a new sas token to be generated.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
```diff
@@ -35,7 +35,7 @@
     protected IotHubSasToken sasToken;
 
     public abstract boolean canRefreshToken();
-    public abstract String getRenewedSasToken(boolean proactivelyRenew) throws IOException, TransportException;
+    public abstract String getRenewedSasToken(boolean proactivelyRenew, boolean forceRenewal) throws IOException, TransportException;
```
jasmineymlo: current references to this function are always with the same values of forceRenewal and proactivelyRenew.  I see this addition of the argument as unnecessary.  But not a blocker...

PR #497, opened 2019-04-24
Add security scope feature

timtay-microsoft:


iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/serviceclient/RegistryManagerTests.java
```diff
@@ -374,7 +375,38 @@ public void apply_configuration_e2e() throws Exception
         registryManager.applyConfigurationContentOnDevice(deviceForTest, content);
     }
 
+    @Test
+    public void deviceCreationWithSecurityScope() throws IOException, InterruptedException, IotHubException
```
jasmineymlo: Should we combine this with the basic crud test above?  seems like some redundancy here...

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/Device.java
```diff
@@ -183,6 +185,16 @@ public void setCapabilities(DeviceCapabilities capabilities)
         this.capabilities = capabilities;
     }
 
+    public String getScope()
```
jasmineymlo: add comments for javadoc?

timtay-microsoft: good idea

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/serviceclient/RegistryManagerTests.java
timtay-microsoft: sure thing

timtay-microsoft: Actually, since the registry manager ops aren't that expensive, I'd rather keep this test separate. It has some edge-specific terminology that I'd rather not mix into a vanilla Iot hub test

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/Device.java
```diff
@@ -183,6 +185,34 @@ public void setCapabilities(DeviceCapabilities capabilities)
...
```
jasmineymlo: duplicate below....

timtay-microsoft: good catch

PR #498, opened 2019-04-24
Allow clients to unsubscribe from twin

timtay-microsoft:
Calling startTwin/startDeviceTwin with null callbacks will tear down any existing twin subscriptions. For MQTT, this was cause an UNSUB packet to be sent on the desired properties topic. For AMQP, this will tear down the twin worker links


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -522,21 +524,22 @@ void startTwinInternal(IotHubEventCallback twinStatusCallback, Object twinStatus
...
-        if (twinStatusCallback == null || genericPropertyCallBack == null)
+        if (this.twin != null && twinStatusCallback == null && genericPropertyCallBack == null)
         {
-            //Codes_SRS_INTERNALCLIENT_34_082: [If either callback is null, this function shall throw an IllegalArgumentException.]
-            throw new IllegalArgumentException("Callback cannot be null");
+            //end twin
+            twin.tearDownTwin();
         }
-        if (this.twin == null)
+        else if (this.twin == null)
```
jasmineymlo: if (this.twin == null) && (twinStatusCallback == null), we should do nothing?

timtay-microsoft: We should probably throw an illegalstateexception or something instead. If the call to a method called "startTwin" returns without error, customers may get confused. It should only return with success if it can build up twin or tear down twin

timtay-microsoft:
Need to follow up with C SDK about how to handle Unsub during reconnection scenarios. Need to document what Cancelled_on_close means if the reconnection fails, for instance

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -446,9 +446,9 @@ public void getDeviceTwin() throws IOException
     /**
      * Starts the device twin.
      *
-     * @param deviceTwinStatusCallback the IotHubEventCallback callback for providing the status of Device Twin operations. Cannot be {@code null}.
+     * @param deviceTwinStatusCallback the IotHubEventCallback callback for providing the status of Device Twin operations. If {@code null}, you will unsubscribe from twin.
      * @param deviceTwinStatusCallbackContext the context to be passed to the status callback. Can be {@code null}.
```
prmathur-microsoft: This is a behavior change.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceTwin/DeviceOperations.java
```diff
@@ -11,8 +11,7 @@
     DEVICE_OPERATION_TWIN_UPDATE_REPORTED_PROPERTIES_RESPONSE,
     DEVICE_OPERATION_TWIN_SUBSCRIBE_DESIRED_PROPERTIES_REQUEST,
     DEVICE_OPERATION_TWIN_SUBSCRIBE_DESIRED_PROPERTIES_RESPONSE,
-    DEVICE_OPERATION_TWIN_UNSUBSCRIBE_DESIRED_PROPERTIES_REQUEST,
-    DEVICE_OPERATION_TWIN_UNSUBSCRIBE_DESIRED_PROPERTIES_RESPONSE,
+    DEVICE_OPERATION_TWIN_TEAR_DOWN,
```
prmathur-microsoft: This enum does not follow iothub semantics for device twin

timtay-microsoft:
Closing this PR due to lack of need from customers

PR #524, opened 2019-05-21
feat(iot-dev): Device Client now periodically checks if messages in queue are expired regardless of connection state

timtay-microsoft:
Currently, the SDK will only check if messages have expired when the client is fully connected. This PR will periodically check for expired messages even when the client is not connected. This also enables messages to be expired if they are in progress, but no ack is received within the message expiry time


jasmineymlo:
>                 if (message != null && this.isMessageValid(packet))

any logic here which we could reuse?

---
Refers to: device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java:380 in 0aaf920. [](commit_id = 0aaf92099a03f4ff4b0db4933cc55572029153ff, deletion_comment = False)

timtay-microsoft:
I don't see any use for that logic in the function I wrote, no

PR #525, opened 2019-05-21
fix(iot-service, deps, e2e): Fix issue where ConnectionState was never propagated up through DeviceTwinDevice object

timtay-microsoft:
service was sending uppercase "Disconnected" and "Connected" when the SDK was expected lowercase "disconnected" and "connected"

The SDK was missing logic to propagate up this Json field to the DeviceTwinDevice object, too.

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinConnectionState.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinConnectionState.java
@@ -10,9 +10,9 @@ import com.google.gson.annotations.SerializedName;
  */
 public enum TwinConnectionState
 {
-    @SerializedName("disconnected")
+    @SerializedName(value = "Disconnected", alternate = "disconnected")
     DISCONNECTED,
 
-    @SerializedName("connected")
+    @SerializedName(value = "Connected", alternate = "connected")
     CONNECTED
 }
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinState.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinState.java
@@ -384,6 +384,20 @@ public class TwinState extends RegisterManager
     }
 
     /**
+     * Get the connection state
+     * @return the connection state
+     */
+    public String getConnectionState()
+    {
+        if (this.connectionState == null)
+        {
+            return null;
+        }
+
+        return this.connectionState.toString();
+    }
+
+    /**
      * Empty constructor
      *
      * <p>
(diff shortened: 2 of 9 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinState.java
```diff
@@ -383,6 +383,16 @@ public static TwinState createFromPropertiesJson(String json)
         return new TwinState(null, result.getDesired(), result.getReported());
     }
 
+    public String getConnectionState()
```
jasmineymlo: add comments for javadoc

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinConnectionState.java
```diff
@@ -10,9 +10,9 @@
  */
 public enum TwinConnectionState
 {
-    @SerializedName("disconnected")
+    @SerializedName("Disconnected")
```
jasmineymlo: could we make this case-insensitive?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
```diff
@@ -495,6 +495,19 @@ public DeviceCapabilities getCapabilities()
...
 
+    /**
+     * @return get the connection state as last reported by the service
+     */
+    public String getConnectionState()
+    {
+        return this.connectionState;
+    }
+
+    protected void setConnectionState(String connectionState)
```
jasmineymlo: add comments

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinConnectionState.java
timtay-microsoft: I can't, but I can make it catch "disconnected" and "Disconnected"

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinState.java
timtay-microsoft: sure thing

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
timtay-microsoft: sure thing

PR #526, opened 2019-05-21
fix(iot-device): Fix issue where provisioned devices don't receive C2D over MQTT

timtay-microsoft:
Device clients created using security providers led to some confusion between null and empty string within MQTT layer that caused the SDK to not subscribe to C2D. Fixing the SDK to set the gatewayhost name to null, rather than empty string. Also adding check within MQTT layer to prevent this bug if gateway hostname ends up as empty string again later.

Removing unnecessary delay in provisioning e2e tests, and fixing a timeout logic bug too

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubConnectionString.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubConnectionString.java
@@ -141,7 +141,7 @@ public class IotHubConnectionString
                                   String sharedAccessKey, String sharedAccessToken)
             throws IllegalArgumentException, URISyntaxException
     {
-        this(hostName, deviceId, sharedAccessKey, sharedAccessToken, "");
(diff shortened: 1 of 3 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubConnectionString.java
```diff
@@ -141,7 +141,7 @@ public IotHubConnectionString(String hostName, String deviceId,
                                   String sharedAccessKey, String sharedAccessToken)
             throws IllegalArgumentException, URISyntaxException
     {
-        this(hostName, deviceId, sharedAccessKey, sharedAccessToken, "");
+        this(hostName, deviceId, sharedAccessKey, sharedAccessToken, null);
```
timtay-microsoft: Within the mqtt layer, there is a check when subscribing to c2d messaging where we do not subscribe to c2d if connecting to an edgehub. That check was just if the gatewayhostname (argument 5 above) is null or not. Since it was empty, mqtt did not subscribe to c2d messaging.

PR #529, opened 2019-05-23
fix(iot-dev): Fix issue where device client created with security pro…

timtay-microsoft:
…vider could not receive c2d messages over mqtt

within mqtt layer, there is a check to subscribe to c2d only if the device is connecting to iot hub (rather than an edge device). That check is if the gateway hostname is null. Since it was being set to empty string within iothubconnectionstring's constructor, the check at the mqtt level believed it to be connecting to an edge device, and would not subscribe to c2d messages

Also fixed a small logical mistake when checking for timeouts in provisioning e2e device registration


timtay-microsoft:
Handling this fix in a different pr

PR #531, opened 2019-05-28
fix(iot-dev): Fix amqp layer not checking for local errors during err…

timtay-microsoft:
…or mapping

The SDK was only checking for remote errors previously

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -1216,6 +1216,58 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
         return this.deviceClientConfig.getIotHubHostname();
     }
 
+    private String getErrorCondition(ErrorCondition condition)
+    {
+        if (condition != null)
+        {
+            if (condition.getCondition() != null)
+            {
+                return condition.getCondition().toString();
+            }
+        }
+
+        return null;
+    }
+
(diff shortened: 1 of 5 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -1229,20 +1229,34 @@ private TransportException getTransportExceptionFromEvent(Event event)
...
-        String connectionErrorDescription = event.getConnection() != null && event.getConnection().getRemoteCondition() != null && event.getConnection().getRemoteCondition().getDescription() != null ? event.getConnection().getRemoteCondition().getDescription() : "";
-        String linkErrorDescription = event.getLink() != null && event.getLink().getRemoteCondition() != null && event.getLink().getRemoteCondition().getDescription() != null ? event.getLink().getRemoteCondition().getDescription() : "";
-        String transportErrorDescription = event.getTransport() != null && event.getTransport().getRemoteCondition() != null && event.getTransport().getRemoteCondition().getDescription() != null ? event.getTransport().getRemoteCondition().getDescription() : "";
-
+        String senderErrorRemote = event.getSender() != null && event.getSender().getRemoteCondition() != null && event.getSender().getRemoteCondition().getCondition() != null ? event.getSender().getRemoteCondition().getCondition().toString() : "";
```
davilu: Please create a function `<T> String nullToBlank(T value)`

timtay-microsoft: I opted to use the Endpoint class that proton defines to generalize searching through these sender/receiver/link/transport classes and their errors

PR #540, opened 2019-06-07
refactor(e2e): Fix SDK so it compiles under java 11

timtay-microsoft:
Still can't run tests due to junit compatibility issues, and javadoc generation seems buggy, too.

However, users can now compile the jar files they need from source calling

```cmd
mvn clean install -DskipTests=true -Dmaven.javadoc.skip=true
```

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpsConnection.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpsConnection.java
@@ -130,6 +130,7 @@ public class AmqpsConnection extends BaseHandler
 
     /**
      * Returns the status of the connection
+     * @throws Exception if any exception occurred during sasl negotiation
      * @return status of the connection
      */
     public boolean isConnected() throws Exception
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
@@ -287,7 +287,7 @@ public class InternalClient
      * be provided alongside the status and reason.
      *
      * @param callback The callback to be fired when the connection status of the device changes. Can be null to
-     *                 unset this listener as long as the provided callbackContext is also null.
+     *                 unset this listener as long as the maven-javadoc-pluginprovided callbackContext is also null.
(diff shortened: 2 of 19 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpsConnection.java
```diff
@@ -130,6 +130,7 @@ public void setListener(AmqpListener listener)
 
     /**
      * Returns the status of the connection
+     * @throws Exception if any exception occurred during sasl negotiation
      * @return status of the connection
      */
     public boolean isConnected() throws Exception
```
prmathur-microsoft: Why is this a generic Exception? Can you throw specific?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -287,7 +287,7 @@ public void sendReportedProperties(Set<Property> reportedProperties, int version
      * be provided alongside the status and reason.
      *
      * @param callback The callback to be fired when the connection status of the device changes. Can be null to
-     *                 unset this listener as long as the provided callbackContext is also null.
+     *                 unset this listener as long as the maven-javadoc-pluginprovided callbackContext is also null.
      * @param callbackContext a context to be passed to the callback. Can be {@code null}.
```
prmathur-microsoft: Typo?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpsConnection.java
jebrando: Will do.

---
In reply to: [316295049](https://github.com/Azure/azure-iot-sdk-java/pull/540#discussion_r316295049) [](ancestors = 316295049)

jebrando: saslListener.getSavedException throws a generic Exception.   So that's why we have to do the same.

---
In reply to: [316295049](https://github.com/Azure/azure-iot-sdk-java/pull/540#discussion_r316295049) [](ancestors = 316295049)

timtay-microsoft:
closing this PR for now. We will revisit full JDK 12 support soon instead

Issue #272, opened 2018-07-09
Provide a way to detect creation of a device that already exists (REST status result 409) in the Registry Manager


chrisgclayton:
- Windows 10
- Java 10.0.1 (build 10.0.1+10)
- JDK 10.0.1

# Description of the issue:
When creating a device using the RegistryManager if the device already exists a generic exception is raised. This exception does not have a good way to detect if the failure is a conflict in device name, short of looking to see if the message starts with a specific text string.

Please catch the REST API response code and if it matches 409 conflict return a derived exception or append an error code to reference

timtay-microsoft:
We are looking into providing this, thanks for filing the issue.

timtay-microsoft:
Thanks for the feature suggestion. As of [this release](https://github.com/Azure/azure-iot-sdk-java/releases/tag/2019-09-30_2), the service client now has a derived exception for the 409 case. As such, I'm closing this issue

az-iot-builder-01:
@chrisgclayton, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #548, opened 2019-06-24
feat(iot-service): Add mapping for 409 status code to IotHubConflictException

timtay-microsoft:
The service SDK had mappings for all other http status codes to IotHubXYZException, but was missing the mapping for 409 errors


timtay-microsoft:
This PR fixes issue #272

PR #553, opened 2019-07-09
Refactor Device Client Mqtt layer to get rid of unnecessary state

timtay-microsoft:
When Mqtt layer receives a message, it first puts that message into a map and alerts a higher level that a message has arrived so that it can pick that message out of the map. It doesn't make sense to do this when the lower level can just pass the message up, rather than putting into a map.

Not ready for review yet, still testing the impact of this refactor


timtay-microsoft:
I'm abandoning this PR for now, it doesn't accomplish much. I may revisit it later, though

PR #563, opened 2019-07-26
Proxy support for HTTP, with e2e fix

timtay-microsoft:
Proxy support for HTTP PR had an unnecessary file upload test that was causing some strain on the test infrastructure since only so many file uploads can be achieved at a time. There was never a need to run the file upload perf test both with proxy, and without, so I removed it


prmathur-microsoft:
Using admin rights as SDL has a known issue and is failing (devops issue WIP)

Issue #578, opened 2019-09-06
DeviceClient support for SSLSocketFactory 


MMaiero:
- **OS and version used:** Raspbian Buster

- **Java runtime used:** openjdk version "1.8.0_181"

- **SDK version used:** 0.22.0
   iot-device-client version 1.16.0

# Description of the issue:
Is there any plan to extend the DeviceClient API to support the creation of a new DeviceClient passing an SSLSocketFactory? In this way, instead of passing the public and private keys as string everything could be managed in a simpler way using java primitives.

# Code sample exhibiting the issue:
https://github.com/Azure/azure-iot-sdk-java/blob/master/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java#L207

# Console log of the issue:

# Need Support?
* Have a feature request for SDKs? Please post it on [User Voice](https://feedback.azure.com/forums/321918-azure-iot) to help us prioritize.
* Have a technical question? Ask on [Stack Overflow](https://stackoverflow.com/questions/tagged/azure-iot-hub) with tag “azure-iot-hub”
* Need Support? Every customer with an active Azure subscription has access to support with guaranteed response time.  Consider submitting a ticket and get assistance from Microsoft support team
* Found a bug? Please help us fix it by thoroughly documenting it and filing an issue on GitHub (C, Java, .NET, Node.js, Python).

timtay-microsoft:
We currently don't have any plans to add this feature. Is there something preventing you from passing in the certificates as strings?

MMaiero:
I see it as somewhat a security issue to have to keep a private key as string instead of having it in a keystore protected with a password. An passing that keystore as SSLSocketFactory to the Device Client.

timtay-microsoft:
Right now, an SSLSocketFactory instance is not enough information for our AMQP stack to work correctly. It requires the SDK to pass along an SSLContext instance instead. Would it still be worthwhile to you if the device client allowed you to pass in an SSLContext instance rather than an SSLSocketFactory instance?

MMaiero:
Hello, I think that would work too, in my case.

timtay-microsoft:
This feature was just released in iot-device-client library version 1.19.0. Now you will be able to pass in an SSLContext to your client's constructor. This constructor allows for either SAS based auth, or x509 based auth depending on if your connection string has a device key or not. If no device key is present, the client will assume you loaded the necessary x509 certificates into the provided SSLContext

[...]

PR #586, opened 2019-09-18
feat(iot-dev): Allow clients to provide SSLContext rather than certif…

timtay-microsoft:
…icate string/certificate paths

some customer's have security concerns with saving their credentials as strings, and other customers may want to use their own library for parsing certificates into SSLContexts, so we should expose this option to the user

issue #578

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
@@ -10,6 +10,7 @@ import com.microsoft.azure.sdk.iot.device.transport.amqps.IoTHubConnectionType;
 import com.microsoft.azure.sdk.iot.provisioning.security.SecurityProvider;
 import lombok.extern.slf4j.Slf4j;
 
+import javax.net.ssl.SSLContext;
 import java.io.Closeable;
 import java.io.IOError;
 import java.io.IOException;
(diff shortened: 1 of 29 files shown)


timtay-microsoft:
I need to reconsider how the user gives us the SSLContext. The SetOption approach isn't great because if the customer wants to do x509 authentication only, they have to use one of the x509 client constructors which require rsa certificates. It may make more sense to just add another x509 auth client constructor that takes an SSLContext and requires the connection string contains "x509=true". Not sure there is a user scenario where customer uses SAS based auth but still needs to be able to set the ssl context.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -215,6 +216,24 @@ public DeviceClient(String connString, IotHubClientProtocol protocol, String pub
         commonConstructorSetup();
     }
 
+    /**
+     * Creates a device client that uses the provided SSLContext for SSL negotiation
+     * @param connString the connection string for the device. May be an x509 connection string (format: "HostName=...;DeviceId=...;x509=true")
```
prmathur-microsoft: Why do we need this change? Adding a new set option complicates our testing. Why do we need our customers to be able setup this API?

timtay-microsoft: Originally, I was going to make this exposed via SetOption like we do with the other certificate settings, but now I just have a constructor that takes the SSLContext. The description of this PR was out of date when you read that it was going to be a new SetOption option.

This new constructor is important for our customers because it allows them to control certificate parsing. Things like ECC certs that we don't have logic for can be done by customer instead. Also, there are security concerns with asking our users to provide certificates in file or string form as mentioned in issue #578

PR #605, opened 2019-10-21

timtay-microsoft:

It also fixes a very small bug in the TransportClient class where it never set its own state to disconnected after calling close.

This PR also adds some basic state locks to the DeviceIO layer such that an open call and a close call cannot be executed simultaneously.

- Verify that a client object can be re-opened after closing it manually
- Verify that a client object can be re-opened from a main thread after a disconnection event exhausts the retry policy
- Verify that a client object can be re-opened from a thread spawned in the connection status callback thread after a disconnection event exhausts the retry policy

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
@@ -10,7 +10,7 @@ import com.microsoft.azure.sdk.iot.device.transport.IotHubTransport;
 import lombok.extern.slf4j.Slf4j;
 
 import java.io.IOException;
-import java.util.LinkedList;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -68,7 +68,7 @@ import java.util.concurrent.TimeUnit;
  * The task scheduler for sending and receiving messages for the Device Client
  */
 @Slf4j
-public final class DeviceIO
+public final class DeviceIO  implements TransportConnectionListener
 {
     /** The state of the IoT Hub client's connection with the IoT Hub. */
     protected enum IotHubClientState
@@ -79,16 +79,16 @@ public final class DeviceIO
     private long sendPeriodInMilliseconds;
     private long receivePeriodInMilliseconds;
 
-    private IotHubTransport transport;
-    private DeviceClientConfig config;
+    private final IotHubTransport transport;
     private IotHubSendTask sendTask = null;
     private IotHubReceiveTask receiveTask = null;
-    private IotHubClientProtocol protocol = null;
+    private final IotHubClientProtocol protocol;
 
     private ScheduledExecutorService taskScheduler;
     private IotHubClientState state;
+    private final Object stateLock = new Object();
 
-    private List<DeviceClientConfig> deviceClientConfigs = new LinkedList<>();
+    private List<DeviceClientConfig> deviceClientConfigs = new ArrayList<>();
 
     /**
      * Constructor that takes a connection string as an argument.
@@ -110,23 +110,17 @@ public final class DeviceIO
 
         /* Codes_SRS_DEVICE_IO_21_001: [The constructor shall store the provided protocol and config information.] */
         deviceClientConfigs.add(config);
-        this.config = config;
-        this.protocol = this.config.getProtocol();
-
-        /* Codes_SRS_DEVICE_IO_21_037: [The constructor shall initialize the `sendPeriodInMilliseconds` with default value of 10 milliseconds.] */
-        this.sendPeriodInMilliseconds = sendPeriodInMilliseconds;
-        /* Codes_SRS_DEVICE_IO_21_038: [The constructor shall initialize the `receivePeriodInMilliseconds` with default value of each protocol.] */
-        this.receivePeriodInMilliseconds = receivePeriodInMilliseconds;
+        this.protocol = config.getProtocol();
 
         /* Codes_SRS_DEVICE_IO_21_006: [The constructor shall set the `state` as `DISCONNECTED`.] */
         this.state = IotHubClientState.CLOSED;
 
         if (protocol == IotHubClientProtocol.AMQPS_WS || protocol == IotHubClientProtocol.MQTT_WS)
         {
-            this.config.setUseWebsocket(true);
+            config.setUseWebsocket(true);
         }
 
-        this.transport = new IotHubTransport(config);
+        this.transport = new IotHubTransport(config, this);
 
         /* Codes_SRS_DEVICE_IO_21_037: [The constructor shall initialize the `sendPeriodInMilliseconds` with default value of 10 milliseconds.] */
         this.sendPeriodInMilliseconds = sendPeriodInMilliseconds;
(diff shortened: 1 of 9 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -110,23 +110,17 @@
 
         /* Codes_SRS_DEVICE_IO_21_001: [The constructor shall store the provided protocol and config information.] */
         deviceClientConfigs.add(config);
-        this.config = config;
-        this.protocol = this.config.getProtocol();
-
-        /* Codes_SRS_DEVICE_IO_21_037: [The constructor shall initialize the `sendPeriodInMilliseconds` with default value of 10 milliseconds.] */
-        this.sendPeriodInMilliseconds = sendPeriodInMilliseconds;
-        /* Codes_SRS_DEVICE_IO_21_038: [The constructor shall initialize the `receivePeriodInMilliseconds` with default value of each protocol.] */
```
prmathur-microsoft: Why is this removed? Isn't it a breaking change?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -10,7 +10,7 @@
 import lombok.extern.slf4j.Slf4j;
 
 import java.io.IOException;
-import java.util.LinkedList;
+import java.util.ArrayList;
 import java.util.List;
```
prmathur-microsoft: Why array list?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -145,26 +139,37 @@
...
+                catch (DeviceClientException e)
+                {
+                    throw new IOException("Could not open the connection", e);
+                }
+            }
+
+            /* Codes_SRS_DEVICE_IO_21_014: [The open shall schedule receive tasks to run every receivePeriodInMilliseconds milliseconds.] */
+            /* Codes_SRS_DEVICE_IO_21_016: [The open shall set the `state` as `CONNECTED`.] */
+            commonOpenSetup();
+        }
```
prmathur-microsoft: is it ok to call commonOpenSetup with openTransport == false>?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -429,4 +445,32 @@ public void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeC
...
+    {
+        try
+        {
+            //only close the device io layer, not the transport layer. The transport layer is already closed when this method is called
+            this.close(false);
+        }
+        catch (IOException e)
+        {
+            log.warn("Encountered an exception while closing device IO layer due to transport disconnection event", e);
+        }
```
prmathur-microsoft: This is bad pattern to catch exception and only log. it is absorbing the exception. What action can SDK user perform in this situation?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/TransportConnectionListener.java
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+package com.microsoft.azure.sdk.iot.device;
+
+public interface TransportConnectionListener
+{
+    public void onTransportConnectionLost();
+
+    public void onTransportConnectionRecovered();
+}
```
prmathur-microsoft: 👍

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: This was duplicate code within this method. These values were set to the same value twice

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: David recommended this change. Arraylist performs better at get/set operations, and linkedlist performs better at add/remove operations. Since we are getting more often then adding, arraylist is the better choice

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: commonOpenSetup doesn't do anything to the transport layer, so it wouldn't be useful as a parameter regardless

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: This exception is only caught when opening/closing the transport layer. Since the transport layer isn't being opened/closed, this exception should never be caught

prmathur-microsoft: In that case would it be safe to rethrow after logging?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothubservices/errorinjection/SendMessagesErrInjTests.java
```diff
@@ -281,6 +281,68 @@ public void sendMessagesWithTcpConnectionDropNotifiesUserIfRetryExpires() throws
...
+    {
+        if (testInstance.protocol == HTTPS || testInstance.useHttpProxy)
+        {
+            return;
+        }
+
+        this.testInstance.setup();
+
+        this.testInstance.client.setRetryPolicy(new NoRetry());
+
```
prmathur-microsoft: Can you also add a test for retrying and recovering by explicitly setting lower retry timeout?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothubservices/errorinjection/SendMessagesErrInjTests.java
```diff
@@ -281,6 +281,68 @@ public void sendMessagesWithTcpConnectionDropNotifiesUserIfRetryExpires() throws
...
+    public void clientCanReopenAfterDisconnectionEvent() throws Exception
+    {
+        if (testInstance.protocol == HTTPS || testInstance.useHttpProxy)
+        {
+            return;
+        }
+
+        this.testInstance.setup();
+
+        this.testInstance.client.setRetryPolicy(new NoRetry());
```
abhipsaMisra: We would also need tests where the device client has a set retry policy; and the application reconnects and recovers on retry policy expiry.

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothubservices/errorinjection/SendMessagesErrInjTests.java
timtay-microsoft: So that the retry policy gets exhausted, and then application level retry is successful? Sure

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothubservices/errorinjection/SendMessagesErrInjTests.java
timtay-microsoft: I'll add that, and a similar test for when the device operation timeout causes the disconnection

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothubservices/errorinjection/SendMessagesErrInjTests.java
timtay-microsoft: Actually, this kind of test isn't possible to write right now. The fault injection framework we use does not allow us to cause a retryable error for X number of seconds. That functionality is necessary for us to test running through the sdk level retry and some application level retry. We only get DISCONNECTED_RETRYING followed by a successful reconnect within the SDK retry logic

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothubservices/errorinjection/SendMessagesErrInjTests.java
timtay-microsoft: Actually, this kind of test isn't possible to write right now. The fault injection framework we use does not allow us to cause a retryable error for X number of seconds. That functionality is necessary for us to test running through the sdk level retry and some application level retry. We only get DISCONNECTED_RETRYING followed by a successful reconnect within the SDK retry logic

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: Upon reworking the DeviceIO layer, this is no longer an issue. No exceptions are being swallowed like this anymore.

timtay-microsoft:
Closing this PR as I had to redesign the DeviceIO layer to fix an issue in this PR where the DeviceIO layer leaked threads. The correct fix for this bug can be seen in #608

Issue #117, opened 2017-07-13
Unable to make a single AMQPS connection to the IOTHub for multiple devices 


bhuvanalakshmi:
- **OS and version used:** 
Mac El Captain 

- **Java runtime used:** 
Java 1.8.0_60

- **SDK version used:**
iot-device-client - 1.3.29

# Description of the issue:
 Unable to make a single connection to the IOTHub for multiple devices for AMQP. Looks like this is supported in the C SDK but not in the java 

# Code sample exhibiting the issue:
String connectionStr = "----";// IotHub connection string
DeviceClient client = new DeviceClient(connectionStr,  IotHubClientProtocol.AMQPS);

# Console log of the issue:
Exception in thread "main" java.lang.IllegalArgumentException: Device ID cannot be null.
      at com.microsoft.azure.sdk.iot.device.IotHubConnectionString.validateTerms(IotHubConnectionString.java:220)
      at com.microsoft.azure.sdk.iot.device.IotHubConnectionString.<init>(IotHubConnectionString.java:98)
      at com.microsoft.azure.sdk.iot.device.DeviceClient.<init>(DeviceClient.java:147)

prmathur-microsoft:
Can you share how are you trying to connect to Multiple devices ? Are you interested in multiple devices over single connection or are you ok with multiple devices over multiple connections ?

bhuvanalakshmi:
I am interested in multiple devices over single connection - just like this sample supported in the C SDK. 
https://github.com/Azure/azure-iot-sdk-c/blob/1ad2c5355a3d9759493eda9b41647c34b845ec20/iothub_client/samples/iothub_client_sample_amqp_shared/iothub_client_sample_amqp_shared.c

I am using the DeviceClient with IOTHub Connection string for AMQPS protocol. 

 String connectionStr = "----";// IotHub connection string
DeviceClient client = new DeviceClient(connectionStr, IotHubClientProtocol.AMQPS);

Is there an other way to do this ?

prmathur-microsoft:
Currently java SDK does not support multiplexing over AMQP. We have this feature in our backlog and will be adding it in future.

bhuvanalakshmi:
@prmathur-microsoft  - What does it take to connect to IOTHUB using amqps without device id ?  I see there is documentation available on how to use MQTT without SDKs (https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support). But, i don't see documentation for AMQP.

[...]

Issue #113, opened 2017-07-04
Device Twin updates cannot be done at a rapid rate


gbidsilva:
**Env. Details:**
SDK Used : **Azure Device SDK**
```
<dependency>
	<groupId>com.microsoft.azure.sdk.iot</groupId>
	<artifactId>iot-device-client</artifactId>
	<version>1.3.29</version>
</dependency>
```
		
On
**Java 1.8 + Cent OS (Linux)**

HI Team,

In my application, I am trying to update/ push device twin data.
I was able to get it working properly but, I have encountered some issues.
In this application I have a method called 'updateDeviceTwinInAzure' and inside that method I am updating the device twin using the azure device sdk.

Following is my code,

```
package com.gbids.azuretwin;

import com.microsoft.azure.sdk.iot.device.DeviceClient;
import com.microsoft.azure.sdk.iot.device.DeviceTwin.Device;
import com.microsoft.azure.sdk.iot.device.DeviceTwin.Property;
import com.microsoft.azure.sdk.iot.device.IotHubClientProtocol;
import com.microsoft.azure.sdk.iot.device.IotHubEventCallback;
import com.microsoft.azure.sdk.iot.device.IotHubStatusCode;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public class AzureTwinUpdate {
    private static final Logger logger = LoggerFactory.getLogger(AzureTwinUpdate.class);

    public void updateDeviceTwinInAzure(final int propertyValue) throws Exception {
        logger.info("Attempting to update device twin...");

        Device homeKit = new Device() {
            @Override
            public void PropertyCall(String propertyKey, Object propertyValue, Object context) {
                System.out.println(propertyKey + " changed to " + propertyValue);
            }
        };

        // Connection String is something like below,
        // final String deviceConnectionString = "HostName=b123.azure-devices.net;DeviceId=" + deviceId + ";SharedAccessKey=***********************";
        DeviceClient client = new DeviceClient(deviceConnectionString, IotHubClientProtocol.MQTT);

        client.open();
        client.startDeviceTwin(new DeviceTwinStatusCallBack(), null, homeKit, null);

        homeKit.setReportedProp(new Property("testProperty", propertyValue));

        client.sendReportedProperties(homeKit.getReportedProp());
        homeKit.clean();
        client.closeNow();
    }

    protected static class DeviceTwinStatusCallBack implements IotHubEventCallback {
[...]

PR #608, opened 2019-10-23

timtay-microsoft:
(I closed the old PR for this issue since it was getting cluttered, and its code was different enough to warrant a new PR)

It also fixes a very small bug in the TransportClient class where it never set its own state to disconnected after calling close. An E2E test has been added for using a closed and reopened transportclient

This PR removes state tracking from the DeviceIO layer, instead deferring to the IotHubTransport layer for isOpen() type checks.

- Verify that a client object can be re-opened after closing it manually
- Verify that a client object can be re-opened from a main thread after a disconnection event exhausts the retry policy
- Verify that a client object can be re-opened from a thread spawned in the connection status callback thread after a disconnection event exhausts the retry policy

### DIFF
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/DeviceTestManager.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/DeviceTestManager.java
@@ -8,8 +8,10 @@ package com.microsoft.azure.sdk.iot.common.helpers;
 import com.microsoft.azure.sdk.iot.device.*;
 import com.microsoft.azure.sdk.iot.device.exceptions.ModuleClientException;
 
+import javax.net.ssl.SSLContext;
 import java.io.IOException;
 import java.net.URISyntaxException;
+import java.security.GeneralSecurityException;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentMap;
 
@@ -75,7 +77,7 @@ public class DeviceTestManager
         deviceEmulator.tearDown();
     }
 
-    public void restartDevice(String connectionString, IotHubClientProtocol protocol, String publicCert, String privateKey) throws InterruptedException, IOException, URISyntaxException, ModuleClientException
+    public void restartDevice(String connectionString, IotHubClientProtocol protocol, String publicCert, String privateKey) throws InterruptedException, IOException, URISyntaxException, ModuleClientException, GeneralSecurityException
     {
         deviceEmulator.tearDown();
 
@@ -87,7 +89,8 @@ public class DeviceTestManager
             }
             else
             {
-                this.client = new DeviceClient(connectionString, protocol, publicCert, false, privateKey, false);
+                SSLContext sslContext = SSLContextBuilder.buildSSLContext(publicCert, privateKey);
+                this.client = new DeviceClient(connectionString, protocol, sslContext);
             }
         }
         else if (this.client instanceof ModuleClient)
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/SSLContextBuilder.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/SSLContextBuilder.java
@@ -1,0 +1,240 @@
+/*
+ *  Copyright (c) Microsoft. All rights reserved.
+ *  Licensed under the MIT license. See LICENSE file in the project root for full license information.
+ */
+
+package com.microsoft.azure.sdk.iot.common.helpers;
+
+import com.microsoft.azure.sdk.iot.deps.auth.IotHubCertificateManager;
+import org.apache.commons.codec.binary.Base64;
+
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManagerFactory;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.security.*;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.security.interfaces.RSAPrivateKey;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.UUID;
+
+public class SSLContextBuilder
+{
+    private static final String SSL_CONTEXT_INSTANCE = "TLSv1.2";
+
+    private static final String CERTIFICATE_ALIAS = "cert-alias";
+    private static final String PRIVATE_KEY_ALIAS = "key-alias";
+
+    /**
+     * Create an SSLContext instance with the provided public and private keys that also trusts the base iot hub certificates
+     */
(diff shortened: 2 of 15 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -108,33 +106,22 @@
...
+        if (this.protocol == IotHubClientProtocol.AMQPS_WS || this.protocol == IotHubClientProtocol.MQTT_WS)
         {
-            this.config.setUseWebsocket(true);
+            config.setUseWebsocket(true);
         }
 
-        this.transport = new IotHubTransport(config);
+        /* Codes_SRS_DEVICE_IO_21_001: [The constructor shall store the provided protocol and config information.] */
+        this.deviceClientConfigs.add(config);
+
```
prmathur-microsoft: Would Line #117 override with defaults and #113 would never be set?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -215,26 +167,14 @@ private void commonOpenSetup()
      */
     public void close() throws IOException
     {
-        /* Codes_SRS_DEVICE_IO_21_017: [The close shall finish all ongoing tasks.] */
-        /* Codes_SRS_DEVICE_IO_21_018: [The close shall cancel all recurring tasks.] */
-        if (taskScheduler != null)
-        {
-            this.taskScheduler.shutdown();
-        }
```
prmathur-microsoft: Where do you shutdown the taskScheduler ?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -269,14 +209,6 @@ public synchronized void sendEventAsync(Message message,
                                Object callbackContext,
                                String deviceId)
     {
-        /* Codes_SRS_DEVICE_IO_21_024: [If the client is closed, the sendEventAsync shall throw an IllegalStateException.] */
-        if (this.state == IotHubClientState.CLOSED)
-        {
-            throw new IllegalStateException(
```
prmathur-microsoft: What happens when transport is not opened and sendEventAsync is triggered?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -429,4 +361,43 @@ public void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeC
...
+
+        if (this.receiveTask == null)
+        {
+            this.receiveTask = new IotHubReceiveTask(this.transport);
+        }
+
+        if (this.taskScheduler == null)
+        {
+            this.taskScheduler = Executors.newScheduledThreadPool(2);
+
```
prmathur-microsoft: Can this be static? Why would you want to spawn new threads everytime transport is opened. Eg - What happens during Retry.

prmathur-microsoft: Executors.newScheduledThreadPool(2); PSF please

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -705,10 +722,6 @@ private void openConnection() throws TransportException
 
         //Codes_SRS_IOTHUBTRANSPORT_34_039: [This function shall open the iotHubTransportConnection object with the saved list of configs.]
         this.iotHubTransportConnection.open(this.deviceClientConfigs, scheduledExecutorService);
-
-        //Codes_SRS_IOTHUBTRANSPORT_34_040: [This function shall invoke the method updateStatus with status CONNECTED,
-        // reason CONNECTION_OK, and a null throwable.]
-        this.updateStatus(IotHubConnectionStatus.CONNECTED, IotHubConnectionStatusChangeReason.CONNECTION_OK, null);
```
prmathur-microsoft: Will this never be triggered now? Are we changing the state diagram for application?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: Sure, I'll make the 2 psf

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
timtay-microsoft: The status update for CONNECTED will now come only from the onConnectionEstablished callback within the same class. Previously, we had this line here only because the Http layer would never call onConnectionEstablished. However, as a result, there would be two calls to updateStatus with CONNECTED around the same time for MQTT and AMQP users. 

I simplified this by making the http layer just call onConnectionEstablished like MQTT and AMQP do.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: When the connection is DISCONNECTED in the transport layer, it calls up to the onTransportConnectionLost() method in DeviceIO which stops sends the sender and receiver threads. This is necessary for the DeviceIO layer to be "closed" when the Transport layer is closed.

Once the client is fully closed, we expect customers to call open again, which will start up the send/receive threads again.

I can make the thread pool static, but the open()  must spawn the send/receive threads, and close() must shut them down.

As for DISCONNECTED_RETRYING happening in the transport layer, nothing happens to the DeviceIO layer as a result. The sender and receiver threads are left running.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: The message gets to the IotHubTransport layer where the same IllegalStateException is thrown. The net result is the same to a customer, but it allows us to just have one place where we check the state before queuing a message

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: When the method onTransportConnectionClosed() is called. That method is called either when a customer calls closeNow() manually, or when the Transport layer loses connection and cannot recover. That way the scheduler shuts down no matter how the SDK gets to a DISCONNECTED state.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: I'm not sure I follow what you mean here. The config passed in will always set the appropriate websocket settings, and then it will save that config to our list

---
In reply to: [338271300](https://github.com/Azure/azure-iot-sdk-java/pull/608#discussion_r338271300) [](ancestors = 338271300)

PR #615, opened 2019-10-30

timtay-microsoft:
(I closed the old PR for this issue since it was getting cluttered, and its code was different enough to warrant a new PR)

It also fixes a very small bug in the TransportClient class where it never set its own state to disconnected after calling close. An E2E test has been added for using a closed and reopened transportclient

This PR removes state tracking from the DeviceIO layer, instead deferring to the IotHubTransport layer for isOpen() type checks.

Verify that a client object can be re-opened after closing it manually
Verify that a client object can be re-opened from a main thread after a disconnection event exhausts the retry policy
Verify that a client object can be re-opened from a thread spawned in the connection status callback thread after a disconnection event exhausts the retry policy


timtay-microsoft:
Still working on making sure that there is no regression around messages that should expire during reconnection, don't merge this PR yet

timtay-microsoft:
closing due to lack of demand

PR #619, opened 2019-11-05
doc(iot-serv): Add documentation to fileuploadNotificationReceiver about QoS

timtay-microsoft:
All receive operations have a QoS of 1, but that wasn't documented before


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotificationReceiver.java
```diff
@@ -75,6 +75,8 @@ public void close() throws IOException
     /**
      * Receive FileUploadNotification with default timeout
```
bikamani: It would be good to document default timeout as well.

PR #635, opened 2019-12-02
fix(vsts): Fix issue with TSA upload task in Azure DevOps

timtay-microsoft:
TSA Upload task changed underneath us, and now we need to update from "Production" to "PROD"


timtay-microsoft:
Closing this PR as it will be a part of a bigger PR

PR #656, opened 2019-12-20
Service client v2

timtay-microsoft:
Deprecating the old feedback receiver and file upload notification receiver APIs since neither of them allowed you to Complete/Abandon/Reject individual messages/notifications.

Added new APIs that take a callback so that the user is notified each time a file upload notification/feedback message is received. This callback also allows the user to Complete/Abandon/Reject that message.

Added some additional error logging to deps/service client/device client AMQP connections

Refactored the file upload e2e tests such that they make use of the Complete/Abandon/Reject feature so that the test runner knows to reject file upload notifications that don't belong to the test process. That way, multiple file upload test runners can run against a single hub at the same time since they won't Complete each other's file upload notifications


timtay-microsoft:
This PR got a little big, so I'm closing it in lieu of two smaller PRs

PR #662, opened 2020-01-06
Refactor Device client and Deps amqp layers to add better error logging

timtay-microsoft:
Created an Error Logging Base Handler class that all amqp connections will eventually extend from. It's purpose is to log whenever a link/session/connection is closed remotely (generally considered an error case) or when a transport error occurs.

The deps amqp layer is used only by provisioning device client currently, but this PR makes the device client use it, too. IoT Service client will use it soon when the v2 changes come.

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/ErrorLoggingBaseHandler.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/ErrorLoggingBaseHandler.java
@@ -1,0 +1,79 @@
+/*
+ *  Copyright (c) Microsoft. All rights reserved.
+ *  Licensed under the MIT license. See LICENSE file in the project root for full license information.
+ */
+
+package com.microsoft.azure.sdk.iot.deps.transport.amqp;
+
+import lombok.extern.slf4j.Slf4j;
+import org.apache.qpid.proton.engine.BaseHandler;
+import org.apache.qpid.proton.engine.Event;
+
+@Slf4j
+public class ErrorLoggingBaseHandler extends BaseHandler
+{
+    protected ProtonJExceptionParser protonJExceptionParser;
+
+    @Override
+    public void onLinkRemoteClose(Event event)
+    {
+        protonJExceptionParser = new ProtonJExceptionParser(event);
+        if (protonJExceptionParser.getError() == null)
+        {
+            log.debug("Amqp link {} was closed remotely", event.getLink().getName());
+        }
+        else
+        {
+            if (event.getLink() != null && event.getLink().getName() != null)
+            {
+                log.warn("Amqp link {} was closed remotely with exception {} with description {}", event.getLink().getName(), protonJExceptionParser.getError(), protonJExceptionParser.getErrorDescription());
+            }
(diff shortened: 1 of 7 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/ProtonJExceptionParser.java
```diff
@@ -0,0 +1,58 @@
...
+    private ErrorCondition getErrorConditionFromEndpoint(Endpoint endpoint)
+    {
+        return endpoint.getCondition() != null && endpoint.getCondition().getCondition() != null ? endpoint.getCondition() : endpoint.getRemoteCondition();
+    }
+
+    private void getTransportExceptionFromProtonEndpoints(Endpoint... endpoints)
+    {
+        for (Endpoint endpoint : endpoints)
+        {
+            if (endpoint == null)
```
abhipsaMisra: Why is this check required? It would be better to check that `endpoints` is not null.

timtay-microsoft: Good point, this check isn't doing anything so I will remove it

timtay-microsoft: This check is required because that collection of endpoints does contain some null endpoints

timtay-microsoft: The collection of endpoints is never null, though

abhipsaMisra: The `Endpoint` spread is over the inputs we get from the constructor; makes sense.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/ErrorLoggingBaseHandler.java
```diff
@@ -0,0 +1,79 @@
...
+                log.warn("Unknown amqp link was closed remotely with exception {} with description {}", protonJExceptionParser.getError(), protonJExceptionParser.getErrorDescription());
+            }
+        }
+    }
+
+    @Override
+    public void onSessionRemoteClose(Event event)
+    {
+        protonJExceptionParser = new ProtonJExceptionParser(event);
+        if (protonJExceptionParser.getError() == null)
```
abhipsaMisra: Do we allow multiple sessions over a single connection? If yes, then it'll be useful to log the session identifier here as well (similar to link).

timtay-microsoft: We do open multiple sessions for multiplexed connections, I believe. However, the proton-j session object doesn't have any information in it here that we could use to correlate back to which session it was. Sessions being closed remotely are a bit of a rare case, from my understanding, so we can hold off on adding that kind of mapping for now

PR #670, opened 2020-01-08
Refactor build gate to cut down on unnecessary testing

timtay-microsoft:
This PR allows the gate to be smart enough to not run e2e tests if the only code changes were .md files

It also allows the gate to be smart enough to only run provisioning tests if only provisioning code was changed.

Changes to IoT Hub code (device client, service client, or deps) will still require all e2e tests to be run.

Some simple assumptions were made in these scripts:
- Any file change that contains "provisioning" in its path constitutes a change in the provisioning SDK
- Any file change that contains "iothub", "deps", "iot-device-client" or "iot-service-client" in its path constitutes a change in the iothub SDK
- Any file change that falls under common test helpers (i.e. proxy logic, android test group declarations) will require all e2e tests to be run because they may be used in any e2e test

All e2e test files were moved in order to fit these assumptions. The majority of the files changed in this PR are tied to this move, and are all contained in one commit. The other two commits are more worth reviewing


iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IntegrationTest.java
```diff
@@ -37,4 +40,27 @@ protected void finished(Description description)
...
+
+    public abstract boolean isProvisioningTest();
+    public abstract boolean isIotHubTest();
+
+    @Before
+    public void checkIfTestShouldBeRun()
+    {
+        if (isIotHubTest())
+        {
+            //Will skip the test if it is an iot hub test when told to skip iot hub tests
```
abhipsaMisra: Nit: this wording is a bit confusing, can be reworded into something along the lines of:
If it is an iot hub test, will skip the test when told to skip iot hub tests

timtay-microsoft: sure thing, I'll use your re-wording here and below

PR #671, opened 2020-01-08
refactor(e2e): Fix e2e test swallowing an exception, remove needless …

timtay-microsoft:
…confirmOpenStabilized calls

confirmOpenStabilized used to be used when open calls were less reliable, but now they are reliable so we can get rid of them

### DIFF
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IotHubServicesCommon.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IotHubServicesCommon.java
@@ -55,10 +55,6 @@ public class IotHubServicesCommon
         try
         {
             client.open();
-            if (statusUpdates != null)
-            {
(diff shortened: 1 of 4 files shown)


iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/IotHubServicesCommon.java
```diff
@@ -55,10 +55,6 @@ public static void sendMessages(InternalClient client,
         try
         {
             client.open();
-            if (statusUpdates != null)
-            {
-                confirmOpenStabilized(statusUpdates, 120000, client);
```
bikamani: I see confirmOpenStabilized being referenced in waitForStabilizedConnection, is that required?

timtay-microsoft: Yes, that is used during our error injection tests so that the test waits for the error injection period to end

PR #672, opened 2020-01-09
Service client v2

timtay-microsoft:
Deprecating the old feedback receiver and file upload notification receiver APIs since neither of them allowed you to Complete/Abandon/Reject individual messages/notifications.

Added new APIs that take a callback so that the user is notified each time a file upload notification/feedback message is received. This callback also allows the user to Complete/Abandon/Reject that message.

Added some additional error logging to service client AMQP connections

Refactored the file upload e2e tests such that they make use of the Complete/Abandon/Reject feature so that the test runner knows to reject file upload notifications that don't belong to the test process. That way, multiple file upload test runners can run against a single hub at the same time since they won't Complete each other's file upload notifications

Also added API to configure service client sas token lifespan


timtay-microsoft:
Need to double check @deprecated annotations to make sure documentation is correct

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackMessageListener.java
```diff
@@ -0,0 +1,52 @@
...
+    private final String sasToken;
+    private AmqpFeedbackListenerHandler amqpReceiveHandler;
+    private IotHubServiceClientProtocol iotHubServiceClientProtocol;
+
+    public FeedbackMessageListener(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol, FeedbackBatchMessageCallback feedbackBatchMessageCallback)
+    {
+        Tools.throwIfNullOrEmpty(hostName, "hostName cannot be null or empty");
+        Tools.throwIfNullOrEmpty(userName, "userName cannot be null or empty");
+        Tools.throwIfNullOrEmpty(sasToken, "sasToken cannot be null or empty");
+        Tools.throwIfNull(iotHubServiceClientProtocol, "iotHubServiceClientProtocol cannot be null");
```
drajput: For Azure services related exceptions, define and use custom exceptions with application specific error codes rather than hand-coding these error messages at each call. This will standardize the error codes across all libraries and provide a better debugging experience.
ex: (error_code.INVALID_IOTHUB_HOSTNAME", "IoTHub hostname cannot be null or empty")

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotificationListener.java
```diff
@@ -0,0 +1,59 @@
...
+     * @param userName The username string to use SASL authentication (example: user@sas.service)
+     * @param sasToken The SAS token string
+     * @param iotHubServiceClientProtocol protocol to use
+     */
+    public FileUploadNotificationListener(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol, FileUploadNotificationCallback fileUploadNotificationCallback)
+    {
+        Tools.throwIfNullOrEmpty(hostName, "hostName cannot be null or empty");
+        Tools.throwIfNullOrEmpty(userName, "userName cannot be null or empty");
+        Tools.throwIfNullOrEmpty(sasToken, "sasToken cannot be null or empty");
+        Tools.throwIfNull(iotHubServiceClientProtocol, "iotHubServiceClientProtocol cannot be null");
```
drajput: Same as above. These should just use error codes.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -85,11 +92,9 @@ protected ServiceClient(IotHubConnectionString iotHubConnectionString, IotHubSer
      */
     public void open() throws IOException
     {
-        // Codes_SRS_SERVICE_SDK_JAVA_SERVICECLIENT_12_008: [The function shall throw IOException if the member AMQP sender object has not been initialized]
-        if (this.amqpMessageSender == null)
-        {
-            throw new IOException("AMQP sender is not initialized");
-        }
+        this.sasToken = new IotHubServiceSasToken(iotHubConnectionString, sasTokenExpiryTime).toString();
```
drajput: "tostring()" (lowercase) in "IotHubServiceSasToken.java" is implemented to return the instance member "token". This is misleading and conflicting with Java's "Object" class's  "toString() (lowerCamelCase) which returns the string representation of the Object itself. 
Suggest to rename this to "getSaSToken()" and provide an overridden toString() method to return custom string representation if/where needed.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
```diff
@@ -0,0 +1,345 @@
...
+                    delivery.settle();
+                }
+                else if (outcome == DeliveryOutcome.Reject)
+                {
+                    //purposefully do nothing. Service will treat unack'd messages as rejected
+                    log.trace("Amqp message was rejected, will not acknowledge it");
+                }
+                else
+                {
+                    log.error("Unknown DeliveryOutcome provided, message will not be acknowledged");
```
drajput: Switch-Case on the enum for better readability?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
```diff
@@ -0,0 +1,345 @@
...
+    {
+        if (this.savedException != null)
+        {
+            throw new IOException("Connection failed to be established", this.savedException);
+        }
+
+        if (this.protonJExceptionParser != null && this.protonJExceptionParser.getError() != null)
+        {
+            throw new IOException("Encountered exception during amqp connection: " + protonJExceptionParser.getError() + " with description " + protonJExceptionParser.getErrorDescription());
+        }
```
drajput: Same as above - use custom exceptions and error codes.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
drajput: please ignore the part of the comment on lowercase tostring - not sure where I saw that.

drajput: This actually looks good. Please ignore my previous comments :(

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
timtay-microsoft: Sure thing

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -65,52 +67,54 @@ protected ServiceClient(IotHubConnectionString iotHubConnectionString, IotHubSer
...
-        this.amqpMessageSender = new AmqpSend(hostName, userName, sasToken, this.iotHubServiceClientProtocol);
+    /**
+     * Set how long newly generated sas tokens will live for, in seconds. By default, tokens live for 1 year.
+     * A sas token is generated each time a connection is opened. This only applies for connections for sending cloud
+     * to device messages. The file upload notification receiver and feedback message receiver have their own sas token
+     * expiry time that can be set. See {@link com.microsoft.azure.sdk.iot.service.FeedbackMessageListener#setSasTokenExpiryTime(long)} and
+     * {@link com.microsoft.azure.sdk.iot.service.FeedbackMessageListener#setSasTokenExpiryTime(long)}
+     * @param sasTokenExpiryTime the number of seconds newly generated sas tokens will live for.
+     */
+    public void setSasTokenExpiryTime(long sasTokenExpiryTime)
```
abhipsaMisra: Does this override the token expiry time calculated by the SDK, if relying on SAS key to generate the auth token?

abhipsaMisra: Are file upload notification receiver and feedback message receiver a part of ServiceClient or are they initialized separately? If they are a part of ServiceClient but have their own token expiry time, this API might get confusing.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -65,52 +67,54 @@ protected ServiceClient(IotHubConnectionString iotHubConnectionString, IotHubSer
...
+        this.sasTokenExpiryTime = sasTokenExpiryTime;
     }
 
     /**
-     * Open AMQP sender
-     * @throws IOException This exception is thrown if the AmqpSender object is not initialized
+     * Does nothing anymore
+     * @throws IOException This method never throws this exception anymore
+     * @deprecated This method no longer does anything. File upload notification listener and feedback message listener
+     * must be opened through their respective open() APIs, and the cloud to device message sender in this client opens
```
abhipsaMisra: Why not have implicit open for all of them?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -134,20 +139,16 @@ public void send(String deviceId, Message message) throws IOException, IotHubExc
...
     public void send(String deviceId, String moduleId, Message message) throws IOException, IotHubException
     {
-        // Codes_SRS_SERVICE_SDK_JAVA_SERVICECLIENT_28_001: [The function shall throw IOException if the member AMQP sender object has not been initialized]
-        if (this.amqpMessageSender == null)
-        {
-            throw new IOException("AMQP sender is not initialized");
-        }
-        // Codes_SRS_SERVICE_SDK_JAVA_SERVICECLIENT_28_002: [The function shall call send() on the member AMQP sender object with the given parameters]
-        this.amqpMessageSender.send(deviceId, moduleId, message);
+        AmqpSend amqpMessageSender = new AmqpSend(hostName, userName, iotHubConnectionString, this.iotHubServiceClientProtocol, this.sasTokenExpiryTime);
```
abhipsaMisra: We are initializing all AMQP resources for every call; what benefit do we get from this?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -134,20 +139,16 @@ public void send(String deviceId, Message message) throws IOException, IotHubExc
...
+        amqpMessageSender.send(deviceId, moduleId, message);
```
abhipsaMisra: Is `close()` a part of `AmqpSend`?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -134,20 +139,16 @@ public void send(String deviceId, Message message) throws IOException, IotHubExc
...
     }
 
     /**
-     * Provide asynchronous access to open()
-     *
+     * Provide asynchronous access to open(), which does nothing anymore
+     * @deprecated This method no longer does anything. See {@link #open()} documentation for more details
```
abhipsaMisra: q - why did we have an async open() to begin with?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -213,39 +218,66 @@ public void send(String deviceId, String moduleId, Message message) throws IOExc
...
     public FileUploadNotificationReceiver getFileUploadNotificationReceiver()
     {
-        return new FileUploadNotificationReceiver(hostName, userName, sasToken, iotHubServiceClientProtocol);
+        return new FileUploadNotificationReceiver(hostName, userName, makeSasToken(), iotHubServiceClientProtocol);
+    }
+
+    /**
+     * Create an instance of a file upload notification listener. This instance allows you to open a connection that will receive
+     * file upload notifications over time. Whenever a file upload notification is received, the provided callback will be executed,
+     * allowing you to complete/abandon/reject each notification. If called multiple times, multiple listener instances will be created.
```
abhipsaMisra: Is there any limit on how many instances can be created? We should also add a note explaining how callbacks are handled when we have multiple listeners attached to a single hub instance.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -213,39 +218,66 @@ public void send(String deviceId, String moduleId, Message message) throws IOExc
...
+     */
+    public FileUploadNotificationListener createFileUploadNotificationListener(FileUploadNotificationCallback fileUploadNotificationCallback)
+    {
+        return new FileUploadNotificationListener(hostName, userName, iotHubConnectionString, iotHubServiceClientProtocol, fileUploadNotificationCallback);
+    }
+
+    /**
+     * Create an instance of a feedback batch message listener. This instance allows you to open a connection that will receive
+     * feedback batch messages over time. Whenever a feedback batch message is received, the provided callback will be executed,
+     * allowing you to complete/abandon/reject each feedback message. If called multiple times, multiple listener instances will be created.
```
abhipsaMisra: Same note as above

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
timtay-microsoft: There is no limit, so I'll add a note about that, sure

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
timtay-microsoft: All of the service client operations had an async flavor, but I don't know why.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
timtay-microsoft: Sure thing

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
timtay-microsoft: It is an API, but it does nothing anymore. Now, the amqpSend class does the open/send/close all under the send() API. I had to do this because the old Open() API didn't declare any thrown exceptions. It didn't declare any thrown exceptions because it actually didn't do any IO operations weirdly

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
timtay-microsoft: Let's split the feedback receiver and file upload notification receiver out into separate clients to avoid this confusion

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackMessageListenerClient.java
```diff
@@ -0,0 +1,108 @@
...
+        log.info("Feedback message listener connection opened");
+    }
+
+    /**
+     * Close the AMQP connection. No feedback messages will be received on this listener until it is opened again.
+     * Closed instances may be re-opened and re-closed any number of times
+     */
+    public synchronized void close()
+    {
+        amqpReceiveHandler.close();
```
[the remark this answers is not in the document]
timtay-microsoft: ```amqpReceiveHandler``` is instantiated in the constructor, and never set to null. I don't see how it could ever be null, so I'd rather not add a check here

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackMessageListenerClient.java
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Open the AMQP connection, and start listening for feedback messages. Any feedback messages that arrive
+     * will be given to the user through the FileUploadNotificationCallback provided in the constructor
+     *
+     * @throws IOException If opening the connection fails
+     * @throws InterruptedException If the thread gets interrupted while waiting for the connection to open
+     */
+    public synchronized void open() throws IOException, InterruptedException
+    {
+        amqpReceiveHandler.open(new IotHubServiceSasToken(iotHubConnectionString, sasTokenExpiryTime).toString());
```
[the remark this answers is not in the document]
timtay-microsoft: It is initialized in the constructor, so if you are calling open on the client, then ```amqpReceiveHandler``` must already have been initialized

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
```diff
@@ -0,0 +1,464 @@
...
+
+    private final IotHubServiceClientProtocol iotHubServiceClientProtocol;
+    private final String webSocketHostName;
+
+    protected Exception savedException;
+
+    protected Connection connection;
+
+    /**
+     * Callback that is executed when the service sent a message to this client
```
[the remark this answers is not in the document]
timtay-microsoft: Good catch

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpFeedbackReceivedHandler.java
```diff
@@ -30,6 +30,7 @@
  * Maintains the layers of AMQP protocol (Link, Session, Connection, Transport)
  * Creates and sets SASL authentication for transport
  */
+@Deprecated
```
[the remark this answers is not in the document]
timtay-microsoft: Sure thing

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpFileUploadNotificationReceive.java
```diff
@@ -22,6 +22,7 @@
  * overriding the events what are needed to handle
  * high level open, close methods and feedback received event.
  */
+@Deprecated
```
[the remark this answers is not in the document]
timtay-microsoft: Sure thing

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpFileUploadNotificationReceivedHandler.java
```diff
@@ -31,6 +31,7 @@
  * Maintains the layers of AMQP protocol (Link, Session, Connection, Transport)
  * Creates and sets SASL authentication for transport
  */
+@Deprecated
```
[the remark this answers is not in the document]
timtay-microsoft: Sure thing

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpReceive.java
```diff
@@ -20,6 +20,7 @@
  * overriding the events what are needed to handle
  * high level open, close methods and feedback received event.
  */
+@Deprecated
```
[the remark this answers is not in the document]
timtay-microsoft: Sure thing

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotificationListener.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.qpid.proton.Proton;
+import org.apache.qpid.proton.engine.BaseHandler;
+import org.apache.qpid.proton.engine.Event;
+import org.apache.qpid.proton.reactor.Reactor;
+
+import java.io.IOException;
+
+public class FileUploadNotificationListener
+{
+    FileUploadNotificationCallback fileUploadNotificationCallback;
```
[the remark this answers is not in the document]
timtay-microsoft: It isn't needed at this level, it can just be saved in the Amqp handler level instead of the Client level. I'll change this

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackMessageListener.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.qpid.proton.engine.Event;
+import org.apache.qpid.proton.reactor.Reactor;
+
+import java.io.IOException;
+
+/**
+ * FeedbackReceiver is a specialized receiver whose ReceiveAsync
+ * method returns a FeedbackBatch instead of a Message.
+ */
+public class FeedbackMessageListener
```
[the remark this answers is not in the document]
timtay-microsoft: The callback provided by the user is handed down into the Amqp handler class (AmqpFeedbackListenerHandler) where it is executed each time a feedback message batch arrives

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackBatchMessageCallback.java
```diff
@@ -0,0 +1,17 @@
...
+
+public interface FeedbackBatchMessageCallback
+{
+    /**
+     * Called each time a file upload notification was received. The listener must return a delivery outcome so that
+     * the amqp connection can acknowledge the message accordingly
+     * @param feedbackBatch The received feedback batch that was received
+     * @return Whether to Complete the message, Abandon it, or Reject it. See {@link DeliveryOutcome} for details
+     */
+    public DeliveryOutcome onFeedbackMessageReceived(FeedbackBatch feedbackBatch);
```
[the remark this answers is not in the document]
timtay-microsoft: Sadly, no. I can add a note about this, but the batch messages come in the form of a single AMQP message, which can only be Accepted/Abandoned/Rejected as a whole

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
timtay-microsoft: I'll redo the C2D flow into a new client after this PR. As we've discussed offline, this is the current expected behavior of the C2D flow, even though it is inefficient

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
```diff
@@ -0,0 +1,487 @@
...
+
+    @Override
+    public void onLinkInit(Event event)
+    {
+        Link link = event.getLink();
+        if (event.getLink().getName().equals(tag))
+        {
+            Target t = new Target();
+            t.setAddress(endpoint);
+            Source source = new Source();
```
abhipsaMisra: For a c2d link, we don't seem to be setting `Source` currently.
Also, `Target` and `Source` both have the same endpoint; what's the difference?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
```diff
@@ -0,0 +1,49 @@
...
+
+        try
+        {
+            log.trace("Starting reactor thread");
+            this.reactor.setTimeout(REACTOR_TIMEOUT);
+            this.reactor.run();
+        }
+        catch (HandlerException e)
+        {
+            log.error("Encountered an exception while running reactor on thread {}", threadName, e);
```
abhipsaMisra: How is this handled (in `OpenAsync()`), since this is running on a different thread?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
```diff
@@ -0,0 +1,487 @@
...
+    {
+        log.debug("Session opened remotely");
+    }
+
+    @Override
+    public void onConnectionRemoteClose(Event event)
+    {
+        super.onConnectionRemoteClose(event);
+        event.getTransport().close_tail();
+        releaseOpenLatch();
```
abhipsaMisra: do we need to clean up reactor reference as well?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
```diff
@@ -0,0 +1,487 @@
...
+    }
+
+    /**
+     * If an exception was encountered while opening the AMQP connection, this function shall throw that saved exception
+     * @throws IOException if an exception was encountered while opening the AMQP connection. The encountered
+     * exception will be the inner exception
+     * @throws IotHubException if an Iot Hub level exception occurs during transport
+     */
+    protected void validateConnectionWasSuccessful() throws IOException, IotHubException
```
abhipsaMisra: This seems to be called from the `SendHandler` only, shouldn't this be called from other clients as well? before performing the operation? Do we do anything with the `savedException`?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
```diff
@@ -0,0 +1,487 @@
...
+    {
+        log.trace("Reactor initialized");
+        event.getReactor().connection(this);
+    }
+
+    @Override
+    public void onReactorFinal(Event event)
+    {
+        log.trace("Reactor finalized, freeing reactor resources");
+        this.reactor.free();
```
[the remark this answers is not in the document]
timtay-microsoft: It should only be called once, so I'll just have the reactor runner call it. Good catch

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    public void run()
+    {
+        Thread.currentThread().setName(this.threadName);
+
+        try
+        {
+            log.trace("Starting reactor thread");
+            this.reactor.setTimeout(REACTOR_TIMEOUT);
+            this.reactor.run();
```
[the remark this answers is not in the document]
timtay-microsoft: The ```this.reactor.run();``` while keep the reactor alive until someone calls ```reactor.stop()```

In this code, ```reactor.stop()``` is called after we close the connection, in the AmqpConnectionHandler method ```onConnectionLocalClose()```

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/telemetry/ReceiveMessagesTests.java
```diff
@@ -10,12 +10,13 @@
 import com.microsoft.azure.sdk.iot.device.DeviceClient;
 import com.microsoft.azure.sdk.iot.device.IotHubClientProtocol;
 import com.microsoft.azure.sdk.iot.device.ModuleClient;
-import com.microsoft.azure.sdk.iot.service.Module;
+import com.microsoft.azure.sdk.iot.service.*;
 import com.microsoft.azure.sdk.iot.service.auth.AuthenticationType;
```
drwill-ms: if line 13 is *, is this explicit import no longer needed?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackMessageListenerClient.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        this.sasTokenExpiryTime = DEFAULT_SAS_TOKEN_EXPIRY_TIME;
+        this.iotHubServiceClientProtocol = iotHubServiceClientProtocol;
+
+        this.hostName = this.iotHubConnectionString.getHostName();
+        this.userName = this.iotHubConnectionString.getUserString();
+
+        Tools.throwIfNullOrEmpty(this.hostName, "The provided connection string is missing a hostName");
+        Tools.throwIfNullOrEmpty(this.hostName, "The provided connection string is missing a userName");
```
drwill-ms: this.userName

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotificationListenerClient.java
```diff
@@ -0,0 +1,104 @@
...
+     * only one will receive a given file upload notification at a time, though. Each instance can abandon any received file upload notifications
+     * if the notification should be received by a different instance.
+     *
+     * @param connectionString The connection string for the IotHub
+     * @param iotHubServiceClientProtocol protocol to use
+     * @param fileUploadNotificationCallback the callback to be executed whenever a file upload notification is received
+     */
+    public FileUploadNotificationListenerClient(String connectionString, IotHubServiceClientProtocol iotHubServiceClientProtocol, FileUploadNotificationCallback fileUploadNotificationCallback)
+    {
+        Tools.throwIfNullOrEmpty(connectionString, "iotHubConnectionString cannot be null");
```
drwill-ms: "connectionString cannot be null"

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotificationListenerClient.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        this.sasTokenExpiryTime = DEFAULT_SAS_TOKEN_EXPIRY_TIME;
+        this.iotHubServiceClientProtocol = iotHubServiceClientProtocol;
+
+        this.hostName = this.iotHubConnectionString.getHostName();
+        this.userName = this.iotHubConnectionString.getUserString();
+
+        Tools.throwIfNullOrEmpty(this.hostName, "The provided connection string is missing a hostName");
+        Tools.throwIfNullOrEmpty(this.hostName, "The provided connection string is missing a userName");
```
drwill-ms: this.userName

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
timtay-microsoft: This should be called by all amqp connections at the end of the open() call, I'll refactor this. Good catch

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotificationListenerClient.java
timtay-microsoft: sure thing

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotificationListenerClient.java
timtay-microsoft: sure thing

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackMessageListenerClient.java
timtay-microsoft: sure thing

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/telemetry/ReceiveMessagesTests.java
timtay-microsoft: Disregard these changes. There was no reason for me to change this file after all, so I've reverted my changes here

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
timtay-microsoft: For all cases where transport error occurs, or if connection/session/link is closed remotely, we respond by starting to close the local links/sessions/connections. At the end of that process, the reactor gets stopped. After the reactor stops, the reactor runner thread will free the reactor's resources. I can talk you through this bit, because it is a bit tricky to follow

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
timtay-microsoft: This catch of HandlerExceptions is only triggered if a an exception is thrown from the reactor thread. Generally, we try not to throw exceptions from that thread, since we don't control it fully. Instead, we catch any possible exception and save it so that the main thread can throw it instead.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
timtay-microsoft: Good question. After digging into this, I found that sender links only need to set the target, and receiver links only need to set the source. I can rewrite this logic to exist only in the AmqpConnectionHandler so that AmqpSendHandler doesn't override this logic

timtay-microsoft:
Closing this PR because it turns out we will likely make turn service client AMQP endpoints into HTTP endpoints, so there isn't much sense in sprucing up this AMQP logic. I'll still bring in logging and some small bug fixes to the existing AMQP stack, but I won't be adding new APIs or deprecating any APIs

PR #678, opened 2020-01-15
refactor(vsts): Allow smart gate to ignore sample code changes

timtay-microsoft:
Changes to sample code don't necessitate running e2e tests

### DIFF
--- a/vsts/determine_tests_to_run.ps1
+++ b/vsts/determine_tests_to_run.ps1
@@ -26,6 +26,10 @@ ForEach ($line in $($GitDiff -split "`r`n"))
 	{
         # These file types are ignored when determining if source code changes require running e2e tests
 	}
+	elseif ($line.toLower().Contains("sample"))
+	{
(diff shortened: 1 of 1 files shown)


vsts/determine_tests_to_run.ps1
```diff
@@ -26,6 +26,10 @@ ForEach ($line in $($GitDiff -split "`r`n"))
 	{
         # These file types are ignored when determining if source code changes require running e2e tests
 	}
+	elseif ($line.toLower().Contains("sample"))
+	{
+	    # Sample changes don't necessitate running e2e tests
```
abhipsaMisra: What about changes to scripts, yaml files, versions.csv etc

timtay-microsoft: Those changes wouldn't be in a folder that contains "iothub", "provisioning", etc so the script already disregards them

abhipsaMisra: cool!

Issue #274, opened 2018-07-15
com.microsoft.azure.sdk.iot.device.exceptions.ModuleClientException: java.net.MalformedURLException: unknown protocol: unix


mkohns:
Hi, I have strange behavior using the sdk with iot edge.

IOT Edge:
OS: ubuntu 18.04
iotedge version: 1.0.0 (52ef77db24126bf473265fc09c53d35290a2dd6b)
moby version:  1.0.0-1
docker version: 18.06.0-dev, build daf021fe
Installations Instructions: https://docs.microsoft.com/de-de/azure/iot-edge/quickstart-linux

azure-iot-sdk-java:
```    
<dependency>
      <groupId>com.microsoft.azure.sdk.iot</groupId>
      <artifactId>iot-device-client</artifactId>
      <version>1.13.1</version>
</dependency>
```

In my code I initialize my test iot edge module with

`ModuleClient module = ModuleClient.createFromEnvironment(IotHubClientProtocol.MQTT);`

and get the Exception

```
com.microsoft.azure.sdk.iot.device.exceptions.ModuleClientException: java.net.MalformedURLException: unknown protocol: unix
        at com.microsoft.azure.sdk.iot.device.ModuleClient.createFromEnvironment(ModuleClient.java:251)
        at com.microsoft.docs.iothub.samples.SimulatedDevice.main(SimulatedDevice.java:172)
Caused by: java.net.MalformedURLException: unknown protocol: unix
        at java.net.URL.<init>(URL.java:600)
        at java.net.URL.<init>(URL.java:490)
        at java.net.URL.<init>(URL.java:439)
        at com.microsoft.azure.sdk.iot.device.hsm.HttpsHsmClient.sign(HttpsHsmClient.java:71)
        at com.microsoft.azure.sdk.iot.device.hsm.HttpHsmSignatureProvider.sign(HttpHsmSignatureProvider.java:84)
        at com.microsoft.azure.sdk.iot.device.hsm.IotHubSasTokenHsmAuthenticationProvider.createNewSasToken(IotHubSasTokenHsmAuthenticationProvider.java:80)
        at com.microsoft.azure.sdk.iot.device.hsm.IotHubSasTokenHsmAuthenticationProvider.create(IotHubSasTokenHsmAuthenticationProvider.java:48)
        at com.microsoft.azure.sdk.iot.device.ModuleClient.createFromEnvironment(ModuleClient.java:234)
        ... 1 more
```

[...]

PR #679, opened 2020-01-16
fix(iot-dev): Fix module client so it doesn't need to set urlStreamFactoryHandler

timtay-microsoft:
setting the factory handler sometimes failed if this library was used alongside other libraries that set the same handler because that handler can only be set once per jvm session.

Most notably, customers trying to use this library alongside springboot would see failures when creating modules from environment as in #274 

Thankfully, we never really needed to set the urlStreamFactoryHandler for our createFromEnvironment flow. It was just one of a few options, so I have altered the code in question avoid it.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/hsm/HttpsHsmClient.java
```diff
@@ -42,48 +44,15 @@
      */
     public HttpsHsmClient(String baseUrl) throws URISyntaxException
```
bikamani: So with new changes, description of this method should also change?

timtay-microsoft: which part of the javadoc header is out of date here?

PR #683, opened 2020-01-18
Refactor according to coding style, compare json objects in E2E

abhipsaMisra:
Expanding all imports, since it increases the readability of the code.
Also modified the E2E tests to always validate against Json Node, instead of relying on String matches.


digital-twin/e2e-tests/digital-twin-e2e-tests-common/src/main/java/com/microsoft/azure/sdk/iot/digitaltwin/e2e/simulator/TestInterfaceInstance2.java
```diff
@@ -83,10 +86,16 @@ public void onRegistered() {
...
     protected void onPropertyUpdate(@NonNull DigitalTwinPropertyUpdate digitalTwinPropertyUpdate) {
         log.trace("Property received: name={}, desired value={}, version={}", digitalTwinPropertyUpdate.getPropertyName(), digitalTwinPropertyUpdate.getPropertyDesired(), digitalTwinPropertyUpdate.getDesiredVersion());
-        propertyUpdatesReceived.put(digitalTwinPropertyUpdate.getPropertyName(), digitalTwinPropertyUpdate.getPropertyDesired());
+        JsonNode desiredProperty = null;
+        try {
+            desiredProperty = objectMapper.readTree(digitalTwinPropertyUpdate.getPropertyDesired());
+        } catch (IOException e) {
+            log.error("Exception thrown while converting desired property update to Json Node: ", e);
+        }
+        propertyUpdatesReceived.put(digitalTwinPropertyUpdate.getPropertyName(), desiredProperty);
```
davilu: move it up into `try` block

abhipsaMisra: Why would we need to wrap a `put` operation within a `try-catch` block?

PR #702, opened 2020-02-06
refactor(e2e): Adjust timing in token renewal tests

timtay-microsoft:
test sometimes fails with case that mqtt connection never lost connection to hub. This implied that the test didn't wait long enough before checking that the grace period had ended. Adding more buffer time to ensure that the grace period ends before we check that mqtt connection was lost

### DIFF
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/TokenRenewalTests.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/TokenRenewalTests.java
@@ -75,9 +75,9 @@ public class TokenRenewalTests extends IotHubIntegrationTest
     @Test
     public void tokenRenewalWorks() throws Exception
     {
-        final long SECONDS_FOR_SAS_TOKEN_TO_LIVE_BEFORE_RENEWAL = 60;
-        final long EXPIRED_SAS_TOKEN_GRACE_PERIOD_SECONDS = 600;
-        final long EXTRA_BUFFER_TO_ENSURE_TOKEN_EXPIRED_SECONDS = 30;
+        final long SECONDS_FOR_SAS_TOKEN_TO_LIVE_BEFORE_RENEWAL = 30;
+        final long EXPIRED_SAS_TOKEN_GRACE_PERIOD_SECONDS = 600; //service extends 10 minute grace period after a token has expired
+        final long EXTRA_BUFFER_TO_ENSURE_TOKEN_EXPIRED_SECONDS = 120; //wait time beyond the expected grace period, just in case
 
         List<InternalClient> clients = createClientsToTest();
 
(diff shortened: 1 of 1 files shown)


iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/TokenRenewalTests.java
```diff
@@ -75,9 +75,9 @@ public static void stopProxy()
...
     public void tokenRenewalWorks() throws Exception
     {
-        final long SECONDS_FOR_SAS_TOKEN_TO_LIVE_BEFORE_RENEWAL = 60;
-        final long EXPIRED_SAS_TOKEN_GRACE_PERIOD_SECONDS = 600;
-        final long EXTRA_BUFFER_TO_ENSURE_TOKEN_EXPIRED_SECONDS = 30;
+        final long SECONDS_FOR_SAS_TOKEN_TO_LIVE_BEFORE_RENEWAL = 30;
+        final long EXPIRED_SAS_TOKEN_GRACE_PERIOD_SECONDS = 600; //service extends 10 minute grace period after a token has expired
+        final long EXTRA_BUFFER_TO_ENSURE_TOKEN_EXPIRED_SECONDS = 120; //wait time beyond the expected grace period, just in case
 
         List<InternalClient> clients = createClientsToTest();
```
bikamani: Is the comment below up to date? Is it 5 minute or 10 minutes?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/TokenRenewalTests.java
```diff
@@ -75,9 +75,9 @@ public static void stopProxy()
     @Test
     public void tokenRenewalWorks() throws Exception
     {
-        final long SECONDS_FOR_SAS_TOKEN_TO_LIVE_BEFORE_RENEWAL = 60;
-        final long EXPIRED_SAS_TOKEN_GRACE_PERIOD_SECONDS = 600;
-        final long EXTRA_BUFFER_TO_ENSURE_TOKEN_EXPIRED_SECONDS = 30;
+        final long SECONDS_FOR_SAS_TOKEN_TO_LIVE_BEFORE_RENEWAL = 30;
+        final long EXPIRED_SAS_TOKEN_GRACE_PERIOD_SECONDS = 600; //service extends 10 minute grace period after a token has expired
+        final long EXTRA_BUFFER_TO_ENSURE_TOKEN_EXPIRED_SECONDS = 120; //wait time beyond the expected grace period, just in case
```
bikamani: so we need 2 extra minutes for the test to succeed?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/TokenRenewalTests.java
timtay-microsoft: The below comment is wrong, I'll update it. It should be 10

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/TokenRenewalTests.java
timtay-microsoft: Yes, this test needs to open a connection, let the first token expire, and then let the service grace period expire. Only after all of that will the test check things like that the MQTT connection was lost.

Issue #699, opened 2020-01-30
TwinCollection Max Depth Issue


bouff:
- **OS and version used:** <VERSION> 
macOS 10.15.2
- **Java runtime used:** <VERSION> 
JDK 1.8
- **SDK version used:** <VERSION> 
iot-service-client version 1.19.0

# Description of the issue:

As per the IoT Hub documentation, the depth limitation for desired properties, reported properties, and tag is 10.

https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-device-twins#tags-and-properties-format

The value is currently hard-coded at 6.

https://github.com/Azure/azure-iot-sdk-java/blob/master/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinCollection.java

This is essentially a reincarnation of the following issue:
https://github.com/Azure/azure-iot-sdk-java/issues/338

Would it be possible to either update the value (or remove that validation, as discussed in the previous issue)?

# Code sample exhibiting the issue:

# Console log of the issue:

# Need Support?
* Have a feature request for SDKs? Please post it on [User Voice](https://feedback.azure.com/forums/321918-azure-iot) to help us prioritize.
* Have a technical question? Ask on [Stack Overflow](https://stackoverflow.com/questions/tagged/azure-iot-hub) with tag “azure-iot-hub”
* Need Support? Every customer with an active Azure subscription has access to support with guaranteed response time.  Consider submitting a ticket and get assistance from Microsoft support team
* Found a bug? Please help us fix it by thoroughly documenting it and filing an issue on GitHub (C, Java, .NET, Node.js, Python).

timtay-microsoft:
We should fix this bug, yes. If you are interested in providing a PR, I'd be happy to review it and merge it in for you

morphvale:
Hi, is there an expected timeline for a new release which will include this fix?
We're also encountering this problem in production with an application which just stops execution and hangs on client connection to azure when we have reported properties with a depth of 7.

timtay-microsoft:
I'm trying to get a new version of the SDK released with this fix by the end of this week. I'll update this thread once it is released

timtay-microsoft:
As of [this release](https://github.com/Azure/azure-iot-sdk-java/releases/tag/2020-02-12), all SDK twin depth validation is removed, so I'm closing this issue

az-iot-builder-01:
[...]

PR #703, opened 2020-02-06
fix(deps, prov-service): Remove SDK-side validation of max twin depth

timtay-microsoft:
Twin depth limits may be arbitrarily changed by services, so there is no reason to validate them on the SDK side. It happened once already when the Iot Hub's max depth went from 5 to 10, and the SDK became inconsistent.

Fixes issue #699

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ParserUtility.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ParserUtility.java
@@ -151,43 +151,6 @@ public class ParserUtility
     }
 
     /**
-     * Helper to validate if the provided string is a valid json key.
-     *
-     * @param key is the string to be validated.
-     * @param isMetadata defines if the key belongs to a metadata, which allows character `$`.
-     * @throws IllegalArgumentException if the string do not fit the criteria.
-     */
-    public static void validateKey(String key, boolean isMetadata) throws IllegalArgumentException
-    {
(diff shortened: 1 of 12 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ParserUtility.java
```diff
@@ -234,17 +226,7 @@ private static void validateMapInternal(Map<String, Object> map, int level, int
...
-                    validateMapInternal((Map<String, Object>) value, level, maxLevel, allowMetadata);
-                }
-                else
-                {
-                    if (!mapOnlyContainsMetaData((Map<String, Object>)value))
-                    {
-                        throw new IllegalArgumentException("Map exceed maximum of " + maxLevel + " levels");
-                    }
-                }
+                validateMapInternal((Map<String, Object>) value, allowMetadata);
```
abhipsaMisra: is the allow metadata still being used for any validation?

abhipsaMisra: there are other validations still being done; valid characters, size etc. This might be a good time to scrub through those and remove them.

timtay-microsoft: There are still a few validations on things like that which we could get rid of, sure. I'll take those out too.

PR #713, opened 2020-02-24
refactor(deps, iot-service): Refactor amqp service client implementation

timtay-microsoft:
-Add substantial logging to service client operations over amqp including happy path and error path
-Abstract some common amqp logic from each of file upload notification receiver, feedback message receiver, and cloud to device sender


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
```diff
@@ -0,0 +1,185 @@
...
+
+        if (!this.sessionOpenedRemotely)
+        {
+            throw new IOException("Amqp Session failed to open");
+        }
+
+        if (!this.linkOpenedRemotely)
+        {
+            throw new IOException("Amqp link never opened");
+        }
```
abhipsaMisra: This doesn't really give any context about why the connection/session/link wasn't opened. I am not sure if simply throwing an IOException here without an inner exception is of any value here. I would expect that if connection/session/links were not opened successfully, then there would be a savedException saved. 
These lines should be simplified accordingly.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpFeedbackReceivedHandler.java
```diff
@@ -126,10 +89,9 @@ public void onDelivery(Event event)
             // Codes_SRS_SERVICE_SDK_JAVA_AMQPFEEDBACKRECEIVEDHANDLER_12_007: [The event handler shall settle the Delivery with the Accepted outcome]
             delivery.disposition(Accepted.getInstance());
             delivery.settle();
-          
-            // Codes_SRS_SERVICE_SDK_JAVA_AMQPFEEDBACKRECEIVEDHANDLER_12_008: [The event handler shall close the Session and Connection (Proton)]
-            recv.getSession().close();
-            recv.getSession().getConnection().close();
+
+            log.debug("Closing amqp feedback receiver link since a feedback message was received");
+            recv.close();
```
abhipsaMisra: Are we relying on link closure closing the session and associate connection? If yes, we should add appropriate comment here.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public void run(long timeoutMs)
+    {
+        Thread.currentThread().setName(this.threadName);
+
+        try
+        {
+            log.trace("Starting reactor thread {}", this.threadName);
+            this.reactor.setTimeout(REACTOR_TIMEOUT);
```
abhipsaMisra: what happens if `timeoutMs` > `REACTOR_TIMEOUT`

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpReceive.java
```diff
@@ -95,26 +75,14 @@ public synchronized FeedbackBatch receive(long timeoutMs) throws IOException, In
...
-            
-            // Codes_SRS_SERVICE_SDK_JAVA_AMQPRECEIVE_12_008: [The function shall stop and free the Proton reactor object]
-            this.reactor.stop();
-            this.reactor.process();
-            this.reactor.free();   
-            this.amqpReceiveHandler.receiveComplete();
+            new ReactorRunner(this.amqpReceiveHandler, "AmqpFeedbackReceiver").run(timeoutMs);
+
+            log.trace("Feedback receiver reactor finished running, verifying that the connection opened correctly");
+            this.amqpReceiveHandler.verifyConnectionOpened();
```
abhipsaMisra: At this point, the reactor has stopped running and the reactor thread has been freed. Are the links/session/connection still open?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpFileUploadNotificationReceive.java
```diff
@@ -167,14 +129,10 @@ public synchronized void onFeedbackReceived(String feedbackJson)
 
             fileUploadNotificationQueue.add(fileUploadNotification);
         }
-        catch (IOException e)
```
abhipsaMisra: Why are we removing this?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpFileUploadNotificationReceivedHandler.java
```diff
@@ -227,39 +149,9 @@ public void onLinkInit(Event event)
         Link link = event.getLink();
         if (event.getLink().getName().equals(FILE_NOTIFICATION_RECEIVE_TAG))
         {
-
-            Target t = new Target();
```
abhipsaMisra: Why are we removing this?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSendHandler.java
```diff
@@ -292,21 +181,22 @@ public void onLinkInit(Event event)
     @Override
     public void onLinkFlow(Event event)
     {
-        if (!messagesToBeSent.isEmpty())
+        if (messageToBeSent != null)
         {
-            org.apache.qpid.proton.message.Message protonMessage = messagesToBeSent.remove();
```
abhipsaMisra: Was the length of `messagesToBeSent` always one, or did we support some sort of queuing mechanism?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSendHandler.java
```diff
@@ -331,75 +221,61 @@ public void onLinkFlow(Event event)
                 snd.send(msgData, 0, length);
 
                 snd.advance();
+
+                this.messageToBeSent = null;
```
abhipsaMisra: Are multiple messages sent over a single connection? Isn't there a 1:1 connection:operation mapping as in others?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public void run()
+    {
+        Thread.currentThread().setName(this.threadName);
+
+        try
+        {
+            log.trace("Starting reactor thread {}", this.threadName);
+            this.reactor.setTimeout(REACTOR_TIMEOUT);
+            this.reactor.run();
```
abhipsaMisra: How long does this run for?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
```diff
@@ -0,0 +1,88 @@
...
+        try
+        {
+            log.trace("Starting reactor thread {}", this.threadName);
+            this.reactor.setTimeout(REACTOR_TIMEOUT);
+
+            long startTime = System.currentTimeMillis();
+            long endTime = startTime + timeoutMs;
+
+            this.reactor.start();
+            while (this.reactor.process())
```
abhipsaMisra: So the channel is open for operations until `timeoutMs`?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSend.java
```diff
@@ -121,17 +105,19 @@ public void send(String deviceId, String moduleId, Message message) throws IOExc
...
+                    log.info("Sending cloud to device module message");
                 }
-                // Codes_SRS_SERVICE_SDK_JAVA_AMQPSEND_28_002: [The function shall initialize the Proton reactor object]
-                this.reactor = Proton.reactor(this);
-                // Codes_SRS_SERVICE_SDK_JAVA_AMQPSEND_28_003: [The function shall start the Proton reactor object]
-                this.reactor.run();
-                this.reactor.free();
+
+                new ReactorRunner(amqpSendHandler, "AmqpSend").run();
+
```
abhipsaMisra: No verifyConnectionWasOpened here?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/ErrorLoggingBaseHandler.java
```diff
@@ -69,11 +79,65 @@ public void onTransportError(Event event)
...
+        }
+
+        log.debug("Closing amqp connection locally since amqp transport error was thrown");
+        event.getConnection().close();
+    }
+
+    @Override
+    public void onLinkRemoteOpen(Event event)
+    {
+        if (event.getLink() != null)
```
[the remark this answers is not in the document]
timtay-microsoft: It should never be null, no. I'll remove this check

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/ErrorLoggingBaseHandler.java
```diff
@@ -69,11 +79,65 @@ public void onTransportError(Event event)
...
+    public void onSessionRemoteOpen(Event event)
+    {
+        log.debug("Amqp Session opened remotely");
+    }
+
+    @Override
+    public void onLinkLocalClose(Event event)
+    {
+        //Reactor should respond to this session close with onSessionLocalClose call, where we close the connection.
+        // Cannot close the connection here because it can only be done when all of its sessions have closed
```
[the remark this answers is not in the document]
timtay-microsoft: Not for provisioning or service client. I'll factor this logic out so that it isn't used by device client though

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnectionTest.java
```diff
@@ -2226,17 +2226,17 @@ public void OnLinkRemoteCloseReportsErrorCodeIfPresent() throws TransportExcepti
                 mockEvent.getReceiver();
                 result = null;
                 mockEvent.getTransport();
-                result = mockTransport;
+                result = null;
```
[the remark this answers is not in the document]
timtay-microsoft: This test is for checking the link for an error condition. If the transport is not null, it will try to check the transport for an error condition instead.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotification.java
```diff
@@ -27,7 +27,7 @@
     private Long blobSizeInBytes;
     private Date enqueuedTimeUtcDate;
 
-    public FileUploadNotification(String deviceId, String blobUri, String blobName, Date lastUpdatedTimeDate, Long blobSizeInBytes, Date enqueuedTimeUtcDate) throws IOException
+    public FileUploadNotification(String deviceId, String blobUri, String blobName, Date lastUpdatedTimeDate, Long blobSizeInBytes, Date enqueuedTimeUtcDate)
```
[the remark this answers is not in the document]
timtay-microsoft: I'll revert this, I just realized that this is considered a public API

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/ErrorLoggingBaseHandler.java
```diff
@@ -8,6 +8,7 @@
 import lombok.extern.slf4j.Slf4j;
 import org.apache.qpid.proton.engine.BaseHandler;
 import org.apache.qpid.proton.engine.Event;
+import org.apache.qpid.proton.engine.Link;
 
 @Slf4j
 public class ErrorLoggingBaseHandler extends BaseHandler
```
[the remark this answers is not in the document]
timtay-microsoft: I think I'll just take this new logic I'm adding and make it into a new class that extends ErrorLoggingBaseHandler. I want the device client amqp layer to use the existing error logging base handler, but I want provisioning device client and iot service client to use this new logic since they both only ever have a single link in a single session in a single connection

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
timtay-microsoft: The problem here is that there isn't necessarily an exception thrown by opening the connection. If we simply time out while opening the connection, proton-j may not give us any exception to include here. If protonJExceptionParser and savedException are both null, then we have no extra context to give here. All we know is that the connection/session/link did not open remotely

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpFeedbackReceivedHandler.java
timtay-microsoft: Sure thing

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
timtay-microsoft: I don't see this REACTOR_TIMEOUT doing anything. I'll go ahead and remove it and remove the reactor.setTimeout(...) calls

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpReceive.java
timtay-microsoft: In this implementation, yes, the links/session/connection were never closed. I've fixed this in the next iteration though. From the reactor runner, we need to signal to close the links first before we stop the reactor.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpFileUploadNotificationReceive.java
timtay-microsoft: I've refactored this a bit for the next iteration. Originally, no IOException was ever being thrown in that try block even though one was declared. In the next iteration, this will just catch any exception instead of an IOException in particular.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpFileUploadNotificationReceivedHandler.java
timtay-microsoft: A link only needs a target if it is a sender link. Receiver links only need a source. It didn't hurt to give a sender link a source or a receiver link a target, but it would be confusing to keep that code

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSendHandler.java
timtay-microsoft: It was always a queue of at most 1 value, weirdly. Since the sender uses one connection per message sent, there was no reason to have a queue here

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSendHandler.java
timtay-microsoft: There is a 1:1 mapping of amqp connection to c2d message here, yes. By setting the messageToBeSent to null, I am signalling that the message doesn't need to be sent anymore if the ```onLinkFlow()``` event happens again for whatever reason

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
timtay-microsoft: It runs until the message is sent, or an exception occurs while opening the connection

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
timtay-microsoft: Yes, this is for use cases like when the customer want's to receive file upload notifications for X number of seconds

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSend.java
timtay-microsoft: It does call ```sendComplete()``` which calls verifyConnectionWasOpened, while also checking the amqp ack code that the sent message got. I've renamed ```sendComplete()``` to ```verifySendSucceeded()``` to make this more clear

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
```diff
@@ -0,0 +1,107 @@
...
+            {
+                if (System.currentTimeMillis() - startTime > CLOSE_REACTOR_GRACEFULLY_TIMEOUT)
+                {
+                    // The connection/session/link may not have been closed from the service's perspective, but we can free up the socket at least
+                    log.trace("Amqp reactor in thread {} failed to close gracefully in expected time frame, forcefully closing it now", this.threadName);
+                    break;
+                }
+            }
+            this.reactor.process();
```
abhipsaMisra: this is not required

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            // onTimerTask event will fire immediately in the basehandler being run. It is the responsibility of that handler
+            // to close it's link/session/connection and stop this reactor. This runner will allow some time for the amqp connection
+            // to be closed gracefully, but will forcefully free the resources if the graceful close takes too long
+            log.trace("Scheduling shutdown event for reactor for thread {}", threadName);
+            this.reactor.schedule(0, this.reactor.getHandler());
+
+            startTime = System.currentTimeMillis();
+            while (this.reactor.process())
```
abhipsaMisra: q - For receiver handlers, we close the local link -> session -> connection as soon as we receive an event. local connection closure stops the reactor from running, and then process() returns false and we exit. We can then free up the reactor thread. 
Shouldn't the `onTimerTask` be optional then? It should be called only if the link/session/connection is still open.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
timtay-microsoft: Removed

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/ReactorRunner.java
timtay-microsoft: You are right. If the reactor closes itself after receiving a message, there is no need to do the timerTask. I'll edit this

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -919,6 +919,8 @@ public void onTransportError(Event event)
         super.onTransportError(event);
         this.state = IotHubConnectionStatus.DISCONNECTED;
 
+        System.out.println("!!!!!!!!!!!!!!!!!!!!!!!!!!onTransportError!!!!!!!!!!!!!!!!!!!!!!!!!!");
```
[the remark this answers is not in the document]
timtay-microsoft: Removed

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/ErrorLoggingBaseHandler.java
```diff
@@ -17,49 +19,73 @@
...
     {
-        protonJExceptionParser = new ProtonJExceptionParser(event);
-        if (protonJExceptionParser.getError() == null)
+        if (event.getSession().getLocalState().equals(EndpointState.ACTIVE))
         {
-            log.warn("Amqp session was closed remotely with an unknown exception");
+            protonJExceptionParser = new ProtonJExceptionParser(event);
+            if (protonJExceptionParser.getError() == null)
+            {
+                log.warn("Amqp session was closed remotely with an unknown exception");
```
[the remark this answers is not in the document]
timtay-microsoft: I can make these log to error, sure

PR #717, opened 2020-02-27
Add support for JobProperties.StorageAuthenticationType property

bikamani:
Adding support for JobProperties.StorageAuthenticationType property


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
```diff
@@ -66,6 +66,11 @@
     @SerializedName(FAILURE_REASON_NAME)
     private String failureReason;
 
+    private static final String STORAGE_AUTHENTICATION_TYPE = "storageAuthenticationType";
+    @Expose(serialize = true, deserialize = true)
+    @SerializedName(STORAGE_AUTHENTICATION_TYPE)
```
drwill-ms: Shouldn't the serialized name be "storageAuthenticationType"?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/StorageAuthenticationType.java
```diff
@@ -0,0 +1,14 @@
+package com.microsoft.azure.sdk.iot.deps.serializer;
+
+import com.google.gson.annotations.SerializedName;
+/* Specifies authentication type being used for connecting to storage account */
+public enum StorageAuthenticationType
+{
+    /** Use a shared access key for authentication */
```
drwill-ms: This is missing the extra comment I added: This means authentication must be supplied in the storage URI(s)

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
```diff
@@ -126,6 +127,21 @@ public void setProgress(int progress) {
...
+
+    /**
+     * @param failureReason the failure reason.
+     */
+    public void setFailureReason(String failureReason) {
+        this.failureReason = failureReason;
+    }
+
     /**
      * @return URI containing SAS token to a blob container that contains registry data to sync.
```
drwill-ms: "containing SAS token" is only true if the storageAuthenticationType is keyBased.

Check all these comments, and compare them against the updated ones in the c# PR.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
bikamani: Yes that's what it is. Did I miss something?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/StorageAuthenticationType.java
bikamani: added

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
drwill-ms: No, I'm just new to Java. That is a string constant that has the value we want. I get it now. :)

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ExportImportTests.java
```diff
@@ -160,8 +213,18 @@ public void export_import_e2e() throws Exception
         {
             try
             {
-                exportJob = registryManager.exportDevices(containerSasUri, excludeKeys);
-                exportJobScheduled = true;
+                if(storageAuthenticationType.isPresent())
```
drwill-ms: minor: space between "if" and "("

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
```diff
@@ -272,4 +290,76 @@ JobPropertiesParser toJobPropertiesParser()
...
+    /**
+     * Creates an instance of JobProperties with parameters ready to start an Export job
+     *
+     * @param outputBlobContainerUri URI containing SAS token to a blob container.  This is used to output the status of the job and the results.
+     * @param excludeKeysInExport Indicates if authorization keys are included in export output
+     * @return An instance of JobProperties
+     */
+    public static JobProperties createForExportJob(
+            String outputBlobContainerUri,
+            Boolean excludeKeysInExport )
```
drwill-ms: minor, extra space before ending parenthesis

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -1439,6 +1579,7 @@ private String CreateImportJobPropertiesJson(String importBlobContainerUri, Stri
         return jobProperties.toJobPropertiesParser().toJson();
     }
 
+
```
drwill-ms: in c# we don't use multiple empty lines between class members. I don't now what Java's style guidelines are, but here's a note for you to check on that.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
```diff
@@ -59,13 +59,18 @@
     private static final String EXCLUDE_KEYS_IN_EXPORT_NAME = "excludeKeysInExport";
     @Expose(serialize = true, deserialize = true)
     @SerializedName(EXCLUDE_KEYS_IN_EXPORT_NAME)
-    private boolean excludeKeysInExport;
+    private Boolean excludeKeysInExport;
```
prmathur-microsoft: Why not use primitive?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ExportImportTests.java
```diff
@@ -160,8 +213,18 @@ public void export_import_e2e() throws Exception
...
+                {
+                    JobProperties exportJobProperties = JobProperties.createForExportJob(containerSasUri, excludeKeys, storageAuthenticationType.get());
+                    exportJob = registryManager.exportDevices(exportJobProperties);
```
drwill-ms: In my experience running several tests at once now, I've run into an JobQuotaExceededException. I thought of different ways of handling this, including using a semaphore, but that would only prevent tests in the same process from running at the same time.

I eventually landed on just guarding against that exception and retrying a certain number of times.

You'll want to do the same, otherwise these tests will occasionally fail and be considered flaky.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
```diff
@@ -156,40 +172,32 @@ public void setOutputBlobContainerUri(String outputBlobContainerUri) {
     }
 
     /**
-     * @return whether the keys are included in export or not.
+     * @return
      */
```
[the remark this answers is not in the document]
prmathur-microsoft: Where ever we expose a public API, we need better description, params etc. This data gets pulled into API reference doc we publish during release.

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ExportImportTests.java
drwill-ms: I see a catch for IotHubTooManyDevicesException, which seems like a weird one to throw, but maybe in Java it is a differently named exception.

If that is the same, you have it covered, although I think a 30 second sleep is a bit much. We don't need tests taking longer than they need to. In my test, I loop for a max of 30 times, sleeping 1 second each time. If I need to increase the iteration count because of failures, I will, but I'd rather know the jobs are taking a long time or that we have too many concurrent ones before I let tests run too long.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
bikamani: validating the get value as NULL or not in new changes. Not possible in boolean.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/StorageAuthenticationType.java
```diff
@@ -0,0 +1,14 @@
+package com.microsoft.azure.sdk.iot.deps.serializer;
+
+import com.google.gson.annotations.SerializedName;
+/* Specifies authentication type being used for connecting to storage account */
+public enum StorageAuthenticationType
+{
+    /* Use a shared access key for authentication */
+    @SerializedName("keyBased")
+    KEY,
+
```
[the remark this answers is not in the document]
bikamani: Can change to CamelCase if that's the standard we are following in java

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ExportImportTests.java
bikamani: done

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ExportImportTests.java
bikamani: yes you are right, it's a same exception, error code: 403. I will change sleep time .

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
bikamani: I looked at your PR and seeing same comment for this context. Did I miss something?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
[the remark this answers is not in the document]
bikamani: Agreed, added comments.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
prmathur-microsoft: Why would excludeKeysInExport be null? what is it set in swagger?

prmathur-microsoft: Ideally we shouldn't validate anything on the client side. Why are we validating this?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
```diff
@@ -270,7 +296,7 @@ public int getProgress()
      *
      * @return The value of excludeKeysInExport
      */
-    public boolean isExcludeKeysInExport()
+    public Boolean isExcludeKeysInExport()
     {
```
prmathur-microsoft: As this is a public API I would be hesistant to make this change. If we absolutely have to then we should follow the deprecation process here. Can you justify why?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/StorageAuthenticationType.java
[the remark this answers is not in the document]
prmathur-microsoft: Please visit our coding guidelines and reconvene

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
bikamani: I see, is there a better way to do this? -> https://github.com/Azure/azure-iot-sdk-java/pull/717/files#diff-85f710f651a02e7437edb830cfaafa4fR737

This is the main reason I had to change it to Boolean for null checking using JobProperties.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
bikamani: Are you sure we shouldn't be validating? I see our Import Export APIs (without storage Auth) validating mandatory parameters. Unit tests are also making sure we do this.

If that's not required, I can remove this change.

prmathur-microsoft: Yeah we were doing that in the very beginning time of SDK. However we got away from that practice  because we can keep track of service changes. So we don't validate anymore and reply on service to validate for us.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
prmathur-microsoft: So if it is boolean and not Boolean then customers can never set it to null and will be forced to provide a value. So I don't see value here in this check. What was the motivation for this check?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
```diff
@@ -272,4 +298,76 @@ JobPropertiesParser toJobPropertiesParser()
...
+
+    /**
+     * Creates an instance of JobProperties with parameters ready to start an Import job
+     *
+     * @param inputBlobContainerUri URI containing SAS token to a blob container that contains registry data to sync
+     * @param outputBlobContainerUri URI containing SAS token to a blob container.  This is used to output the status of the job and the results.
+     * @return An instance of JobProperties
+     */
+    public static JobProperties createForImportJob(
+            String inputBlobContainerUri,
```
prmathur-microsoft: why static? wouldn't it vary per instance of the job?

bikamani: This is static as customer does not need to create object of class before accessing this method. I am not sure what you mean as per instance of the job here?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
bikamani: That's only true for earlier APIs. In new API it is JobProperties and if excludeKeys is not set, we are sending the request to service without mandatory requirements. Motivation of this check is to make sure these two necessary parameters are provided at least.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/StorageAuthenticationType.java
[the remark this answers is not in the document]
bikamani: Doesn't say anything about enums.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
bikamani: As per our conversation offline, removing the validation for import export API and reverting back changes for boolean -> Boolean

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
drwill-ms: Static because one doesn't need an instance to call it, rather it returns an instance. It is a factory method.

drwill-ms: I changed mine to:

/// <param name="inputBlobContainerUri">URI to a blob container that contains registry data to sync. Including a SAS token is dependent on the <see cref="StorageAuthenticationType" /> parameter.</param>
/// <param name="outputBlobContainerUri">URI to a blob container.  This is used to output the status of the job and the results. Including a SAS token is dependent on the <see cref="StorageAuthenticationType" /> parameter.</param>

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
bikamani: Done

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
prmathur-microsoft: factory in java...cool. Haven't seen  that in a while

service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/RegistryManagerTest.java
```diff
@@ -1162,7 +1163,74 @@ public Device getStatistics() throws IOException, IotHubException
...
+                mockIotHubExceptionManager.httpResponseVerification((HttpResponse) any);
+                mockHttpResponse.getBody();
+                result = jobPropertiesJson.getBytes();
+            }
+        };
+
+        RegistryManager registryManager = RegistryManager.createFromConnectionString(connectionString);
+        JobProperties exportJobProperties =
+                JobProperties.createForExportJob("blob1", true, StorageAuthenticationType.IDENTITY);
+        JobProperties jobProperties = registryManager.exportDevices(exportJobProperties);
```
prmathur-microsoft: So where are you setting the env for this to work?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
prmathur-microsoft: won't make you go back and redesign, but in java we usually use builder in these cases. factory is frowned upon.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
```diff
@@ -156,40 +175,32 @@ public void setOutputBlobContainerUri(String outputBlobContainerUri) {
...
+        this.storageAuthenticationType = storageAuthenticationType;
     }
 
     /**
-     * @return System generated. Ignored at creation.
-     * If status == failure, this represents a string containing the reason.
+     * @return whether the keys are included in export or not.
      */
-    public String getFailureReason() {
-        return failureReason;
```
prmathur-microsoft: Maybe its a github thing but I see this API missing.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/StorageAuthenticationType.java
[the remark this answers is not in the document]
prmathur-microsoft: We should leave it as is. If I rem enums should be caps. wanted to confirm that.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobPropertiesParser.java
drwill-ms: A quick check of the Java repo shows you have about half of your enums as all caps, and the other half camelCased.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/StorageAuthenticationType.java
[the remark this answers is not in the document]
drwill-ms: A quick check of the Java repo shows you have about half of your enums as all caps, and the other half camelCased.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/IotHubConnectionString.java
```diff
@@ -632,7 +647,19 @@ public URL getUrlCreateExportImportJob() throws MalformedURLException
         stringBuilder.append(URL_SEPARATOR_0);
         stringBuilder.append("create");
         stringBuilder.append(URL_SEPARATOR_1);
-        stringBuilder.append(URL_API_VERSION);
+        // The new api-version is only available in a few initial regions
+        // Control access via an environment variable. If a user wishes to try it out,
+        // they can set "EnabledStorageIdentity" to "1". Otherwise, the SDK will still
+        // default to the latest, broadly-supported api-version used in this SDK.
+        IotHubConnectionString iotHubConnectionString = new IotHubConnectionString();
+        if (iotHubConnectionString.IsStorageIdentityEnabled)
```
drwill-ms: Does this set the new api-version globally, or just for import/export devices job?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
drwill-ms: What do people have against factory pattern? Genuinely curious.

service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/RegistryManagerTest.java
drwill-ms: For .NET, I'm setting it locally when I test.

For our pipeline, we can't set the env var because this api-version will not be rolled to west us, so the test would fail. We'll have to wait for the feature to make it to west us, and enable our hubs with identity (still a TODO for me, so I created this PBI: https://msazure.visualstudio.com/One/_boards/board/t/IoT-Developers%20and%20Devices-Managed/Backlog%20items/?workitem=6454493).

For now, we're relying on local testing to validate the new functionality.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -718,6 +718,69 @@ public JobProperties exportDevices(String exportBlobContainerUri, Boolean exclud
...
+    {
+        // CODES_SRS_SERVICE_SDK_JAVA_REGISTRYMANAGER_15_062: [The function shall get the URL for the bulk export job creation]
+        URL url = iotHubConnectionString.getUrlCreateExportImportJob();
+
+        // CODES_SRS_SERVICE_SDK_JAVA_REGISTRYMANAGER_15_063: [The function shall create a new SAS token for the bulk export job]
+        String sasTokenString = new IotHubServiceSasToken(this.iotHubConnectionString).toString();
+
+        // CODES_SRS_SERVICE_SDK_JAVA_REGISTRYMANAGER_15_064: [The function shall create a new HttpRequest for the bulk export job creation ]
+        exportDevicesParameters.setType(JobProperties.JobType.EXPORT);
+        String jobPropertiesJson = exportDevicesParameters.toJobPropertiesParser().toJson();
```
drwill-ms: If this feature is off, you need to make sure the storageAuthenticationType is not serialized into the json, otherwise the service will reject the request.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/IotHubConnectionString.java
bikamani: Thanks for catching this. It is setting only for import export currently, will change it to all.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
bikamani: It is checked in toJobPropertiesParser. It won't be serialized into json

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/IotHubConnectionString.java
drwill-ms: No, we want it only for import/export job. Not all. Sorry for the confusion.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
bikamani: +1,I also want to know.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/IotHubConnectionString.java
bikamani: As discussed offline, this is only required for import export which is what it is currently. No changes required, resolving

PR #727, opened 2020-03-11
feat(iot-dev): Add support for CA signed X509 over AMQPS_WS, MQTT_WS, and HTTPS

timtay-microsoft:
Previously, these changes were not sufficient, but something changed on the service side to make them work as expected. We no longer need the query string "?iothub-no-client-cert=true"


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -132,17 +133,21 @@ public void open(Queue<DeviceClientConfig> deviceClientConfigs, ScheduledExecuto
...
-                        //Codes_SRS_MQTTIOTHUBCONNECTION_34_027: [If this function is called while using websockets and x509 authentication, an UnsupportedOperationException shall be thrown.]
-                        throw new UnsupportedOperationException("X509 authentication is not supported over MQTT_WS");
-                    }
-
                     this.log.trace("MQTT connection will use X509 certificate based auth");
 
                     this.iotHubUserPassword = null;
                 }
 
+                if (this.config.getAuthenticationType() != DeviceClientConfig.AuthType.X509_CERTIFICATE)
```
bikamani: isn't this "else" case basically?

timtay-microsoft: Yeah, I can simplify this, sure

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -132,17 +133,21 @@ public void open(Queue<DeviceClientConfig> deviceClientConfigs, ScheduledExecuto
...
+                {
+                    //X509 auth websocket should not include the iothub-no-client-cert flag
+                    this.webSocketQueryString = NO_CLIENT_CERT_QUERY_STRING;
```
bikamani: Do we need "this" any more? I know it is there in legacy code but this going forward if we don't need it why use?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -129,17 +130,15 @@ public void open(Queue<DeviceClientConfig> deviceClientConfigs, ScheduledExecuto
...
                 {
-                    if (this.config.isUseWebsocket())
-                    {
-                        //Codes_SRS_MQTTIOTHUBCONNECTION_34_027: [If this function is called while using websockets and x509 authentication, an UnsupportedOperationException shall be thrown.]
-                        throw new UnsupportedOperationException("X509 authentication is not supported over MQTT_WS");
-                    }
-
                     this.log.trace("MQTT connection will use X509 certificate based auth");
 
+                    this.webSocketQueryString = "";
```
bikamani: So this needs to be set to empty string? Or should there be conditional check when appending this?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -129,17 +130,15 @@ public void open(Queue<DeviceClientConfig> deviceClientConfigs, ScheduledExecuto
...
+
+                    //X509 auth over websocket should not include the iothub-no-client-cert flag
```
bikamani: Did you want to move this before line 139?

timtay-microsoft: Whoops, yes I do

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
timtay-microsoft: Sure, I can leave it null for SAS cases, and just check when building the string later

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
timtay-microsoft: C# guidelines suggest avoiding "this." when possible, but Java guidelines do not

bikamani: Sorry about that, I added this comment thinking about C#

PR #732, opened 2020-03-13
document EnabledStorageIdentity for new APIs

bikamani:


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -721,7 +721,9 @@ public JobProperties exportDevices(String exportBlobContainerUri, Boolean exclud
     /**
      * Create a bulk export job.
      *
-     * @param exportDevicesParameters A JobProperties object containing input parameters for export Devices job
+     * @param exportDevicesParameters A JobProperties object containing input parameters for export Devices job.
+     *                                This API initially has a limited availability and is only is implemented in a few regions.
```
abhipsaMisra: We should clarify that authentication based on Identity has limited availability. There are other fields in JobProperties as well that do not have any restrictions on their availability.

bikamani: Do you think it is really required? There are no other parameters in JobProperties other than what's in legacy APIs which will be helpful to provide and use this API instead of legacy with old API version. Or in other words the only reason I see someone wanting to use this API currently is to add storageAuthType for the request.

abhipsaMisra: Not really, our motivation for adding JobProperties was to enable users to pass in a single object having the necessary fields, instead of multiple fields individually to the API. IdentityBased authentication just happens to be one of those multiple fields.

abhipsaMisra: Also, why not have this documentation in the class wherever the enum lives? Any particular reason for adding this here?

bikamani: I would expect user to find relevant documentation where the API is defined instead of where the enum is. Also this is temporary description until we have this feature available in all regions, we will be removing it later.

abhipsaMisra: > I would expect user to find relevant documentation where the API is defined instead of where the enum is. Also this is temporary description until we have this feature available in all regions, we will be removing it later.

That's reasonable. 
In that case you need to update the comment to indicate what feature it is that has limited availability, the export API as a whole does not have limited availability.

bikamani: > Not really, our motivation for adding JobProperties was to enable users to pass in a single object having the necessary fields, instead of multiple fields individually to the API. IdentityBased authentication just happens to be one of those multiple fields.

You are correct, that is one of the motivation going forward as we add new parameters, also we have factory methods to create JsonProperties based on parameter user would like to set. I agree with you that clarifying it would be better. However for now I do not want to go through build pipeline again and would like to get this included in the release.

bikamani: We have added it in the release notes.

> > I would expect user to find relevant documentation where the API is defined instead of where the enum is. Also this is temporary description until we have this feature available in all regions, we will be removing it later.
> 
> That's reasonable.
> In that case you need to update the comment to indicate what feature it is that has limited availability, the export API as a whole does not have limited availability.

We have added it in the release notes

abhipsaMisra: "This API initially has a limited availability and is only is implemented in a few regions."

This is the API in question, does it hardcode the auth mechanism to be identity based? If not, then this API does not have limited availability.
```
public JobProperties exportDevices(JobProperties exportDevicesParameters) throws IllegalArgumentException, IOException, IotHubException, JsonSyntaxException
```

PR #733, opened 2020-03-16
Do not accept HTTP urls in Https connection constructor

azabbasi:
# Checklist

[HTTP plaintext communication Java SDK](https://msazure.visualstudio.com/One/_workitems/edit/5867818)

# Description of the problem
IoT Hub Java SDK allows unencrypted (HTTP) communication within an https constructor

# Description of the solution
We will not accept an HTTP method in an HTTPS connection constructor

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsConnection.java
@@ -70,10 +70,10 @@ public class HttpsConnection
     {
         // Codes_SRS_HTTPSCONNECTION_11_022: [If the URI given does not use the HTTPS or HTTP protocol, the constructor shall throw an IllegalArgumentException.]
         final String protocol = url.getProtocol();
-        if (!protocol.equalsIgnoreCase("HTTPS") && !protocol.equalsIgnoreCase("HTTP"))
+        if (!protocol.equalsIgnoreCase("HTTPS"))
         {
             String errMsg = String.format("Expected URL that uses protocol "
-                            + "HTTPS or HTTP but received one that uses "
+                            + "HTTPS but received one that uses "
                             + "protocol '%s'.%n",
                     protocol);
             throw new IllegalArgumentException(errMsg);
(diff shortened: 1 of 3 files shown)


device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/transport/https/HttpsConnectionTest.java
```diff
@@ -394,27 +394,21 @@ public void setSSLContextSetsContextWithProxy(@Mocked final SSLContext mockedCon
     }
 
     //Tests_SRS_HTTPSCONNECTION_34_026: [If this object uses HTTP, this function shall throw an UnsupportedOperationException.]
-    @Test (expected = UnsupportedOperationException.class)
+    @Test (expected = IllegalArgumentException.class)
     public void setSSLContextThrowsIfHttp(@Mocked final SSLContext mockedContext) throws IOException, TransportException
```
timtay-microsoft: Since this test method now checks that the constructor throws, you can rename this test method

azabbasi: I totally spaced out on the name ! sure thing

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsConnection.java
```diff
@@ -70,10 +70,10 @@ public HttpsConnection(URL url, HttpsMethod method, final ProxySettings proxySet
     {
         // Codes_SRS_HTTPSCONNECTION_11_022: [If the URI given does not use the HTTPS or HTTP protocol, the constructor shall throw an IllegalArgumentException.]
         final String protocol = url.getProtocol();
-        if (!protocol.equalsIgnoreCase("HTTPS") && !protocol.equalsIgnoreCase("HTTP"))
+        if (!protocol.equalsIgnoreCase("HTTPS"))
         {
```
prmathur-microsoft: If we make this change we need to let PM know before drop it off. lets chat about this

azabbasi: The method signature doesn't change and if someone uses http instead of https, the SSL context will throw an exception. so I don't think there is a behavior change. The exception we throw will be slightly different though.

timtay-microsoft: I took a look through the code, and there are no instances where the SDK would ever pass "http" instead of "https", so there is no impact on our customers here. We aren't taking away http support because we never had http support

timtay-microsoft:
Before you merge this in, ping me on Teams so I can walk you through general git guidelines we maintain on this team

PR #735, opened 2020-03-16
Improve API documentation

bikamani:


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -722,6 +722,11 @@ public JobProperties exportDevices(String exportBlobContainerUri, Boolean exclud
      * Create a bulk export job.
      *
      * @param exportDevicesParameters A JobProperties object containing input parameters for export Devices job
+     *                                This API also supports identity bases storage authentication, identity authentication
```
abhipsaMisra: type: This API also supports identity ~bases~ based storage authenticatio

bikamani: done

PR #743, opened 2020-03-25
fix(iot-device): Fix issue where SDK didn't validate application properties correctly

### LINKED ISSUE #724: usesValidChars method of MessageProperty does not work, always returning true
- **OS and version used:** Custom build Linux, not relevant
- **Java runtime used:** 1.8.0, not relevant
- **SDK version used:** 1.20.0

# Description of the issue:
The MessageProperty documentation https://azure.github.io/azure-iot-sdk-java/device/com/microsoft/azure/iothub/MessageProperty.html states that usesValidChars method should return true only if the string contains US-ASCII printable chars, and does not contain ()<>@,;:\"/[]?={} (space) (horizontal tab). The regex used for this validation does not seem to work, and the method always returns true, however messages are never sent to Azure.

Also, if the string contains invalid characters (in our case, this was a space) there is no exception or information logged when attempting to send the invalid message, and the callback of sendEventAsync is not entered. However, when we close our application and the resources are disposed, all previously sent messages will be logged with MESSAGE_CANCELLED_ONCLOSE status.

timtay-microsoft:
Normally the SDK wouldn't validate something like this, but if a user tries to send an application property with an invalid character in it, the different protocols react in different and unintuitive ways. MQTT will sever the connection with no error message while AMQPS and HTTPS will allow it.

I'm also taking this opportunity to fix a hack tied to URL encoding. URLEncoder doesn't handle encoding spaces correctly, but URIUtil from apache commons does

Lastly, the application property key should be encoded just like the value is when using mqtt, so I've fixed that too

fixes #724

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MessageProperty.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MessageProperty.java
@@ -31,6 +31,11 @@ public final class MessageProperty {
     public static final String IOTHUB_SECURITY_INTERFACE_ID = "iothub-interface-id";
     public static final String IOTHUB_SECURITY_INTERFACE_ID_VALUE = "urn:azureiot:Security:SecurityAgent:1";
 
+    //Note this list includes the characters for tab and space.
+    private static final String BANNED_NON_ALPHANUMERIC_CHARACTERS = "()<>@,;:\"\\[]?={} \t";
+    private static final int ASCII_LOWER_BOUND = 32;
+    private static final int ASCII_UPPER_BOUND = 127;
+
 
     static {
         HashSet<String> reservedPropertyNames = new HashSet<>();
@@ -89,7 +94,7 @@ public final class MessageProperty {
 
         // Codes_SRS_MESSAGEPROPERTY_11_002: [If the name contains a character that is not in US-ASCII, the function shall throw an IllegalArgumentException.]
         if (!usesValidChars(name)) {
-            String errMsg = String.format("%s is not a valid IoT Hub message property name. %n", name);
+            String errMsg = String.format("%s is not a valid IoT Hub message property name. Characters in property keys must fall within ASCII value range %d and %d but excluding the set %s", name, ASCII_LOWER_BOUND, ASCII_UPPER_BOUND, BANNED_NON_ALPHANUMERIC_CHARACTERS);
             throw new IllegalArgumentException(errMsg);
         }
 
@@ -102,7 +107,7 @@ public final class MessageProperty {
         // Codes_SRS_MESSAGEPROPERTY_11_003: [If the value contains a character that is not in US-ASCII, the function shall throw an IllegalArgumentException.]
         if (!usesValidChars(value))
         {
-            String errMsg = String.format("%s is not a valid IoT Hub message property value.%n", value);
+            String errMsg = String.format("%s is not a valid IoT Hub message property value. Characters in property values must fall within ASCII value range %d and %d but excluding the set %s", value, ASCII_LOWER_BOUND, ASCII_UPPER_BOUND, BANNED_NON_ALPHANUMERIC_CHARACTERS);
             throw new IllegalArgumentException(errMsg);
         }
 
@@ -192,22 +197,23 @@ public final class MessageProperty {
         return propertyIsValid;
     }
 
-    /**
-     * Returns true if the string only uses US-ASCII 
-     *
-     * @param s the string.
-     *
-     * @return whether the string only uses US-ASCII 
-     */
     private static boolean usesValidChars(String s) {
-        boolean isValid = false;
-
-        if (s.matches("\\p{ASCII}*"))
+        for (char c : s.toCharArray())
         {
-            isValid = true;
+            //Range of characters includes US alphanumeric characters, as well as a few special characters like '+' and ','
+            // Any characters outside of this range are banned
+            if (!(c > ASCII_LOWER_BOUND && c < ASCII_UPPER_BOUND))
+            {
+                return false;
+            }
+
+            if (BANNED_NON_ALPHANUMERIC_CHARACTERS.indexOf(c) != -1)
+            {
+                return false;
+            }
         }
 
-        return isValid;
+        return true;
     }
 
     @SuppressWarnings("unused")
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttMessaging.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttMessaging.java
@@ -102,24 +102,24 @@ public class MqttMessaging extends Mqtt
         //Codes_SRS_MqttMessaging_34_032: [If the message has a content type, this method shall append that to publishTopic before publishing using the key name `$.ct`.]
         //Codes_SRS_MqttMessaging_34_032: [If the message has a content encoding, this method shall append that to publishTopic before publishing using the key name `$.ce`.]
         //Codes_SRS_MqttMessaging_34_034: [If the message has a creation time utc, this method shall append that to publishTopic before publishing using the key name `$.ctime`.]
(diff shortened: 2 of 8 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -138,10 +140,8 @@ else if (this.config.getAuthenticationType() == DeviceClientConfig.AuthType.X509
                     this.iotHubUserPassword = null;
                 }
 
-                //URLEncoder follows HTML spec for encoding urls, which includes substituting space characters with '+'
-                // We want "%20" for spaces, not '+', however, so replace them manually after utf-8 encoding
                 String userAgentString = this.config.getProductInfo().getUserAgentString();
-                String clientUserAgentIdentifier = "DeviceClientType=" + URLEncoder.encode(userAgentString, "UTF-8").replaceAll("\\+", "%20");
+                String clientUserAgentIdentifier = "DeviceClientType=" + URIUtil.encodePath(userAgentString, StandardCharsets.UTF_8.name());
```
timtay-microsoft: URIUtil encodePath handles spaces correctly for our purposes, whereas URLEncoder did not, hence the switch

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttMessaging.java
```diff
@@ -152,18 +154,18 @@ private boolean appendPropertyIfPresent(StringBuilder stringBuilder, boolean sep
                     stringBuilder.append(MESSAGE_PROPERTY_SEPARATOR);
                 }
 
-                stringBuilder.append(propertyKey);
+                stringBuilder.append(URIUtil.encodePath(propertyKey, StandardCharsets.UTF_8.name()));
```
timtay-microsoft: We weren't encoding the application property keys, which could be an issue. If a customer gives us a key with odd characters in it, we need to encode in order to keep the mqtt topic string valid

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MessageProperty.java
```diff
@@ -31,6 +31,11 @@
     public static final String IOTHUB_SECURITY_INTERFACE_ID = "iothub-interface-id";
     public static final String IOTHUB_SECURITY_INTERFACE_ID_VALUE = "urn:azureiot:Security:SecurityAgent:1";
 
+    //Note this list includes the characters for tab and space.
+    private static final String BANNED_NON_ALPHANUMERIC_CHARACTERS = "()<>@,;:\"/[]?={} \t";
```
azabbasi: Is this a final set of characters? where did you get them from? is it defined somewhere? can we have a link as a comment?

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/MessagePropertyTest.java
```diff
@@ -46,32 +46,6 @@ public void constructorSavesPropertyValue()
 		
     }
 
-	 @Test
-    public void constructorSavesPropertyValueSlash()
```
timtay-microsoft: These tests were incorrect, as '/' characters are not allowed in application properties

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/MessagePropertyTest.java
```diff
@@ -46,32 +46,6 @@ public void constructorSavesPropertyValue()
...
-        MessageProperty property = new MessageProperty(name, value);
-        String testValue = property.getValue();
-
-        final String expectedValue = value;
-        assertThat(testValue, is(expectedValue));
-    }
-    
-    @Test
-    public void constructorSavesPropertyValueTab()
-    {
```
timtay-microsoft: Same here. Tab is an illegal character, so we shouldn't have had a test like this

device/iot-device-client/pom.xml
```diff
@@ -107,6 +107,11 @@
             <artifactId>slf4j-api</artifactId>
             <version>1.7.25</version>
         </dependency>
+        <dependency>
+            <groupId>commons-httpclient</groupId>
+            <artifactId>commons-httpclient</artifactId>
+            <version>3.1</version>
```
prmathur-microsoft: How huge is this dependency>

prmathur-microsoft: Does it work on android?

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsDeviceTelemetryTest.java
```diff
@@ -544,6 +549,15 @@ public void protonMessageToIoTHubMessageWithInputName(@Mocked final MessageImpl
...
                 result = mockedAnnotations;
 
+                mockMessage.getProperties();
+                result = mockProperties;
+
+                mockProperties.getUserId();
+                result = mockBinary;
+
+                mockBinary.toString();
+                result = "someUserId";
```
timtay-microsoft: The userId wasn't being mocked properly, and it was hitting the codepath in MessageProperty.java that complains about the contents of the userId not being valid. Without this change, it was using a memory address with several illegal characters

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTest.java
```diff
@@ -1253,8 +1253,8 @@ public void receiveSuccessCustomPropertyHasUnusualCharacters() throws TransportE
         assertEquals("$", receivedMessage.getProperties()[0].getValue());
         assertEquals("&", receivedMessage.getProperties()[1].getValue());
         assertEquals("%", receivedMessage.getProperties()[2].getName());
-        assertEquals("\"", receivedMessage.getProperties()[2].getValue());
```
timtay-microsoft: This test was bad because these characters should never appear in an application property

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -7,10 +7,12 @@
 import com.microsoft.azure.sdk.iot.device.exceptions.TransportException;
 import com.microsoft.azure.sdk.iot.device.transport.*;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.httpclient.util.URIUtil;
 
 import javax.net.ssl.SSLContext;
 import java.io.IOException;
 import java.net.URLEncoder;
```
barustum: is this still needed?

device/iot-device-client/pom.xml
timtay-microsoft: 300kb, so it's not that bad. The gate will let us know if android will be an issue, but I don't expect it will be

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MessageProperty.java
timtay-microsoft: I got this requirement from our own documentation here: https://azure.github.io/azure-iot-sdk-java/device/com/microsoft/azure/iothub/MessageProperty.html

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
timtay-microsoft: Yes, we need to use the URIUtil to encode some strings in this class

Issue #574, opened 2019-08-27
Service client apparently can send an invalid eTag and it still updates the twin.


anthonyvercolano:
I tried running a java service client with an invalid eTag and it worked.

Do a patch with an eTag of 'abc' (which can never be correct).

timtay-microsoft:
The SDK needs to surround the etag values with quotes for the iot hub to interpret them correctly

barustum:
Synced up with a couple of team members. We decided to hold off on this fix as it might introduce some complexity for our developers, not to mention it might introduce breaking changes. We gave this issue a lower priority, hence we will address this fix as an enhancement once it's at a higher priority.

timtay-microsoft:
Just a quick status update: We plan on doing a complete revamp of the iothub service client in the coming months. During that process, we will fix this issue

timtay-microsoft:
As per the above, this issue will not be fixed in this codebase. We will have a separate codebase for this revamp of our service client.

az-iot-builder-01:
@anthonyvercolano, @timtay-microsoft, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #749, opened 2020-03-31
fix(iot-service): Send if-match header in the request to updateTwin

barustum:
This pull request is a fix for [this]( https://github.com/Azure/azure-iot-sdk-java/issues/574) issue.

This change ensures an if-match header is sent along with the request to update the twin. This header ensures that the etag a user sets explicitly matches the one that is set in the twin. If both etags are different, the service should throw a IotHubPreconditionFailedException exception. Without the if-match header, a user can make updates to the device twin without knowing whether another update has been performed from a different client.

### DIFF
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/twin/UpdateTwinTests.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/twin/UpdateTwinTests.java
@@ -5,18 +5,20 @@
 
 package com.microsoft.azure.sdk.iot.common.tests.iothub.twin;
 
-import com.microsoft.azure.sdk.iot.common.helpers.ClientType;
 import com.microsoft.azure.sdk.iot.common.helpers.ConditionalIgnoreRule;
 import com.microsoft.azure.sdk.iot.common.helpers.StandardTierOnlyRule;
-import com.microsoft.azure.sdk.iot.common.setup.iothub.DeviceTwinCommon;
-import com.microsoft.azure.sdk.iot.device.IotHubClientProtocol;
+import com.microsoft.azure.sdk.iot.common.helpers.Tools;
+import com.microsoft.azure.sdk.iot.common.tests.iothub.serviceclient.ServiceClientTests;
 import com.microsoft.azure.sdk.iot.device.exceptions.ModuleClientException;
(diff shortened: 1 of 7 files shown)


abhipsaMisra:
Note regarding naming commits: https://github.com/Azure/azure-iot-sdk-java/blob/master/.github/CODING_GUIDELINES.md#Commit-Guidelines

timtay-microsoft:
You will need to add an Android runner class to mirror the JVMRunner class you made and a new android test group to run it. You will also need to edit our windowsLinuxAndAndroidBuildMatrixConfig.yaml file to use the new android test group. We can chat about this step if you need help.

You can edit the yaml file to just add a new line for testGroup40 like we have for [testGroup39](https://github.com/Azure/azure-iot-sdk-java/blob/8f1fd3c34cc4937273564ef6160c80b6894c306c/vsts/windowsLinuxAndAndroidBuildMatrixConfig.yaml#L258)

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/twin/UpdateTwinTests.java
```diff
@@ -0,0 +1,83 @@
...
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.security.GeneralSecurityException;
+import java.util.*;
+
+/**
+ * Test class containing all tests to be run on JVM and android pertaining to Queries. Class needs to be extended
+ * in order to run these tests as that extended class handles setting connection strings and certificate generation
+ */
+public class UpdateTwinTests extends DeviceTwinCommon
```
timtay-microsoft: Extending DeviceTwinCommon will make your tests in this file run with parameterized inputs on device client protocols, which doesn't seem relevant to a service client only test. Could you remove this extension and add in the parts that you need from it manually instead?

barustum: sure

barustum:
@abhipsaMisra . Thanks for letting me know about the commit naming guidelines. I will fix the name when I squash and merge.

barustum:
Synced up with a couple of team members. We decided to hold off on this fix as it might introduce some complexity for our developers, not to mention it might introduce breaking changes. We gave this issue a lower priority, hence we will address this fix as an enhancement once it's at a higher priority.

Issue #739, opened 2020-03-22
[Module Client] Throughput Issues When Receiving Messages From OPC Publisher 2.6.x


jerickcolendres:
- **OS and version used:** <VERSION> 
Host: Ubuntu 18.04.2 LTS

- **Java runtime used:** <VERSION> 
openjdk version "1.8.0_212"
OpenJDK Runtime Environment (IcedTea 3.12.0) (Alpine 8.212.04-r0)
OpenJDK 64-Bit Server VM (build 25.212-b04, mixed mode)

- **SDK version used:** <VERSION> 
1.20.2

- **IOT Edge runtime**
iotedge 1.0.8 (208b2204fd30e856d00b280112422130c104b9f0)

# Description of the issue:

Hi,

We currently have a custom IOT edge module consuming messages from  **OPC Publisher 2.6.x** module using **MQTT** protocol. Since the new version of OPC Publisher 2.6.x no longer support message batching like the previous 2.5.x version, messages are sent one by one and we are experiencing throughput problems. I look at the code and the current implementation uses a scheduler that runs a **IotHubReceiveTask** which takes a message from the **receivedMessagesQueue** every **receivedPeriodInMillis**. For MQTT, the RECEIVE_PERIOD_MILLIS_MQTT value is set to 10 so it can only process 100 message per second. We need to process over 1000 OPC UA values per second and on our largest sites we might have around 80,000+ telemetry tags configured with an average of 15 seconds between updates.

We already contacted the OPC Publisher team about the urgent need to support batched messages but is there any other way to increase the message throughput? Would it be possible to modify the code to poll and process messages in realtime by having a long running thread consuming from a blocking queue? If possible, could you apply that to the "sending" code as well? We have another custom IOT Edge module that had the same message throughput problem when sending messages  using the Java IOT SDK and we worked around the problem by not using the SDK and directly send the messages to IOT Hub via REST API.

# Code sample exhibiting the issue:

ModuleClient.java
```java
public class ModuleClient extends InternalClient
{
    ...
    private static long RECEIVE_PERIOD_MILLIS_MQTT = 10;
    ...
}
```

DeviceIO.java
```java
public final class DeviceIO
{
    ...
    private void commonOpenSetup()
    {
        this.sendTask = new IotHubSendTask(this.transport);
        this.receiveTask = new IotHubReceiveTask(this.transport);

        this.taskScheduler = Executors.newScheduledThreadPool(2);
    
[...]

PR #752, opened 2020-04-01
fix(iot-device):Introduce a new option for receive interval

azabbasi:
# Checklist

https://github.com/Azure/azure-iot-sdk-java/issues/739

# Description of the problem
The SDK needs a way to allow users to configure the receive interval for their connection oriented protocols (MQTT and AMQP)

# Description of the solution
We will expose a new option name that specifically calls out receive interval on the internal client.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
@@ -24,7 +24,10 @@ import static com.microsoft.azure.sdk.iot.device.IotHubClientProtocol.*;
 @Slf4j
 public class InternalClient
 {
+    // SET_MINIMUM_POLLING_INTERVAL is used for setting the interval for https polling task.
     static final String SET_MINIMUM_POLLING_INTERVAL = "SetMinimumPollingInterval";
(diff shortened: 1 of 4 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -353,6 +356,10 @@ public DeviceClientConfig getConfig()
      *	      option specifies the interval in milliseconds between calls to
      *	      the service checking for availability of new messages. The value
      *	      is expected to be of type {@code long}.
+     *	    - <b>SetReceiveInterval</b> - this option is applicable to all protocols
+     *	      in case of HTTPS protocol, this option acts the same as {@code SetMinimumPollingInterval}
+     *	      in case of MQTT and AMQP protocols, this option specifies the interval in millisecods between
+     *	      spawning up threads that would read a message off of the connection and invoking the message callback.
      *	    - <b>SetCertificatePath</b> - this option is applicable only
```
timtay-microsoft, line 363: "... between spawning a thread that dequeues a message from the SDK's queue of received messages..." Just to be clear that these messages aren't being taken directly from the connection, but from the SDK's transport layer that already received the message

PR #755, opened 2020-04-10
feat(iot-dev): Add setOption API for setting https request read and connect timeouts

timtay-microsoft:
Customer's don't like how we originally hard coded the read timeout to 4 minutes, so let's make these levers available to the customer. This PR does not change the default behavior of the SDK, but it does allow users to set new values for the read and connect timeouts

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
@@ -11,6 +11,8 @@ import com.microsoft.azure.sdk.iot.provisioning.security.SecurityProvider;
 import com.microsoft.azure.sdk.iot.provisioning.security.SecurityProviderSymmetricKey;
 import com.microsoft.azure.sdk.iot.provisioning.security.SecurityProviderTpm;
 import com.microsoft.azure.sdk.iot.provisioning.security.SecurityProviderX509;
+import lombok.Getter;
+import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 
 import javax.net.ssl.SSLContext;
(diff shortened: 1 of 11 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
```diff
@@ -25,8 +27,9 @@
 @Slf4j
 public final class DeviceClientConfig
 {
-    /** The default value for readTimeoutMillis. */
-    private static final int DEFAULT_READ_TIMEOUT_MILLIS = 240000;
+    private static final int DEFAULT_HTTPS_READ_TIMEOUT_MILLIS = 240000;
```
drwill-ms, line 30: 4 minute? would you mind adding a comment to that effect

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsConnection.java
```diff
@@ -182,12 +182,21 @@ public void setRequestHeader(String field, String value)
      *
      * @param timeout the read timeout.
      */
-    public void setReadTimeoutMillis(int timeout)
+    public void setReadTimeout(int timeout)
```
drwill-ms, line 185: Doesn't this change a public interface?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsRequest.java
```diff
@@ -174,10 +180,23 @@ public HttpsRequest setHeaderField(String field, String value)
      *
      * @return itself, for fluent setting.
      */
-    public HttpsRequest setReadTimeoutMillis(int timeout)
+    public HttpsRequest setReadTimeout(int timeout)
```
drwill-ms, line 183: Same here, changed public API?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsConnection.java
timtay-microsoft: Java doesn't consider this class part of the public interface, no.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsRequest.java
timtay-microsoft: See above, this isn't a public API

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
timtay-microsoft: sure thing

PR #756, opened 2020-04-13
fix(iot-dev): Upgrade mqtt Paho dependency version

### LINKED ISSUE #596: [Java 1.7] MQTT runtime exception due to paho 1.2.2
Details on the paho issue [here](https://github.com/eclipse/paho.mqtt.java/issues/633)

The latest release of the java device client/deps packages takes a dependency on paho 1.2.2 which encounters a runtime exception while opening a connection using mqtt.

This exception seems to be limited to mqtt users below Android 7, as well as Java SDK version 1.7 and below. JDK 1.8+ works as expected.

timtay-microsoft:
Fixes #596

Here are the release notes from this new Paho version: https://github.com/eclipse/paho.mqtt.java/releases/tag/v1.2.3

### DIFF
--- a/device/iot-device-client/pom.xml
+++ b/device/iot-device-client/pom.xml
@@ -47,12 +47,6 @@
             <version>0.23</version>
         </dependency>
         <dependency>
-            <groupId>org.apache.qpid</groupId>
-            <artifactId>proton-j</artifactId>
-            <version>0.30.0</version>
-            <type>jar</type>
-        </dependency>
-        <dependency>
             <groupId>com.microsoft.azure.sdk.iot</groupId>
             <artifactId>${iot-deps-artifact-id}</artifactId>
             <version>${iot-deps-version}</version>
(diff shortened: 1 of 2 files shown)


device/iot-device-client/pom.xml
```diff
@@ -65,7 +65,7 @@
         <dependency>
             <groupId>org.eclipse.paho</groupId>
             <artifactId>org.eclipse.paho.client.mqttv3</artifactId>
-            <version>1.2.2</version>
```
abhipsaMisra, line 68: Question - doesn't device-client take a dependency on deps? Why do we have this added in both?

timtay-microsoft: Good question! We don't need it declared in device client's pom, so I'll get rid of it.

timtay-microsoft:
Trying out this library locally, and I'm seeing some unexpected behavior when running Java 7 or lower still. Don't merge this in yet

timtay-microsoft:
Closing this PR as the upgraded paho version didn't fix any bugs that we had

PR #762, opened 2020-04-21
fix(iot-dev): Fix issue with setOption for setting https settings

timtay-microsoft:
Worked for module clients, but not for device clients


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -598,6 +598,8 @@ else if (value == null)
             // Codes_SRS_DEVICECLIENT_02_016: ["SetMinimumPollingInterval" - time in milliseconds between 2 consecutive polls.]
             case SET_MINIMUM_POLLING_INTERVAL:
             case SET_RECEIVE_INTERVAL:
+            case SET_HTTPS_CONNECT_TIMEOUT:
+            case SET_HTTPS_READ_TIMEOUT:
```
prmathur-microsoft, line 602: How are we verifying this works?

Issue #742, opened 2020-03-25
[Device Client] AMQPS_WS cannot handle large messages


pizerg:
- **OS and version used:** Windows 10 + Ubuntu (18.04) + Raspbian (latest version)

- **Java runtime used:** Oracle's JDK 11 + OpenJDK 11

- **SDK version used:** 1.20.1

# Description of the issue:
When using AMQPS_WS as protocol and trying to send a "big" message (around 12KB after json encoding) the IoT Hub backend closes the connection which translates to a MESSAGE_EXPIRED in the client. 

MQTT_WS protocol handles this scenario without problems.

For the connection on both cases, a connection string is used.

# Code sample exhibiting the issue:
Just start a hub, set any desired callbacks, generate a large enough message and send it.

# Console log of the issue:
```2020-03-25 12:14:33.846 DEBUG AmqpsIotHubConnection:1223 - No gateway hostname was present in config, connecting directly to hub
2020-03-25 12:14:33.919 DEBUG AmqpsIotHubConnection:182 - Opening AMQP connection
2020-03-25 12:14:33.937 DEBUG AmqpsIotHubConnection:1223 - No gateway hostname was present in config, connecting directly to hub
2020-03-25 12:14:34.680 DEBUG AmqpsDeviceOperations:154 - Opening cbs sender link with link correlation id [REMOVED]
2020-03-25 12:14:34.680 DEBUG AmqpsDeviceOperations:176 - Opening cbs receiver link with link correlation id [REMOVED]
2020-03-25 12:14:34.721 DEBUG AmqpsDeviceOperations:396 - cbs sender link with link correlation id [REMOVED] was successfully opened {}
2020-03-25 12:14:34.721 DEBUG AmqpsDeviceOperations:404 - cbs receiver link with link correlation id [REMOVED] was successfully opened
2020-03-25 12:14:34.748 DEBUG AmqpsDeviceAuthenticationCBS:110 - Received amqp message on cbs receiver link with link correlation id [REMOVED] and message correlation id [REMOVED]
2020-03-25 12:14:34.748 DEBUG AmqpsDeviceAuthenticationCBS:153 - CBS authentication message was acknowledged with status 200, authentication for one cbs link pair was successful
2020-03-25 12:14:34.748 DEBUG AmqpsIotHubConnection:998 - Handling a received message with amqp message correlation id [REMOVED]
[...]

PR #767, opened 2020-05-01
fix(iot-dev): Fix bug in amqpws, raise max message size from 1 kb to 16 kb

timtay-microsoft:
Max IoT Hub messsage size is 256 kb, so this is still a bug that you can only send 16 kb, but this is a good incremental change that will benefit customers

related to issue #742, but does not completely solve that issue. We'll need to continue investigating why amqps_ws still cannot send 256 kb messages


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -66,6 +63,7 @@
     //sending messages is done on reactor thread, but we don't want to hog that thread indefinitely, so there is a limit
     // on how many messages to send per reactor callback
     private final static int MAX_MESSAGES_TO_SEND_PER_CALLBACK = 1000;
+    private final static int MAX_MESSAGE_PAYLOAD_SIZE = 256*1000; //max IoT Hub message size is 256 kb, so amqp websocket layer should buffer at least that much space
```
drwill-ms, line 66: Should this be *1024?

timtay-microsoft: It should be, good catch

Issue #766, opened 2020-04-30
Corporate proxy for iot-service-client


alam-R:
- **OS and version used:** Centos 7

- **Java runtime used:** 1.8

- **SDK version used:** 1.21

Hey there,

I couldn't find an answer so my question is:
Is it possible to define an unauthenticated internal proxy for iot-service-client? If so, is there a sample or a reference?

Thanks!

timtay-microsoft:
What features do you need proxy support for on the service client? Is this for registry manager (device identity management) or service client (messaging, file upload notification)?

alam-R:
This is for the registry manager. I was wondering if I could specify my company's internal network when I am trying to connect to my IoT Hub.

timtay-microsoft:
Currently the registry manager does not have proxy support, but I can look into adding it

alam-R:
It would be most helpful for our application which runs behind a corporate proxy, thanks!

alam-R:
Thank you very much @timtay-microsoft for your effort. 
But we will need also proxy support for the service client for messaging. Is this also possible?

timtay-microsoft:
That depends, what kind of proxy do you use, and which protocol do you use for messaging (amqps or amqps_ws)?

timtay-microsoft:
I could feasibly add proxy support for typical HTTP proxies if you are using amqps_ws, similar to how we have it on the device client side.

alam-R:
Yes we are using the websocket version of the AMQPS, so again we would be grateful if you could add support for HTTP proxies.
Thank you!

timtay-microsoft:
Yeah, that's feasible. I'll try to get that feature in soon. Do you need authentication for these proxies?

alam-R:
No, not at the moment at least!

alam-R:
Thank you @timtay-microsoft, I've used the updates and works like a charm!
Do you know when it will be released?

timtay-microsoft:
Glad to hear it! We're hoping to get the release out within the next week

timtay-microsoft:
Proxy support has now been released as of service client version 1.22.0, so I'm closing this issue.

az-iot-builder-01:
[...]

PR #768, opened 2020-05-01
feat(iot-service): Add proxy support for registry manager

timtay-microsoft:
Not for service client, just for registry manager

see #766


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -41,6 +44,10 @@
     private ExecutorService executor;
     private IotHubConnectionString iotHubConnectionString;
 
+    @Setter
```
drwill-ms, line 47: Is there a standard order in which to list these? I'd have assumed get would be first, if there was.

timtay-microsoft: There isn't a standard that I know of, unlike in C#. I can switch them if it's bugging you though

PR #770, opened 2020-05-04
feat(device;service): Support Twin Array properties

azabbasi:


timtay-microsoft:
Don't forget to add new tests that send json arrays as property values for desired and reported properties

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -235,6 +237,10 @@ public void open(Queue<DeviceClientConfig> deviceClientConfigs, ScheduledExecuto
                 {
                     closeConnectionWithException("Timed out waiting for worker links to open", true);
                 }
+
+                // In case this is a recovery from a prematurely closed connection, we will re-queue subscribe messages that were sent within this connection.
+                requeueSubscribeMessages();
```
abhipsaMisra, line 242: Does this resend the device method subscription as well?

azabbasi: no, only Device twin at this point.

azabbasi: from what I understand device methods do not require a message to be sent to subscribe to properties, they just need the link to be opened. but if in the future they require messages to be sent, we will ad them to the subscribe list and send those along with twin subscriptions

abhipsaMisra: Right, that's true!

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -251,6 +257,12 @@ public void open(Queue<DeviceClientConfig> deviceClientConfigs, ScheduledExecuto
         this.log.debug("AMQP connection opened successfully");
     }
 
+    private void requeueSubscribeMessages()
+    {
+        messagesToSend.addAll(subscribeMessages);
+        subscribeMessages.clear();
```
timtay-microsoft, line 263: Is there any reason why we need to save the message, instead of just creating a new one each time? I'm worried about these two queues getting out of sync. I'd feel safer if we always added a new subscribe message to the messagesToSend queue instead of relying on subscribeMessages being in the right state

timtay-microsoft: This layer should know how to build the message, after all

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsDeviceOperations.java
```diff
@@ -427,6 +427,11 @@ protected boolean onLinkRemoteOpen(String linkName)
      */
     abstract protected AmqpsConvertToProtonReturnValue convertToProton(Message message) throws TransportException;
 
+    protected MessageImpl buildSubscribeToDesiredPropertiesProtonMessage()
```
timtay-microsoft, line 430: Since this is a twin only concept, you don't need an abstract method for this

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsDeviceTwin.java
```diff
@@ -191,6 +192,30 @@ protected AmqpsConvertToProtonReturnValue convertToProton(Message message) throw
         }
     }
 
+    @Override
+    protected MessageImpl buildSubscribeToDesiredPropertiesProtonMessage()
+    {
```
timtay-microsoft, line 197: Nice, I like this method. Can you reference this method in the iotHubMessageToProtonMessage method in the same class since it uses this logic, too?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -33,6 +33,7 @@
 import java.util.Queue;
 import java.util.UUID;
 import java.util.concurrent.*;
+import java.util.ArrayList;
```
timtay-microsoft, line 36: "ctrl + shift + o"  can cleanup unused imports

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsDeviceOperations.java
azabbasi: amqpsDeviceOperationMap value is an AmqpDeviceOperation, so I think I need to have it in there right?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionDeviceOperation.java
```diff
@@ -337,6 +337,11 @@ boolean onLinkRemoteOpen(String linkName)
             {
                 if (entry.getValue().onLinkRemoteOpen(linkName))
                 {
+                    if (entry.getKey() == MessageType.DEVICE_TWIN)
+                    {
+                        sendMessage(entry.getValue().buildSubscribeToDesiredPropertiesProtonMessage(), entry.getKey(), deviceClientConfig.getDeviceId());
```
timtay-microsoft, line 342: If you make the '''buildSubscribeToDesiredPropertiesProtonMessage''' method concrete only on the twin class, you can safely cast entry.getValue() as a AmqpsDeviceTwin class to access it here

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsDeviceTwin.java
azabbasi: sure

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsDeviceOperations.java
timtay-microsoft: That map has a mix of classes that extend AmqpDeviceOperation. You can cast the twin one as a AmqpsDeviceTwin instance so that you can access the method that is twin specific

azabbasi: FINE ! I will cast it :)

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionDeviceOperation.java
```diff
@@ -337,6 +337,13 @@ boolean onLinkRemoteOpen(String linkName)
             {
                 if (entry.getValue().onLinkRemoteOpen(linkName))
                 {
+                    if (entry.getKey() == MessageType.DEVICE_TWIN)
```
timtay-microsoft, line 340: actually, you do need a check here that the opened link was a sender link. If you try to send this message when only the receiver link is open, an exception will get thrown. Plus you don't want to send this subscribe message once when the sender link opens, and again once the receiver link opens

PR #773, opened 2020-05-08
feat(iot-service): Add proxy support for service client operations

timtay-microsoft:
Now that registryManager has proxy support, this PR adds proxy support to the AMQP side of the service client, too.

### DIFF
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
@@ -140,8 +140,7 @@ public class ServiceClientTests extends IotHubIntegrationTest
         {
             Proxy testProxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(testProxyHostname, testProxyPort));
             ProxyOptions proxyOptions = new ProxyOptions(testProxy);
-            serviceClientOptions = new ServiceClientOptions();
-            serviceClientOptions.setProxyOptions(proxyOptions);
+            serviceClientOptions = ServiceClientOptions.builder().proxyOptions(proxyOptions).build();
         }
 
         ServiceClient serviceClient = ServiceClient.createFromConnectionString(iotHubConnectionString, testInstance.protocol, serviceClientOptions);
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
@@ -10,10 +10,16 @@ import com.microsoft.azure.sdk.iot.common.helpers.IotHubIntegrationTest;
 import com.microsoft.azure.sdk.iot.common.helpers.StandardTierOnlyRule;
 import com.microsoft.azure.sdk.iot.common.helpers.Tools;
 import com.microsoft.azure.sdk.iot.service.*;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runners.Parameterized;
+import org.littleshoot.proxy.HttpProxyServer;
+import org.littleshoot.proxy.impl.DefaultHttpProxyServer;
 
 import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.net.Proxy;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
@@ -34,6 +40,10 @@ public class ServiceClientTests extends IotHubIntegrationTest
     private static String deviceIdPrefix = "java-service-client-e2e-test";
     private static String content = "abcdefghijklmnopqrstuvwxyz1234567890";
     private static String hostName;
+
+    protected static HttpProxyServer proxyServer;
+    protected static String testProxyHostname = "127.0.0.1";
+    protected static int testProxyPort = 8869;
 
     private static final long MAX_TEST_MILLISECONDS = 1 * 60 * 1000;
 
@@ -75,9 +85,42 @@ public class ServiceClientTests extends IotHubIntegrationTest
         return inputs;
     }
 
+    @BeforeClass
+    public static void startProxy()
+    {
+        proxyServer = DefaultHttpProxyServer.bootstrap()
+                        .withPort(testProxyPort)
+                        .start();
+    }
+
+
+    @AfterClass
+    public static void stopProxy()
+    {
+        proxyServer.stop();
+    }
+
     @Test (timeout=MAX_TEST_MILLISECONDS)
     @ConditionalIgnoreRule.ConditionalIgnore(condition = StandardTierOnlyRule.class)
     public void cloudToDeviceTelemetry() throws Exception
+    {
+        cloudToDeviceTelemetry(false);
+    }
+
+    @Test (timeout=MAX_TEST_MILLISECONDS)
+    @ConditionalIgnoreRule.ConditionalIgnore(condition = StandardTierOnlyRule.class)
+    public void cloudToDeviceTelemetryWithProxy() throws Exception
+    {
+        if (testInstance.protocol != IotHubServiceClientProtocol.AMQPS_WS)
+        {
+            //Proxy support only exists for AMQPS_WS currently
+            return;
+        }
+
+        cloudToDeviceTelemetry(true);
+    }
+
+    public void cloudToDeviceTelemetry(boolean withProxy) throws Exception
     {
         // Arrange
 
@@ -92,7 +135,14 @@ public class ServiceClientTests extends IotHubIntegrationTest
         // Act
 
         // Create service client
-        ServiceClient serviceClient = ServiceClient.createFromConnectionString(iotHubConnectionString, testInstance.protocol);
+        ProxyOptions proxyOptions = null;
+        if (withProxy)
+        {
+            Proxy testProxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(testProxyHostname, testProxyPort));
+            proxyOptions = new ProxyOptions(testProxy);
+        }
+
+        ServiceClient serviceClient = ServiceClient.createFromConnectionString(iotHubConnectionString, testInstance.protocol, proxyOptions);
         CompletableFuture<Void> futureOpen = serviceClient.openAsync();
         futureOpen.get();
 
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
@@ -135,13 +135,14 @@ public class ServiceClientTests extends IotHubIntegrationTest
         // Act
 
         // Create service client
-        ServiceClientOptions serviceClientOptions = null;
+        ProxyOptions proxyOptions = null;
         if (withProxy)
         {
             Proxy testProxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(testProxyHostname, testProxyPort));
-            ProxyOptions proxyOptions = new ProxyOptions(testProxy);
-            serviceClientOptions = ServiceClientOptions.builder().proxyOptions(proxyOptions).build();
+            proxyOptions = new ProxyOptions(testProxy);
         }
+
+        ServiceClientOptions serviceClientOptions = ServiceClientOptions.builder().proxyOptions(proxyOptions).build();
 
         ServiceClient serviceClient = ServiceClient.createFromConnectionString(iotHubConnectionString, testInstance.protocol, serviceClientOptions);
         CompletableFuture<Void> futureOpen = serviceClient.openAsync();
--- a/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
+++ b/iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
@@ -135,14 +135,16 @@ public class ServiceClientTests extends IotHubIntegrationTest
         // Act
 
         // Create service client
-        ProxyOptions proxyOptions = null;
+        ServiceClientOptions serviceClientOptions = null;
         if (withProxy)
         {
             Proxy testProxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(testProxyHostname, testProxyPort));
-            proxyOptions = new ProxyOptions(testProxy);
+            ProxyOptions proxyOptions = new ProxyOptions(testProxy);
+            serviceClientOptions = new ServiceClientOptions();
+            serviceClientOptions.setProxyOptions(proxyOptions);
         }
 
-        ServiceClient serviceClient = ServiceClient.createFromConnectionString(iotHubConnectionString, testInstance.protocol, proxyOptions);
+        ServiceClient serviceClient = ServiceClient.createFromConnectionString(iotHubConnectionString, testInstance.protocol, serviceClientOptions);
         CompletableFuture<Void> futureOpen = serviceClient.openAsync();
         futureOpen.get();
 
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
@@ -56,8 +56,6 @@ public class FeedbackReceiver extends Receiver
             throw new IllegalArgumentException("deviceId cannot be null or empty");
         }
         
-               
-        
         // Codes_SRS_SERVICE_SDK_JAVA_FEEDBACKRECEIVER_12_002: [The constructor shall store deviceId]
         this.deviceId = deviceId;
         // Codes_SRS_SERVICE_SDK_JAVA_FEEDBACKRECEIVER_12_003: [The constructor shall create a new instance of AmqpReceive object]
@@ -72,11 +70,26 @@ public class FeedbackReceiver extends Receiver
      * @param userName The iot hub user name
      * @param sasToken The iot hub SAS token for the given device
      * @param iotHubServiceClientProtocol protocol to be used
-     * 
+     *
      */
     public FeedbackReceiver(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol)
     {
-        // Codes_SRS_SERVICE_SDK_JAVA_FEEDBACKRECEIVER_12_001: [The constructor shall throw IllegalArgumentException if any the input string is null or empty]
+        this(hostName, userName, sasToken, iotHubServiceClientProtocol, (ProxyOptions) null);
+    }
+
+    /**
+     * Constructor to verify initialization parameters
+     * Create instance of AmqpReceive
+     *
+     * @param hostName The iot hub host name
+     * @param userName The iot hub user name
+     * @param sasToken The iot hub SAS token for the given device
+     * @param iotHubServiceClientProtocol protocol to be used
+     * @param proxyOptions the proxy options to tunnel through, if a proxy should be used.
+     * 
+     */
+    public FeedbackReceiver(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol, ProxyOptions proxyOptions)
+    {
         if (Tools.isNullOrEmpty(hostName))
         {
             throw new IllegalArgumentException("hostName cannot be null or empty");
@@ -96,7 +109,7 @@ public class FeedbackReceiver extends Receiver
         }
                 
         // Codes_SRS_SERVICE_SDK_JAVA_FEEDBACKRECEIVER_12_003: [The constructor shall create a new instance of AmqpReceive object]
-        this.amqpReceive = new AmqpReceive(hostName, userName, sasToken, iotHubServiceClientProtocol);
+        this.amqpReceive = new AmqpReceive(hostName, userName, sasToken, iotHubServiceClientProtocol, proxyOptions);
     }
         
     /**
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotificationReceiver.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotificationReceiver.java
@@ -30,7 +30,20 @@ public class FileUploadNotificationReceiver extends Receiver
      */
     FileUploadNotificationReceiver(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol)
     {
-        // Codes_SRS_SERVICE_SDK_JAVA_FILEUPLOADNOTIFICATIONRECEIVER_25_001: [** The constructor shall throw IllegalArgumentException if any the input string is null or empty **]**
+        this(hostName, userName, sasToken, iotHubServiceClientProtocol, null);
+    }
+
+    /**
+     * Constructor to verify initialization parameters
+     * Create instance of AmqpReceive
+     * @param hostName The iot hub host name
+     * @param userName The iot hub user name
+     * @param sasToken The iot hub SAS token for the given device
+     * @param iotHubServiceClientProtocol The iot hub protocol name
+     * @param proxyOptions the proxy options to tunnel through, if a proxy should be used.
+     */
+    FileUploadNotificationReceiver(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol, ProxyOptions proxyOptions)
+    {
         if (Tools.isNullOrEmpty(hostName))
         {
             throw new IllegalArgumentException("hostName cannot be null or empty");
@@ -49,7 +62,7 @@ public class FileUploadNotificationReceiver extends Receiver
         }
 
         // Codes_SRS_SERVICE_SDK_JAVA_FILEUPLOADNOTIFICATIONRECEIVER_25_002: [** The constructor shall create a new instance of AmqpFileUploadNotificationReceive object **]**
-        this.amqpFileUploadNotificationReceive = new AmqpFileUploadNotificationReceive(hostName, userName, sasToken, iotHubServiceClientProtocol);
+        this.amqpFileUploadNotificationReceive = new AmqpFileUploadNotificationReceive(hostName, userName, sasToken, iotHubServiceClientProtocol, proxyOptions);
     }
 
     /**
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
@@ -1,0 +1,67 @@
+package com.microsoft.azure.sdk.iot.service;
+
+import lombok.Getter;
+
+import java.net.InetSocketAddress;
+import java.net.Proxy;
+
+/**
+ * The settings supported by this SDK when communicating to IoT Hub through a proxy. HTTP proxies are supported by this SDK,
+ * but only if the proxy does not require authentication.
+ */
+public class ProxyOptions
+{
+    @Getter
+    private Proxy proxy;
+
+    /**
+     * Create the proxy options for connecting to a proxy
+     * @param proxy the proxy to communicate through
+     */
+    public ProxyOptions(Proxy proxy)
+    {
+        if (proxy == null)
+        {
+            throw new IllegalArgumentException("Proxy cannot be null");
+        }
+
+        if (proxy.type() != Proxy.Type.HTTP)
+        {
+            throw new IllegalArgumentException("Service proxy options only support using HTTP proxies");
+        }
+
+        this.proxy = proxy;
+    }
+
+    /**
+     * @return the saved proxy hostname
+     */
+    public String getHostName()
+    {
+        if (this.proxy.address() instanceof InetSocketAddress)
+        {
+            InetSocketAddress addr = (InetSocketAddress) this.proxy.address();
+            return addr.getHostName();
+        }
+        else
+        {
+            //Should only happen if proxy is of type DIRECT, which isn't supported
+            throw new UnsupportedOperationException("Unsupported proxy type, could not get host name. Proxy address was not an instance of java.net.InetSocketAddress.");
+        }
+    }
+
+    public int getPort()
+    {
+        if (this.proxy.address() instanceof InetSocketAddress)
+        {
+            InetSocketAddress addr = (InetSocketAddress) this.proxy.address();
+            return addr.getPort();
+        }
+        else
+        {
+            //Should only happen if proxy is of type DIRECT, which isn't supported
+            throw new UnsupportedOperationException("Unsupported proxy type, could not get port. Proxy address was not an instance of java.net.InetSocketAddress.");
+        }
+    }
+}
+
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
@@ -23,6 +23,7 @@ import javax.json.JsonObject;
 import javax.json.JsonReader;
 import java.io.IOException;
 import java.io.StringReader;
+import java.net.Proxy;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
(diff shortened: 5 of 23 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -44,9 +41,7 @@
     private ExecutorService executor;
     private IotHubConnectionString iotHubConnectionString;
 
-    @Getter
-    @Setter
-    private Proxy proxy;
```
timtay-microsoft, line 49: Note that this would be a breaking change, but this getter/setter hasn't been in a released jar file yet, so it is still safe to change

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
```diff
@@ -0,0 +1,67 @@
+package com.microsoft.azure.sdk.iot.service;
+
+import lombok.Getter;
+
+import java.net.InetSocketAddress;
+import java.net.Proxy;
+
+/**
+ * The settings supported by this SDK when communicating to IoT Hub through a proxy. Only HTTP Proxies are supported.
+ * Authentication with the HTTP proxy is not supported.
```
abhipsaMisra, line 10: I am not sure what this means:
Only HTTP Proxies are supported. Authentication with the HTTP proxy is not supported.

timtay-microsoft: It means you can use HTTP proxies, but only if that HTTP proxy doesn't require authentication. I can reword this

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
```diff
@@ -0,0 +1,67 @@
...
+    {
+        if (this.proxy.address() instanceof InetSocketAddress)
+        {
+            InetSocketAddress addr = (InetSocketAddress) this.proxy.address();
+            return addr.getHostName();
+        }
+        else
+        {
+            //Should only happen if proxy is of type DIRECT, which isn't supported
+            throw new UnsupportedOperationException("Could not get port from saved proxy");
```
abhipsaMisra, line 49: text needs to be: Could not get ~port~ hostname from saved proxy

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -76,9 +102,15 @@ protected ServiceClient(IotHubConnectionString iotHubConnectionString, IotHubSer
         this.userName = iotHubConnectionString.getUserString();
         this.sasToken = iotHubServiceSasToken.toString();
         this.iotHubServiceClientProtocol = iotHubServiceClientProtocol;
+        this.proxyOptions = proxyOptions;
+
+        if (this.proxyOptions != null && this.proxyOptions.getProxy() != null && this.iotHubServiceClientProtocol != IotHubServiceClientProtocol.AMQPS_WS)
```
abhipsaMisra, line 107: This is a protected method call, `this.proxyOptions != null` check should be one level higher. Also, having proxy options null over amqp_ws will throw an exception with the message saying `Proxies are only supported over AMQPS_WS`, which is then confusing.

abhipsaMisra, line 107: Is `this.proxyOptions.getProxy()` check required? There is only 1 public constructor for initializing the `ProxyOptions` class, and that takes in a `Proxy` object.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -38,6 +40,19 @@
...
+
+    /**
+     * Create ServiceClient from the specified connection string
+     * @param iotHubServiceClientProtocol  protocol to use
+     * @param connectionString The connection string for the IotHub
+     * @param proxyOptions The proxy options to use when connecting to the service. May be null if no proxy will be used.
+     * @return The created ServiceClient object
+     * @throws IOException This exception is thrown if the object creation failed
+     */
+    public static ServiceClient createFromConnectionString(String connectionString, IotHubServiceClientProtocol iotHubServiceClientProtocol, ProxyOptions proxyOptions) throws IOException
```
abhipsaMisra, line 55: What is the preferable way to set transport configurations for this SDK? Do we take in any transport settings? How do we set the keep-alive time etc?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
timtay-microsoft: Good point, just checking if proxyoptions is null should be sufficient here and everywhere else

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
timtay-microsoft: We currently don't have any other options beyond proxy settings. I don't see us extending other options like the C# SDK does right now. I see that as something we can address when this SDK goes track 2 instead.

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
```diff
@@ -75,9 +85,42 @@ public static Collection inputsCommon() throws IOException
         return inputs;
     }
 
+    @BeforeClass
+    public static void startProxy()
+    {
+        proxyServer = DefaultHttpProxyServer.bootstrap()
+                .withPort(testProxyPort)
```
drwill-ms, line 92: are 8 spaces of indentation normal for Java?

timtay-microsoft: Not quite, I can fix this

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
```diff
@@ -0,0 +1,67 @@
...
+            throw new IllegalArgumentException("Service proxy options only support using HTTP proxies");
+        }
+
+        this.proxy = proxy;
+    }
+
+    /**
+     * @return the saved proxy hostname
+     */
+    public String getHostname()
```
drwill-ms, line 39: InetSocketAddress uses a capital N in getHostName. Not sure if that is the preferred casing to stick with and match for this method.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
timtay-microsoft: No one has asked for any settings beyond proxy support, for reference

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
timtay-microsoft: Ah, good catch. We do use HostName and Hostname very interchangeably in this SDK. HostName seems more appropriate considering it is a pair of words

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
```diff
@@ -0,0 +1,67 @@
...
+    {
+        if (this.proxy.address() instanceof InetSocketAddress)
+        {
+            InetSocketAddress addr = (InetSocketAddress) this.proxy.address();
+            return addr.getPort();
+        }
+        else
+        {
+            //Should only happen if proxy is of type DIRECT, which isn't supported
+            throw new UnsupportedOperationException("Could not get port from saved proxy");
```
drwill-ms, line 63: Perhaps message should be "Unsupported proxy type" then

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -1604,7 +1612,7 @@ private JobProperties ProcessJobResponse(HttpResponse response) throws IotHubExc
 
     private HttpRequest CreateRequest(URL url, HttpMethod method, byte[] payload, String sasToken) throws IOException
     {
-        HttpRequest request = new HttpRequest(url, method, payload, proxy);
+        HttpRequest request = new HttpRequest(url, method, payload, this.proxyOptions != null ? this.proxyOptions.getProxy(): null);
```
drwill-ms, line 1615: space before : ?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -30,6 +30,8 @@
     protected IotHubConnectionString iotHubConnectionString;
     private IotHubServiceClientProtocol iotHubServiceClientProtocol;
 
+    private ProxyOptions proxyOptions;
```
drwill-ms, line 33: I don't see where this is being set. We seem to pass it into ServiceClient ctor though.

drwill-ms: And AmqpSend

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
```diff
@@ -0,0 +1,67 @@
...
+public class ProxyOptions
+{
+    @Getter
+    private Proxy proxy;
+
+    /**
+     * Create the proxy options for connecting to a proxy
+     * @param proxy the proxy to communicate through
+     */
+    public ProxyOptions(Proxy proxy)
```
abhipsaMisra, line 21: More of a question that PR comments: Does java support only setting proxies explicitly via the application, or does it also automatically tunnel traffic through a proxy set on the OS directly?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
```diff
@@ -75,9 +85,42 @@ public static Collection inputsCommon() throws IOException
...
+    }
+
+    @Test (timeout=MAX_TEST_MILLISECONDS)
+    @ConditionalIgnoreRule.ConditionalIgnore(condition = StandardTierOnlyRule.class)
+    public void cloudToDeviceTelemetryWithProxy() throws Exception
+    {
+        if (testInstance.protocol != IotHubServiceClientProtocol.AMQPS_WS)
+        {
+            //Proxy support only exists for AMQPS_WS currently
+            return;
```
prmathur-microsoft, line 117: What are the hurdles with AMQP?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
abhipsaMisra: What I mean is, if I have both set, who wins?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
```diff
@@ -36,7 +36,7 @@
      * @param iotHubServiceClientProtocol The iot hub protocol name
      * @param deviceId The device id
      */
-    public @Deprecated FeedbackReceiver(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol, String deviceId)
+    public @Deprecated FeedbackReceiver(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol, String deviceId, ProxyOptions proxyOptions)
```
prmathur-microsoft, line 39: Why are we adding support to deprecated API?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
```diff
@@ -74,7 +72,7 @@
      * @param iotHubServiceClientProtocol protocol to be used
      * 
      */
-    public FeedbackReceiver(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol)
+    public FeedbackReceiver(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol, ProxyOptions proxyOptions)
```
prmathur-microsoft, line 75: Breaking change?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotificationReceiver.java
```diff
@@ -28,7 +28,7 @@
      * @param sasToken The iot hub SAS token for the given device
      * @param iotHubServiceClientProtocol The iot hub protocol name
      */
-    FileUploadNotificationReceiver(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol)
+    FileUploadNotificationReceiver(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol, ProxyOptions proxyOptions)
     {
```
prmathur-microsoft, line 32: Isn't this a breaking change?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
timtay-microsoft: Java doesn't allow you to force a proxy like that at the OS level as far as I know. Since we use transport libraries for HTTP and AMQP, we have to set the proxy at the http/amqp client level

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
timtay-microsoft: It may be possible, but no one has asked for this yet. The customer I'm writing this for just wants AMQPS_WS to support tunneling through HTTP proxies

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
timtay-microsoft: Good question, give me a sec to check

service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/FeedbackReceiverTest.java
```diff
@@ -36,7 +37,7 @@ public void constructor_input_hostName_null() throws Exception
         String deviceId = "xxx";
         IotHubServiceClientProtocol iotHubServiceClientProtocol = IotHubServiceClientProtocol.AMQPS;
         // Act
-        FeedbackReceiver feedbackReceiver = new FeedbackReceiver(hostName, userName, sasToken, iotHubServiceClientProtocol, deviceId);
+        FeedbackReceiver feedbackReceiver = new FeedbackReceiver(hostName, userName, sasToken, iotHubServiceClientProtocol, deviceId, null);
```
drwill-ms, line 40: If you have to pass null as an additional parameter, is that a breaking change?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
prmathur-microsoft, line 55: Why not group this under connection options which has protocol, proxyOptions etc and lets you add future connection options?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
drwill-ms: What is normal?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
timtay-microsoft: I can roll back this change actually. I'll just need to throw an unsupportedOperationException if the user tries to make proxy support work with these deprecated APIs

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
timtay-microsoft: Good catch, here and on the other sub-service client classes. I'll add an overload constructor that takes the proxy options instead

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotificationReceiver.java
timtay-microsoft: As per above, I'll add an overload to take the proxy instead

service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/FeedbackReceiverTest.java
timtay-microsoft: Looks like it, yeah. I've added an overloaded constructor so that it won't break now

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
timtay-microsoft: That depends on how long we plan on supporting this particular SDK. I didn't envision bringing any other options here since we will be track 2 soon. But what do you think?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/serviceclient/ServiceClientTests.java
timtay-microsoft: One tab over from the method that started the fluent chain:

Although github won't let me demonstrate exactly how this looks, the latest view of this file should show you.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
abhipsaMisra: I see, it's better to start off with a config file, rather than placing them in the client initializer; but I agree, now that we are this close to going track 2, we should do it there directly.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
abhipsaMisra: So if I understand correctly, if my OS would have some default proxy configured, that would never be honored?

timtay-microsoft: Yes, hence why customer's want proxy support on this SDK, instead of just setting it on their own devices

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
prmathur-microsoft: I don't forsee either (As it is hard to see for 10 years down the line). I would still prefer to group because we have to support this for 10 years even after we move to track-2 ;just to be on safer side.

timtay-microsoft: Do you picture a class "ConnectionOptions" being reused from AMQP to HTTP, though? Or do you picture an HttpConnectionOptions class for registry manager, and an AmqpConnectionOptions class for service client, where each class has just the proxy to set?

timtay-microsoft: Well now I am building RegistryManagerOptions and ServiceClientOptions classes as per Priyanka, in case you are still interested in this point

timtay-microsoft: I pushed a commit that shows one way to have a more general "RegistryManagerOptions" and "ServiceClientOptions" class structure. Can you take a look?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
abhipsaMisra: Interesting!

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
```diff
@@ -0,0 +1,67 @@
...
+    {
+        if (proxy == null)
+        {
+            throw new IllegalArgumentException("Proxy cannot be null");
+        }
+
+        if (proxy.type() != Proxy.Type.HTTP)
```
abhipsaMisra, line 28: Q - this is for proxy support over Amqp_ws, then why don't we use socks proxy?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -1604,7 +1612,7 @@ private JobProperties ProcessJobResponse(HttpResponse response) throws IotHubExc
 
     private HttpRequest CreateRequest(URL url, HttpMethod method, byte[] payload, String sasToken) throws IOException
     {
-        HttpRequest request = new HttpRequest(url, method, payload, proxy);
+        HttpRequest request = new HttpRequest(url, method, payload, this.options != null && this.options.getProxyOptions() != null ? this.options.getProxyOptions().getProxy() : null);
```
abhipsaMisra, line 1615: there is a whole lot happening here, can we move this to the line above:
this.options != null && this.options.getProxyOptions() != null ? this.options.getProxyOptions().getProxy() : null

abhipsaMisra: The way I see, registry manager options should never be null, but have some default initial value (proxy = null)

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
```diff
@@ -0,0 +1,17 @@
...
+ * Configurable options for all registry manager operations
+ */
+public class RegistryManagerOptions
+{
+    /**
+     * The options that specify what proxy to tunnel through. If null, no proxy will be used
+     */
+    @Getter
+    @Setter
+    private ProxyOptions proxyOptions;
```
abhipsaMisra, line 16: Why not add `@AllArgsConstructor` for 1 line initialization.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
timtay-microsoft: The customer who prompted all this only needs support for HTTP proxies. I don't know offhand if amqps or amqps_ws work with socks proxies, or if proton-j has support for them

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
timtay-microsoft: Yeah, good call. This is pretty busy

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
timtay-microsoft: I don't want this to be construed as an essential option. As such, there should be a constructor that doesn't take this

timtay-microsoft: For instance, if we ever add something like HTTP readTimeout/connectTimeout, I wouldn't want customers to need a proxy here

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -76,9 +102,15 @@ protected ServiceClient(IotHubConnectionString iotHubConnectionString, IotHubSer
...
+        this.options = options;
+
+        if (this.options != null && this.options.getProxyOptions() != null && this.iotHubServiceClientProtocol != IotHubServiceClientProtocol.AMQPS_WS)
+        {
+            throw new UnsupportedOperationException("Proxies are only supported over AMQPS_WS");
+        }
 
         // Codes_SRS_SERVICE_SDK_JAVA_SERVICECLIENT_12_007: [The constructor shall create a new instance of AmqpSend object]
-        this.amqpMessageSender = new AmqpSend(hostName, userName, sasToken, this.iotHubServiceClientProtocol);
+        this.amqpMessageSender = new AmqpSend(hostName, userName, sasToken, this.iotHubServiceClientProtocol, options != null ? options.getProxyOptions() : null);
```
abhipsaMisra, line 113: same, here as well

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
abhipsaMisra: yeah, good point.

abhipsaMisra: But I still feel the options object in itself should not be null, but have a default value

timtay-microsoft: I'm not sure I see the value in holding a non-null default proxyOptions class. Each of those null checks would just turn into checks on if the proxy is of type DIRECT (no proxy) or type HTTP.

timtay-microsoft: For AMQP in particular, it isn't just a straight pass through. We have to add a transport layer manually if there is a proxy to go through

timtay-microsoft: What about a builder pattern here instead? Since all of the hypothetical fields in this class are optional, and we don't know how many options there may be in the future, a builder could fit in well here

abhipsaMisra: Suppose tomorrow we add a parameter for Http connect timeout, which has a default value of 1min, but settable. So now, the `RegistryManagerOptions` class is always non-null (irrespective of proxy). 
So something like this, which we have below:
`options != null ? options.getProxyOptions() : null`
would want to get proxy options, even though we did not set it/ want to use it.

abhipsaMisra: Yeah, a builder would work here. We can set @NonNull for required params

timtay-microsoft: Okay, I pushed a commit that makes these two options classes builder based

timtay-microsoft: I can make the RegistryManagerOptions and ServiceClientOptions always non-null, for simplicity, sure. The default proxyOptions will still be null, though

abhipsaMisra: Yes, of course, we don't want to force/ have a default for proxy

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -229,6 +261,11 @@ public void send(String deviceId, String moduleId, Message message) throws IOExc
      */
     @Deprecated public FeedbackReceiver getFeedbackReceiver(String deviceId)
     {
+        if (options != null)
```
abhipsaMisra, line 264: this now needs to be `options.getProxyOptions() != null`

timtay-microsoft: Good catch, you are right

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -44,9 +42,16 @@
...
-    @Getter
-    @Setter
-    private Proxy proxy;
+    private RegistryManagerOptions options;
+
+    /**
+     * As of release 1.22.0, replaced by {@link #createFromConnectionString()}
+     */
+    @Deprecated
+    public RegistryManager()
```
abhipsaMisra, line 51: what is this for? did we allow creation of registry manager without having to pass in even a connection string?

timtay-microsoft: Java has a funny quirk where a class with no public constuctor automatically has a public constructor. This commit just makes that constuctor visible. I believe it was an accident that we didn't make the only constuctor protected

abhipsaMisra: Ah, the default constructor, yeah. Can we not make it private now? Also, can we modify the comment to say that this is simply the default java constructor. the `replaced` text makes it seem that we did allow a no args initialization of registry manager.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
```diff
@@ -0,0 +1,18 @@
+package com.microsoft.azure.sdk.iot.service;
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
```
abhipsaMisra, line 5: nit: no longer required: ctrl+shift+o :)

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClientOptions.java
```diff
@@ -0,0 +1,18 @@
+package com.microsoft.azure.sdk.iot.service;
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
```
abhipsaMisra, line 5: same, no longer required

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
abhipsaMisra: I see, what does our device client currently use?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
timtay-microsoft: Since it was a public constructor, we can't just make it private now. That would be a breaking change. We did technically allow users to build a registry manager through that constructor. I'll add to the header here that it was once the default constructor, though

abhipsaMisra: Oh, I am confused, if this worked, why are we deprecating this now? It is because it doesn't really do anything useful?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ProxyOptions.java
timtay-microsoft: HTTP proxies work for device clients using MQTT_WS, AMQPS_WS, and HTTP. In theory, SOCKs proxies work for device clients using MQTT

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -38,18 +40,36 @@
...
+        {
+            throw new IllegalArgumentException("ServiceClientOptions cannot be null for this constructor");
+        }
+
         // Codes_SRS_SERVICE_SDK_JAVA_SERVICECLIENT_12_002: [The constructor shall create IotHubConnectionString object using the IotHubConnectionStringBuilder]
         IotHubConnectionString iotHubConnectionString = IotHubConnectionStringBuilder.createConnectionString(connectionString);
 
         // Codes_SRS_SERVICE_SDK_JAVA_SERVICECLIENT_12_003: [The constructor shall create a new instance of ServiceClient using the created IotHubConnectionString object and return with it]
-        ServiceClient iotServiceClient = new ServiceClient(iotHubConnectionString, iotHubServiceClientProtocol);
+        ServiceClient iotServiceClient = new ServiceClient(iotHubConnectionString, iotHubServiceClientProtocol, options);
```
prmathur-microsoft, line 72: Does service client works on anything other than AMQP ? If I re-call it doesn't. Then what purpose does iotHubServiceClientProtocol serve?

timtay-microsoft: It can be either AMQPS or AMQPS_WS

PR #775, opened 2020-05-12
fix(iot-dev): Fix issue where MQTT didn't url encode application properties correctly

timtay-microsoft:
URLEncoder.Encode encodes the space character ' ' as '+' instead of the expected '%20', so I've added a manual replace step for this character. This is most important for MQTT because that protocol doesn't allow topic strings with the '+' character

Also removing all character validation for application property keys and values now that this MQTT bug is fixed. There is no longer a device-side reason to validate these values as all protocols are encoding them correctly now such that the protocol doesn't throw an exception. Iot Hub will be the validator, not the client.

Hub does allow these previously banned characters, provided they are encoded correctly for each protocol.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MessageProperty.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MessageProperty.java
@@ -31,12 +31,6 @@ public final class MessageProperty {
     public static final String IOTHUB_SECURITY_INTERFACE_ID = "iothub-interface-id";
     public static final String IOTHUB_SECURITY_INTERFACE_ID_VALUE = "urn:azureiot:Security:SecurityAgent:1";
 
-    //Note this list includes the characters for tab and space.
-    private static final String BANNED_NON_ALPHANUMERIC_CHARACTERS = "()<>@,;:\"\\[]?={} \t";
-    private static final int ASCII_LOWER_BOUND = 32;
-    private static final int ASCII_UPPER_BOUND = 127;
-
-
     static {
         HashSet<String> reservedPropertyNames = new HashSet<>();
         reservedPropertyNames.add("iothub-enqueuedtime");
@@ -92,22 +86,9 @@ public final class MessageProperty {
             throw new IllegalArgumentException("Property argument 'value' cannot be null.");
         }
 
-        // Codes_SRS_MESSAGEPROPERTY_11_002: [If the name contains a character that is not in US-ASCII, the function shall throw an IllegalArgumentException.]
-        if (!usesValidChars(name)) {
-            String errMsg = String.format("%s is not a valid IoT Hub message property name. Characters in property keys must fall within ASCII value range %d and %d but excluding the set %s", name, ASCII_LOWER_BOUND, ASCII_UPPER_BOUND, BANNED_NON_ALPHANUMERIC_CHARACTERS);
-            throw new IllegalArgumentException(errMsg);
-        }
-
         // Codes_SRS_MESSAGEPROPERTY_11_008: [If the name is a reserved property name, the function shall throw an IllegalArgumentException.]
         if (RESERVED_PROPERTY_NAMES.contains(name)) {
             String errMsg = String.format("%s is a reserved IoT Hub message property name.%n", name);
-            throw new IllegalArgumentException(errMsg);
-        }
-
-        // Codes_SRS_MESSAGEPROPERTY_11_003: [If the value contains a character that is not in US-ASCII, the function shall throw an IllegalArgumentException.]
-        if (!usesValidChars(value))
-        {
-            String errMsg = String.format("%s is not a valid IoT Hub message property value. Characters in property values must fall within ASCII value range %d and %d but excluding the set %s", value, ASCII_LOWER_BOUND, ASCII_UPPER_BOUND, BANNED_NON_ALPHANUMERIC_CHARACTERS);
             throw new IllegalArgumentException(errMsg);
         }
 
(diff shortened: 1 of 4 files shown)


device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/MessagePropertyTest.java
```diff
@@ -131,20 +111,6 @@ public void isValidAppPropertyReturnsTrueForValidAppProperty()
         assertThat(testIsValidAppProperty, is(expectedIsValidAppProperty));
     }
 
-    // Tests_SRS_MESSAGEPROPERTY_11_007: [The function shall return true if and only if the name and value only use characters in: US-ASCII and the name is not a reserved property name.]
-    @Test
-    public void isValidAppPropertyReturnsFalseForInvalidAppProperty()
-    {
-        final String name = "test-name";
-        final String illegalValue = "परीक्षण"; // Unicode is not supported in MessageProperty value
```
abhipsaMisra, line 139: whoa, who added hindi in here!

timtay-microsoft: Haha, I'm not sure! @prmathur-microsoft perhaps?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MessageProperty.java
```diff
@@ -163,17 +144,9 @@ public boolean hasSameName(String name) {
...
-        if (!RESERVED_PROPERTY_NAMES.contains(name)
-                && usesValidChars(name)
-                && usesValidChars(value)) {
-            propertyIsValid = true;
-        }
-
-        return propertyIsValid;
+    public static boolean isValidAppProperty(String name, String value)
+    {
+        return !RESERVED_PROPERTY_NAMES.contains(name);
```
abhipsaMisra, line 149: Do we need to validate even this?

timtay-microsoft: Maybe not, but that kind of change is a bit out of scope for the customer issue I'm working to solve. Maybe if a customer complains later, we can re-evaluate

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttMessaging.java
```diff
@@ -154,15 +154,15 @@ private boolean appendPropertyIfPresent(StringBuilder stringBuilder, boolean sep
 
                 if (isApplicationProperty)
                 {
-                    stringBuilder.append(URLEncoder.encode(propertyKey, StandardCharsets.UTF_8.name()));
+                    stringBuilder.append(URLEncoder.encode(propertyKey, StandardCharsets.UTF_8.name()).replaceAll("\\+", "%20"));
```
abhipsaMisra, line 157: What's the limitation with `URLEncoder` being unable to encode `+`?

timtay-microsoft: https://www.urlencoder.io/java/

timtay-microsoft: "Note that Java’s URLEncoder class encodes space character(" ") into a + sign. This is contrary to other languages like Javascript that encode space character into %20."

timtay-microsoft: Just updated my explanation above, the problem is actually that URLEncoder.Encode(...) encodes spaces as '+' which is unusual for url encoding. For MQTT in particular, you can't have '+' characters in the topic string, but you can have '%20'

abhipsaMisra: So are we encoding space or "+"?

timtay-microsoft: We are fixing how ' ' gets encoded by replacing the incorrect '+' with the correct '%20'

For example:
```
String urlEncoded = URLEncoder.encode("This string has spaces", StandardCharsets.UTF_8.displayName());
String actuallyUrlEncoded = urlEncoded.replaceAll("\\+", "%20");
```

urlEncoded = "This+string+has+spaces"
actuallyUrlEncoded = "This%20string%20has%20spaces"

timtay-microsoft: C# does this correctly: https://docs.microsoft.com/en-us/dotnet/api/system.web.httputility.urlencode?view=netcore-3.1

abhipsaMisra: yes, this is what I was looking for: We are fixing how ' ' gets encoded by replacing the incorrect '+' with the correct '%20'
The initial wording of the PR confused me.

I would suggest adding this as a comment where the manual replace is done.

timtay-microsoft: Sure thing, I can add some comments to clarify this .replaceAll call

Issue #772, opened 2020-05-08
Java SDK 1.21.1 - startDeviceTwin() method does not provide changes in Desired properties and issue with its JavaDoc


cursu:
- **OS and version used:** <VERSION> 
Windows 10 - Build 19041

- **Java runtime used:** <VERSION> 
PS C:\apps\IntelliJIDEA\jbr\bin> .\java.exe --version
openjdk 11.0.5 2019-10-15
OpenJDK Runtime Environment (build 11.0.5+10-b520.38)
OpenJDK 64-Bit Server VM (build 11.0.5+10-b520.38, mixed mode, sharing)

- **SDK version used:** <VERSION> 
SDK 1.21.1

# Description of the issue:

When a backend service changes one of the Desired properties for a certain device connected to an IoT Hub, when the device runs client.startDeviceTwin() it's supposed to receive only the changes in the Desired properties like described in the JavaDoc from here - https://docs.microsoft.com/en-us/java/api/com.microsoft.azure.sdk.iot.device.deviceclient.startdevicetwin?view=azure-java-stable  
"genericPropertyCallBack  PropertyCallBack<Type1,Type2>
the PropertyCallBack callback for providing any changes in desired properties. Cannot be null" 

In reality, the behavior of startDeviceTwin() looks that it provides all Desired and previously Reported properties for that specific device. For the Desired properties, it seems to provide a wrong version as well. This is the bug reported by this issue + the fact that the JavaDoc for method startDeviceTwin() should be changed to describe that Reported properties are received as well alongside Desired properties.

To reproduce this issue I used the following:

1. Device side - AMQPS protocol - client.startDeviceTwin() 
https://github.com/Azure/azure-iot-sdk-java/blob/master/device/iot-device-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java

Modified it so that I'm only using:
client.startDeviceTwin(new DeviceTwinStatusCallBack(), null, new onProperty(), null);

2. Service backend side - I only used changeDesiredProperties(twinClient, device); to change a desired property on a simulated device for an IoTHub:
https://github.com/Azure/azure-iot-sdk-java/blob/master/service/iot-service-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java

After running this service backend and changing only Desired property "TEST2", the twin from the IoT Hub looks like following:

{
  "deviceId": "device4",
[...]

PR #776, opened 2020-05-12
doc(iot-dev): Add clarification to startDeviceTwin() API docs

timtay-microsoft:
See #772 

Customer's need a bit more help understanding how startDeviceTwin works, so I've beefed up the javadocs for it

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
@@ -463,7 +463,16 @@ public final class DeviceClient extends InternalClient implements Closeable
     }
 
     /**
-     * Starts the device twin.
+     * Starts the device twin. This device client will receive a callback with the current state of the full twin, including
+     * reported properties and desired properties. After that callback, this device client will receive a callback
+     * each time a desired property is updated. That callback will either contain the full desired properties set, or
+     * only the updated desired property depending on how the desired property was changed. IoT Hub supports a PUT and a PATCH
+     * on the twin, and the PATCH will cause this device client to receive the full desired properties set. Similarly, the version
+     * of each desired property will be incremented from a PATCH call, and only the actually updated desired property will
+     * have its version incremented from a PUT call.
+     *
+     * See <a href="https://docs.microsoft.com/en-us/rest/api/iothub/service/twin/replacedevicetwin">PUT</a> and
+     * <a href="https://docs.microsoft.com/en-us/rest/api/iothub/service/twin/updatedevicetwin">PATCH</a>
      *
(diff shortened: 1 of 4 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -463,7 +463,16 @@ public void getDeviceTwin() throws IOException
     }
 
     /**
-     * Starts the device twin.
+     * Starts the device twin. This device client will receive a callback with the current state of the full twin, including
+     * reported properties and desired properties. After that callback, this device client will receive a callback
+     * each time a desired property is updated. That callback will either contain the full desired properties set, or
+     * only the updated desired property depending on how the desired property was changed. IoT Hub supports a PUT and a PATCH
+     * on the twin, and the PATCH will cause this device client to receive the full desired properties set. Similarly, the version
```
abhipsaMisra, line 470: Shouldn't PATCH return only the patched update?

abhipsaMisra: From the github issue:

```
Initial desired state:
{ “a” : 1, “b” : 2 }

If customer sends PATCH { “a” : -1} device will receive { “a” : -1 }. If customer sends PATCH { “a”: 1, “b”: 2 } device will receive { “a”: 1, “b”: 2 }

Having said that, if customer sends PUT { “a” : 1, “c” : 100 }, device will get { “a”: 1, “b”: null, “c” : 100 }
```

timtay-microsoft: Interesting, that is what the service team said. I double checked to see how our SDK does desired property updates, and it is a PATCH that causes the device side to get the full desired property update. However, it looks like our service SDK sends the full desired properties set even when patching just one property. That's why our SDK's PATCH behaved like the PUT.

I'll update the javadoc here and the issue accordingly

abhipsaMisra:
You should add your finding to the documentation for property update as well (that the API we call is a PATCH, but the SDK will send the entire twin object, instead of sending only the patch object)

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -463,7 +463,19 @@ public void getDeviceTwin() throws IOException
     }
 
     /**
-     * Starts the device twin.
+     * Starts the device twin. This device client will receive a callback with the current state of the full twin, including
+     * reported properties and desired properties. After that callback, this device client will receive a callback
```
abhipsaMisra, line 467: nit: After that callback _is received_ ...

PR #778, opened 2020-05-13
fix(iot-dev): Fix issue where amqp layer didn't recognize an ack to a subscription message that it sent

timtay-microsoft:
The recent twin array support changes introduced sending a desired properties subscription message each time the twin sender link opened. However, once the service acknowledges that message, our SDK logged an exception as it didn't correlate the ack back to the subscription message it sent. This PR adds a callback so that the top AMQP layer can be notified each time a lower level sends a subscription message

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsDeviceOperations.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsDeviceOperations.java
@@ -394,7 +394,7 @@ public abstract class AmqpsDeviceOperations
         if (linkName.equals(this.getSenderLinkTag()))
         {
             this.amqpsSendLinkState = AmqpsDeviceOperationLinkState.OPENED;
-            this.log.debug("{} sender link with link correlation id {} was successfully opened {}", getLinkInstanceType(), this.linkCorrelationId);
+            this.log.debug("{} sender link with link correlation id {} was successfully opened", getLinkInstanceType(), this.linkCorrelationId);
             return true;
         }
 
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -692,7 +692,7 @@ public final class AmqpsIotHubConnection extends ErrorLoggingBaseHandler impleme
                     else if (this.inProgressSubscriptionMessages.containsKey(deliveryTag))
                     {
                         SubscriptionType subscriptionType = this.inProgressSubscriptionMessages.remove(deliveryTag);
-                        log.debug("Successfully re-sent amqp subscription message of type {}", subscriptionType);
+                        log.debug("Successfully sent amqp subscription message of type {}", subscriptionType);
                     }
                     else
                     {
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -43,7 +43,7 @@ import static com.microsoft.azure.sdk.iot.device.MessageType.DEVICE_TWIN;
  * a message, and logic to re-establish the connection with the IoTHub in case it gets lost.
  */
 @Slf4j
-public final class AmqpsIotHubConnection extends ErrorLoggingBaseHandler implements IotHubTransportConnection
+public final class AmqpsIotHubConnection extends ErrorLoggingBaseHandler implements IotHubTransportConnection, SubscriptionMessageRequestSentCallback
 {
     private static final int MAX_WAIT_TO_CLOSE_CONNECTION = 60 * 1000; // 60 second timeout
     private static final int MAX_WAIT_TO_OPEN_CBS_LINKS = 20 * 1000; // 20 second timeout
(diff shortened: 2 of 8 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -688,6 +689,11 @@ else if (remoteState instanceof Modified || remoteState instanceof Released || r
                             this.listener.onMessageSent(inProgressMessages.remove(deliveryTag), transportException);
                         }
                     }
+                    else if (this.inProgressSubscriptionMessages.containsKey(deliveryTag))
+                    {
+                        SubscriptionType subscriptionType = this.inProgressSubscriptionMessages.remove(deliveryTag);
+                        log.debug("Successfully re-sent amqp subscription message of type {}", subscriptionType);
+                    }
                     else
```
timtay-microsoft, line 697: Without the above changes, this logic would fall to the else block and log an error

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -66,6 +66,7 @@
     private final static int MAX_MESSAGE_PAYLOAD_SIZE = 256*1024; //max IoT Hub message size is 256 kb, so amqp websocket layer should buffer at least that much space
     private final Boolean useWebSockets;
     private final Map<Integer, com.microsoft.azure.sdk.iot.device.Message> inProgressMessages = new ConcurrentHashMap<>();
+    private final Map<Integer, SubscriptionType> inProgressSubscriptionMessages = new ConcurrentHashMap<>();
```
timtay-microsoft, line 69: This new map will track outstanding subscription messages. I couldn't re-use the inProgressMessages map because there is no transport agnostic message to add to that map since we are creating the amqp message directly. Plus, messages in that map are expected to have a matching message in the IotHubTransport layer, which isn't the case for this special subscription message

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsDeviceOperations.java
```diff
@@ -394,7 +394,7 @@ protected boolean onLinkRemoteOpen(String linkName)
         if (linkName.equals(this.getSenderLinkTag()))
         {
             this.amqpsSendLinkState = AmqpsDeviceOperationLinkState.OPENED;
-            this.log.debug("{} sender link with link correlation id {} was successfully opened {}", getLinkInstanceType(), this.linkCorrelationId);
+            this.log.debug("{} sender link with link correlation id {} was successfully opened", getLinkInstanceType(), this.linkCorrelationId);
```
timtay-microsoft, line 397: This is orthogonal to the actual PR, but this log message had one too many spots for parameters, so I got rid of it

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionDeviceOperation.java
```diff
@@ -29,14 +29,16 @@
 
     private List<UUID> cbsCorrelationIdList = Collections.synchronizedList(new ArrayList<UUID>());
 
+    private SubscriptionMessageRequestSentCallback subscriptionMessageRequestSentCallback;
+
     /**
      * Create logical device entity to handle all operation.
      *
      * @param deviceClientConfig the configuration of teh device.
      * @param amqpsDeviceAuthentication the authentication object associated with the device.
```
barustum, line 38: Update constructor comments to reflect new parameter.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/SubscriptionMessageRequestSentCallback.java
```diff
@@ -0,0 +1,13 @@
+package com.microsoft.azure.sdk.iot.device.transport.amqps;
+
+import com.microsoft.azure.sdk.iot.device.DeviceTwin.DeviceOperations;
+
+public interface SubscriptionMessageRequestSentCallback
+{
+    void onSubscriptionMessageSent(int deliveryTag, SubscriptionType deviceOperationType);
```
azabbasi, line 7: subscriptionType instead of deviceOperationType?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionManager.java
```diff
@@ -24,14 +24,15 @@
...
     private AmqpsDeviceAuthentication amqpsDeviceAuthentication;
     private ArrayList<AmqpsSessionDeviceOperation> amqpsDeviceSessionList = new ArrayList<>();
+    private SubscriptionMessageRequestSentCallback subscriptionMessageRequestSentCallback;
 
     /**
      * Constructor that takes a device configuration.
      *
      * @param deviceClientConfig the device configuration to use for 
      *                           session management.
      */
```
barustum, line 34: Update comment header to include new parameter

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionDeviceOperation.java
timtay-microsoft: Good catch

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/SubscriptionMessageRequestSentCallback.java
timtay-microsoft: Good catch

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
abhipsaMisra: Why does the logs say: "re-sent amqp subscription message"?
I thought what we do here is -> if we receive an ack from service for the subscription message, we remove it from our internal queue, so as to not resend it. Is that not correct?

abhipsaMisra: @timtay-microsoft : A question :raised_hand:

timtay-microsoft: This log statement gets executed when we receive an ack for an outstanding subscription message that we sent. It does get removed from the queue of outstanding messages as a result. The only odd thing about this subscription message is that it is sent at the session level. That's why this new queue needed to be added for messages like it

timtay-microsoft:
I talked with Abhipsa about this, and she pointed out there is a small problem with this code. I'll get the fix checked in tomorrow. Don't merge this PR for now

PR #781, opened 2020-05-18
doc(iot-dev): Add missing javadoc explaining SetSendInterval for setOption API

timtay-microsoft:
This was missing from the list of options, even though it is a valid option that users can set

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
@@ -549,7 +549,7 @@ public final class DeviceClient extends InternalClient implements Closeable
      *	      is expected to be of type {@code long}.
      *
      *	    - <b>SetSendInterval</b> - this option is applicable to all protocols.
-     *	      This value sets the period (in milliseconds) that this SDK spawns a thread to send a queued message.
+     *	      This value sets the period (in milliseconds) that this SDK spawns threads to send queued messages.
      *	      Even if no message is queued, this thread will be spawned.
      *
      *	    - <b>SetReceiveInterval</b> - this option is applicable to all protocols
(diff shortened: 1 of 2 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -548,6 +548,10 @@ public void subscribeToDeviceMethod(DeviceMethodCallback deviceMethodCallback, O
      *	      the service checking for availability of new messages. The value
      *	      is expected to be of type {@code long}.
      *
+     *	    - <b>SetSendInterval</b> - this option is applicable to all protocols.
+     *	      This value sets the period (in milliseconds) that this SDK spawns a thread to send a queued message.
```
vinagesh, line 552: time instead of period?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -360,6 +360,10 @@ public DeviceClientConfig getConfig()
      *	      the service checking for availability of new messages. The value
      *	      is expected to be of type {@code long}.
      *
+     *	    - <b>SetSendInterval</b> - this option is applicable to all protocols.
+     *	      This value sets the period (in milliseconds) that this SDK spawns a thread to send a queued message.
```
vinagesh, line 364: time?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
timtay-microsoft: Period implies that it spawns periodically, so I like it here

vinagesh: I see. I did not infer that initially but makes sense now.

PR #782, opened 2020-05-21
Add support for fully qualified error codes in service client HTTP exceptions

timtay-microsoft:
Previously, the service client ignored the fully qualified status codes embedded in the error message for cases where the service threw a 404/500/etc.

Customer's need to have some granularity in treating a 404001 (deviceNotFound) from a 404010 (moduleNotFound) from a 404103 (deviceNotOnline) as these are all potential error cases when invoking a method on a module.

Now all of our service side exception types have dedicated fields for error code and error code description.

The error code will be the 6 digit code (404001, for instance) if the service gave the SDK that code, or it will be the 3 digit error code that we always get in the http response object (404, for instance)

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
@@ -1,0 +1,193 @@
+package com.microsoft.azure.sdk.iot.deps.serializer;
+
+/**
+ * See https://docs.microsoft.com/en-us/rest/api/iothub/common-error-codes for additional details
+ */
+public enum ErrorCodeDescription
+{
+    // These descriptions belong to fully qualified status codes, such as 404001
+    UnclassifiedErrorCode,
+    InvalidProtocolVersion,
+    DeviceInvalidResultCount,
+    InvalidOperation,
+    ArgumentInvalid,
+    ArgumentNull,
+    IotHubFormatError,
+    DeviceStorageEntitySerializationError,
+    BlobContainerValidationError,
+    ImportWarningExistsError,
+    InvalidSchemaVersion,
+    DeviceDefinedMultipleTimes,
+    DeserializationError,
+    BulkRegistryOperationFailure,
+    CannotRegisterModuleToModule,
+    IotHubNotFound,
+    IotHubUnauthorizedAccess,
+    IotHubUnauthorized,
+    IotHubSuspended,
+    IotHubQuotaExceeded,
+    JobQuotaExceeded,
+    DeviceMaximumQueueDepthExceeded,
+    IotHubMaxCbsTokenExceeded,
+    DeviceNotFound,
+    JobNotFound,
+    PartitionNotFound,
+    ModuleNotFound,
+    DeviceAlreadyExists,
+    ModuleAlreadyExistsOnDevice,
+    DeviceMessageLockLost,
+    MessageTooLarge,
+    TooManyDevices,
+    TooManyModulesOnDevice,
+    ThrottleBacklogLimitExceeded,
+    InvalidThrottleParameter,
+    ServerError,
+    JobCancelled,
+    ConnectionForcefullyClosedOnNewConnection,
+    DeviceNotOnline,
+    DeviceConnectionClosedRemotely,
+
+    // These descriptions belong to 3 digit http response codes such as 404.
+    BadFormat,
+    Unauthorized,
+    Forbidden,
+    NotFound,
+    Conflict,
+
+    /**
+     * Represents status code for 429 and for status code 429001
+     */
+    PreconditionFailed,
+
+    RequestEntityTooLarge,
+
+    /**
+     * Represents status code for 429 and 429001
+     */
+    ThrottlingException,
+    InternalServerError,
+
+    /**
+     * Defined for status code 503 and 503001
+     */
+    ServiceUnavailable;
+
+
+    /**
+     * Get the ErrorCodeDescription tied to the provided errorCode
+     * @param errorCode the service error code, such as 404, or 429001
+     * @return the corresponding ErrorCodeDescription
+     */
+    public static ErrorCodeDescription Parse(int errorCode)
+    {
+        switch (errorCode)
+        {
+            case(400001):
+                return InvalidProtocolVersion;
+            case(400002):
+                return DeviceInvalidResultCount;
+            case(400003):
+                return InvalidOperation;
+            case(400004):
+                return ArgumentInvalid;
+            case(400005):
+                return ArgumentNull;
+            case(400006):
+                return IotHubFormatError;
+            case(400007):
+                return DeviceStorageEntitySerializationError;
+            case(400008):
+                return BlobContainerValidationError;
+            case(400009):
+                return ImportWarningExistsError;
+            case(400010):
+                return InvalidSchemaVersion;
+            case(400011):
+                return DeviceDefinedMultipleTimes;
+            case(400012):
+                return DeserializationError;
+            case(400013):
+                return BulkRegistryOperationFailure;
+            case (400027):
+                return ConnectionForcefullyClosedOnNewConnection;
+            case(400301):
+                return CannotRegisterModuleToModule;
+            case(401001):
+                return IotHubNotFound;
+            case(401002):
+                return IotHubUnauthorizedAccess;
+            case(401003):
+                return IotHubUnauthorized;
+            case(403001):
+                return IotHubSuspended;
+            case(403002):
+                return IotHubQuotaExceeded;
+            case(403003):
+                return JobQuotaExceeded;
+            case(403004):
+                return DeviceMaximumQueueDepthExceeded;
+            case(403005):
+                return IotHubMaxCbsTokenExceeded;
+            case(404001):
+                return DeviceNotFound;
+            case(404002):
+                return JobNotFound;
+            case(404003):
+                return PartitionNotFound;
+            case(404010):
+                return ModuleNotFound;
+            case (404103):
+                return DeviceNotOnline;
+            case (404104):
+                return DeviceConnectionClosedRemotely;
+            case(409001):
+                return DeviceAlreadyExists;
+            case(409301):
+                return ModuleAlreadyExistsOnDevice;
+            case(412001):
+                return PreconditionFailed;
+            case(412002):
+                return DeviceMessageLockLost;
+            case(413001):
+                return MessageTooLarge;
+            case(413002):
+                return TooManyDevices;
+            case(413003):
+                return TooManyModulesOnDevice;
+            case(429001):
+                return ThrottlingException;
+            case(429002):
+                return ThrottleBacklogLimitExceeded;
+            case(429003):
+                return InvalidThrottleParameter;
+            case(500001):
+                return ServerError;
+            case(500002):
+                return JobCancelled;
+            case(503001):
+                return ServiceUnavailable;
+            case (400):
+                return BadFormat;
+            case (401):
+                return Unauthorized;
+            case (403):
+                return Forbidden;
+            case (404):
+                return NotFound;
+            case (409):
+                return Conflict;
+            case (412):
+                return PreconditionFailed;
+            case (413):
+                return RequestEntityTooLarge;
+            case (429):
+                return ThrottlingException;
+            case (500):
+                return InternalServerError;
+            case (503):
+                return ServiceUnavailable;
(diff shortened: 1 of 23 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSendHandler.java
```diff
@@ -266,7 +266,15 @@ public void onDelivery(Event event)
...
-                log.debug("Closing amqp cloud to device message sender link since the message was delivered");
+                if (remoteState.getClass().equals(Accepted.class))
+                {
+                    log.debug("Closing AMQP cloud to device message sender link since the message was delivered");
+                }
+                else
+                {
+                    log.debug("Closing AMQP cloud to device message sender link since the message failed to be delivered");
+                }
+
```
timtay-microsoft, line 277: This change is completely orthogonal to the error code work that this PR is for, but it is a small change that is worth fixing despite that

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
```diff
@@ -0,0 +1,193 @@
+package com.microsoft.azure.sdk.iot.deps.serializer;
+
+/**
+ * See https://docs.microsoft.com/en-us/rest/api/iothub/common-error-codes for additional details
+ */
+public enum ErrorCodeDescription
```
abhipsaMisra, line 6: This will be a good addition to the C# SDK as well

timtay-microsoft: I was planning on doing the same over there, too, yeah

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
```diff
@@ -0,0 +1,193 @@
...
+            case (413):
+                return RequestEntityTooLarge;
+            case (429):
+                return ThrottlingException;
+            case (500):
+                return InternalServerError;
+            case (503):
+                return ServiceUnavailable;
+            default:
+                return UnclassifiedErrorCode;
```
prmathur-microsoft, line 190: Have these been validated by service ? i.e. do they support the above for what do they mean?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/methods/DeviceMethodTests.java
```diff
@@ -43,7 +44,7 @@ public void cleanToStart() throws Exception
     {
         super.cleanToStart();
     }
-
+/*
     @Test
```
prmathur-microsoft, line 48: Why is this commented?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
timtay-microsoft: These are all taken directly from the C# sdk [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/service/src/Common/Exceptions/ErrorCode.cs)

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/methods/DeviceMethodTests.java
timtay-microsoft: Whoops, my mistake. I can uncomment these now

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * Get the ErrorCodeDescription tied to the provided errorCode
+     * @param errorCode the service error code, such as 404, or 429001
+     * @return the corresponding ErrorCodeDescription
+     */
+    public static ErrorCodeDescription Parse(int errorCode)
+    {
+        switch (errorCode)
+        {
+            case(400001):
```
abhipsaMisra, line 85: Add a link to the doc that has these mappings. Is it on some MSDN doc?

timtay-microsoft: Unfortunately, I couldn't find a MSDN doc for this. These codes are taken directly from the C# SDK [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/service/src/Common/Exceptions/ErrorCode.cs)

timtay-microsoft: In our work with them over the track 2 SDK, we can push for a document like that, though

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/methods/DeviceMethodTests.java
```diff
@@ -43,7 +44,7 @@ public void cleanToStart() throws Exception
     {
         super.cleanToStart();
     }
-
+/*
```
abhipsaMisra: what is this for?

timtay-microsoft: I forgot to undo that, but now it is gone

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
prmathur-microsoft: 👍

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
```diff
@@ -0,0 +1,193 @@
...
+    TooManyModulesOnDevice,
+    ThrottleBacklogLimitExceeded,
+    InvalidThrottleParameter,
+    ServerError,
+    JobCancelled,
+    ConnectionForcefullyClosedOnNewConnection,
+    DeviceNotOnline,
+    DeviceConnectionClosedRemotely,
+
+    // These descriptions belong to 3 digit http response codes such as 404.
```
abhipsaMisra, line 50: nit: need consistency in using single-line vs block comments

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
```diff
@@ -0,0 +1,193 @@
...
+    Forbidden,
+    NotFound,
+    Conflict,
+
+    /**
+     * Represents status code for 429 and for status code 429001
+     */
+    PreconditionFailed,
+
+    RequestEntityTooLarge,
```
abhipsaMisra, line 62: Does this comment mean PreconditionFailed maps to 429 and RequestEntityTooLarge to 429001?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * Represents status code for 429 and 429001
+     */
+    ThrottlingException,
+    InternalServerError,
+
+    /**
+     * Defined for status code 503 and 503001
+     */
+    ServiceUnavailable;
```
abhipsaMisra, line 73: Both of these codes map to ServiceUnavailable?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
timtay-microsoft: The single line comments are for us developers to see. The block comments will show up when a user has intellisense

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
```diff
@@ -0,0 +1,193 @@
+package com.microsoft.azure.sdk.iot.deps.serializer;
+
+/**
+ * See https://docs.microsoft.com/en-us/rest/api/iothub/common-error-codes for additional details
+ */
+public enum ErrorCodeDescription
+{
+    // These descriptions belong to fully qualified status codes, such as 404001
+    UnclassifiedErrorCode,
```
abhipsaMisra, line 9: Is there a better name for this enum than "Description"? This isn't really a description.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
timtay-microsoft: The javadoc is only visible when using intellisense for the PreconditionFailed enum value. It means that the service may return a 429, and it may return a 429001, but in both cases it maps to this enum value

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
timtay-microsoft: The enum is called "ErrorCodeDescription" because it describes what the error code (404001, etc) mean. I can't think of a better name.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
abhipsaMisra: really? I didn't know that!
So when I hover over `PreconditionFailed`, I should be seeing
```
/**
     * Represents status code for 429 and for status code 429001
     */
```

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
timtay-microsoft: Yep, weirdly

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
abhipsaMisra: Ah, the spacing and block comments confused me.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
timtay-microsoft: Without the slashes or stars, but yeah!

timtay-microsoft: It's the java equivalent of how C# does 
```csharp
// Comment 
```
vs
```csharp
/// <summary>Comment</summary>
```

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
timtay-microsoft: Registry manager ops gets 503, whereas a device method invocation likely gets a 503001 based on my research so far. In both cases, though, the service is telling the SDK that the service is unavailable

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
timtay-microsoft: Just to be more clear, though, if the block comments start with only one *, then they behave like the single line comments.

```java
/**
 * This is a javadoc header
 */

/*
 * This is just a comment. Users won't see it
 */
```

PR #787, opened 2020-05-29
refactor(iot-dev): Clean up AMQP layer redundancies and useless code

timtay-microsoft:
This is a rather large refactor, but I couldn't find a way to break this into smaller chunks, since all the code changes are localized to the AMQP layer.

Broadly speaking, the AMQP layer had a lot of redundant code, confusing abstractions, and out of date comments. This PR re-organizes the AMQP layer to cut down on all of that. Ideally, this layer should follow the abstraction that is defined in AMQP as a protocol. For instance, connections shouldn't be aware of links, but sessions should.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsAuthenticationLinkHandler.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.qpid.proton.ProtonUnsupportedOperationException;
+import org.apache.qpid.proton.engine.SslDomain;
+import org.apache.qpid.proton.engine.Transport;
+import org.apache.qpid.proton.message.impl.MessageImpl;
+
+import javax.net.ssl.SSLContext;
+import java.util.UUID;
+
+@Slf4j
+public abstract class AmqpsAuthenticationLinkHandler extends AmqpsLinksHandler
```
drwill-ms, line 18: Why the trailing 's' in Amqps?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -333,57 +283,60 @@ boolean handleAuthenticationMessage(AmqpsMessage amqpsMessage)
...
-            msgType = DEVICE_TELEMETRY;
-        }
-        else
-        {
-            msgType = message.getMessageType();
+            message.setMessageType(DEVICE_TELEMETRY);
         }
 
-        if (amqpsDeviceOperationsMap.get(msgType) != null)
+        if (amqpsLinkMap.get(message.getMessageType()) != null)
```
drwill-ms, line 337: seems like before the if statement it should get the value, then compare it to null, and if not null, use the cached value. Right now we have 2 look ups.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsAuthenticationLinkHandler.java
timtay-microsoft: AMQPS is to AMQP what HTTPS is to HTTP. It just signifies that we are doing ssl handshakes when connecting

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
timtay-microsoft: Sure, this can be simplified

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsAuthenticationLinkHandler.java
abhipsaMisra, line 18: Why don't we call out that this is the cbs link?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsAuthenticationLinkHandler.java
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * @param amqpsMessage the message to evaluate.
+     * @param authenticationCorrelationId the expected correlation ID.
+     * @return true if the message acknowledge the authentication, false otherwise.
+     */
+    abstract protected boolean handleAuthenticationMessage(AmqpsMessage amqpsMessage, UUID authenticationCorrelationId);
+
+    protected IotHubTransportMessage convertFromProton(AmqpsMessage amqpsMessage, DeviceClientConfig deviceClientConfig)
+    {
+        log.warn("AmqpsAuthenticationLinkHandler called to convertFromProton, but this should never occur");
```
abhipsaMisra, line 62: Why should this never be called? Can you add a comment explaining it.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsAuthenticationLinkHandler.java
abhipsaMisra: Ah, this is the abstract class; is this for grouping method calls that are relevant to the cbs link?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsAuthenticationLinkHandlerX509.java
```diff
@@ -7,31 +7,23 @@
 import org.apache.qpid.proton.engine.SslDomain;
 import org.apache.qpid.proton.engine.Transport;
 
-import java.io.IOException;
+import javax.net.ssl.SSLContext;
 import java.util.UUID;
 
 @Slf4j
-public class AmqpsDeviceAuthenticationX509 extends AmqpsDeviceAuthentication
+public class AmqpsAuthenticationLinkHandlerX509 extends AmqpsAuthenticationLinkHandler
```
abhipsaMisra, line 14: What does this link do? I thought X509 certificate info is handled by the tcp connection directly (from the ssl context supplied)

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsConnectionStateCallback.java
```diff
@@ -0,0 +1,19 @@
...
+{
+    /**
+     * Called when a device's amqp session has successfully been opened. For multiplexing scenarios, this callback will
+     * be fired once per device.
+     * @param deviceId the id of the device that had its session opened successfully
+     */
+    public void onDeviceSessionOpened(String deviceId);
+
+    /**
+     * Called when the authentication session has successfully been opened. Never called for x509 auth.
```
abhipsaMisra, line 16: For multiplexing scenarios with multiple devices, we will have 1 auth link associated with the connection, so this will be fired once, correct?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -116,51 +117,30 @@ public AmqpsIotHubConnection(DeviceClientConfig config)
...
+        DeviceClientConfig.AuthType authType = config.getAuthenticationType();
+        if (authType.equals(DeviceClientConfig.AuthType.SAS_TOKEN))
         {
-            // Codes_SRS_AMQPSIOTHUBCONNECTION_12_019: [The function shall call AmqpsSessionManager.addDeviceOperationSession with the given deviceClientConfig.]
-            this.amqpsSessionManager.addDeviceOperationSession(deviceClientConfig);
-
-            this.log.trace("Added device to session list");
+            this.amqpsAuthenticationLinkHandler = new AmqpsAuthenticationLinkHandlerCBS();
+        }
+        else if (authType == DeviceClientConfig.AuthType.X509_CERTIFICATE)
```
abhipsaMisra, line 134: Q - who sets `DeviceClientConfig.AuthType`? Is it the sdk or the user?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -178,61 +158,47 @@ public void addDeviceOperationSession(DeviceClientConfig deviceClientConfig) thr
...
                 if (this.savedException != null)
                 {
                     throw this.savedException;
                 }
 
-                if (authenticationLinksOpenTimedOut)
+                if (authenticationSessionOpenedTimedOut)
                 {
-                    closeConnectionWithException("Timed out waiting for authentication links to open", true);
+                    closeConnectionWithException("Timed out waiting for authentication session to open", true);
```
abhipsaMisra, line 190: does this throw an exception?

abhipsaMisra: I cannot seem to find the impl for `closeConnectionWithException`

abhipsaMisra: Got it!

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -178,61 +158,47 @@ public void addDeviceOperationSession(DeviceClientConfig deviceClientConfig) thr
...
+                this.log.trace("Waiting for device sessions to open...");
+                boolean deviceSessionsOpenTimedOut = !this.deviceSessionsOpenedLatch.await(MAX_WAIT_TO_OPEN_WORKER_LINKS, TimeUnit.MILLISECONDS);
 
                 if (this.savedException != null)
                 {
                     throw this.savedException;
                 }
 
-                if (workerLinksOpenTimedOut)
+                if (deviceSessionsOpenTimedOut)
```
abhipsaMisra, line 201: what happens if some sessions open, and some time out - how are they closed? 
do we rely on client connection closure -> remote connection closure -> remote session/ links closure -> client session/ link closure?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -396,23 +333,21 @@ private void executorServicesCleanup() throws TransportException
...
      */
     private void closeAsync()
     {
-        this.log.trace("CloseAsync called on amqp connection");
-        // Codes_SRS_AMQPSIOTHUBCONNECTION_15_013: [The function shall closeNow the AmqpsSessionManager and the AMQP connection.]
-        if (this.amqpsSessionManager != null)
-        {
-            this.amqpsSessionManager.closeNow();
-        }
+        this.log.debug("Closing amqp connection...");
```
abhipsaMisra, line 340: is this invoked only on explicit deviceclient.closeAsync() call, or during reconnect as well?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -421,158 +356,95 @@ private void closeAsync()
...
-        }
 
+        reactor.connectionToHost(hostName, port, this);
         reactor.schedule(SEND_MESSAGES_PERIOD_MILLIS, this);
 
         if (this.deviceClientConfig.getAuthenticationProvider() instanceof IotHubSasTokenAuthenticationProvider)
         {
             int sasTokenRenewalPeriod = this.deviceClientConfig.getSasTokenAuthentication().getMillisecondsBeforeProactiveRenewal();
-            reactor.schedule(sasTokenRenewalPeriod, sasTokenRenewalHandler);
+            reactor.schedule(sasTokenRenewalPeriod, new AmqpSasTokenRenewalHandler(this, this.deviceClientConfig));
```
abhipsaMisra, line 387: what is the behavior here for scenarios where user supplied the sas token directly (which is not auto-renewable)?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -421,158 +356,95 @@ private void closeAsync()
...
-                this.amqpsSessionManager.onConnectionBound(transport);
-            }
-            catch (TransportException e)
-            {
-                this.savedException = e;
-                log.error("Encountered an exception while reacting to onConnectionBound event within amqp session", this.savedException);
-            }
+        try
+        {
+            this.amqpsAuthenticationLinkHandler.setSslDomain(transport, this.deviceClientConfig.getAuthenticationProvider().getSSLContext());
```
abhipsaMisra, line 428: Add a note here specifying that this will always be 1 per connection, and that is why it is initialized here (even when multiplexed devices exist).

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -599,168 +471,164 @@ private void addProxyLayer(Transport transport, String hostName)
...
-     */
     @Override
     public void onDelivery(Event event)
     {
         Link link = event.getLink();
 
         if (link instanceof Sender)
         {
-            //ack received for a message that this SDK sent earlier
+            handleMessageAcknowledgmentDelivery(event.getDelivery(), (Sender) link);
```
abhipsaMisra, line 481: is this callback received on the connection, or on a link?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -1336,6 +1022,207 @@ public void onSubscriptionMessageSent(int deliveryTag, SubscriptionType subscrip
...
+        this.authenticationSessionOpenedLatch.countDown();
+    }
+
+    final void addDeviceSession(DeviceClientConfig deviceClientConfig, SubscriptionMessageRequestSentCallback subscriptionMessageRequestSentCallback, boolean afterOpen) throws TransportException {
+        if (deviceClientConfig == null)
+        {
+            throw new IllegalArgumentException("deviceClientConfig cannot be null.");
+        }
+
+        // Check if the device session still exists from a previous connection
```
abhipsaMisra, line 1042: What does "previous connection" mean here?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -1336,6 +1022,207 @@ public void onSubscriptionMessageSent(int deliveryTag, SubscriptionType subscrip
...
+            log.debug("Sending authentication message with correlationId {} to authenticate device {}", authenticationCorrelationId, sessionHandler.getDeviceId());
+            this.amqpsAuthenticationLinkHandler.authenticate(sessionHandler.deviceClientConfig, authenticationCorrelationId);
+        }
+    }
+
+    public void authenticate() throws TransportException
+    {
+        for (AmqpsSessionHandler sessionHandler : this.sessionHandlerList)
+        {
+            authenticate(sessionHandler);
```
abhipsaMisra, line 1085: authenticate for each session, why? multiplexed sessions will still have a single cbs auth link, right?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsAuthenticationLinkHandler.java
timtay-microsoft: Yes, this is the abstract class that the CBS link handler extends from

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsAuthenticationLinkHandlerX509.java
timtay-microsoft: This class doesn't do much, but it does differ from the CBS implementation of the AmqpsAuthenticationLinkHandler in how it sets the SSL domain. For CBS, during the SSL domain setup, you enable SASL auth with type "anonymous". X509 doesn't do that. Really, its a bit of an odd class overall since x509 auth has no links to manage. What do you think about it?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsConnectionStateCallback.java
timtay-microsoft: Correct, yes. Even for single device connections, that callback only gets fired once.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
timtay-microsoft: The SDK sets it, but only after the user has provided us their certificates or connection string so that we know what type they are using.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
timtay-microsoft: If some links don't open, then we do shut down the connection, all sessions in that connection, and all links in those sessions. This is handled by the retry logic in the IotHubTransport layer where the amqp layer is closed, then re-opened

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
timtay-microsoft: In the IotHubTransport layer, there is some retry logic that recycles the instance of AmqpsIotHubConnection.java between retry attempts. As such, this layer was written with the understanding that there may be states (like device foo being subscribed to twin) that should carry over from one connection to another

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
timtay-microsoft: Each session uses the same CBS links, but each session has to send a message over those links to authenticate its corresponding device. This method just says to authenticate each device over the open CBS links.

timtay-microsoft:
I'm finally getting the hang of how proton-j uses different handlers to create the abstraction of responsibility that one would expect from our AMQP classes. Expect another refactor to come along to this PR soon with that

timtay-microsoft:
Closing this in lieu of #793 since a lot has changed since this PR was put up

PR #793, opened 2020-06-04
refactor(iot-dev): Clean up AMQP layer redundancies and useless code (redo)

timtay-microsoft:
Opening a new PR since a lot has changed since the last PR, so it doesn't make sense to reuse it

see #787 for the basics, but this time the abstraction of responsibility is significantly better thanks to using multiple reactor handlers instead of just one.


timtay-microsoft:
Note that there are still some TODO's left for me to resolve, but the broad strokes of this PR are ready for review

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
```diff
@@ -301,7 +301,6 @@ public void setProperty(String name, String value)
             this.properties.remove(messageProperty);
         }
 
-        log.trace("Setting message property");
```
timtay-microsoft, line 304: This log statement was annoying while I was debugging, and I imagine it would be for users, too

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
```diff
@@ -120,7 +120,7 @@ public boolean shouldRefreshToken(boolean proactivelyRenew)
 
     public int getMillisecondsBeforeProactiveRenewal()
     {
-        return (int) this.tokenValidSecs * this.timeBufferPercentage;
```
timtay-microsoft, line 123: This was a bug where the token renewal time was much more proactive than it needed to be. Note that it was doing (time in seconds) X (85) which yielded a unit of time between seconds and milliseconds. But the return value was supposed to be in milliseconds

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransportConnection.java
```diff
@@ -60,7 +60,7 @@
      * @return true if the acknowledgement was sent successfully, and false otherwise
      * @throws TransportException if an exception occurred while sending the acknowledgement
      */
-    boolean sendMessageResult(Message message, IotHubMessageResult result) throws TransportException;
```
timtay-microsoft, line 63: Every call to this API was already passing in an IotHubTransportMessage type (which is a derived type from Message), and it made the amqp refactor easier for this interface to just take a transport message (which has callback info that Message does not)

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
drwill-ms: Won't the value be 10x larger now?

Is it a count down or count up?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransportConnection.java
drwill-ms: Is this change contract breaking?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
```diff
@@ -120,7 +120,7 @@ public boolean shouldRefreshToken(boolean proactivelyRenew)
...
+        return (int) ((this.tokenValidSecs * 1000) * (this.timeBufferPercentage / 100.0));
```
bikamani, line 123: What is divide by 100 for?

timtay-microsoft: The field ```timeBufferPercentage``` is supposed to represent 85%, but it is just 85 instead of .85

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsMethodsSenderLinkHandler.java
```diff
@@ -0,0 +1,124 @@
...
+
+            protonMessage.setProperties(properties);
+
+            Map<String, Object> userProperties = new HashMap<>();
+            if (deviceMethodMessage.getStatus() != null)
+            {
+                userProperties.put(APPLICATION_PROPERTY_KEY_IOTHUB_STATUS, Integer.parseInt(deviceMethodMessage.getStatus()));
+            }
+
+            Map<String, Object> applicationPropertiesMap;
```
drwill-ms, line 109: if this isn't used outside of the if statement below, can it be moved into it?

timtay-microsoft: It can be moved in there, good catch

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
drwill-ms: Worth a code comment.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsCbsReceiverLinkHandler.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.qpid.proton.engine.Event;
+import org.apache.qpid.proton.engine.Receiver;
+
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Every SAS token based authentication over AMQP requires a CBS session with a sender and receiver link. This
+ * class defines the receiver link which receives authentication status codes corresponding to each authentication attempt.
```
bikamani, line 20: Is this a new line or appears this way on github?

timtay-microsoft: I'm not sure that I see what you are seeing. This class javadoc is two lines for me

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsCbsReceiverLinkHandler.java
```diff
@@ -0,0 +1,117 @@
...
+
+    static String getCbsTag()
+    {
+        return RECEIVER_LINK_TAG_PREFIX;
+    }
+
+    @Override
+    public String getLinkInstanceType()
+    {
+        return "cbs";
```
bikamani, line 48: should this be defined as private static final String

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsCbsReceiverLinkHandler.java
```diff
@@ -0,0 +1,117 @@
...
+@Slf4j
+public final class AmqpsCbsReceiverLinkHandler extends AmqpsReceiverLinkHandler
+{
+    private final static String APPLICATION_PROPERTY_STATUS_CODE = "status-code";
+    private final static String APPLICATION_PROPERTY_STATUS_DESCRIPTION = "status-description";
+
+    private static final String RECEIVER_LINK_ENDPOINT_PATH = "$cbs";
+    private static final String RECEIVER_LINK_TAG_PREFIX = "cbs-receiver";
+
+    private Map<UUID, AuthenticationMessageCallback> correlationMap = new ConcurrentHashMap<>();
```
bikamani, line 31: what is correlationmap?

timtay-microsoft: I can add some comments to explain this. When the CBS receiver link gets a message from iothub, it needs to correlate that message's correlation id back to a callback so that a device session knows that it was authenticated

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsCbsSenderLinkHandler.java
```diff
@@ -0,0 +1,120 @@
...
+
+    static String getCbsTag()
+    {
+        return SENDER_LINK_TAG_PREFIX;
+    }
+
+    @Override
+    public String getLinkInstanceType()
+    {
+        return "cbs";
```
bikamani, line 62: you can use CBS_REPLY  here

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsCbsSenderLinkHandler.java
```diff
@@ -0,0 +1,120 @@
...
+
+        Section section;
+        try
+        {
+            section = new AmqpValue(deviceClientConfig.getSasTokenAuthentication().getRenewedSasToken(true, true));
+            outgoingMessage.setBody(section);
+        }
+        catch (IOException e)
+        {
+            log.error("getRenewedSasToken has thrown exception while building new cbs authentication message", e);
```
bikamani, line 114: getRenewedSasToken?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsCbsSenderLinkHandler.java
```diff
@@ -0,0 +1,120 @@
...
+
+        properties.setTo(CBS_TO);
+        properties.setReplyTo(CBS_REPLY);
+        outgoingMessage.setProperties(properties);
+
+        Map<String, Object> userProperties = new HashMap<>(3);
+        userProperties.put(OPERATION_KEY, OPERATION_VALUE);
+        userProperties.put(TYPE_KEY, TYPE_VALUE);
+
+        String host = deviceClientConfig.getGatewayHostname();
```
bikamani, line 96: In which case is this value null?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsCbsSenderLinkHandler.java
timtay-microsoft: I'll add a private static final variable called LINK_TYPE on all these sender/receiver links instead, so they all feel similar

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsCbsSenderLinkHandler.java
timtay-microsoft: I'll change this to "Failed to renew sas token while building..."

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsCbsSenderLinkHandler.java
timtay-microsoft: Gateway host name is only not null when connecting to edgehub. Typically, it is null

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsCbsReceiverLinkHandler.java
timtay-microsoft: I'm not sure I see what you are seeing

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsMessage.java
```diff
@@ -19,96 +17,18 @@
...
-        // Codes_SRS_AMQPSMESSAGE_12_002: [Setter for the MessageType.]
-        this.amqpsMessageType = amqpsMessageType;
-    }
-
-    /**
-     * Get the deviceClientConfig
-     *
-     * @return The type of the message
-     */
-    public DeviceClientConfig getDeviceClientConfig()
```
bikamani, line 98: So all these is not being used?

timtay-microsoft: correct

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
drwill-ms: @timtay-microsoft ?

timtay-microsoft: The value will be 10 times larger, yes. The bug here was that the sas token was previously being renewed at 8.5% of its life spent, instead of the intended 85%. This was due to an incorrect mixing of seconds and milliseconds, and the timeBufferPercentage being treated as 85 instead of .85

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransportConnection.java
timtay-microsoft: No, this is an internal interface

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
drwill-ms: Please add code comments.

PR #796, opened 2020-06-05
refactor(iot-dev): Redesign constructor with sslContext for modelId changes

bikamani:
Review constructor changes as per design discussed. Also have fixed and refactored some unit tests which were failing


device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/DeviceClientTest.java
```diff
@@ -1995,7 +2041,7 @@ private void deviceClientInstanceExpectation(final IotHubClientProtocol protocol
                 break;
         }
 
-        new NonStrictExpectations()
+        new Expectations()
         {
```
bikamani, line 2045: NonStrictExpectations is deprecated and guidance says to use Expectations instead.

PR #806, opened 2020-06-11
fix(e2e): Fix multithreaded test not waiting for threads to finish correctly

timtay-microsoft:
This test always waited a full 3 minutes because the countdown latch waiting didn't work right. No need for a countdown latch though since we can just wait for all the threads to finish before checking if each send succeeded. 

Speeds up this test from 3 minutes to ~5 seconds


iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/TransportClientTests.java
```diff
@@ -50,8 +47,8 @@
     private static final int NUM_MESSAGES_PER_CONNECTION = 3;
 
     private static final long RETRY_MILLISECONDS = 100; //.1 seconds
-    private static final long SEND_TIMEOUT_MILLISECONDS = 5 * 60 * 1000; // 5 minutes
-    private static final long RECEIVE_MESSAGE_TIMEOUT = 5 * 60 * 1000; // 5 minutes
+    private static final long SEND_TIMEOUT_MILLISECONDS = 20 * 1000; // 20 seconds
```
drwill-ms, line 50: Nothing like TimeSpan in Java?

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/TransportClientTests.java
```diff
@@ -251,20 +248,21 @@ public void receiveMessagesIncludingProperties() throws Exception
...
-                    new MultiplexRunnable(
-                            testInstance.devicesList[i], testInstance.clientArrayList.get(i), NUM_MESSAGES_PER_CONNECTION, NUM_KEYS_PER_MESSAGE, SEND_TIMEOUT_MILLISECONDS, cdl, testInstance.succeed))
-                    .start();
+            threads[i] = new Thread(new MultiplexRunnable(testInstance.devicesList[i], testInstance.clientArrayList.get(i), testInstance.succeed));
+            threads[i].start();
+        }
+
+        for (int i = 0; i < testInstance.clientArrayList.size(); i++)
+        {
+            threads[i].join(SEND_TIMEOUT_MILLISECONDS);
```
prmathur-microsoft, line 264: 👍

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/tests/iothub/TransportClientTests.java
timtay-microsoft: Not that I'm familiar with, no. Operations like .await are always just in milliseconds (long or int)

PR #808, opened 2020-06-12
feat(iot-dev): Deprecate existing upload to blob APIs in favor of more granular APIs

timtay-microsoft:
Customers should be free to use whatever version of the Azure Storage SDK that they want, and to configure it however they want. Because of that, the existing file upload API has been split into the three service calls: Get SAS uri from Iot Hub, upload to Azure Storage blob using iothub provided credentials, and then notify Iot hub that a file upload has completed.

Same as the recent effort on C# SDK that can be seen here: https://github.com/Azure/azure-iot-sdk-csharp/pull/1403


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        this.correlationId = fileUploadCompletionNotification.getCorrelationId();
+        this.isSuccess = fileUploadCompletionNotification.getSuccess();
+        this.statusCode = fileUploadCompletionNotification.getStatusCode();
+        this.statusDescription = fileUploadCompletionNotification.getStatusDescription();
+    }
+
+    /**
+     * Update the status information in the collection, and return the new json.
```
barustum, line 94: Comment says the method return the new job. Method signature does not not return anything.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
```diff
@@ -0,0 +1,195 @@
...
+     *
+     * @param isSuccess is a Boolean representing whether the file was uploaded successfully.
+     * @param statusCode is the status for the upload of the file to storage.
+     * @param statusDescription is the description of the status code.
+     * @throws IllegalArgumentException if one of the parameters is null, empty, or not valid.
+     */
+    private void updateStatus(Boolean isSuccess, Integer statusCode, String statusDescription) throws IllegalArgumentException
```
barustum, line 101: if the first constructor is used to create this object, the statusDescription is null. So wouldn't this method throw since statusDescription is null?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
```diff
@@ -0,0 +1,195 @@
...
+    {
+        this.isSuccess = isSuccess;
+        this.statusCode = statusCode;
+        this.statusDescription = statusDescription;
+    }
+
+    /**
+     * Convert this class in a valid json.
```
barustum, line 109: 'to' a valid json.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
```diff
@@ -0,0 +1,195 @@
...
+     * for this upload from IoT Hub. Must equal {@link FileUploadSasUriResponse#getCorrelationId()}.
+     * @param correlationId the unique identifier that correlates this file upload status to a SAS URI that IoT Hub retreived from Azure Storage earlier.
+     */
+    public void setCorrelationId(String correlationId)
+    {
+        this.correlationId = correlationId;
+    }
+
+    /**
+     * @return Get the correlationId that correlates this FileUploadCompletionNotification to the earlier request to get the SAS URI
```
barustum, line 131: Seems like this description describes the API and not the @return value. Maybe remove the 'Get'?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
```diff
@@ -0,0 +1,195 @@
...
+     * @param statusDescription The status description associated with this file upload request
+     */
+    public void setStatusDescription(String statusDescription)
+    {
+        this.statusDescription = statusDescription;
+    }
+
+    /**
+     * Empty constructor: Used only to keep GSON happy.
+     */
```
barustum, line 191: do you need a @SuppressWarnings("unused") for this constructor?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * Create an instance of the FileUploadCompletionNotification for a single file upload operation using Azure Storage.
+     *
+     * @param correlationId the correlationId that correlates this FileUploadCompletionNotification to the earlier request to get the SAS URI
+     *                      for this upload from IoT Hub. This field is mandatory. Must equal {@link FileUploadSasUriResponse#getCorrelationId()}.
+     * @param isSuccess whether the file was uploaded successfully. This field is mandatory.
+     * @throws IllegalArgumentException if one of the parameters is null, empty, or not valid.
+     */
+    public FileUploadCompletionNotification(String correlationId, Boolean isSuccess)
+            throws IllegalArgumentException
```
barustum, line 46: is this 'throws' needed since the method doesn't throw anything?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
```diff
@@ -0,0 +1,195 @@
...
+     *
+     * @param correlationId the correlationId that correlates this FileUploadCompletionNotification to the earlier request to get the SAS URI
+     *                      for this upload from IoT Hub. This field is mandatory. Must equal {@link FileUploadSasUriResponse#getCorrelationId()}.
+     * @param isSuccess whether the file was uploaded successfully. This field is mandatory.
+     * @param statusCode is the status for the upload of the file to storage.
+     * @param statusDescription is the description of the status code.
+     * @throws IllegalArgumentException if one of the parameters is null, empty, or not valid.
+     */
+    public FileUploadCompletionNotification(String correlationId, Boolean isSuccess, Integer statusCode, String statusDescription)
+            throws IllegalArgumentException
```
barustum, line 63: is this 'throws' needed since the method doesn't throw anything?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
```diff
@@ -0,0 +1,195 @@
...
+        setCorrelationId(correlationId);
+        updateStatus(isSuccess, statusCode, statusDescription);
+    }
+
+    /**
+     * Construct this notification with json
+     * @param json the json to parse.
+     */
+    public FileUploadCompletionNotification(String json)
+    {
```
barustum, line 74: do we need a `throws IllegalArgumentException` for thie constructor?

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotificationTest.java
```diff
@@ -77,39 +77,18 @@ private static String createJson(String correlationId, Boolean isSuccess, Intege
...
-                assert false;
-            }
-            catch (IllegalArgumentException expected)
-            {
-                // Don't do anything, expected throw.
-            }
-        }
-    }
-
     /* Tests_SRS_FILE_UPLOAD_STATUS_21_004: [The toJson shall return a string with a json that represents the contend of the FileUploadStatusParser.] */
```
barustum, line 86: `FileUploadStatusParser` -> `FileUploadCompletionNotification`

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadSasUriRequestTest.java
```diff
@@ -0,0 +1,57 @@
...
+    public void constructor_succeed()
+    {
+        // act
+        FileUploadSasUriRequest fileUploadSasUriRequest = new FileUploadSasUriRequest(VALID_BLOB_NAME);
+
+        // assert
+        assertFileUploadRequest(fileUploadSasUriRequest, VALID_BLOB_NAME);
+    }
+
+    /* Tests_SRS_FILE_UPLOAD_REQUEST_21_004: [The toJson shall return a string with a json that represents the contend of the FileUploadSasUriResponse.] */
```
barustum, line 43: contend -> content

barustum, line 43: FileUploadSasUriResponse -> FileUploadSasUriRequest

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
timtay-microsoft: I'm not sure why we had javadoc for a private method, so I'll just delete the whole thing

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
timtay-microsoft: See above, I'll delete this javadoc

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
timtay-microsoft: This repo doesn't really pay much attention to warnings, so no

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
timtay-microsoft: Good catch, there is nothing thrown here anymore so I'll update the method signature and the javadocs

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
timtay-microsoft: No, because IllegalArgumentException is a runtime exception, so java methods don't need to declare them. https://www.tutorialspoint.com/how-to-handle-the-runtime-exception-in-java#:~:text=The Runtime Exception is the,Runtime Exceptions are never checked.

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotificationTest.java
timtay-microsoft: Actually, I should just delete these comments. They are a relic of an out-dated practice we used to do. They aren't worth keeping around anymore

PR #810, opened 2020-06-16
E2E Test categorization

timtay-microsoft:
This PR adds tags for test categorization as iothub tests, dps tests, CI build tests, flakey tests, and refactors the existing tags for standard tier and basic tier tests.

This PR also categorizes all the e2e tests in this repo with the relevant tags.


iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/BasicTierOnlyRule.java
```diff
@@ -1,16 +0,0 @@
...
- *  Copyright (c) Microsoft. All rights reserved.
- *  Licensed under the MIT license. See LICENSE file in the project root for full license information.
- */
-
-package com.microsoft.azure.sdk.iot.common.helpers;
-
-public class BasicTierOnlyRule implements ConditionalIgnoreRule.IgnoreCondition
-{
-
-    @Override
```
timtay-microsoft, line 11: The concept of this rule still exists, but I renamed it and refactored it to match the other tags

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/rules/BasicTierHubOnlyTestRule.java
```diff
@@ -0,0 +1,40 @@
...
+        private final Statement base;
+        private final Description description;
+
+        public IgnorableStatement(Statement base, Description description) {
+            this.base = base;
+            this.description = description;
+        }
+
+        @Override
+        public void evaluate() throws Throwable {
```
timtay-microsoft, line 29: This method is called prior to running any test method, and can be used to determine if the test should be run or skipped at runtime

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/rules/BasicTierHubOnlyTestRule.java
```diff
@@ -0,0 +1,40 @@
...
+            BasicTierHubOnlyTest annotation = description.getAnnotation(BasicTierHubOnlyTest.class);
+            BasicTierHubOnlyTest classAnnotation = description.getTestClass().getAnnotation(BasicTierHubOnlyTest.class);
+            if (annotation != null || classAnnotation != null)
```
timtay-microsoft, line 32: If the test method or its class is tagged with this annotation, then check if this test run is against a basic tier hub or not

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/rules/BasicTierHubOnlyTestRule.java
```diff
@@ -0,0 +1,40 @@
...
+            {
+                Assume.assumeTrue("Test is ignored", IntegrationTest.isBasicTierHub);
```
timtay-microsoft, line 34: assumeTrue is not the same as assertTrue. AssumeTrue will return a status code equivalent to "Ignored" while assertTrue would return a status code equivalent to "Failed"

timtay-microsoft: Or to simplify it even further, the test won't run if any assumptions are violated

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/rules/BasicTierHubOnlyTestRule.java
```diff
@@ -0,0 +1,40 @@
...
+            }
+
+            base.evaluate();
```
timtay-microsoft, line 37: This will call into all the other rules to see if the test method or its class are tagged with other annotations that could determine if the test should run or not.

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/annotations/DeviceProvisioningServiceTest.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Tests that use the Device Provisioning Service
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Inherited
+public @interface DeviceProvisioningServiceTest
+{
+
```
timtay-microsoft, line 13: No special rules for DPS tests to be defined here, but the DeviceProvisioningServiceTestRule class will check to see if a method/class is tagged with this annotation and run/skip the test accordingly

timtay-microsoft: Same for all these other annotations (other than ContinuousIntegrationTest which has a timeout as an additional rule)

iot-e2e-tests/common/src/main/java/com/microsoft/azure/sdk/iot/common/helpers/annotations/ContinuousIntegrationTest.java
```diff
@@ -0,0 +1,21 @@
...
+ * run longer than most tests, cover niche scenarios, or test failure case scenarios that are atypical for users.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Inherited
+public @interface ContinuousIntegrationTest {
+    int E2E_TEST_TIMEOUT_MS = 17 * 60 * 1000;
+
+    // Each CI test must finish in under 17 minutes
+    @Rule
```
timtay-microsoft, line 19: It's a long timeout, but only because we have a token renewal test that takes at least 10 minutes, and often runs for 16 minutes.

PR #811, opened 2020-06-17
feat(iot-dev): Adding client options for module client

bikamani:
This PR does not include any changes for createFromEnvironment, TBD whether it is needed.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
```diff
@@ -86,6 +86,30 @@ public ModuleClient(String connectionString, IotHubClientProtocol protocol) thro
...
+     *                         xxxx;moduleId=xxxx;HostNameGateway=xxxx
+     * @param protocol The protocol to use when communicating with the module
+     * @param clientOptions The options that allow configuration of the module client instance during initialization
+     * @throws ModuleClientException if an exception is encountered when parsing the connection string
+     * @throws UnsupportedOperationException if using any protocol besides MQTT, if the connection string is missing
+     * the "moduleId" field, or if the connection string uses x509
+     * @throws IllegalArgumentException if the provided connection string is null or empty, or if the provided protocol is null
+     * @throws URISyntaxException if the connection string cannot be parsed for a valid hostname
+     */
+    public ModuleClient(String connectionString, IotHubClientProtocol protocol, ClientOptions clientOptions) throws ModuleClientException, IllegalArgumentException, UnsupportedOperationException, URISyntaxException
```
abhipsaMisra, line 107: why don't other overloads of Module Client constructor have ClientOptions as well; eg one that takes in x509 certificate keys, or one that takes in the ssl context?

timtay-microsoft:
+1 to Abhipsa's point. I'd rather we not add these options to the constructor that takes certificates as strings though. I want to deprecate that constructor, not add an overload. Customer's who need to do x509 auth should use the constructor that takes the SSLContext

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
bikamani: As I had learned from Tim, this is one of those constructors we would like to deprecate going forward. @timtay-microsoft Can you add more insight to this? Or correct me if that's not right.

timtay-microsoft: The constructor that takes the certificates as strings is the one I plan to deprecate. The one that takes SSLContext should have an overload that takes the client options. See my comment below for details

bikamani: That one is already covered by this overload: public ModuleClient(String connectionString, IotHubClientProtocol protocol, ClientOptions clientOptions)

sslcontext is part of clientOptions.

PR #813, opened 2020-06-17
refactor(iot-dev): Deprecate the client constructors that don't take SSLContext for x509 authentication

timtay-microsoft:
Refactor our x509 sample to use the non-deprecated constructor, and demonstrate how to remove the bouncycastle dependency when you aren't using the deprecated constructor


abhipsaMisra:
IMO both the string keys and SSL context constructors should be deprecated, and users should set `SSLContext` via `ClientOptions` only.

timtay-microsoft:
> IMO both the string keys and SSL context constructors should be deprecated, and users should set `SSLContext` via `ClientOptions` only.

Agreed. I just noticed your suggestion on Binal's PR, and I like that approach

device/iot-device-samples/send-event-x509/src/main/java/samples/com/microsoft/azure/sdk/iot/SSLContextBuilder.java
```diff
@@ -0,0 +1,246 @@
...
+import java.security.cert.X509Certificate;
+import java.security.interfaces.RSAPrivateKey;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.UUID;
+
+/**
+ * Helper class that demonstrates how to build an SSLContext for x509 authentication from your public and private certificates,
+ * or how to build an SSLContext for SAS authentication from the default IoT Hub public certificates
```
abhipsaMisra, line 29: is the use-case here - devices that don't have the root certificates (digicert, baltimore etc) installed on their device, but need to trust those certs?

timtay-microsoft: For SAS based auth? Yes. Users of this SDK always have to have their SSLContext trust iothub's certs. This isn't something that the JVM just picks up from the physical device's certificate store

timtay-microsoft: For x509 auth, you would need to load the iothub public cert, your device's public cert, and your device's private key when building the SSLContext

abhipsaMisra: >Users of this SDK always have to have their SSLContext trust iothub's certs. This isn't something that the JVM just picks up from the physical device's certificate store

So this is something mandatory that needs to be passed by the users, and set by the SDK for every single TLS negotiation that the SDK will execute. In that case, why isn't SSLContext a mandatory argument?

timtay-microsoft: It's mandatory for establishing a connection to Iot hub, regardless of if it is x509 or SAS authentication. For SAS cases, the SDK can just build the SSLContext for the user which makes it optional at the constructor level

abhipsaMisra: ok, so for sas based authentication, what is the use-case for the customer wanting to pass this SSLContext to the SDK?

timtay-microsoft: Perhaps they have some gateway scenario set up and they need to trust a few more machines than just the iothub. Or perhaps they need to configure the TLS handshake to use a different version or something. By taking the SSLContext directly, we also let them do things like upgrade to TLS1.3 when the time comes without an SDK update

abhipsaMisra: Makes sense. I would add the note in comments for `ClientOptions.SSLConext` that the SDK will automatically create an SSL context to trust IoT hub certs, and if they pass in their custom SSLContext instance, they need to make sure to add IoT Hub's certificate info in there.

abhipsaMisra: So now a different question - the certs info that I see below in the sample, are these the public key info? How does the actual negotiation take place if the JVM cannot look up private key info in my device's certificate store/ I don't have the private key info for these certs?

timtay-microsoft: Good idea. I'll add a note to that effect in the javadoc for the clientOptions.sslContext

timtay-microsoft: The certs in the sample code's SSLContextBuilder class are indeed IoT Hub's public certificates. All SSLcontexts used must trust those certs to communicate with IoT Hub.

The user has to load their certs into the SSLContext from file or from string. Once that instance of SSLContext has the iothub public cert, their device's private key, and their device's public cert, then the SSL handshake will work for x509 auth.

PR #815, opened 2020-06-18
feat(iot-service): Add support for setting http read and connect timeouts for registry manager operations

timtay-microsoft:
Adding them to the RegistryManagerOptions class so that they are set at constructor time

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
@@ -589,7 +589,7 @@ public class RegistryManager
         String sasToken = new IotHubServiceSasToken(this.iotHubConnectionString).toString();
 
         // Codes_SRS_SERVICE_SDK_JAVA_REGISTRYMANAGER_12_049: [The function shall create a new HttpRequest for removing the device from IotHub]
-        HttpRequest request = CreateRequest(url, HttpMethod.DELETE, new byte[0], sasToken); //TODO
+        HttpRequest request = CreateRequest(url, HttpMethod.DELETE, new byte[0], sasToken);
         request.setHeaderField("If-Match", etag);
 
         // Codes_SRS_SERVICE_SDK_JAVA_REGISTRYMANAGER_12_050: [The function shall send the created request and get the response]
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
@@ -37,7 +37,6 @@ import java.util.concurrent.Executors;
  */
 public class RegistryManager
 {
-    private final Integer DEFAULT_HTTP_TIMEOUT_MS = 24000;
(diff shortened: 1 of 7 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -582,9 +589,7 @@ private void removeDeviceOperation(String deviceId, String etag) throws IOExcept
         String sasToken = new IotHubServiceSasToken(this.iotHubConnectionString).toString();
 
         // Codes_SRS_SERVICE_SDK_JAVA_REGISTRYMANAGER_12_049: [The function shall create a new HttpRequest for removing the device from IotHub]
-        HttpRequest request = new HttpRequest(url, HttpMethod.DELETE, new byte[0]);
-        request.setReadTimeoutMillis(DEFAULT_HTTP_TIMEOUT_MS);
-        request.setHeaderField("authorization", sasToken);
+        HttpRequest request = CreateRequest(url, HttpMethod.DELETE, new byte[0], sasToken); //TODO
```
abhipsaMisra, line 592: TODO?

timtay-microsoft: good catch, I'll delete this

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
```diff
@@ -9,9 +9,18 @@
 @Builder
 public class RegistryManagerOptions
 {
+    protected static final Integer DEFAULT_HTTP_READ_TIMEOUT_MS = 24000; // 24 seconds
+    protected static final Integer DEFAULT_HTTP_CONNECT_TIMEOUT_MS = 0; // infinite
```
abhipsaMisra, line 13: why infinite connect timeout?

timtay-microsoft: This was the default behavior since the SDK never set an http connect timeout before.

timtay-microsoft: By default, http connections in java have no read or connect timeouts. However, our SDK did manually set the read timeout to 24 seconds. This PR doesn't change the default behavior of the SDK

abhipsaMisra: Sure, but should the SDK set a timeout value for http connect? If the http connection cannot be established (network issue/ proxy hasn't been set up etc), then when would this actually time out and throw an exception? Would we reply on the TCP layer underneath throwing an exception (which can take minutes)?

timtay-microsoft: One could argue it wouldn't be a breaking change to start adding connect timeouts by default. The same exception would be thrown for a connect timeout as a read timeout. The only difference is the exception message. I'll add the same default of 24 seconds for the connect timeout

PR #819, opened 2020-06-24
fix(e2e): Fix issue where transport client twin tests sometimes threw a null pointer exception

timtay-microsoft:
The twin service client can be relegated to the test instance. Having a static instance was causing issues as it was sometimes prematurely disposed and would cause the NPE in a different thread.


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/TransportClientTests.java
```diff
@@ -968,8 +967,6 @@ private void tearDownTwin(TransportClient transportClient)
             // No need to fail the test just because cleanup failed.
             e.printStackTrace();
         }
-
-        sCDeviceTwin = null;
```
timtay-microsoft, line 972: Setting a shared resource to null caused NPEs, so don't do this

PR #820, opened 2020-06-24
Fix flakey twin e2e tests

timtay-microsoft:
A lot of scenarios had a 3 second timeout which is rather strict. Opening that timeout up to 1 minute so that the test isn't as flakey

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
@@ -103,9 +103,9 @@ public class DeviceTwinCommon extends IntegrationTest
 
     // Max time to wait to see it on Hub
     protected static final long PERIODIC_WAIT_TIME_FOR_VERIFICATION = 1000; // 1 sec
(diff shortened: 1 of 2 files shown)


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -103,9 +103,9 @@ public static Collection inputs() throws Exception
 
     // Max time to wait to see it on Hub
     protected static final long PERIODIC_WAIT_TIME_FOR_VERIFICATION = 1000; // 1 sec
-    protected static final long MAX_WAIT_TIME_FOR_VERIFICATION = 3000; // 30 sec
```
timtay-microsoft, line 106: This was labeled as 30 seconds, but it was actually only 3 seconds

azabbasi: And I assume 3 seconds wasn't long enough and extending the wait period will fix the flakiness?

timtay-microsoft: Yeah

PR #821, opened 2020-06-24
fix(e2e): Fix transportclient tests not waiting for devices to subscribe to methods before invoking methods

timtay-microsoft:
subscribeToDeviceMethod(...) is asynchronous (with a callback), but it was being used as if it was synchronous


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/TransportClientTests.java
```diff
@@ -389,7 +403,20 @@ public void invokeMethodInvokeParallelSucceed() throws Exception
 
         for (int i = 0; i < testInstance.clientArrayList.size(); i++)
         {
-            ((DeviceClient)testInstance.clientArrayList.get(i)).subscribeToDeviceMethod(new SampleDeviceMethodCallback(), null, new DeviceMethodStatusCallBack(), null);
+            DeviceMethodStatusCallBack subscribedCallback = new DeviceMethodStatusCallBack();
```
bikamani, line 406: just wondering why is this one tagged as _ContinuousIntegrationTest_ and the other one above is not?

timtay-microsoft: The CI test is just a parallelized flavor of the other test. For our check in builds, the simpler test is sufficient for coverage

PR #822, opened 2020-06-24
feat(iot-service): Add model id support for PnP

bikamani:


service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinTest.java
```diff
@@ -946,20 +946,19 @@ public void nextRetrievesCorrectlyWithoutModuleId() throws IotHubException, IOEx
         final String connectionString = "testString";
         DeviceTwin testTwin = DeviceTwin.createFromConnectionString(connectionString);
         final String expectedString = "testJsonAsNext";
+        final String modelId = "testModelId";
```
bikamani, line 949: Unit Test to check whether this value is set correctly for Twin

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/RegisterManager.java
```diff
@@ -123,6 +125,17 @@
...
+     * The Digital Twin model id of the device and module
+     * The value will be null for a non-pnp device.
+     * The value will be null for a pnp device until the device connects and registers with the model id.
+     */
+    private static final String MODEL_ID = "digital-twin-model-id";
+    @Expose(serialize = true, deserialize = true)
+    @SerializedName(MODEL_ID)
+    @Setter
+    @Getter
+    protected String modelId = null;
```
prmathur-microsoft, line 138: Is this needed now that modelid sits on device twin?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
```diff
@@ -536,6 +542,10 @@ public String toString()
         {
             thisDevice.append("Version: " + this.getVersion() + "\n");
         }
+        if(this.modelId != null && !this.modelId.isEmpty())
+        {
```
prmathur-microsoft, line 546: Why is this check needed? What would happen if a null or empty is placed on parser.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/RegisterManager.java
timtay-microsoft: +1 on this comment. This new field should only exist on twin, not identity

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/RegisterManager.java
```diff
@@ -8,6 +8,8 @@
 import com.microsoft.azure.sdk.iot.deps.serializer.ConfigurationContentParser;
```
prmathur-microsoft, line 8: Questions : 1. Where is this verified to be working?
2. Will e2e be another PR? Is the task captured?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
```diff
@@ -504,6 +505,11 @@ private DeviceTwinDevice jsonToDeviceTwinDevice(String json) throws IOException
         deviceTwinDevice.setConnectionState(twinState.getConnectionState());
         deviceTwinDevice.setConfigurations(twinState.getConfigurations());
 
+        if (twinState.getModelId() != null && !twinState.getModelId().isEmpty())
+        {
```
prmathur-microsoft, line 509: Same comment here

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
```diff
@@ -504,6 +505,11 @@ private DeviceTwinDevice jsonToDeviceTwinDevice(String json) throws IOException
...
+            deviceTwinDevice.setModelId(twinState.getModelId());
```
abhipsaMisra, line 510: unrelated to this PR, but why did we name this `DeviceTwinDevice`? Is it to separate a device twin from a module twin? Is it public? Can we rename it?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/RegisterManager.java
bikamani: Synced offline on this. This class is used for Twin and this field exist on twin only.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/RegisterManager.java
bikamani: Verified using a sample.
I added e2e task for this PBI, thanks for reminding

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
bikamani: synced offline on this.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
bikamani: I would like to understand if there is a reasoning behind why should we set model id when null/empty?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
timtay-microsoft: This is a toString function. There's no reason to print the modelId if there is no modelId

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
prmathur-microsoft: To avoid SDK's having any check. The point I was making is we should leave it to service to throw incase there is no model id and not customize it here.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
prmathur-microsoft: I see that this pattern is followed everywhere. Looking back I would have just printed the data as is i.e. along with null to represent what has actually come on the wire from service. However now I don't think we can make that change and should follow the suit.

bikamani: If we want to print out null/empty values, this check is not needed. I followed the pattern similar to module id and others as it doesn't seem to provide any value printing out null/empty

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
bikamani: ok. done

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
```diff
@@ -504,11 +504,7 @@ private DeviceTwinDevice jsonToDeviceTwinDevice(String json) throws IOException
         deviceTwinDevice.setCapabilities(twinState.getCapabilities());
         deviceTwinDevice.setConnectionState(twinState.getConnectionState());
         deviceTwinDevice.setConfigurations(twinState.getConfigurations());
-
```
prmathur-microsoft, line 506: Have you verified that this works for non PnP device before making the change?

bikamani: yes for non pnp this value is set to null, when printed it will look like below:

Device ID: <id>
ETag: <etag>
**Model ID:** 
Tags:{}
Reported Properties{}
Desired Properties: {}

prmathur-microsoft: 👍

PR #823, opened 2020-06-25
feat(iot-dev): Adding client options for modules (edge)

bikamani:


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
```diff
@@ -185,13 +185,36 @@ public static ModuleClient createFromEnvironment() throws ModuleClientException
         return createFromEnvironment(IotHubClientProtocol.AMQPS);
     }
 
+    /**
+     * Create a module client instance from your environment variables
+     * @param clientOptions The options that allow configuration of the module client instance during initialization
+     * @return the created module client instance
+     * @throws ModuleClientException if the module client cannot be created
+     */
+    public static ModuleClient createFromEnvironment(ClientOptions clientOptions) throws ModuleClientException
```
timtay-microsoft, line 194: This is a tricky constructor in practice. Right now, AMQP is the default protocol, and the modelId changes are only applicable for MQTT. ClientOptions does encapsulate the SSLContext though, which may make this constructor worthwile. Can you double check that the SSLContext is used correctly from this constructor?

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/ModuleClientTest.java
```diff
@@ -135,6 +138,38 @@ public void constructorThrowsForHTTP() throws URISyntaxException, ModuleClientEx
         new ModuleClient(connectionString, IotHubClientProtocol.HTTPS);
     }
 
+    @Test
+    public void constructorWithModelIdSuccess(final @Mocked System mockedSystem) throws URISyntaxException, IOException, ModuleClientException {
+        //arrange
+        final IotHubClientProtocol protocol = IotHubClientProtocol.AMQPS;
```
abhipsaMisra, line 144: use mqtt here instead

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
timtay-microsoft: Especially check the case when the user provides certs in the environment variables

bikamani: yes right, I'll change default to MQTT. Also checked, ssl context will be set correctly with this constructor by DeviceClientConfig

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/ModuleClientTest.java
bikamani: will do

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
timtay-microsoft: What happens if I pass in an SSLContext in the clientOptions and have extra trusted certs in my environment variables? Does the SSLContext take precedence? If so, you'll need to add a note in the javadoc about that

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/ModuleClientTest.java
bikamani: Rethinking about this, it is actually fine to leave it as it is. Here we just want to make sure model id is set correctly with this constructor no matter which protocol

PR #825, opened 2020-06-26
Fix a few flakey tests, add more context when logging transport messages

timtay-microsoft:


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransportMessage.java
```diff
@@ -293,6 +293,11 @@ public String toString()
             base.append("Request Id [").append(this.requestId).append("] ");
         }
 
+        if (this.getDeviceOperationType() != null && this.getDeviceOperationType() != DeviceOperations.DEVICE_OPERATION_UNKNOWN)
+        {
+            base.append("Device Operation Type [").append(this.getDeviceOperationType()).append("] ");
```
timtay-microsoft, line 298: This information has been useful to me when debugging these flakey tests, and I suspect customer's will appreciate this too

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/FileUploadTests.java
```diff
@@ -304,7 +304,8 @@ private void verifyNotification(FileUploadNotification fileUploadNotification, F
         try (InputStream inputStream = u.openStream())
         {
             byte[] testBuf = new byte[(int)fileUploadState.fileLength];
-            int testLen = inputStream.read(testBuf,  0, (int)fileUploadState.fileLength);
+            inputStream.read(testBuf,  0, (int)fileUploadState.fileLength);
+            int testLen = (int)fileUploadState.fileLength;
```
timtay-microsoft, line 308: The commit message explains this further, but Android SDKs return -1 when inputStream.read(...) reads no bytes. JDKs return 0 though. For consistency between JDK and android SDK, we shouldn't use this return value

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -159,25 +160,20 @@ public static Collection inputs() throws Exception
     // How much to wait until a message makes it to the server, in milliseconds
     protected static final Integer SEND_TIMEOUT_MILLISECONDS = 60000;
 
-    public enum STATUS
```
timtay-microsoft, line 162: This status class was used to simplify twin callbacks, but it hid useful information for failure cases. Knowing why a callback fired back with an unsuccessful state is important to log when the test fails

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -230,13 +226,18 @@ public void PropertyCall(String propertyKey, Object propertyValue, Object contex
             }
         }
 
-        public synchronized void createNewReportedProperties(int maximumPropertiesToCreate)
+        public synchronized Set<Property> createNewReportedProperties(int maximumPropertiesToCreate)
         {
+            Set<Property> newProperties = new HashSet<>();
```
timtay-microsoft, line 231: only return the newly created property

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/twin/ReportedPropertiesTests.java
```diff
@@ -81,14 +85,14 @@ public void run()
                 {
                     try
                     {
-                        deviceUnderTest.dCDeviceForTwin.createNewReportedProperties(1);
-                        internalClient.sendReportedProperties(deviceUnderTest.dCDeviceForTwin.getReportedProp());
+                        Set<Property> newReportedProperties = deviceUnderTest.dCDeviceForTwin.createNewReportedProperties(1);
+                        internalClient.sendReportedProperties(newReportedProperties);
```
timtay-microsoft, line 89: It's important to only send this thread's new reported property. If it sends the device's full reported property set, then iothub will receive conflicting sets of properties. 

We would send two twin messages to hub. one with:
```
{
  "key1":"oldValue1",
  "key2":"someUpdatedValue2"
}
```

and another with 

```
{
  "key1":"someUpdatedValue1",
  "key2":"someUpdatedValue2"
}
```

but iothub may process them out of order, and the resulting twin representation on the service side could be either

```
{
  "key1":"someUpdatedValue1",
  "key2":"someUpdatedValue2"
}
```

or the incorrect representation

```
{
  "key1":"oldValue1",
  "key2":"someUpdatedValue2"
}
```

timtay-microsoft: The test would fail in the latter case because it was expecting a final state of 
```
{
  "key1":"someUpdatedValue1",
  "key2":"someUpdatedValue2"
}
```

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/twin/ReportedPropertiesTests.java
```diff
@@ -153,31 +157,31 @@ public void testUpdateReportedPropertiesMultiThreaded() throws IOException, Inte
...
-                }
-                catch (IOException | InterruptedException e)
+                @Override
+                public void run()
                 {
-                    fail(CorrelationDetailsLoggingAssert.buildExceptionMessage("Unexpected exception occurred during sending reported properties: " + Tools.getStackTraceFromThrowable(e), internalClient));
+                    try
+                    {
+                        Set<Property> updatedProperties = deviceUnderTest.dCDeviceForTwin.updateExistingReportedProperty(index);
+                        internalClient.sendReportedProperties(updatedProperties);
```
timtay-microsoft, line 175: Again, only send the reported properties that belong to this thread. See above for why

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/twin/ReportedPropertiesTests.java
timtay-microsoft: So the fix here is to just send one property at a time, as it gets updated.

```
{
  "key1":"someUpdatedValue1"
}
```

and

```
{
  "key2":"someUpdatedValue2"
}
```

so that iothub's representation of this twin will always be 

```
{
  "key1":"someUpdatedValue1",
  "key2":"someUpdatedValue2"
}
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransportMessage.java
timtay-microsoft: This will log if a message is a twin subscription message, method subscription message, etc.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/FileUploadTests.java
azabbasi: that is unfortunate ... good thinking.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -159,25 +160,20 @@ public static Collection inputs() throws Exception
...
-        SUCCESS, FAILURE, UNKNOWN
-    }
-
     public class DeviceTwinStatusCallBack implements IotHubEventCallback
     {
         public void execute(IotHubStatusCode status, Object context)
         {
             DeviceState state = (DeviceState) context;
 
-            //On failure, Don't update status any further
```
azabbasi, line 173: how is this going to affect the callback? is this considered a breaking change?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -259,8 +260,9 @@ public synchronized void updateAllExistingReportedProperties()
             }
         }
 
-        public synchronized void updateExistingReportedProperty(int index)
+        public synchronized Set<Property> updateExistingReportedProperty(int index)
```
azabbasi, line 263: this method is no longer doing what its name implies does it? it's creating and returning a new set

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
timtay-microsoft: This is in the test code, so it isn't a breaking change, no

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
timtay-microsoft: It still does what it says. It only returns the property that was updated, but that returned property did already exist

timtay-microsoft: By the way, this and other methods always return Set<Property> even when only one property is in the set. The reason for this is so that the test method can use this returned value directly when calling "sendReportedProperties(Set<Property>)" instead of building the set after calling this method

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
azabbasi: ah, the file path threw me off

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
azabbasi: I see, so it updates the existing and also returns a new set as a return value. fair.

PR #826, opened 2020-06-26
refactor(iot-dev): Use wait and notify to cut down on useless thread creation in transport layer

timtay-microsoft:
The current design of the device client library periodically spawns threads to handle sending and receiving messages as well as invoking callbacks. The device client will spawn these threads regardless of if any messages actually need to be sent/received, though.

With this change, the send/receive threads will instead wait until a message requires them to wake up. This will cut down on thread creation during less busy times without sacrificing performance during busy times

The second commit for this PR makes our DeviceIO layer more responsive to connection change events. Previously, if the transport layer went into reconnecting mode, the device IO layer would still spawn send/receive threads repeatedly even though they would be unable to send/receive anything since the transport isn't connected.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
@@ -191,6 +191,11 @@ public final class DeviceIO
         this.receiveTask = new IotHubReceiveTask(this.transport);
 
         this.taskScheduler = Executors.newScheduledThreadPool(2);
+
+        // Note that even though these threads are scheduled at a fixed interval, the sender/receiver threads will wait
+        // if no messages are available to process. These waiting threads will still count against the pool size defined above,
+        // so threads will not be needlessly scheduled during times when this SDK has no messages to process.
+
         // the scheduler waits until each execution is finished before
         // scheduling the next one, so executions of a given task
         // will never overlap.
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
@@ -4,6 +4,7 @@
 package com.microsoft.azure.sdk.iot.device;
 
 import com.microsoft.azure.sdk.iot.device.exceptions.DeviceClientException;
+import com.microsoft.azure.sdk.iot.device.transport.IotHubConnectionStatus;
 import com.microsoft.azure.sdk.iot.device.transport.IotHubReceiveTask;
 import com.microsoft.azure.sdk.iot.device.transport.IotHubSendTask;
 import com.microsoft.azure.sdk.iot.device.transport.IotHubTransport;
@@ -68,14 +69,8 @@ import java.util.concurrent.TimeUnit;
  * The task scheduler for sending and receiving messages for the Device Client
  */
 @Slf4j
-public final class DeviceIO
+public final class DeviceIO implements IotHubConnectionStatusChangeCallback
 {
-    /** The state of the IoT Hub client's connection with the IoT Hub. */
-    protected enum IotHubClientState
-    {
-        OPEN, CLOSED
-    }
-
     private long sendPeriodInMilliseconds;
     private long receivePeriodInMilliseconds;
 
@@ -86,7 +81,7 @@ public final class DeviceIO
     private IotHubClientProtocol protocol = null;
 
     private ScheduledExecutorService taskScheduler;
-    private IotHubClientState state;
+    private IotHubConnectionStatus state;
 
     private List<DeviceClientConfig> deviceClientConfigs = new LinkedList<>();
 
@@ -119,14 +114,14 @@ public final class DeviceIO
         this.receivePeriodInMilliseconds = receivePeriodInMilliseconds;
 
         /* Codes_SRS_DEVICE_IO_21_006: [The constructor shall set the `state` as `DISCONNECTED`.] */
-        this.state = IotHubClientState.CLOSED;
+        this.state = IotHubConnectionStatus.DISCONNECTED;
 
         if (protocol == IotHubClientProtocol.AMQPS_WS || protocol == IotHubClientProtocol.MQTT_WS)
         {
             this.config.setUseWebsocket(true);
         }
 
-        this.transport = new IotHubTransport(config);
+        this.transport = new IotHubTransport(config, this);
 
         /* Codes_SRS_DEVICE_IO_21_037: [The constructor shall initialize the `sendPeriodInMilliseconds` with default value of 10 milliseconds.] */
         this.sendPeriodInMilliseconds = sendPeriodInMilliseconds;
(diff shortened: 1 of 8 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -68,14 +69,8 @@
  * The task scheduler for sending and receiving messages for the Device Client
  */
 @Slf4j
-public final class DeviceIO
+public final class DeviceIO implements IotHubConnectionStatusChangeCallback
 {
-    /** The state of the IoT Hub client's connection with the IoT Hub. */
-    protected enum IotHubClientState
```
timtay-microsoft, line 74: This type was pretty redundant, and couldn't capture the same state as the IotHubTransport layer had, so I removed this type and just re-used the IotHubConnectionStatus class for this purpose

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -119,22 +114,22 @@
...
-        this.state = IotHubClientState.CLOSED;
+        this.state = IotHubConnectionStatus.DISCONNECTED;
 
         if (protocol == IotHubClientProtocol.AMQPS_WS || protocol == IotHubClientProtocol.MQTT_WS)
         {
             this.config.setUseWebsocket(true);
         }
 
-        this.transport = new IotHubTransport(config);
+        this.transport = new IotHubTransport(config, this);
```
timtay-microsoft, line 124: Passing ```this``` in as a parameter so that the transpor layer can callback up to this layer when the connection status changes

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -161,10 +156,6 @@ void open() throws IOException
         {
             throw new IOException("Could not open the connection", e);
         }
-
-        /* Codes_SRS_DEVICE_IO_21_014: [The open shall schedule receive tasks to run every receivePeriodInMilliseconds milliseconds.] */
-        /* Codes_SRS_DEVICE_IO_21_016: [The open shall set the `state` as `CONNECTED`.] */
-        commonOpenSetup();
```
timtay-microsoft, line 167: This logic was moved to the connection status callback. That way, this method (which starts up the send/receive thread pool) gets called every time the client goes from DISCONNECTED or DISCONNECTED_RETRYING to CONNECTED. This allows the thread pool to not be active during DISCONNECTED_RETRYING

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -429,4 +425,25 @@ public void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeC
         //Codes_SRS_DEVICE_IO_34_020: [This function shall register the callback with the transport.]
         this.transport.registerConnectionStatusChangeCallback(statusChangeCallback, callbackContext);
     }
+
+    /*
+     * IotHubTransport layer will notify this layer when the connection is established and when it is lost. This layer should start/stop
+     * the send/receive threads accordingly
+     */
+    @Override
+    public void execute(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable, Object callbackContext)
```
timtay-microsoft, line 434: this method will be called back from the transport layer every time the state changes. This class should start/stop the send/receive thread pool and then update its own state to be consistent with the transport state

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubReceiveTask.java
```diff
@@ -32,11 +40,24 @@ public void run()
...
+            // wait on the IoTHubTransport layer to notify this thread that a received message is ready to be handled.
+            if (this.transport.getProtocol() != IotHubClientProtocol.HTTPS)
+            {
+                synchronized (this.receiveThreadLock)
+                {
+                    if (!this.transport.hasReceivedMessagesToHandle() && !this.transport.isClosed())
+                    {
+                        // AMQP and MQTT layers will notify the IoTHubTransport layer once a message arrives, and at
+                        // that time, this thread will be notified to handle them.
+                        this.receiveThreadLock.wait();
```
timtay-microsoft, line 53: See [here](https://www.baeldung.com/java-wait-notify#the-wait-method) for a more in-depth explanation on wait/notify, but the short of it is that this thread goes dormant until notified. When dormant, this thread has minimal performance impact on the machine running the SDK

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -185,24 +177,30 @@ void addClient(DeviceClientConfig config)
     /**
      * Handles logic common to all open functions.
      */
-    private void commonOpenSetup()
+    private void startWorkerThreads()
     {
         this.sendTask = new IotHubSendTask(this.transport);
         this.receiveTask = new IotHubReceiveTask(this.transport);
 
-        this.taskScheduler = Executors.newScheduledThreadPool(2);
```
timtay-microsoft, line 193: Splitting this into two separate schedulers so that there is no chance of this thread pool of size 2 having two send tasks active at once, and both waiting for work to arrive. If that were to happen, then no receive threads will spawn, so any received messages will be stuck.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -146,7 +142,7 @@
     void open() throws IOException
     {
         /* Codes_SRS_DEVICE_IO_21_007: [If the client is already open, the open shall do nothing.] */
-        if (this.state == IotHubClientState.OPEN)
+        if (this.state == IotHubConnectionStatus.CONNECTED || this.state == IotHubConnectionStatus.DISCONNECTED_RETRYING)
```
abhipsaMisra, line 145: how are these managed when calls are made in parallel - i.e. I call open() over one thread, and then close() over another thread, at the same time.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -429,4 +432,33 @@ public void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeC
...
+                this.receiveTaskScheduler.shutdown();
+            }
+        }
+        else if (status == IotHubConnectionStatus.CONNECTED)
+        {
+            // Restart the task scheduler so that send/receive tasks start spawning again
+            this.startWorkerThreads();
+        }
+
+        this.state = status;
```
abhipsaMisra, line 462: shouldn't these status updates be made over a lock?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: I can add a state lock for that, sure. Right now, we just hope customers don't do that

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: Sure, I'll protect this with the same lock that I'll add for open/close

PR #828, opened 2020-06-26
Run TPM tests in serial

timtay-microsoft:
~~Two small e2e test changes~~. The TPM tests need to run in serial since they each need exclusive access to the tpm simulator in order to pass.

~~The twin tear down logic should never throw an exception that causes the test to fail since the teardown logic is part of what's being tested~~

### DIFF
--- a/iot-e2e-tests/common/pom.xml
+++ b/iot-e2e-tests/common/pom.xml
@@ -93,6 +93,12 @@
             <version>1.18.8</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>net.jcip</groupId>
+            <artifactId>jcip-annotations</artifactId>
+            <version>1.0</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
@@ -373,7 +373,7 @@ public class DeviceTwinCommon extends IntegrationTest
         }
     }
 
-    public void tearDownTwin(DeviceState deviceState) throws IOException
+    public void tearDownTwin(DeviceState deviceState)
     {
         // tear down twin on device client
         if (deviceState != null)
@@ -394,7 +394,14 @@ public class DeviceTwinCommon extends IntegrationTest
         }
         if (internalClient != null)
         {
-            internalClient.closeNow();
+            try
+            {
+                internalClient.closeNow();
+            }
+            catch (IOException e)
+            {
+                // Ignore, the test isn't about being able to close the client
+            }
             internalClient = null;
         }
     }
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
@@ -373,7 +373,7 @@ public class DeviceTwinCommon extends IntegrationTest
         }
     }
 
-    public void tearDownTwin(DeviceState deviceState)
+    public void tearDownTwin(DeviceState deviceState) throws IOException
     {
         // tear down twin on device client
         if (deviceState != null)
@@ -394,14 +394,7 @@ public class DeviceTwinCommon extends IntegrationTest
         }
         if (internalClient != null)
         {
-            try
-            {
-                internalClient.closeNow();
-            }
-            catch (IOException e)
-            {
-                // Ignore, the test isn't about being able to close the client
-            }
+            internalClient.closeNow();
             internalClient = null;
         }
     }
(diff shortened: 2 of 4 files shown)


iot-e2e-tests/common/pom.xml
```diff
@@ -93,6 +93,12 @@
             <version>1.18.8</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>net.jcip</groupId>
+            <artifactId>jcip-annotations</artifactId>
+            <version>1.0</version>
+            <scope>test</scope>
+        </dependency>
```
timtay-microsoft, line 101: This dependency gives us the NotThreadSafe annotation

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -394,7 +394,14 @@ public void tearDownTwin(DeviceState deviceState) throws IOException
...
         if (internalClient != null)
         {
-            internalClient.closeNow();
+            try
+            {
+                internalClient.closeNow();
+            }
+            catch (IOException e)
+            {
+                // Ignore, the test isn't about being able to close the client
```
azabbasi, line 403: why does it throw in those flaky instances? doesn't that indicate a problem? doesn't swallowing that exception mean we are ignoring a potential issue?

timtay-microsoft: We have tests where closing the client is part of the code under test, but this isn't one of them

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/setup/ProvisioningCommon.java
```diff
@@ -141,17 +141,29 @@ public static Collection inputsCommon(AttestationType attestationType)
...
-                                    {ProvisioningDeviceClientTransportProtocol.AMQPS, attestationType},
-                                    {ProvisioningDeviceClientTransportProtocol.AMQPS_WS, attestationType}
-
-                                    //MQTT/MQTT_WS does not support tpm attestation
-                                    //{ProvisioningDeviceClientTransportProtocol.MQTT, attestationType},
-                                    //{ProvisioningDeviceClientTransportProtocol.MQTT_WS, attestationType},
-                            });
+            if (!isPullRequest)
+            {
+                //TODO TPM tests are flakey, so only run them for CI and nightly builds
```
bikamani, line 146: did you forget to remove this?

timtay-microsoft: No, this is more of an explanation of why we don't run these tests for pull request builds

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/setup/ProvisioningCommon.java
```diff
@@ -141,17 +141,29 @@ public static Collection inputsCommon(AttestationType attestationType)
...
+                return Arrays.asList(
+                        new Object[][]
+                                {
+                                        {ProvisioningDeviceClientTransportProtocol.HTTPS, attestationType},
```
bikamani, line 150: So was this not referenced earlier?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/setup/ProvisioningCommon.java
timtay-microsoft: The TODO is just me being optimistic that we will eventually fix these tests to not be flakey

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/setup/ProvisioningCommon.java
timtay-microsoft: The TPM tests were being run previously. They were just returned as inputs above alongside x509 and symmetric key. This PR just splits the TPM tests into their own class so they can be run in serial while still running the x509 and symmetric key tests in parallel

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
azabbasi: that doesn't explain why it throws here though

timtay-microsoft: I can take this out for now. I'm more interested in the TPM changes anyways

timtay-microsoft: I was under the impression that we generally don't fail a test if the cleanup fails, but this may be worth more discussion

PR #831, opened 2020-07-01
fix(iot-dev): Fix issue where clients couldn't queue messages while reconnecting

timtay-microsoft:
InternalClient methods such as subscribeToTwinDesiredProperties check this ```isOpen()``` method before queueing messages.

The previous behavior was to allow users to queue messages and send subscription messages even when the client is in a reconnecting state.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
@@ -146,7 +146,7 @@ public final class DeviceIO implements IotHubConnectionStatusChangeCallback
     {
         synchronized (this.stateLock)
         {
-            if (this.state == IotHubConnectionStatus.CONNECTED || this.state == IotHubConnectionStatus.DISCONNECTED_RETRYING)
+            if (this.isOpen())
             {
                 return;
             }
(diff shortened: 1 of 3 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -400,12 +400,13 @@ public IotHubClientProtocol getProtocol()
     /**
      * Getter for the connection state.
      *
-     * @return a boolean true if the connection is open, or false if it is closed.
+     * @return a boolean true if the connection is open or reconnecting, and false otherwise.
      */
```
prmathur-microsoft, line 404: 👍

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -146,7 +146,7 @@ void open() throws IOException
     {
         synchronized (this.stateLock)
         {
-            if (this.state == IotHubConnectionStatus.CONNECTED || this.state == IotHubConnectionStatus.DISCONNECTED_RETRYING)
+            if (this.isOpen())
```
prmathur-microsoft, line 149: Do we already have e2e/UT reflecting this behavior?

timtay-microsoft: Just added a unit test for this

timtay-microsoft: Some of our e2e tests were already hitting this issue sometimes which is why I caught this in the first place

prmathur-microsoft: 👍

PR #832, opened 2020-07-01
fix(e2e): Fix fault injection message tests

timtay-microsoft:
Iot Hub responds to fault injection messages differently for AMQP and MQTT. For AMQP, the fault injection message will be ack'd before the fault takes effect, but this isn't the case for MQTT. To make it more reliable that the fault injection message gets sent, AMQP tests should wait for this FI message to be ack'd. MQTT tests still need to provide an expiry time, but that time has been bumped up from .1 or .2 seconds to 1 second to increase the chances that it gets delivered at least once

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ErrorInjectionHelper.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ErrorInjectionHelper.java
@@ -31,7 +31,7 @@ public class ErrorInjectionHelper
     public static final String FaultCloseReason_Boom = "Boom";
     public static final String FaultCloseReason_Bye = "byebye";
 
-    public static final int DefaultDelayInSec = 1;
+    public static final int DefaultDelayInSec = 5;
     public static final int DefaultDurationInSec = 5;
 
     public static Message tcpConnectionDropErrorInjectionMessage(int delayInSecs, int durationInSecs)
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/IotHubServicesCommon.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/IotHubServicesCommon.java
@@ -23,6 +23,11 @@ import static tests.integration.com.microsoft.azure.sdk.iot.helpers.CorrelationD
  */
 public class IotHubServicesCommon
 {
+    protected static final Integer RETRY_MILLISECONDS = 100;
+    protected static final Integer SEND_TIMEOUT_MILLISECONDS = 60000;
+
+    public static final int ERROR_INJECTION_MESSAGE_TIMEOUT_MILLISECONDS = 200;
+
     //if error injection message has not taken effect after 1 minute, the test will timeout
     private final static long ERROR_INJECTION_MESSAGE_EFFECT_TIMEOUT_MILLISECONDS = 1 * 60 * 1000;
     private final static String TEST_ASC_SECURITY_MESSAGE = "{ \"AgentVersion\": \"0.0.1\", "
@@ -60,7 +65,7 @@ public class IotHubServicesCommon
                     //error injection message is not guaranteed to be ack'd by service so it may be re-sent. By setting expiry time,
                     // we ensure that error injection message isn't resent to service too many times. The message will still likely
                     // be sent 3 or 4 times causing 3 or 4 disconnections, but the test should recover anyways.
-                    messagesToSend.get(i).message.setExpiryTime(200);
+                    messagesToSend.get(i).message.setExpiryTime(ERROR_INJECTION_MESSAGE_TIMEOUT_MILLISECONDS);
                 }
 
                 sendMessageAndWaitForResponse(client, messagesToSend.get(i), RETRY_MILLISECONDS, SEND_TIMEOUT_MILLISECONDS, protocol);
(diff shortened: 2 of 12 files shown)


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/IotHubServicesCommon.java
```diff
@@ -23,6 +23,11 @@
  */
 public class IotHubServicesCommon
 {
+    protected static final Integer RETRY_MILLISECONDS = 100;
+    protected static final Integer SEND_TIMEOUT_MILLISECONDS = 60000;
+
+    public static final int ERROR_INJECTION_MESSAGE_TIMEOUT_MILLISECONDS = 200;
```
timtay-microsoft, line 29: We set a short expiry time on these fault injection messages because they often fail to be acknowledged by the service before the connection is lost. With this expiry time, our retry logic wont send this message more than a few times. At the same time though, we risk them never being sent. That's why I added some retry logic to prevent that

abhipsaMisra: What would cause the service to drop the connection before ack'ing the receipt of this fault injection request? Afaik, service will drop the connection on the next message that they receive (after the FI request message). So this should mean that the FI request message would always be ack'ed.
In java, do we not wait until message 1 is ack'ed, before sending message 2? Is that because of the way the send message queue works?

timtay-microsoft: > Afaik, service will drop the connection on the next message that they receive (after the FI request message)

That hasn't been my experience. The Java tests don't even send another message until after the connection is dropped. From my experience, the service may or may not ack the fault injection message before dropping the connection.

timtay-microsoft: I'll spend some time investigating if we can rely on the service ack'ing these fault injection messages before merging this in

timtay-microsoft:
It turns out that fault injection tests will behave differently for MQTT and AMQP. For MQTT, the fault is immediately injected upon the service receiving the FI message, with no sending of an ACK. The C# SDK does use a timeout trick like the Java SDK to prevent an infinite loop of sending the FI message: https://github.com/Azure/azure-iot-sdk-csharp/blob/d0332ab3dc3c0b00d80483c4a7edeb0449fc95c2/e2e/test/FaultInjection.cs#L89

timtay-microsoft:
For AMQP though, IotHub does seem to send an ack for the FI message, and then respects the delayInSeconds that the FI message defines. I'll try to follow the C# approach since the C# builds don't see fault injection failures as often

abhipsaMisra:
> It turns out that fault injection tests will behave differently for MQTT and AMQP. For MQTT, the fault is immediately injected upon the service receiving the FI message, with no sending of an ACK. The C# SDK does use a timeout trick like the Java SDK to prevent an infinite loop of sending the FI message: https://github.com/Azure/azure-iot-sdk-csharp/blob/d0332ab3dc3c0b00d80483c4a7edeb0449fc95c2/e2e/test/FaultInjection.cs#L89

Good point, yeah! I remembered this after our meeting, Mqtt does terminate the connection immediately for all "kill connection" FI requests.

abhipsaMisra:
> For AMQP though, IotHub does seem to send an ack for the FI message, and then respects the delayInSeconds that the FI message defines. I'll try to follow the C# approach since the C# builds don't see fault injection failures as often

Fwiw I wouldn't rely on the `delayInSeconds`; afaik they respect it only for hub level faults (unauth, quota exceeded etc), and isn't super reliable.

timtay-microsoft:
> Azure Pipelines successfully started running 4 pipeline(s).

It waits about a second before faulting, and that should be long enough, thankfully

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ErrorInjectionHelper.java
```diff
@@ -31,7 +31,7 @@
     public static final String FaultCloseReason_Boom = "Boom";
     public static final String FaultCloseReason_Bye = "byebye";
 
-    public static final int DefaultDelayInSec = 1;
```
timtay-microsoft, line 34: C# SDK uses 5 here, but the service doesn't appear to respect this value. Just in case though, I want to unify with the C# SDK

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/IotHubServicesCommon.java
```diff
@@ -55,12 +55,16 @@ public static void sendMessages(InternalClient client,
 
             for (int i = 0; i < messagesToSend.size(); ++i)
             {
-                if (isErrorInjectionMessage(messagesToSend.get(i)))
+                if ((protocol == IotHubClientProtocol.MQTT || protocol == IotHubClientProtocol.MQTT_WS) && isErrorInjectionMessage(messagesToSend.get(i)))
```
abhipsaMisra, line 58: Q - why do we need these changes here, and in `sendErrorInjectionMessageAndWaitForResponse` both?

timtay-microsoft: I don't think there is a good reason. There's some logic here that could certainly be centralized. I'll take that in a separate PR though

PR #833, opened 2020-07-02
fix(prov-dev, prov-serv): Fix provisioning clients not sending content length with http requests

timtay-microsoft:
DPS would occasionally return a 411 error because this wasn't being sent on registration requests and enrollment creations

It doesn't appear to be a hard requirement since tests usually pass without them, but just to be safe we should send them. The 411 failures seem to disappear once this change is made.

### DIFF
--- a/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/contract/http/ContractAPIHttp.java
+++ b/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/contract/http/ContractAPIHttp.java
@@ -46,6 +46,7 @@ public class ContractAPIHttp extends ProvisioningDeviceClientContract
     private static final String ACCEPT_VALUE = "application/json";
     private static final String ACCEPT_CHARSET = "charset=utf-8";
     private static final String CONTENT_TYPE = "Content-Type";
+    private static final String CONTENT_LENGTH = "Content-Length";
     private static final Integer DEFAULT_HTTP_TIMEOUT_MS = Integer.MAX_VALUE;
     private static final Integer ACCEPTABLE_NONCE_HTTP_STATUS = 401;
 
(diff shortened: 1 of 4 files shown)


provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/contract/ContractApiHttp.java
```diff
@@ -183,6 +184,7 @@ private HttpRequest createRequest(URL url, HttpMethod method, Map<String, String
         request.setHeaderField(HEADER_FIELD_NAME_ACCEPT, HEADER_FIELD_VALUE_ACCEPT);
         request.setHeaderField(HEADER_FIELD_NAME_CONTENT_TYPE, HEADER_FIELD_VALUE_CONTENT_TYPE);
         request.setHeaderField(HEADER_FIELD_NAME_CHARSET, HEADER_FIELD_VALUE_CHARSET);
+        request.setHeaderField(HEADER_FIELD_NAME_CONTENT_LENGTH, payload != null ? String.valueOf(payload.length) : "0");
```
bikamani, line 187: If this is not a hard requirement, why was it throwing failure only occasionally?

provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/contract/http/ContractAPIHttp.java
```diff
@@ -126,6 +127,7 @@ private HttpRequest prepareRequest(
         request.setHeaderField(USER_AGENT, userAgentValue);
         request.setHeaderField(ACCEPT, ACCEPT_VALUE);
         request.setHeaderField(CONTENT_TYPE, ACCEPT_VALUE + "; " + ACCEPT_CHARSET);
+        request.setHeaderField(CONTENT_LENGTH, payload != null ? String.valueOf(payload.length) : "0");
```
bikamani, line 130: Do we not have any e2e tests for this change?

timtay-microsoft: Our e2e tests already hit both the device and service side of this change. The fact that I've stopped seeing those 411 errors is about as close as I can get to verifying that this works.

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/contract/ContractApiHttp.java
timtay-microsoft: I'm not sure. Service folks didn't answer me when I asked if it was a hard requirement. This seems like a safe change to make regardless, though

PR #838, opened 2020-07-06
refactor(e2e): Fix query tests

timtay-microsoft:
Now every query test will re-run the query until the expected conditions are met or a timeout occurs. This helps to account for any propagation delays on the service side between device creation and when the device is queryable.

By inheriting from DeviceTwinCommon, each query test was run for every device side protocol and for every device side authentication. These query tests are entirely service client tests, though. It doesn't make sense to parameterize them on device client settings


vsts/windowsLinuxAndAndroidBuildMatrixConfig.yaml
```diff
@@ -241,7 +243,7 @@ jobs:
       - task: Bash@3
         condition: eq(variables['task.android.needToRunTestGroup'], 'yes')
         displayName: 'Start Android Emulator'
-        timeoutInMinutes: 10
```
timtay-microsoft, line 244: The emulator tasks have hit this 10 minute mark and failed even though the emulator is ready to use in the next step, so I've bumped up the timeout to allow for slower emulator start up times

timtay-microsoft:
> Why are we creating devices and then running these queries? Why not maintain a testDevice on the hub, and query for that in these tests. We can then avoid these propagation delay workarounds.

In general I want the test to setup up any state needed on the hub. That way we don't have special exclusions in our cleanup scripts and the device doesn't get accidentally deleted by someone either

PR #840, opened 2020-07-06
feat(sample): Add plug and play Thermostat sample (no-component)

abhipsaMisra:
Interface implemented: https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/Thermostat.json

PR #842, opened 2020-07-07
feat(sample): Add pnp TemperatureController (2 component) sample

abhipsaMisra:


device/iot-device-samples/pnp-device-sample/TemperatureController/src/main/java/samples/com/microsoft/azure/sdk/iot/TemperatureController.java
```diff
@@ -0,0 +1,382 @@
...
+
+        sendDeviceMemory();
+        sendDeviceSerialNumber();
+
+        final AtomicBoolean temperatureReset = new AtomicBoolean(true);
+        maxTemperature.put(THERMOSTAT_1, 0.0d);
+        maxTemperature.put(THERMOSTAT_2, 0.0d);
+
+        new Thread(new Runnable() {
+            @SneakyThrows({InterruptedException.class, IOException.class})
```
abhipsaMisra, line 105: SneakyThrows: https://github.com/Azure/azure-iot-sdk-java/pull/840#discussion_r450468891

pnp/src/main/java/com/microsoft/azure/sdk/iot/pnphelpers/PnpHelper.java
```diff
@@ -0,0 +1,214 @@
...
+    /**
+     * Create a plug and play compatible telemetry message.
+     * @param telemetryName The name of the telemetry, as defined in the DTDL interface. Must be 64 characters or less. For more details refer <see href="https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/dtdlv2.md#telemetry"
+     * @param telemetryValue The telemetry payload, in the format defined in the DTDL interface.
+     * @param componentName (optional) The name of the component in which the telemetry is defined. Can be null for telemetry defined under the root interface.
+     * @return A plug and play compatible telemetry message, which can be sent to IoT Hub.
+     */
+    public static Message createIotHubMessageUtf8(@NonNull String telemetryName, @NonNull Object telemetryValue, String componentName) {
+        Map<String, Object> payload = singletonMap(telemetryName, telemetryValue);
+        String messageId = java.util.UUID.randomUUID().toString();
```
timtay-microsoft, line 52: For what its worth, the SDK will assign a message id for you if you don't explicitly set one like this

abhipsaMisra: Ah, good to know! I'll get of this, then.

device/iot-device-samples/pnp-device-sample/TemperatureController/src/main/java/samples/com/microsoft/azure/sdk/iot/TemperatureController.java
```diff
@@ -0,0 +1,382 @@
...
+        private final int value;
+        StatusCode(int value) {
+            this.value = value;
+        }
+    }
+
+    // DTDL interface used: https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/TemperatureController.json
+    private static final String deviceConnectionString = System.getenv("IOTHUB_DEVICE_CONNECTION_STRING");
+    private static final String MODEL_ID = "dtmi:com:example:TemperatureController;1";
+    private static final String THERMOSTAT_1 = "thermostat1";
```
bikamani, line 45: I would rather call this Component1

device/iot-device-samples/pnp-device-sample/TemperatureController/src/main/java/samples/com/microsoft/azure/sdk/iot/TemperatureController.java
```diff
@@ -0,0 +1,382 @@
...
+    private static final String THERMOSTAT_2 = "thermostat2";
+    private static final String SERIAL_NO = "SR-123456";
```
bikamani, line 47: call it Property to update? or add a comment?

device/iot-device-samples/pnp-device-sample/TemperatureController/src/main/java/samples/com/microsoft/azure/sdk/iot/TemperatureController.java
abhipsaMisra: This is an identifier that you would tend to use in a sample, and model the properties and commands accordingly. So, in this case, it made sense to call this `Thermostat1`, since it signifies that the component getting modelled is a `Thermostat`, and the index is 1. 
If I call this `Component`1, then my property update handlers, command handlers etc will lose context as to what interface `Component1` actually implements.
Does that make sense?

device/iot-device-samples/pnp-device-sample/TemperatureController/src/main/java/samples/com/microsoft/azure/sdk/iot/TemperatureController.java
abhipsaMisra: This value can ideally be used anywhere, property update/ telemetry/ command response etc. I don't see the value in tying it to the usage here.

device/iot-device-samples/pnp-device-sample/Thermostat/src/main/java/samples/com/microsoft/azure/sdk/iot/Thermostat.java
```diff
@@ -38,6 +37,7 @@
         }
     }
 
+    // DTDL interface used: https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/Thermostat.json
```
vinagesh, line 40: Looks like we are adding this file to our own repo as well. So why not point it to our version of it? This repo is not owned by us and it can change so it might be safer to point to our own. Or we don't have a copy on our end so that there is no multi master problem

PR #848, opened 2020-07-08
refactor(deps, iot-dev, prov-serv): Move glassfish dependency out of …

timtay-microsoft:
…deps since device client doesn't need it

Deps didn't need a dependency on glassfish, and neither did device client. Only provisioning service client and iot service client still depend on this package.

This helps make our device client library just a little bit lighter


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/util/Tools.java
```diff
@@ -36,236 +30,4 @@ public static Boolean isNullOrEmpty(String value)
...
         return retVal;
     }
-
-    /**
-     * Helper function to check if the input string is null or contains only whitespace(s)
-     *
-     * @param value The string to check
-     * @return The value true if the input string is empty or contains only whitespace(s)
-     */
-    public static Boolean isNullOrWhiteSpace(String value)
```
timtay-microsoft, line 46: All of these helper methods were only being called from test code, so there is no reason to keep it

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/util/ToolsTest.java
```diff
@@ -63,464 +54,4 @@ public void isNullOrEmptyInputNotEmpty()
         // Assert
         assertEquals(expResult, result);
     }
-
-    // Tests_SRS_SDK_JAVA_TOOLS_12_004: [The function shall return true if the input is null]
-    @Test
```
timtay-microsoft, line 68: Again, these tests covered code that wasn't being used anywhere, so they are safe to delete

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/hsm/parser/SignResponse.java
```diff
@@ -8,9 +8,7 @@
 import com.google.gson.GsonBuilder;
 import com.google.gson.annotations.Expose;
 import com.google.gson.annotations.SerializedName;
-import com.google.gson.stream.JsonToken;
 
-import javax.json.Json;
```
timtay-microsoft, line 13: Despite being imported, these weren't being used anywhere in this file

provisioning/provisioning-service-client/pom.xml
```diff
@@ -43,6 +43,12 @@
             <artifactId>${iot-deps-artifact-id}</artifactId>
             <version>${iot-deps-version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>javax.json</artifactId>
+            <version>1.0.4</version>
+            <type>jar</type>
+        </dependency>
```
timtay-microsoft, line 51: Iot service client got this dependency through deps before, but now that deps does not depend on glassfish, we need to declare it here

timtay-microsoft:
Closing this PR and bringing its changes to #1039

PR #853, opened 2020-07-09
fix(iot-dev): Fix race condition that caused send/receive task deadlock

timtay-microsoft:
Our invoke method tests have been failing lately likely due to a deadlock happening on the device side upon receiving a method invocation.

This commit limits the scope of where the synchronization on send/receive locks takes place to where the wait and notify calls occur. Previoiusly, the queues of the transport layer were in these synchronization blocks, but these queues are already thread safe.

While not necessary, this commit also adds a failsafe wakeup to the send/receive threads every 10 seconds in case other race conditions arise in the future. It's a little wasteful when everything works correctly though.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubReceiveTask.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubReceiveTask.java
@@ -16,6 +16,10 @@ public final class IotHubReceiveTask implements Runnable
     private static final String THREAD_NAME = "azure-iot-sdk-IotHubReceiveTask";
     private final IotHubTransport transport;
 
+    // This thread should be woken up whenever a new message is received and needs to be handled. But just to be safe,
+    // this thread will wakeup periodically.
+    private static final int FAILSAFE_WAKEUP_PERIOD_MILLISECONDS = 10 * 1000;
+
     // This lock is used to communicate state between this thread and the IoTHubTransport layer. This thread will
     // wait until a message has been received in that layer before continuing. This means that if the transport layer
     // has no received messages to handle, then this thread will do nothing and cost nothing. This is useful
@@ -46,11 +50,11 @@ public final class IotHubReceiveTask implements Runnable
             {
                 synchronized (this.receiveThreadLock)
                 {
-                    if (!this.transport.hasReceivedMessagesToHandle() && !this.transport.isClosed())
+                    while (!this.transport.hasReceivedMessagesToHandle() && !this.transport.isClosed())
                     {
                         // AMQP and MQTT layers will notify the IoTHubTransport layer once a message arrives, and at
                         // that time, this thread will be notified to handle them.
-                        this.receiveThreadLock.wait();
+                        this.receiveThreadLock.wait(FAILSAFE_WAKEUP_PERIOD_MILLISECONDS);
                     }
                 }
             }
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubSendTask.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubSendTask.java
@@ -15,6 +15,10 @@ public final class IotHubSendTask implements Runnable
     private static final String THREAD_NAME = "azure-iot-sdk-IotHubSendTask";
     private final IotHubTransport transport;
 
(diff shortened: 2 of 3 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubReceiveTask.java
```diff
@@ -46,11 +50,11 @@ public void run()
             {
                 synchronized (this.receiveThreadLock)
                 {
-                    if (!this.transport.hasReceivedMessagesToHandle() && !this.transport.isClosed())
+                    while (!this.transport.hasReceivedMessagesToHandle() && !this.transport.isClosed())
                     {
                         // AMQP and MQTT layers will notify the IoTHubTransport layer once a message arrives, and at
                         // that time, this thread will be notified to handle them.
-                        this.receiveThreadLock.wait();
+                        this.receiveThreadLock.wait(FAILSAFE_WAKEUP_PERIOD_MILLISECONDS);
```
timtay-microsoft, line 57: Adding this failsafe actually wasn't what solved the deadlock. But it is a small cost for a safeguard against other potential deadlocks

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubSendTask.java
```diff
@@ -42,11 +46,11 @@ public void run()
         {
             synchronized (this.sendThreadLock)
             {
-                if (!this.transport.hasMessagesToSend() && !this.transport.hasCallbacksToExecute() && !this.transport.isClosed())
+                while (!this.transport.hasMessagesToSend() && !this.transport.hasCallbacksToExecute() && !this.transport.isClosed())
```
timtay-microsoft, line 49: Changing from if (...) to while (...) so that this thread can continue to wait if no work was available after it was woken up. This thread may be woken up by a message being queued, the periodic failsafe wakeup, or by spurious wakeups. In the latter two cases, the thread will usually just go back to sleep since no work is available to process.

For an explanation on spurious wakeups, see [this post](https://stackoverflow.com/questions/13148001/spurious-wakeups-wait-and-notifyall)

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -117,26 +117,17 @@ public Object getReceiveThreadLock()
 
     public boolean hasMessagesToSend()
     {
-        synchronized (sendThreadLock)
```
timtay-microsoft, line 120: This method is only called by the send thread when it has grabbed this lock. No need to try to grab it again here

timtay-microsoft: Same here and below

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1220,10 +1214,10 @@ private void addToCallbackQueue(IotHubTransportPacket packet)
         //Codes_SRS_IOTHUBTRANSPORT_28_002: [This function shall add the packet to the callback queue if it has a callback.]
         if (packet.getCallback() != null)
         {
+            this.callbackPacketsQueue.add(packet);
```
timtay-microsoft, line 1217: Deliberately adding the packet outside of the synchronized block. Doing so here and below fixes the deadlock that we've been seeing. These queues are concurrent queues and may block on add/remove if another thread is using them. For safety, we don't want this queue addition to be done inside of another synchronization block as that may lead to deadlock

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubReceiveTask.java
timtay-microsoft: For comparison, we used to spawn a new thread every 10 milliseconds when a client had no traffic. Now it will just wake up 2 threads and go back to sleep every 10 seconds. This is significantly more efficient on devices even with this failsafe

timtay-microsoft:
This actually didn't fix the deadlock that I thought it would. Don't merge this in just yet

PR #858, opened 2020-07-10
fix(iot-dev): Fix issue with ordering of queue within amqp layer

timtay-microsoft:
If two messages are queued, one for subscribing to twin, and one for getting a twin, and the latter is queued first, then the SDK gets stuck trying to send the get twin request repeatedly without trying to send the subscribe message first.

This change ensures that any message that fails to send is placed at the back of the queue and that the next message stays at the front of the queue. This helps to protect the SDK from states where some messages need to be sent before others, but never are

One of our edge e2e tests was occasionally failing due to this bug.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -530,18 +530,18 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
     private void sendQueuedMessages()
     {
         int messagesAttemptedToBeProcessed = 0;
-        boolean lastSendSucceeded = true;
         Message message = messagesToSend.poll();
-        while (message != null && messagesAttemptedToBeProcessed < MAX_MESSAGES_TO_SEND_PER_CALLBACK && lastSendSucceeded)
+        while (message != null && messagesAttemptedToBeProcessed < MAX_MESSAGES_TO_SEND_PER_CALLBACK)
         {
             messagesAttemptedToBeProcessed++;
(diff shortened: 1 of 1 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -530,18 +530,18 @@ private void addProxyLayer(Transport transport, String hostName)
...
             messagesAttemptedToBeProcessed++;
-            lastSendSucceeded = sendQueuedMessage(message);
+            boolean lastSendSucceeded = sendQueuedMessage(message);
 
             if (!lastSendSucceeded)
             {
                 //message failed to send, likely due to lack of link credit available. Re-queue and try again later
                 log.trace("Amqp message failed to send, adding it back to messages to send queue ({})", message);
                 messagesToSend.add(message);
+                return;
```
timtay-microsoft, line 544: It's important to return here before the next call to messagesToSend.poll() as it keeps the next message to send at the front of the queue. Before this change, if exactly two messages are queued, and the first fails to send (if twin links aren't up yet, for instance), then the second message will never send.

bikamani: Did you verify only with the e2e test not failing or something else?

timtay-microsoft: I verified by manually causing that 2 message scenario and verifying that the SDK eventually sends the subscription message. There isn't a way to test this reliably because we don't have control over when each message will be queued from the e2e tests level. There is some level of randomness

PR #859, opened 2020-07-10
refactor(e2e): Add sdk logging to all test runs

timtay-microsoft:
While system.out will get very busy thanks to our tests running in parallel, Azure Devops is smart enough to allow you to view the logs for each individual test run without other test run's logs getting in the mix.

We'll need this level of logging in place to understand the recent transient issues that we've been seeing and haven't been able to reproduce


iot-e2e-tests/common/src/test/resources/log4j.properties
```diff
@@ -5,4 +5,10 @@ log4j.rootLogger=ERROR, stdout
 log4j.appender.stdout=org.apache.log4j.ConsoleAppender
 log4j.appender.stdout.Target=System.out
 log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
-log4j.logger.tests.integration.com.microsoft.azure.sdk.* = DEBUG 
\ No newline at end of file
+log4j.logger.tests.integration.com.microsoft.azure.sdk.* = DEBUG
```
timtay-microsoft, line 8: We don't really do different log levels in our e2e test packages, so I'm leaving this as DEBUG

PR #868, opened 2020-07-24
fix(e2e): Tests failing with "Runnable returns null without exception information"

bikamani:
So it turns out that tests running in parallel within a class is closing device client for the same instance used in other tests within devicemethod class. This impacts our device method tests, it may be impacting some other tests too.

Check line 295 in DeviceMethodCommon.java which is called after each test
@ After
    public void afterTest()
    {
        this.testInstance.dispose();
    }

### DIFF
--- a/iot-e2e-tests/common/pom.xml
+++ b/iot-e2e-tests/common/pom.xml
@@ -144,13 +144,6 @@
                         <!--By default, only test files that end in ___Test.java are recognized as tests. This statement extends that to ___Tests.java files-->
                         <include>*Tests.java</include>
                     </includes>
-                    <forkCount>1</forkCount>
-
-                    <!--Tests can be run in parallel with tests within their own test class as well as in parallel with tests from other test classes-->
-                    <parallel>both</parallel>
-
-                    <!--Maven will spawn up as many threads as it wants/can while running tests in parallel-->
-                    <useUnlimitedThreads>true</useUnlimitedThreads>
                 </configuration>
             </plugin>
             <plugin>
--- a/iot-e2e-tests/common/pom.xml
+++ b/iot-e2e-tests/common/pom.xml
@@ -144,6 +144,13 @@
                         <!--By default, only test files that end in ___Test.java are recognized as tests. This statement extends that to ___Tests.java files-->
                         <include>*Tests.java</include>
                     </includes>
+                    <forkCount>1</forkCount>
+
+                    <!--Tests can be run in parallel with tests within their own test class as well as in parallel with tests from other test classes-->
+                    <parallel>classes</parallel>
+
+                    <!--Maven will spawn up as many threads as it wants/can while running tests in parallel-->
+                    <useUnlimitedThreads>true</useUnlimitedThreads>
                 </configuration>
             </plugin>
             <plugin>


iot-e2e-tests/common/pom.xml
```diff
@@ -147,7 +147,7 @@
                     <forkCount>1</forkCount>
 
                     <!--Tests can be run in parallel with tests within their own test class as well as in parallel with tests from other test classes-->
-                    <parallel>both</parallel>
+                    <parallel>classes</parallel>
```
bikamani, line 150: "both" is deprecated now, it basically means both methods and class tests run in parallel

abhipsaMisra:
> tests running in parallel within a class is closing device client for the same instance used in other tests within devicemethod class

So for a class having 10 tests, are those 10 tests using the same device client instance? I don't quite understand what this means "closing device client for the same instance used in other tests".

bikamani:
> > tests running in parallel within a class is closing device client for the same instance used in other tests within devicemethod class
> 
> So for a class having 10 tests, are those 10 tests using the same device client instance? I don't quite understand what this means "closing device client for the same instance used in other tests".

That's true. Take a look at setup in tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceMethodCommon.java

iot-e2e-tests/common/pom.xml
bikamani: But with this change I am deleting methods meaning tests within a class will run sequentially.

abhipsaMisra:
I still don't understand, junit will create a new instance of the test class for each `@Test` being executed. 
So, in this case, for method tests, the `testInstance` is initialized by [DeviceMethodCommon.cleanToStart()](https://github.com/Azure/azure-iot-sdk-java/blob/master/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceMethodCommon.java#L281), which is in turn invoked by [DeviceMethodTests.cleanToStart()](https://github.com/Azure/azure-iot-sdk-java/blob/master/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/methods/DeviceMethodTests.java#L51) on a per test basis (`@Before` annotation). 
Subsequently, it is disposed after every test: https://github.com/Azure/azure-iot-sdk-java/blob/master/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceMethodCommon.java#L296

So how does disposing the device client after test1 interfere with the device client in test2?

Also, if we are reusing the same device clients, then our test cases change almost entirely (based on the scenario as well, of course). 
In that case, it makes more sense to decouple each of the tests, rather than have then run serially.

bikamani:
> I still don't understand, junit will create a new instance of the test class for each `@Test` being executed.
> So, in this case, for method tests, the `testInstance` is initialized by [DeviceMethodCommon.cleanToStart()](https://github.com/Azure/azure-iot-sdk-java/blob/master/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceMethodCommon.java#L281), which is in turn invoked by [DeviceMethodTests.cleanToStart()](https://github.com/Azure/azure-iot-sdk-java/blob/master/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/methods/DeviceMethodTests.java#L51) on a per test basis (`@Before` annotation).
> Subsequently, it is disposed after every test: https://github.com/Azure/azure-iot-sdk-java/blob/master/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceMethodCommon.java#L296
> 
> So how does disposing the device client after test1 interfere with the device client in test2?
> 
> Also, if we are reusing the same device clients, then our test cases change almost entirely (based on the scenario as well, of course).
> In that case, it makes more sense to decouple each of the tests, rather than have then run serially.

@Before annotation is for a class and not test basis. However the dispose one is per test basis

bikamani:
> > I still don't understand, junit will create a new instance of the test class for each `@Test` being executed.
> > So, in this case, for method tests, the `testInstance` is initialized by [DeviceMethodCommon.cleanToStart()](https://github.com/Azure/azure-iot-sdk-java/blob/master/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceMethodCommon.java#L281), which is in turn invoked by [DeviceMethodTests.cleanToStart()](https://github.com/Azure/azure-iot-sdk-java/blob/master/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/methods/DeviceMethodTests.java#L51) on a per test basis (`@Before` annotation).
> > Subsequently, it is disposed after every test: https://github.com/Azure/azure-iot-sdk-java/blob/master/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceMethodCommon.java#L296
> > So how does disposing the device client after test1 interfere with the device client in test2?
> > Also, if we are reusing the same device clients, then our test cases change almost entirely (based on the scenario as well, of course).
> > In that case, it makes more sense to decouple each of the tests, rather than have then run serially.
> 
> @ before annotation is for a class and not test basis. However the dispose one is per test basis

abhipsaMisra:
> @before annotation is for a class and not test basis. However the dispose one is per test basis

Well, actually no, `@Before` is per-test basis, `@BeforeClass` is per-class basis. Here is a doc explaining the differences: https://www.baeldung.com/junit-before-beforeclass-beforeeach-beforeall

timtay-microsoft:
Test instances are per test method per parameter set, so there shouldn't be any closing of a client that is in use by a different test. You can verify this by adding a log to each test method that states which device id is tied to the client being closed. Can you double check that before going any further on this?

PR #869, opened 2020-07-27
refactor(iot-dev): Change thread scheduling pattern in deviceIO layer

timtay-microsoft:
The difference between scheduleAtFixedRate and scheduleWithFixedDelay is that the former will attempt to queue new threads even when the previous send/receive thread is not done yet. Now that our SDK has these send/receive threads wait to be notified when a message is ready to be sent/received, we don't want to try to schedule new threads every X milliseconds regardless of previous thread completion. We want to schedule a new thread every X milliseconds after the previous thread has completed


abhipsaMisra:
So this was potentially causing new threads to be spawned even when the previous thread was in-progress. With our thread pool size of 1, would that not stop addition of new threads when the current thread is in progress?

timtay-microsoft:
> So this was potentially causing new threads to be spawned even when the previous thread was in-progress. With our thread pool size of 1, would that not stop addition of new threads when the current thread is in progress?

I believe it would stop them, yes. This is more of a formality than a functional change. If we ever increase the size of the send/receive pools, then we will want this change

Issue #870, opened 2020-07-28
[Bug Report] DeviceClient does not send user agent string with proxy connect requests


timtay-microsoft:
When sending a proxy tunnel request, the SDK should send the same user agent string that the SDK sends to iothub for consistency. Right now, the SDK doesn't send a user agent header in the tunnel request, so the user agent string ends up only describing the JVM version that the tunnel request was sent from

AB#7822159

timtay-microsoft:
Fixed by #871

timtay-microsoft:
The fix for this issue has been released as of Device Client 1.25.0, so I'm closing this issue

az-iot-builder-01:
@timtay-microsoft, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #871, opened 2020-07-28
fix(iot-dev): Fix device client not sending user agent string with proxy tunnel request

timtay-microsoft:
#870


timtay-microsoft:
> Why not define a message template and use that? It'll work the same, but would be easier to maintain?

I'm not sure I follow what a message template is in this context. What is that exactly?

abhipsaMisra:
```java
private static final String messageFormat = "CONNECT %s %s\r\nHost: %s\r\nUser-Agent: %s\r\n\r\n";
private static final String messageFormatWithProxy = messageFormat  + "Proxy-Authorization: Basic %s\r\n";

connectMessage = String.format(messageFormat, hostWithPort, HTTP_VERSION_1_1, hostWithPort, TransportUtils.USER_AGENT_STRING)
OR
connectMessage = String.format(messageFormatWithProxy, hostWithPort, HTTP_VERSION_1_1, hostWithPort, TransportUtils.USER_AGENT_STRING, base64EncodedCredentials);
```

PR #873, opened 2020-08-03
refactor(e2e): Reduce unnecessary operations, simplify setup/teardown for methods tests

timtay-microsoft:
Each deviceMethods client only allows one method invocation at a time, but some tests deliberately have the method invocation take multiple seconds. This commit splits the static device method client so that each test instance has its own direct method client instead. That minimizes any cross-test interference that the above synchronization could cause. Note that I suspect this is one of the possible causes of the device method test timeouts that we've been seeing lately.

Fixed the device method tests not waiting for the methods subscription message to be acknowledged on the device before invoking the method

Previously test setup always created and opened a device client instance regardless of if the test actually needs it. Now that setup is opt-in with a call to super.openDeviceClientAndSubscribeToMethods(...)

Also, test setup used to include calls to close the device client under test even though the client is never open before the test. I have removed that to reduce unnecessary code.

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/DeviceEmulator.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/DeviceEmulator.java
@@ -31,6 +31,9 @@ public class DeviceEmulator
     public static final int METHOD_THROWS = 403;
     public static final int METHOD_NOT_DEFINED = 404;
 
+    private static final int DEVICE_METHOD_SUBSCRIBE_TIMEOUT_MILLISECONDS = 60 * 1000;
+    private static final int METHOD_SUBSCRIBE_CHECK_PERIOD_MILLISECONDS = 1000;
+
     private InternalClient client;
     private DeviceStatus deviceStatus = new DeviceStatus();
     private ConcurrentMap<String, ConcurrentLinkedQueue<Object>> twinChanges = new ConcurrentHashMap<>();
@@ -54,26 +57,9 @@ public class DeviceEmulator
         this.client = client;
     }
 
-    void setup() throws IOException
+    void open() throws IOException
     {
-        try
-        {
-            if (this.client != null)
-            {
-                this.client.closeNow();
-            }
-        }
-        catch (IOException e)
-        {
-            e.printStackTrace();
-        }
-
-        clearStatistics();
-
-        if (this.client != null)
-        {
-            this.client.open();
-        }
+        this.client.open();
     }
 
     /**
@@ -92,9 +78,9 @@ public class DeviceEmulator
      * Enable device method on this device using the local callbacks.
      * @throws IOException if the client failed to subscribe on the device method.
      */
-    void enableDeviceMethod() throws IOException
+    void subscribeToDeviceMethod() throws IOException, InterruptedException
     {
-        enableDeviceMethod(null, null, null, null);
+        subscribeToDeviceMethod(null, null, null, null);
     }
 
     /**
@@ -110,10 +96,10 @@ public class DeviceEmulator
      *                                    deviceMethodStatusCallback is not null.
      * @throws IOException if the client failed to subscribe on the device method.
      */
-    void enableDeviceMethod(
+    void subscribeToDeviceMethod(
             DeviceMethodCallback deviceMethodCallback, Object deviceMethodCallbackContext,
             IotHubEventCallback deviceMethodStatusCallback, Object deviceMethodStatusCallbackContext)
-            throws IOException
+            throws IOException, InterruptedException
     {
         if(deviceMethodCallback == null)
         {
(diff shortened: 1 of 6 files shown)


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/DeviceEmulator.java
```diff
@@ -54,26 +57,9 @@
...
     }
 
-    void setup() throws IOException
+    void open() throws IOException
     {
-        try
-        {
-            if (this.client != null)
-            {
-                this.client.closeNow();
```
timtay-microsoft, line 63: The client was never open at this point, so we can remove this

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/DeviceEmulator.java
```diff
@@ -54,26 +57,9 @@
...
-            }
-        }
-        catch (IOException e)
-        {
-            e.printStackTrace();
-        }
-
-        clearStatistics();
```
timtay-microsoft, line 71: Each test instance has its own statistics, so there is no need to clear them between test runs

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/DeviceEmulator.java
```diff
@@ -149,6 +135,22 @@ else if (client instanceof ModuleClient)
                     this.deviceMethodCallback, this.deviceMethodCallbackContext,
                     this.deviceMethodStatusCallback, this.deviceMethodStatusCallbackContext);
         }
+
+        long startTime = System.currentTimeMillis();
+        while (deviceStatus.statusOk == 0)
```
timtay-microsoft, line 140: Adding this step to ensure that the device methods subscription message gets acknowledged before continuing.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/DeviceTestManager.java
```diff
@@ -50,29 +50,27 @@ public void waitIotHub(int numberOfEvents, long timeoutInSeconds) throws Interru
         }
     }
 
-    public void clearDevice()
+    public void clearStatistics()
     {
         this.deviceEmulator.clearStatistics();
     }
 
-    public void setup(boolean enableMethod, boolean enableTwin) throws IOException, InterruptedException
```
timtay-microsoft, line 58: Renaming to "subscribe" to be more descriptive

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/methods/DeviceMethodTests.java
```diff
@@ -47,16 +45,11 @@ public DeviceMethodTests(IotHubClientProtocol protocol, AuthenticationType authe
         super(protocol, authenticationType, clientType, publicKeyCert, privateKey, x509Thumbprint);
     }
 
-    @Before
-    public void cleanToStart() throws Exception
-    {
-        super.cleanToStart();
```
timtay-microsoft, line 53: This creates a device/module client for this test and opens it. Not every test needs this though, so I moved this logic to the ```super.openDeviceClientAndSubscribeToMethods()``` call instead

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceMethodCommon.java
```diff
@@ -59,7 +59,6 @@ public static Collection inputs() throws Exception
 
     protected static String iotHubConnectionString = "";
 
-    protected static DeviceMethod methodServiceClient;
```
timtay-microsoft, line 62: Having a static device method client is an issue since each client only allows one method invocation at a time ([the method invocation method is synchronized](https://github.com/Azure/azure-iot-sdk-java/blob/72b92781a529daa7bf75dd4dc87892e1c775f59b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethod.java#L64))

In theory, if one test's method invocation takes 10 seconds, that adds a 10 second delay to all concurrently running tests, too. 

I removed this static client in favor of one client per test instance so that no test can interfere with a different test through this.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceMethodCommon.java
```diff
@@ -255,36 +254,19 @@ public static void classSetup()
...
-        {
-            e.printStackTrace();
-        }
-
-        this.testInstance.deviceTestManager.clearDevice();
-
-        try
-        {
-            this.testInstance.deviceTestManager.setup(true, false);
-            IotHubServicesCommon.confirmOpenStabilized(actualStatusUpdates, 120000, this.testInstance.deviceTestManager.client);
```
timtay-microsoft, line 282: All of the above logic isn't necessary since the test client starts off closed, are opened elsewhere, and we no longer have connection stability issues that require us to use this ```confirmOpenStabilized(...)``` anymore.

timtay-microsoft: As a backstory for this ```confirmOpenStabilized()```, our AMQP layer used to be unstable and our tests would have to wait a bit for the instability to go away before continuing. I fixed that instability a while back, so we really don't need this anymore

PR #875, opened 2020-08-03
fix(e2e): Fix small logging issues

timtay-microsoft:
log4j.properties doesn't recurse with star notation, it just recurses in this notation

The start/stop logs should be to our logger, not system.out. That way we get the logger timestamp for free on these two logs

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
@@ -92,7 +92,7 @@ public class Tools
         }
         catch (ClassNotFoundException e)
         {
-            log.debug("Likely running the JVM tests, ignoring ClassNotFoundException");
+            log.debug("Likely running the JVM tests, ignoring ClassNotFoundException\n");
         }
 
         return envVariables;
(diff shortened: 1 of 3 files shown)


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -92,7 +92,7 @@ public static String retrieveEnvironmentVariableValue(String environmentVariable
         }
         catch (ClassNotFoundException e)
         {
-            log.debug("Likely running the JVM tests, ignoring ClassNotFoundException");
+            log.debug("Likely running the JVM tests, ignoring ClassNotFoundException\n");
```
timtay-microsoft, line 95: For some reason, this log never had a newline character printed even though you don't normally have to manually add one. I went ahead and added so it definitely would

iot-e2e-tests/common/src/test/resources/log4j.properties
```diff
@@ -5,7 +5,7 @@ log4j.rootLogger=ERROR, stdout
 log4j.appender.stdout=org.apache.log4j.ConsoleAppender
 log4j.appender.stdout.Target=System.out
 log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
-log4j.logger.tests.integration.com.microsoft.azure.sdk.* = DEBUG
```
timtay-microsoft, line 8: Removing the star here allows for the expected recursion to pick up logs from this package and subpackages

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
abhipsaMisra: interesting that we don't add this elsewhere but it still adds a newline at the end of each log

PR #878, opened 2020-08-04
feat(iot-serv): Make http connect and read timeouts configurable on twin and methods clients

timtay-microsoft:
Basically the same as how the registry manager takes these options.

Making the read/connect timeouts configurable will help with speeding up our e2e tests, and this is a useful feature for customers anyways

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/methods/DeviceMethodTests.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/methods/DeviceMethodTests.java
@@ -10,8 +10,10 @@ import com.microsoft.azure.sdk.iot.deps.serializer.ErrorCodeDescription;
 import com.microsoft.azure.sdk.iot.device.IotHubClientProtocol;
 import com.microsoft.azure.sdk.iot.service.Device;
 import com.microsoft.azure.sdk.iot.service.Module;
+import com.microsoft.azure.sdk.iot.service.ProxyOptions;
 import com.microsoft.azure.sdk.iot.service.auth.AuthenticationType;
 import com.microsoft.azure.sdk.iot.service.devicetwin.DeviceMethod;
+import com.microsoft.azure.sdk.iot.service.devicetwin.DeviceMethodClientOptions;
 import com.microsoft.azure.sdk.iot.service.devicetwin.MethodResult;
 import com.microsoft.azure.sdk.iot.service.exceptions.IotHubGatewayTimeoutException;
 import com.microsoft.azure.sdk.iot.service.exceptions.IotHubNotFoundException;
@@ -19,12 +21,16 @@ import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
+import org.littleshoot.proxy.HttpProxyServer;
+import org.littleshoot.proxy.impl.DefaultHttpProxyServer;
 import tests.integration.com.microsoft.azure.sdk.iot.helpers.*;
 import tests.integration.com.microsoft.azure.sdk.iot.helpers.annotations.ContinuousIntegrationTest;
 import tests.integration.com.microsoft.azure.sdk.iot.helpers.annotations.IotHubTest;
 import tests.integration.com.microsoft.azure.sdk.iot.helpers.annotations.StandardTierHubOnlyTest;
 import tests.integration.com.microsoft.azure.sdk.iot.iothub.setup.DeviceMethodCommon;
 
+import java.net.InetSocketAddress;
+import java.net.Proxy;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
@@ -481,6 +487,41 @@ public class DeviceMethodTests extends DeviceMethodCommon
                 Assert.assertEquals(404001, actualException.getErrorCode());
                 Assert.assertEquals(ErrorCodeDescription.DeviceNotFound, actualException.getErrorCodeDescription());
             }
+        }
+    }
+
+    @Test
+    @StandardTierHubOnlyTest
+    public void invokeMethodWithServiceSideProxy() throws Exception
+    {
+        if (testInstance.protocol != IotHubClientProtocol.MQTT || testInstance.authenticationType != AuthenticationType.SAS || testInstance.clientType != ClientType.DEVICE_CLIENT)
+        {
+            // This test doesn't really care about the device side protocol or authentication, so just run it once
+            // when the device is using MQTT with SAS auth
+            return;
+        }
+
+        String testProxyHostname = "127.0.0.1";
+        int testProxyPort = 8894;
+        HttpProxyServer proxyServer = DefaultHttpProxyServer.bootstrap()
+                .withPort(testProxyPort)
+                .start();
+
+        try
+        {
+            Proxy serviceSideProxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(testProxyHostname, testProxyPort));
+
+            ProxyOptions proxyOptions = new ProxyOptions(serviceSideProxy);
+            DeviceMethodClientOptions options = DeviceMethodClientOptions.builder().proxyOptions(proxyOptions).build();
+
+            this.testInstance.methodServiceClient = DeviceMethod.createFromConnectionString(iotHubConnectionString, options);
+
+            super.openDeviceClientAndSubscribeToMethods();
+            super.invokeMethodSucceed();
+        }
+        finally
+        {
+            proxyServer.stop();
         }
     }
 }
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
@@ -59,10 +59,6 @@ public class DeviceTwinCommon extends IntegrationTest
         String privateKey = certificateGenerator.getPrivateKey();
         String x509Thumbprint = certificateGenerator.getX509Thumbprint();
 
-        sCDeviceTwin = DeviceTwin.createFromConnectionString(iotHubConnectionString);
-        registryManager = RegistryManager.createFromConnectionString(iotHubConnectionString);
-        scRawTwinQueryClient = RawTwinQuery.createFromConnectionString(iotHubConnectionString);
-
         List inputs = new ArrayList();
         for (ClientType clientType : ClientType.values())
         {
@@ -143,10 +139,7 @@ public class DeviceTwinCommon extends IntegrationTest
     protected static String moduleIdPrefix = "java-twin-e2e-test-module";
 
     // States of SDK
-    protected static RegistryManager registryManager;
     protected InternalClient internalClient;
-    protected static RawTwinQuery scRawTwinQueryClient;
-    protected static DeviceTwin sCDeviceTwin;
     protected DeviceState deviceUnderTest = null;
 
     protected DeviceState[] devicesUnderTest;
@@ -296,8 +289,8 @@ public class DeviceTwinCommon extends IntegrationTest
             String id = "java-device-twin-e2e-test-" + this.testInstance.protocol.toString() + UUID.randomUUID().toString();
             devicesUnderTest[i].sCDeviceForRegistryManager = com.microsoft.azure.sdk.iot.service.Device.createFromId(id, null, null);
             devicesUnderTest[i].sCModuleForRegistryManager = com.microsoft.azure.sdk.iot.service.Module.createFromId(id, "module", null);
-            devicesUnderTest[i].sCDeviceForRegistryManager = Tools.addDeviceWithRetry(registryManager, devicesUnderTest[i].sCDeviceForRegistryManager);
-            devicesUnderTest[i].sCModuleForRegistryManager = Tools.addModuleWithRetry(registryManager, devicesUnderTest[i].sCModuleForRegistryManager);
+            devicesUnderTest[i].sCDeviceForRegistryManager = Tools.addDeviceWithRetry(testInstance.registryManager, devicesUnderTest[i].sCDeviceForRegistryManager);
+            devicesUnderTest[i].sCModuleForRegistryManager = Tools.addModuleWithRetry(testInstance.registryManager, devicesUnderTest[i].sCModuleForRegistryManager);
             setUpTwin(devicesUnderTest[i], openDeviceClients);
         }
 
@@ -309,7 +302,7 @@ public class DeviceTwinCommon extends IntegrationTest
         for (int i = 0; i < numberOfDevices; i++)
         {
             tearDownTwin(devicesUnderTest[i]);
-            registryManager.removeDevice(devicesUnderTest[i].sCDeviceForRegistryManager.getDeviceId());
+            testInstance.registryManager.removeDevice(devicesUnderTest[i].sCDeviceForRegistryManager.getDeviceId());
         }
     }
 
@@ -366,7 +359,7 @@ public class DeviceTwinCommon extends IntegrationTest
         }
 
         // set up twin on ServiceClient
-        if (sCDeviceTwin != null)
+        if (testInstance.twinServiceClient != null)
         {
             if (testInstance.clientType == ClientType.DEVICE_CLIENT)
             {
@@ -377,7 +370,7 @@ public class DeviceTwinCommon extends IntegrationTest
                 deviceState.sCDeviceForTwin = new DeviceTwinDevice(deviceState.sCDeviceForRegistryManager.getDeviceId(), deviceState.sCModuleForRegistryManager.getId());
             }
 
-            sCDeviceTwin.getTwin(deviceState.sCDeviceForTwin);
+            testInstance.twinServiceClient.getTwin(deviceState.sCDeviceForTwin);
             Thread.sleep(DELAY_BETWEEN_OPERATIONS);
         }
     }
@@ -408,7 +401,7 @@ public class DeviceTwinCommon extends IntegrationTest
         }
     }
 
-    public DeviceTwinCommon(IotHubClientProtocol protocol, AuthenticationType authenticationType, ClientType clientType, String publicKeyCert, String privateKey, String x509Thumbprint)
+    public DeviceTwinCommon(IotHubClientProtocol protocol, AuthenticationType authenticationType, ClientType clientType, String publicKeyCert, String privateKey, String x509Thumbprint) throws IOException
     {
         this.testInstance = new DeviceTwinTestInstance(protocol, authenticationType, clientType, publicKeyCert, privateKey, x509Thumbprint);
     }
@@ -422,8 +415,12 @@ public class DeviceTwinCommon extends IntegrationTest
         public String x509Thumbprint;
         public String uuid;
         public ClientType clientType;
+        public DeviceTwin twinServiceClient;
+        public RegistryManager registryManager;
+        public RawTwinQuery rawTwinQueryClient;
 
-        public DeviceTwinTestInstance(IotHubClientProtocol protocol, AuthenticationType authenticationType, ClientType clientType, String publicKeyCert, String privateKey, String x509Thumbprint)
+
+        public DeviceTwinTestInstance(IotHubClientProtocol protocol, AuthenticationType authenticationType, ClientType clientType, String publicKeyCert, String privateKey, String x509Thumbprint) throws IOException
         {
             this.protocol = protocol;
             this.authenticationType = authenticationType;
@@ -431,31 +428,10 @@ public class DeviceTwinCommon extends IntegrationTest
             this.privateKey = privateKey;
             this.x509Thumbprint = x509Thumbprint;
             this.clientType = clientType;
-        }
-    }
+            this.twinServiceClient = DeviceTwin.createFromConnectionString(iotHubConnectionString);
 
-    @BeforeClass
-    public static void classSetup()
-    {
-        try
-        {
-            sCDeviceTwin = DeviceTwin.createFromConnectionString(iotHubConnectionString);
             registryManager = RegistryManager.createFromConnectionString(iotHubConnectionString);
-            scRawTwinQueryClient = RawTwinQuery.createFromConnectionString(iotHubConnectionString);
-        }
-        catch (IOException e)
-        {
-            e.printStackTrace();
-            fail("Unexpected exception encountered");
-        }
-    }
-
-    @AfterClass
-    public static void classTearDown()
-    {
-        if (registryManager != null)
-        {
-            registryManager.close();
+            rawTwinQueryClient = RawTwinQuery.createFromConnectionString(iotHubConnectionString);
         }
     }
 
@@ -485,11 +461,11 @@ public class DeviceTwinCommon extends IntegrationTest
         }
 
 
-        deviceUnderTest.sCDeviceForRegistryManager = Tools.addDeviceWithRetry(registryManager, deviceUnderTest.sCDeviceForRegistryManager);
+        deviceUnderTest.sCDeviceForRegistryManager = Tools.addDeviceWithRetry(testInstance.registryManager, deviceUnderTest.sCDeviceForRegistryManager);
 
         if (deviceUnderTest.sCModuleForRegistryManager != null)
         {
-            Tools.addModuleWithRetry(registryManager, deviceUnderTest.sCModuleForRegistryManager);
+            Tools.addModuleWithRetry(testInstance.registryManager, deviceUnderTest.sCModuleForRegistryManager);
         }
 
         Thread.sleep(2000);
@@ -503,7 +479,7 @@ public class DeviceTwinCommon extends IntegrationTest
         try
         {
             tearDownTwin(deviceUnderTest);
-            registryManager.removeDevice(deviceUnderTest.sCDeviceForRegistryManager.getDeviceId());
+            testInstance.registryManager.removeDevice(deviceUnderTest.sCDeviceForRegistryManager.getDeviceId());
         }
         catch (Exception e)
         {
@@ -528,7 +504,7 @@ public class DeviceTwinCommon extends IntegrationTest
             }
 
             actualCount = 0;
-            sCDeviceTwin.getTwin(deviceState.sCDeviceForTwin);
+            testInstance.twinServiceClient.getTwin(deviceState.sCDeviceForTwin);
             Set<Pair> repProperties = deviceState.sCDeviceForTwin.getReportedProperties();
 
             for (Pair p : repProperties)
@@ -561,7 +537,7 @@ public class DeviceTwinCommon extends IntegrationTest
             }
 
             actualCount = 0;
-            sCDeviceTwin.getTwin(deviceState.sCDeviceForTwin);
+            testInstance.twinServiceClient.getTwin(deviceState.sCDeviceForTwin);
             Set<Pair> repProperties = deviceState.sCDeviceForTwin.getReportedProperties();
 
             for (Pair p : repProperties)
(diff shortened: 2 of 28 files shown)


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/methods/DeviceMethodTests.java
```diff
@@ -483,4 +489,39 @@ public void invokeMethodOnUnregisteredDevice() throws Exception
...
+                .start();
+
+        try
+        {
+            Proxy serviceSideProxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(testProxyHostname, testProxyPort));
+
+            ProxyOptions proxyOptions = new ProxyOptions(serviceSideProxy);
+            DeviceMethodClientOptions options = DeviceMethodClientOptions.builder().proxyOptions(proxyOptions).build();
+
+            this.testInstance.methodServiceClient = DeviceMethod.createFromConnectionString(iotHubConnectionString, options);
```
timtay-microsoft, line 517: By default, the test instance already has a methodServiceClient, but it isn't setup to use a proxy, so this will create a proxy enabled client for this test in particular

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/methods/DeviceMethodTests.java
```diff
@@ -483,4 +489,39 @@ public void invokeMethodOnUnregisteredDevice() throws Exception
...
+
+            super.openDeviceClientAndSubscribeToMethods();
+            super.invokeMethodSucceed();
```
timtay-microsoft, line 520: This test method uses the above testInstance.methodServiceClient with the proxy configured

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -146,7 +145,6 @@ public static Collection inputs() throws Exception
     protected static RegistryManager registryManager;
     protected InternalClient internalClient;
     protected static RawTwinQuery scRawTwinQueryClient;
-    protected static DeviceTwin sCDeviceTwin;
```
timtay-microsoft, line 149: In order for some tests to have the twin service client use a proxy, and others to not, each test must have its own

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -431,6 +431,17 @@ public DeviceTwinTestInstance(IotHubClientProtocol protocol, AuthenticationType
...
+
+            try
+            {
+                this.sCDeviceTwin = DeviceTwin.createFromConnectionString(iotHubConnectionString);
+            }
+            catch (IOException e)
+            {
+                // A little odd to do this, but the createFromConnectionString call doesn't ever actually throw this exception, and
+                // this constructor shouldn't throw
+                throw new RuntimeException("Unexpected exception encountered while building device twin service client", e);
```
timtay-microsoft, line 443: I did look into this IOException and verified that it isn't actually ever thrown. I'd remove the exception from the method signature, but that would be a breaking change

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethod.java
```diff
@@ -22,9 +23,9 @@
 {
     private IotHubConnectionString iotHubConnectionString = null;
     private Integer requestId = 0;
-    private static final int DEFAULT_RESPONSE_TIMEOUT = 30; // default response timeout is 30 seconds
-    private static final int DEFAULT_CONNECT_TIMEOUT = 0;
```
timtay-microsoft, line 26: These default values have changed to 24 seconds each to follow the RegistryManager pattern

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -422,6 +420,8 @@ public DeviceTwinCommon(IotHubClientProtocol protocol, AuthenticationType authen
         public String x509Thumbprint;
         public String uuid;
         public ClientType clientType;
+        public DeviceTwin sCDeviceTwin;
+
```
sharmasejal, line 424: nit: extra line

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethod.java
```diff
@@ -33,17 +34,36 @@
...
+                .build());
+    }
+
+    /**
+     * Create a DeviceMethod instance from the information in the connection string.
+     *
+     * @param connectionString is the IoTHub connection string.
+     * @param options the configurable options for each operation on this client. May not be null.
+     * @return an instance of the DeviceMethod.
+     * @throws IOException This exception is thrown if the object creation failed
```
sharmasejal, line 50: add stop punctuation for consistency

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceOperations.java
```diff
@@ -140,6 +143,91 @@ public static HttpResponse request(
         return response;
     }
 
+    /**
+     * Send a http request to the IoTHub using the Twin/Method standard, and return its response.
+     *
+     * @param iotHubConnectionString is the connection string for the IoTHub
```
sharmasejal, line 149: nit: best if we can be consistent with stoppages

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
```diff
@@ -33,22 +34,36 @@
...
+                .httpConnectTimeout(DeviceTwinClientOptions.DEFAULT_HTTP_CONNECT_TIMEOUT_MS)
+                .httpReadTimeout(DeviceTwinClientOptions.DEFAULT_HTTP_READ_TIMEOUT_MS)
+                .build());
+    }
+
+    /**
+     * Static constructor to create instance from connection string
+     *
+     * @param connectionString The iot hub connection string
+     * @param options the configurable options for each operation on this client. May not be null.
```
sharmasejal, line 48: nit: same as above

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinClientOptions.java
```diff
@@ -0,0 +1,39 @@
...
+ * Configurable options for all twin client operations
+ */
+@Builder
+public class DeviceTwinClientOptions
+{
+    protected static final Integer DEFAULT_HTTP_READ_TIMEOUT_MS = 24000; // 24 seconds
+    protected static final Integer DEFAULT_HTTP_CONNECT_TIMEOUT_MS = 24000; // 24 seconds
+
+    /**
+     * The options that specify what proxy to tunnel through. If null, no proxy will be used
```
sharmasejal, line 17: nit: stop punctuation

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinClientOptions.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.microsoft.azure.sdk.iot.service.ProxyOptions;
+import lombok.Builder;
+import lombok.Getter;
+
+/**
+ * Configurable options for all twin client operations.
+ */
+@Builder
+public class DeviceTwinClientOptions
```
abhipsaMisra, line 11: Can we not use a common client options for both methods and twins? They seem to be containing settings for the same properties.

timtay-microsoft: Right now they are the same, yes, but this separation allows us to add other options that are specific to method/twin/registry later

timtay-microsoft: Plus the one that exists right now is called "RegistryManagerOptions" so it would be a bit odd to recycle it for the twin and method clients

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -422,6 +420,7 @@ public DeviceTwinCommon(IotHubClientProtocol protocol, AuthenticationType authen
         public String x509Thumbprint;
         public String uuid;
         public ClientType clientType;
+        public DeviceTwin sCDeviceTwin;
```
bikamani, line 423: Now is a good time to rename this one like we did for some other tests

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -431,6 +430,17 @@ public DeviceTwinTestInstance(IotHubClientProtocol protocol, AuthenticationType
...
             this.x509Thumbprint = x509Thumbprint;
             this.clientType = clientType;
+
+            try
+            {
+                this.sCDeviceTwin = DeviceTwin.createFromConnectionString(iotHubConnectionString);
+            }
+            catch (IOException e)
+            {
+                // A little odd to do this, but the createFromConnectionString call doesn't ever actually throw this exception, and
```
bikamani, line 440: so why are we doing it?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -146,7 +145,6 @@ public static Collection inputs() throws Exception
     protected static RegistryManager registryManager;
     protected InternalClient internalClient;
     protected static RawTwinQuery scRawTwinQueryClient;
```
bikamani, line 147: Since you are moving this one to test instance. Are others here not used as a common resource?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -716,7 +725,7 @@ public void execute(IotHubConnectionStatus status, IotHubConnectionStatusChangeR
         this.internalClient.registerConnectionStatusChangeCallback(connectionStatusUpdateCallback, null);
     }
 
-    protected void testGetDeviceTwin() throws IOException, InterruptedException, IotHubException
+    protected void testGetDeviceTwin() throws IOException, InterruptedException, IotHubException, GeneralSecurityException, ModuleClientException, URISyntaxException
```
bikamani, line 728: When do we see this one - GeneralSecurityException ?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/twin/GetTwinTests.java
```diff
@@ -38,16 +47,47 @@ public GetTwinTests(IotHubClientProtocol protocol, AuthenticationType authentica
...
     {
-        super.testGetDeviceTwin();
+        if (testInstance.protocol != IotHubClientProtocol.MQTT || testInstance.authenticationType != AuthenticationType.SAS || testInstance.clientType != ClientType.DEVICE_CLIENT)
+        {
+            // This test doesn't really care about the device side protocol or authentication, so just run it once
+            // when the device is using MQTT with SAS auth
+            return;
+        }
+
+        super.setUpNewDeviceAndModule();
```
bikamani, line 69: So if we are using in both the tests, why are we removing @Before block?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethod.java
```diff
@@ -22,28 +23,47 @@
...
      * Create a DeviceMethod instance from the information in the connection string.
      *
      * @param connectionString is the IoTHub connection string.
      * @return an instance of the DeviceMethod.
-     * @throws IOException This exception is thrown if the object creation failed
+     * @throws IOException This exception is thrown if the object creation failed.
      */
     public static DeviceMethod createFromConnectionString(String connectionString) throws IOException
+    {
+        return createFromConnectionString(connectionString, DeviceMethodClientOptions.builder()
```
abhipsaMisra, line 38: nit: can we have the tabbing differentiate between the builder args and createFromConnectionString args. Eg:
```java
public static DeviceMethod createFromConnectionString(String connectionString) throws IOException
{
    return createFromConnectionString(
            connectionString,
            DeviceMethodClientOptions.builder()
                    .httpConnectTimeout(DeviceMethodClientOptions.DEFAULT_HTTP_CONNECT_TIMEOUT_MS)
                    .httpReadTimeout(DeviceMethodClientOptions.DEFAULT_HTTP_READ_TIMEOUT_MS)
                    .build());
}
```

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethodClientOptions.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.microsoft.azure.sdk.iot.service.ProxyOptions;
+import lombok.Builder;
+import lombok.Getter;
+
+/**
+ * Configurable options for all device method operations.
+ */
+@Builder
+public class DeviceMethodClientOptions
```
bikamani, line 11: Is this a customer request?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethodClientOptions.java
```diff
@@ -0,0 +1,39 @@
...
+{
+    protected static final Integer DEFAULT_HTTP_READ_TIMEOUT_MS = 24000; // 24 seconds
+    protected static final Integer DEFAULT_HTTP_CONNECT_TIMEOUT_MS = 24000; // 24 seconds
```
bikamani, line 14: Why are these values 24?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethod.java
```diff
@@ -127,74 +147,43 @@ public synchronized MethodResult invoke(String deviceId, String moduleId, String
...
-        /* Codes_SRS_DEVICEMETHOD_21_011: [The invoke shall add a HTTP body with Json created by the `serializer.MethodParser`.] */
         String json = methodParser.toJson();
         if(json == null)
         {
-            /* Codes_SRS_DEVICEMETHOD_21_012: [If `MethodParser` return a null Json, the invoke shall throw IllegalArgumentException.] */
             throw new IllegalArgumentException("MethodParser return null Json");
         }
 
-        long  responseTimeout, connectTimeout;
+        Proxy proxy = options.getProxyOptions() != null ? options.getProxyOptions().getProxy() : null;
```
abhipsaMisra, line 165: If proxy is presesnt as a part of options, then in order to invoke methods and update twins, I'd have to set proxy to their specific "options" types, correct?
Wouldn't that be a bit of overhead? What do think about having a ClientOptions type instead (with proxy, and anything else that would be common across the client), and have these operations specific options inherit from that?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethodClientOptions.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * The options that specify what proxy to tunnel through. If null, no proxy will be used.
+     */
+    @Getter
+    private ProxyOptions proxyOptions;
+
+    /**
+     * The http read timeout to a specified timeout, in milliseconds. A non-zero value specifies the timeout when reading from
+     * Input stream after a connection is established to a resource. If the timeout expires before there is data available
+     * for read, a java.net.SocketTimeoutException is raised. A timeout of zero is interpreted as an infinite timeout.
```
bikamani, line 25: What happens if these are negative values?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceOperations.java
```diff
@@ -141,9 +144,94 @@ public static HttpResponse request(
...
-     * @param httpHeaders non null and non empty custom headers
-     * @throws IllegalArgumentException This exception is thrown if headers were null or empty
+     * Send a http request to the IoTHub using the Twin/Method standard, and return its response.
+     *
+     * @param iotHubConnectionString is the connection string for the IoTHub.
+     * @param url is the Twin URL for the device ID.
+     * @param method is the HTTP method (GET, POST, DELETE, PATCH, PUT).
+     * @param payload is the array of bytes that contains the payload.
+     * @param requestId is an unique number that identify the request.
+     * @param connectTimeout the http connect timeout to use.
```
bikamani, line 154: Add a comment about this timeout values being in seconds

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceOperations.java
```diff
@@ -141,9 +144,94 @@ public static HttpResponse request(
...
+        if (proxy != null)
+        {
+            request = new HttpRequest(url, method, payload, proxy);
+        }
+        else
+        {
+            request = new HttpRequest(url, method, payload);
+        }
+
+        request.setReadTimeoutMillis(readTimeout);
```
bikamani, line 203: should we not check these values >= 0? What's the behavior otherwise?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceOperations.java
```diff
@@ -141,9 +144,94 @@ public static HttpResponse request(
...
+        {
+            for(Map.Entry<String, String> header : headers.entrySet())
+            {
+                request.setHeaderField(header.getKey(), header.getValue());
+            }
+
+            headers = null;
+        }
+
+        HttpResponse response = request.send();
```
bikamani, line 226: should we not catch exception here?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
timtay-microsoft: I'll get rid of this and just have every twin test constructor declare that they throw this instead

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
timtay-microsoft: When something goes wrong when creating the SSL context

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/twin/GetTwinTests.java
timtay-microsoft: The @Before runs before any of the short circuiting we do at line 66 above. Basically, there is no need to setup for a test that doesn't need to run

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethod.java
timtay-microsoft: I'll do it here and for the same block of code in DeviceTwin.java sure

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethodClientOptions.java
timtay-microsoft: The builder pattern is just for consistency with our RegistryManagerOptions class. These options are a customer ask to configure, yes

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethodClientOptions.java
timtay-microsoft: These were the defaults before my changes here

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethod.java
timtay-microsoft: Customers can re-use the same ```ProxyOptions``` instance, but you are right that they would have to create separate methods/twin/registry options classes to house that. It is a bit of overhead, sure, but I think this is the easiest design for customer's to understand.

I'm not sure inheritance here would make for the best user experience. If the constructor for a twin/method service client took a concrete class named ```ClientOptions```, then how am I supposed to know that there is also some ```TwinClientOptions``` class that inherits from it that has other options that I may want? I'd rather just have several concrete classes so that the API is quick to pick up and customer's won't miss important configurable options.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceOperations.java
timtay-microsoft: milliseconds, but yeah I'll add a note

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceOperations.java
timtay-microsoft: The http library we use does this check for us and throws an appropriate exception, so I don't see the need to add any checking up here.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethodClientOptions.java
timtay-microsoft: The http library will throw an exception saying "timeout cannot be negative"

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceOperations.java
timtay-microsoft: No, what would we do with the exception if we caught it? For IOExceptions, they will get bubbled up to the user as expected. For hub level exceptions (404, 429, etc) the line below this one check the http response and will generate and throw any appropriate exceptions

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/twin/GetTwinTests.java
timtay-microsoft: It helps to cut down on device identity creation, which has been a bottleneck lately in our builds

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethod.java
abhipsaMisra: Yes, I agree; inheritance will not be solving anything here from customer's perspective, but it might make code maintenance easier for us though.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethod.java
abhipsaMisra: nit: If we are changing the default values, can we default to a round number instead? 😄

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceOperations.java
```diff
@@ -141,9 +144,94 @@ public static HttpResponse request(
...
+            throws IOException, IotHubException, IllegalArgumentException
+    {
+        if (iotHubConnectionString == null)
+        {
+            throw new IllegalArgumentException("Null ConnectionString");
+        }
+
+        if (url == null)
+        {
+            throw new IllegalArgumentException("Null URL");
```
abhipsaMisra, line 179: nit: can we add some more context to these exception messages - http method is null, endpoint connecting to is null etc?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/QueryCollection.java
```diff
@@ -116,14 +127,86 @@ protected QueryCollection(int pageSize, QueryType requestQueryType, IotHubConnec
...
+        this.url = url;
+        this.responseQueryType = QueryType.UNKNOWN;
+
+        this.isSqlQuery = true;
+
+        this.isInitialQuery = true;
+    }
+
+    /**
+     * Constructor for non-sql based queries.
```
abhipsaMisra, line 168: Q - haven't seen this before! Do we have a sample/ test on how these non-sql queries are used?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
timtay-microsoft: I can move these to be per-test instance, sure

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethod.java
timtay-microsoft: But what is more important, customer usability or code maintenance? I don't think that we are imposing too much on ourselves here

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceOperations.java
timtay-microsoft: Sure

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/QueryCollection.java
timtay-microsoft: Honestly, I'm not sure how this API gets used. "Non-sql query" seems contradictory to me

PR #879, opened 2020-08-04
refactor(e2e) adding throttling resiliency to tests, allow more registry operations in parallel

timtay-microsoft:
This PR removes some unnecessary SDK side bottle necking around adding devices/modules to the registry. See comments below for an explanation on that.

This PR also adds a test rule that watches out for throttling from the service. Now that we are sending more registry requests at once, we are hitting throttling pretty often on S1 hubs in canary and sometimes in S1 hubs in prod. This new test rule basically re-runs any test that fails due to a throw 429 exception. This allows us to generally not care if a test hits a throttling limit or not without implementing retry logic on each service request (which would take a very long time)

This PR also moves some static registry/method/twin clients to live at a per-test-instance level to keep each test more isolated from one another

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/IntegrationTest.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/IntegrationTest.java
@@ -53,6 +53,9 @@ public abstract class IntegrationTest
     @Rule
     public FlakeyTestRule flakeyTestRule = new FlakeyTestRule();
 
+    @Rule
+    public ThrottleResistantTestRule throttleResistantTestRule = new ThrottleResistantTestRule();
+
     int E2E_TEST_TIMEOUT_MILLISECONDS = 5 * 60 * 1000;
 
     // Each test must finish in under 5 minutes. Only the token renewal test should last longer,
@@ -67,4 +70,6 @@ public abstract class IntegrationTest
     public static boolean runIotHubTests = Boolean.parseBoolean(Tools.retrieveEnvironmentVariableValue("RUN_IOTHUB_TESTS", "true"));
     public static boolean runProvisioningTests = Boolean.parseBoolean(Tools.retrieveEnvironmentVariableValue("RUN_PROVISIONING_TESTS", "true"));
 
+    // Infinite read timeout for all http operations
+    public static int HTTP_READ_TIMEOUT = 0;
 }
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
@@ -1,7 +1,7 @@
 /*
-*  Copyright (c) Microsoft. All rights reserved.
-*  Licensed under the MIT license. See LICENSE file in the project root for full license information.
-*/
+ *  Copyright (c) Microsoft. All rights reserved.
+ *  Licensed under the MIT license. See LICENSE file in the project root for full license information.
+ */
 
 package tests.integration.com.microsoft.azure.sdk.iot.helpers;
 
@@ -14,6 +14,7 @@ import org.apache.commons.lang3.exception.ExceptionUtils;
 
 import java.io.IOException;
 import java.net.SocketException;
+import java.net.SocketTimeoutException;
 import java.net.UnknownHostException;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -109,7 +110,7 @@ public class Tools
         return possibleExceptionCause.isInstance(exceptionToSearch) || (exceptionToSearch != null && isCause(possibleExceptionCause, exceptionToSearch.getCause()));
     }
 
-    public synchronized static Device addDeviceWithRetry(RegistryManager registryManager, Device device) throws IotHubException, IOException, InterruptedException
+    public static Device addDeviceWithRetry(RegistryManager registryManager, Device device) throws IotHubException, IOException, InterruptedException
     {
         long startTime = System.currentTimeMillis();
         Device ret = null;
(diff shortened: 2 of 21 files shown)


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -109,20 +110,22 @@ public static boolean isCause(Class<? extends Throwable> possibleExceptionCause,
         return possibleExceptionCause.isInstance(exceptionToSearch) || (exceptionToSearch != null && isCause(possibleExceptionCause, exceptionToSearch.getCause()));
     }
 
-    public synchronized static Device addDeviceWithRetry(RegistryManager registryManager, Device device) throws IotHubException, IOException, InterruptedException
```
timtay-microsoft, line 112: This method being synchronized meant that only one thread can enter this static method at a time. Because of this, tests have been failing lately just by getting stuck waiting in line to enter this method. In most cases, this method is called during each test setup, so it is called a lot. 

By removing the synchronization, the SDK test side doesn't slow itself down in making these requests, but IoT Hub often responds slower because of it, or responds with a throttle. The added throttleResistantTestRule covers the latter, and using infinite read timeouts covers the former. As a result, our tests scale better if we ever upgrade to an S2 for testing, but still work against an S1

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -136,20 +139,22 @@ public synchronized static Device addDeviceWithRetry(RegistryManager registryMan
         return ret;
     }
 
-    public synchronized static Module addModuleWithRetry(RegistryManager registryManager, Module module) throws IotHubException, IOException, InterruptedException
```
timtay-microsoft, line 139: Same as above. This synchronization placed a bottle neck on the SDK tests, so I'm removing it

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/rules/ThrottleResistantTestRule.java
```diff
@@ -0,0 +1,55 @@
...
+        private final Statement base;
+        private final Description description;
+
+        public ThrottleResistantStatement(Statement base, Description description) {
+            this.base = base;
+            this.description = description;
+        }
+
+        @Override
+        public void evaluate() throws Throwable {
```
timtay-microsoft, line 38: This test rule makes each test run (```base.evaluate()``` runs the test) until it passes or until it fails with a non-throttling exception

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/rules/ThrottleResistantTestRule.java
```diff
@@ -0,0 +1,55 @@
...
+            boolean ranWithoutThottling = false;
+            while (!ranWithoutThottling)
+            {
+                try
+                {
+                    base.evaluate();
+                    ranWithoutThottling = true;
+                }
+                catch (IotHubTooManyRequestsException e)
```
timtay-microsoft, line 47: This exception gets thrown by any HTTP request (registryManager.addDevice, deviceMethod.invokeMethod, etc.)

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/methods/DeviceMethodTests.java
```diff
@@ -340,7 +339,7 @@ public void invokeMethodDefaultConnectionTimeoutSucceed() throws Exception
         Assert.assertEquals(buildExceptionMessage("Unexpected status errors occurred", testInstance.deviceTestManager.client), 0, deviceTestManger.getStatusError());
     }
 
-    @Test (expected = IotHubGatewayTimeoutException.class)
```
timtay-microsoft, line 343: With the throttle resistant test rule, we can no longer use this method of checking for an exception upon executing some code. If the test encounters a 429 exception, that 429 gets compared to this GatewayTimeoutException, and since they don't match, the test throws an AssertionFailedException. Once that AssertionFailedException bubbles up to the throttle resistant test rule, the test won't be re-run since the test didn't seem to fail from a 429.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceMethodCommon.java
```diff
@@ -59,8 +58,6 @@ public static Collection inputs() throws Exception
 
     protected static String iotHubConnectionString = "";
 
-    protected static RegistryManager registryManager;
```
timtay-microsoft, line 62: Making each test have their own registry manager instead of sharing one static instance

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/IntegrationTest.java
```diff
@@ -67,4 +70,6 @@ protected void finished(Description description)
     public static boolean runIotHubTests = Boolean.parseBoolean(Tools.retrieveEnvironmentVariableValue("RUN_IOTHUB_TESTS", "true"));
     public static boolean runProvisioningTests = Boolean.parseBoolean(Tools.retrieveEnvironmentVariableValue("RUN_PROVISIONING_TESTS", "true"));
 
+    // Infinite read timeout for all http operations
+    public static int HTTP_READ_TIMEOUT = 0;
```
timtay-microsoft, line 74: Note that only the read timeout will be infinite. The tests will still timeout if they fail to connect to iothub. With this change, they just won't throw an exception if iothub takes a while to finish the operation and return the http response

PR #881, opened 2020-08-05
refactor(e2e): Remove shorter, test specific timeouts

timtay-microsoft:
These tests are timing out since they rely on registry manager operations which may take a while to execute since only one thread can do a registry manager operation at a time and most tests need to do at least one registry manager operation.

For instance: https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=65369&view=ms.vss-test-web.build-test-results-tab&runId=1579266&resultId=103226&paneView=debug

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/serviceclient/RegistryManagerTests.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/serviceclient/RegistryManagerTests.java
@@ -104,7 +104,7 @@ public class RegistryManagerTests extends IntegrationTest
         proxyServer.stop();
     }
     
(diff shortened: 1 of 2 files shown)


bikamani:
> These tests are timing out since they rely on registry manager operations which may take a while to execute since only one thread can do a registry manager operation at a time and most tests need to do at least one registry manager operation.

How long is it taking for one thread now in your experiments? & what would be the timeout limit with this change?

timtay-microsoft:
> How long is it taking for one thread now in your experiments? 
I can't say how long these tests take since the multithreading of our tests makes it random which tests will execute first and which tests will get stuck in line. Judging by how only these tests are timing out now, removing their strict timeout should make them stop timing out

> What would be the timeout limit with this change?

By default, all tests in this repo have a 5 minute timeout, so these tests would go from 1 minute timeout to 5 minutes.

PR #882, opened 2020-08-05
refactor(vsts): Add retry count to test results title in Azure Devops

timtay-microsoft:
This is useful if you run a pipeline once, it fails on some tests, you rerun the job, and it fails again. With this change, you can correlate the test results drop down with the rerun #


vsts/windowsLinuxAndAndroidBuildMatrixConfig.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
         displayName: 'Publish Test Results'
         inputs:
           mergeTestResults: true
-          testRunTitle: "Windows"
+          testRunTitle: "Windows (Attempt $(System.JobAttempt))"
```
bikamani, line 61: 👍

Issue #887, opened 2020-08-07
[Bug Report] Device Client Amqp layer doesn't check delivery state for sent messages


timtay-microsoft:
When the amqp library notifies our SDK of a delivery on a sender link (aka when iothub acknowledges a message sent on that link), our code currently does not look at whether the message was actually accepted by the hub or not. There aren't many cases where the sent message is rejected/released, but the SDK should still log when the message is unexpected rejected/released and should propagate this info up to the user.

https://github.com/Azure/azure-iot-sdk-java/blob/40f93c895e0ef1ab3e762bd3fc5458afe96c422a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSenderLinkHandler.java#L73

AB#7923246

timtay-microsoft:
Fixed by #888

timtay-microsoft:
The fix for this issue has been released as of Device Client 1.25.0, so I'm closing this issue

az-iot-builder-01:
@timtay-microsoft, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #888, opened 2020-08-07
fix(iot-dev): Fix issue where amqp layer ignored delivery state of sent messages

timtay-microsoft:
fixes issue #887 

If the delivery state was anything other than Accepted, the SDK needs to either retry sending the message or bubble up that exception to the user

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -12,11 +12,13 @@ import com.microsoft.azure.proton.transport.proxy.impl.ProxyImpl;
 import com.microsoft.azure.proton.transport.ws.impl.WebSocketImpl;
 import com.microsoft.azure.sdk.iot.device.*;
 import com.microsoft.azure.sdk.iot.device.auth.IotHubSasTokenAuthenticationProvider;
+import com.microsoft.azure.sdk.iot.device.exceptions.ProtocolException;
 import com.microsoft.azure.sdk.iot.device.exceptions.TransportException;
 import com.microsoft.azure.sdk.iot.device.transport.*;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.qpid.proton.Proton;
 import org.apache.qpid.proton.amqp.messaging.Accepted;
+import org.apache.qpid.proton.amqp.messaging.Modified;
 import org.apache.qpid.proton.amqp.messaging.Rejected;
 import org.apache.qpid.proton.amqp.messaging.Released;
 import org.apache.qpid.proton.amqp.transport.DeliveryState;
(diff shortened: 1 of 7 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -468,9 +470,29 @@ public void onAuthenticationSessionOpened()
...
-        this.listener.onMessageSent(message, null);
+        if (deliveryState == Accepted.getInstance())
+        {
+            this.listener.onMessageSent(message, null);
+        }
+        else if (deliveryState instanceof Rejected)
+        {
+            // The message was not accepted by the server, and the reason why is found within the nested error
+            TransportException ex = AmqpsExceptionTranslator.convertFromAmqpException(((Rejected) deliveryState).getError());
+            this.listener.onMessageSent(message, ex);
```
timtay-microsoft, line 483: The exception translator above does the hard work of figuring out if the message can be retried or not

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -468,9 +470,29 @@ public void onAuthenticationSessionOpened()
...
+        }
+        else if (deliveryState == Released.getInstance())
```
abhipsaMisra, line 485: Does the service ever send a delivery state of "released"? What do they expect the client to do, in this case?

timtay-microsoft: I've never seen it, but the protocol defines it as a possible delivery state. As a delivery state, it just means that the service has made the message available to be re-delivered for whatever reason. So on the client side, we just need to create a retryable exception here so that the transport layer re-sends it

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -468,9 +470,29 @@ public void onAuthenticationSessionOpened()
...
     @Override
-    public void onMessageAcknowledged(Message message)
+    public void onMessageAcknowledged(Message message, DeliveryState deliveryState)
     {
-        this.listener.onMessageSent(message, null);
+        if (deliveryState == Accepted.getInstance())
+        {
+            this.listener.onMessageSent(message, null);
+        }
+        else if (deliveryState instanceof Rejected)
```
azabbasi, line 479: why is the check for rejected look different than the other delivery states? 
Shouldn't it be 
`deliveryState == Rejected.getInstance()` ?

timtay-microsoft: Rejected is the one delivery state that comes with additional information in the ack. For Accepted/Released, there is no payload telling why the message was Accepted/Released. For Rejected, there should be some error details explaining why the message was rejected

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -468,9 +470,29 @@ public void onAuthenticationSessionOpened()
...
+            this.listener.onMessageSent(message, ex);
+        }
+        else if (deliveryState == Released.getInstance())
+        {
+            // As per AMQP spec, this state means the message should be re-delivered to the server at a later time
+            ProtocolException protocolException = new ProtocolException("Message was released by the amqp server");
+            protocolException.setRetryable(true);
+            this.listener.onMessageSent(message, protocolException);
+        }
+        else
```
azabbasi, line 492: There are at least 4 more delivery state types that are defined in the transport layer that will fall into the else statement here, can you point me to the service documentation that indicates what delivery states they may return? I am just curious.

timtay-microsoft: There isn't any documentation like that, unfortunately. For Modified/Received, I'm not sure what the SDK is supposed to do. I've never seen those delivery states returned, though

timtay-microsoft: From the official AMQP spec, though: 

• accepted: indicates successful processing at the receiver.
• rejected: indicates an invalid and unprocessable message.
• released: indicates that the message was not (and will not be) processed.
• modified: indicates that the message was modified, but not processed.
• received: indicates partial message data seen by the receiver as well as the starting point for a resumed transfer.

PR #889, opened 2020-08-10
feat(device): support batch messages for HTTPS

azabbasi:


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
```diff
@@ -165,6 +175,30 @@ public Message()
         initialize();
     }
 
+    /**
+     * Internal constructor to initialize a bulk message instance.
+     * @param messages List of nested messages
+     */
+    Message(Set<Message> messages)
```
azabbasi, line 182: This constructor is internal to the iot.device package.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -184,6 +184,33 @@ public void sendEventAsync(Message message, IotHubEventCallback callback, Object
         deviceIO.sendEventAsync(message, callback, callbackContext, this.config.getDeviceId());
     }
 
+    /**
+     * ASynchronously sends a batch of messages to the IoT Hub
+     * HTTPS messages will be sent in a single batch and MQTT and AMQP messages will be sent individually.
```
abhipsaMisra, line 189: MQTT and AMQP messages will be sent individually - protocol library restriction?

timtay-microsoft:
So how does IoTHub accept these batch messages if one is malformed? Is it an all-or-nothing type API?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -184,6 +184,33 @@ public void sendEventAsync(Message message, IotHubEventCallback callback, Object
...
+     * @param callback the callback to be invoked when a response is received.
+     * Can be {@code null}.
+     * @param callbackContext a context to be passed to the callback. Can be
+     * {@code null} if no callback is provided.
+     *
+     * @throws IllegalArgumentException if the message provided is {@code null}.
+     * @throws IllegalStateException if the client has not been opened yet or is
+     * already closed.
+     */
+    public void sendEventBatchAsync(Set<Message> messages, IotHubEventCallback callback, Object callbackContext)
```
abhipsaMisra, line 202: The ordering of the messages needs to be preserved, so maybe we should use a `List` instead? Or add a comment that the customer should supply an ordered implementation of `Set`?

timtay-microsoft, line 202: Does this have to be a set? What about a more generic type like Collection or Iterable?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -401,6 +401,17 @@ public void addMessage(Message message, IotHubEventCallback callback, Object cal
...
         // callback, and context and then add that packet to the waiting queue.]
+
+        if (message.isBulk() && !(this.iotHubTransportConnection instanceof HttpsIotHubConnection))
+        {
+            for (Message singleMessage : message.getNestedMessages())
+            {
+                waitingPacketsQueue.add(new IotHubTransportPacket(message, callback, callbackContext,null, System.currentTimeMillis()));
+            }
+
+            return;
```
abhipsaMisra, line 412: nit: add a log statement here

timtay-microsoft:
If we plan on adding bulk message support to AMQP later, shouldn't we not allow users to call this API over AMQP? Since HTTP is the only protocol that actually currently supports bulk messages, I'd expect to see 

```java
if (this.protocol != HTTPS)
{
    throw new UnsupportedOperationException("Bulk messages are only supported over HTTPS");
}
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
timtay-microsoft: Does the ordering need to be preserved? I thought the whole point was that they were all delivered simultaneously?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
```diff
@@ -212,6 +246,15 @@ public Message(String body)
     
     // ----- Public Methods -----
 
+    /**
+     * Indicates whether or not this instance is a bulk message instance.
+     * @return Boolean value to indicate whether or not this is a bulk message instance.
+     */
+    public boolean isBulk()
+    {
+        return this.isBulk;
```
timtay-microsoft, line 255: What about having a class ```BulkMessage``` that extends from this message class instead? That way in the lower layers, you can have code like

```java
if (message instanceof BulkMessage)
{
// bulk message logic
}
else
{
//normal message logic
}

File: device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsIotHubConnection.java [hunk 6]
```diff
@@ -80,7 +80,29 @@ public IotHubStatusCode sendMessage(Message message) throws TransportException
...
+            if (message.isBulk())
+            {
+                HttpsBatchMessage batchMessage = new HttpsBatchMessage();
+                for (Message singleMessage : message.getNestedMessages())
+                {
+                    try
+                    {
+                        batchMessage.addMessage(HttpsSingleMessage.parseHttpsMessage(singleMessage));
+                    }
+                    catch (Exception e)
```
timtay-microsoft, line 95: You can make this catch more specific since only a ```IotHubSizeExceededException``` exception is thrown here

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
abhipsaMisra: yes, they would be delivered simultaneously, but still as a batch containing the individual messages in some order -> if I add them as 1->2->3 into a batch; then on service send, once I receive a batch and iterate over it, the order should still be 1->2->3.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsIotHubConnection.java
```diff
@@ -80,7 +80,29 @@ public IotHubStatusCode sendMessage(Message message) throws TransportException
...
+                    {
+                        throw new TransportException(e);
```
timtay-microsoft, line 97: ```throw new TransportException("Failed to create http batch message", e);``` to add context

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/telemetry/SendMessagesTests.java
```diff
@@ -67,6 +67,14 @@ public void sendMessages() throws Exception
...
     }
 
+    @Test
+    public void sendBulkMessages() throws Exception
+    {
+        this.testInstance.setup();
+
+        IotHubServicesCommon.sendBulkMessages(testInstance.client, testInstance.protocol, NORMAL_MESSAGES_TO_SEND, RETRY_MILLISECONDS, SEND_TIMEOUT_MILLISECONDS, 0, null);
+    }
+
```
timtay-microsoft, line 77: I think a good CI-only test to complement the above test is to try a bulk message made up of a large number of small messages. What do you think?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
timtay-microsoft: That way getNestedMessages() can only be called on a bulk message

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
azabbasi: yes, service does not support MQTT batch messages, and our current AMQP library (proton) doesn't have built in batch message capabilities so we have to implement our own. That will be done separately.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
azabbasi: I will make a note of that in the comments.

azabbasi: That is an excellent question. I don't think it has to necessarily be a Set, I tried to follow the other APIs we have in the internal client and used Sets. is there a drawback in using Sets here?

azabbasi:
> If we plan on adding bulk message support to AMQP later, shouldn't we not allow users to call this API over AMQP? Since HTTP is the only protocol that actually currently supports bulk messages, I'd expect to see
> 
> ```java
> if (this.protocol != HTTPS)
> {
>     throw new UnsupportedOperationException("Bulk messages are only supported over HTTPS");
> }
> ```

We sure can, We use the same API in our C# SDK and I tried to stay consistent with that. we can chat offline about this.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
azabbasi: that is the first approach I tried, If we decide to create a BulkMessage class (which I did initially) we have to expose it (which I am not a huge fan of doing). 
The reason is that this class is used in other packages device.Transport for example. so we can't make it internal (not a public/private class). So I baked it into the same Message class.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsIotHubConnection.java
azabbasi: the inner exception already has this message. I will add this on top of it for more clarity

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/telemetry/SendMessagesTests.java
azabbasi: I like that. I will add a negative test as well to make sure the correct exception is thrown if the size of the message is above 255k

azabbasi:
> So how does IoTHub accept these batch messages if one is malformed? Is it an all-or-nothing type API?

This is an all or nothing API for HTTPS, I will add an e2e test to catch that and have coverage for those cases.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -401,6 +401,17 @@ public void addMessage(Message message, IotHubEventCallback callback, Object cal
 
         //Codes_SRS_IOTHUBTRANSPORT_34_042: [This function shall build a transport packet from the provided message,
         // callback, and context and then add that packet to the waiting queue.]
+
+        if (message.isBulk() && !(this.iotHubTransportConnection instanceof HttpsIotHubConnection))
+        {
+            for (Message singleMessage : message.getNestedMessages())
+            {
+                waitingPacketsQueue.add(new IotHubTransportPacket(message, callback, callbackContext,null, System.currentTimeMillis()));
```
timtay-microsoft, line 409: Use ```singleMessage``` when constructing the iothubtransportpacket, not ```message```

timtay-microsoft: Otherwise you'll just be sending the bulk message over and over, not its children.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
azabbasi: I made it into a List to maintain the orders in the collection as Abhipsa mentioned.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/telemetry/SendMessagesTests.java
azabbasi: it turns out we already have unit tests for the BatchHTTPmessages class the failure scenario

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
```diff
@@ -35,6 +35,16 @@
      */
     private String messageId;
 
+    /**
+     * Whether or not there is a list of nested messages included in this instance.
+     */
+    private boolean isBulk;
+
+    /**
+     * List of nested messages.
```
bikamani, line 44: what do nested messages mean here?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
```diff
@@ -35,6 +35,16 @@
...
+     */
+    private Set<Message> nestedMessages;
+
```
bikamani, line 47: so you can use lombok getter here right instead of line 194?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -401,6 +401,17 @@ public void addMessage(Message message, IotHubEventCallback callback, Object cal
 
         //Codes_SRS_IOTHUBTRANSPORT_34_042: [This function shall build a transport packet from the provided message,
         // callback, and context and then add that packet to the waiting queue.]
+
+        if (message.isBulk() && !(this.iotHubTransportConnection instanceof HttpsIotHubConnection))
+        {
+            for (Message singleMessage : message.getNestedMessages())
+            {
```
bikamani, line 408: may be add some logging here?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
azabbasi: True, but then I will have to change all the getters in this class no?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
timtay-microsoft: As long as they need to be ordered, then a List should be fine. In general, if the API can take a more generic collection type, then it should. That way users who like sets can pass in a set, and users who like lists can pass in a list. But that's all moot if we have to maintain order here.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
timtay-microsoft: I'm not sure I follow why making a hypothetical ```BulkMessage``` class public would be bad when it extends an already public class.

azabbasi: I made a separate class and made the constructor internal.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
azabbasi: I have moved it to batched messages. That makes the batch message a wrapper around the same object type of Message.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/BatchMessages.java
```diff
@@ -0,0 +1,30 @@
+package com.microsoft.azure.sdk.iot.device;
+
+import java.util.List;
+
+public class BatchMessages extends Message
```
timtay-microsoft, line 5: A javadoc at the class level would be helpful. It should explain what it is, and link it to the relevant device client method by using "{@ link \<classname\>#\<methodName\>}"

timtay-microsoft: I can't do the spacing right on that string above due to github, but autocomplete should help you out here

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/BatchMessages.java
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * List of nested messages.
+     */
+    private List<Message> nestedMessages;
+
+    /**
+     * Creates a batch message
+     * This constructor is internal and can only be called within the com.microsoft.azure.sdk.iot.device package.
+     * @param messages
```
timtay-microsoft, line 15: "@ param messages" could use a description. Something like "The messages that will be sent as a batch"

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -184,6 +185,33 @@ public void sendEventAsync(Message message, IotHubEventCallback callback, Object
...
     }
 
+    /**
+     * ASynchronously sends a batch of messages to the IoT Hub
+     * HTTPS messages will be sent in a single batch and MQTT and AMQP messages will be sent individually.
+     * Maximum payload size for HTTPS is 255KB
+     *
+     * @param messages the list of message to be sent.
+     * @param callback the callback to be invoked when a response is received.
+     * Can be {@code null}.
```
timtay-microsoft, line 195: Clarify that this callback is all-or-nothing for each batch message, so it will only be fired once

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -401,6 +401,18 @@ public void addMessage(Message message, IotHubEventCallback callback, Object cal
 
         //Codes_SRS_IOTHUBTRANSPORT_34_042: [This function shall build a transport packet from the provided message,
         // callback, and context and then add that packet to the waiting queue.]
+
+        if (message instanceof BatchMessages && !(this.iotHubTransportConnection instanceof HttpsIotHubConnection))
```
timtay-microsoft, line 405: If you want to allow amqp/mqtt to use this batch API, then at least add a comment here explaining why only https doesn't split the batch message here

device/iot-device-samples/send-batch-events/README.md
```diff
@@ -0,0 +1,60 @@
+# Send Event Sample
```
timtay-microsoft, line 1: Is this a copy-paste error? It should be "Send Batch Events Sample"

device/iot-device-samples/send-batch-events/README.md
```diff
@@ -0,0 +1,60 @@
...
+$> mvn install -DskipTests
+```
+
+## Run the sample
+
+Listed below is the command to launch the sample along with a description of its arguments
+
+```
+$> cd {sample root}/send-event
+$> java -jar target/send-event-{version}-with-deps.jar "{connection string}" {number of messages to send} {protocol}  {path to trusted root ca}
```
timtay-microsoft, line 20: Again, make sure you update this readme to reflect the different jar file that gets generated when building this sample

device/iot-device-samples/send-batch-events/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+If everything was correctly provided via the CLI arguments, the following should be observed on stdout
+
+```
+...
+IoT Hub responded to message d630cd17-4b84-49f1-878a-901f53ac4038 with status OK_EMPTY
+IoT Hub responded to message e45e5998-5ea1-49bd-978c-5a149e2012e8 with status OK_EMPTY
+IoT Hub responded to message f12eb242-b314-4ded-98a6-9f2229a9c2b7 with status OK_EMPTY
+IoT Hub responded to message 033cc1fd-b3fe-43a8-b05a-5e1e188d2b8c with status OK_EMPTY
+IoT Hub responded to message 1695f37f-455a-4386-a46b-57fe4e4fae44 with status OK_EMPTY
```
timtay-microsoft, line 58: I imagine that the bulk API callback only gets fired once, so this expected output looks wrong to me

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -184,6 +185,33 @@ public void sendEventAsync(Message message, IotHubEventCallback callback, Object
...
+     * @param callbackContext a context to be passed to the callback. Can be
+     * {@code null} if no callback is provided.
+     *
+     * @throws IllegalArgumentException if the message provided is {@code null}.
+     * @throws IllegalStateException if the client has not been opened yet or is
+     * already closed.
+     */
+    public void sendEventBatchAsync(List<Message> messages, IotHubEventCallback callback, Object callbackContext)
```
timtay-microsoft, line 203: By adding this to the InternalClient rather than device client, you are allowing ModuleClients to use this API, too. I don't think that is an issue, but I wanted to make sure you think about that.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -184,6 +185,33 @@ public void sendEventAsync(Message message, IotHubEventCallback callback, Object
...
+    {
+        for (Message message: messages)
+        {
+            message.setConnectionDeviceId(this.config.getDeviceId());
+        }
+
+        Message message = new BatchMessages(messages);
```
timtay-microsoft, line 210: nit: The class should be "BatchMessage" not "BatchMessages" since it is ultimately just one http request

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/telemetry/SendMessagesTests.java
```diff
@@ -67,6 +69,28 @@ public void sendMessages() throws Exception
...
+    @Test
+    public void sendManySmallMessagesAsBatch() throws Exception
```
timtay-microsoft, line 81: I see this as a good CI test. If you'd like, you can tag it with @ContinuousIntegrationTest. That will make it only run during nightly builds and CI builds

device/iot-device-samples/send-event/README.md
```diff
@@ -1,8 +1,8 @@
-# Send Event Sample
+# Send Events In Batch Sample
```
timtay-microsoft, line 1: It looks like you edited the wrong readme. Double check the path of this file

device/iot-device-samples/send-batch-events/README.md
timtay-microsoft: See below, but it looks like you edited the wrong readme file

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
timtay-microsoft: The tests you wrote in SendMessagesTests automatically run the scenarios for ModuleClients and DeviceClients, so don't worry about test coverage on this one

device/iot-device-samples/send-batch-events/README.md
azabbasi: yes it is

device/iot-device-samples/send-event/README.md
azabbasi: yeah :)))

device/iot-device-samples/send-batch-events/README.md
azabbasi: this is due to the edition of the wrong file. fixed.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -184,6 +185,34 @@ public void sendEventAsync(Message message, IotHubEventCallback callback, Object
         deviceIO.sendEventAsync(message, callback, callbackContext, this.config.getDeviceId());
     }
 
+    /**
+     * ASynchronously sends a batch of messages to the IoT Hub
```
timtay-microsoft, line 189: Async... not ASync

PR #890, opened 2020-08-11
Update IotHubServiceSasToken.java

lucasm997:
# Checklist

# Description of the problem
Calling String.valueOf may return "java.nio.charset.CharsetICU[UTF-8]" depending on java version, which causes an IllegalCharsetNameException. This happened to me on Android 5.1.1
# Description of the solution
Calling .name() instead solves the problem


timtay-microsoft:
Looks good to me, thanks for the fix!

lucasm997:
Glad i could help!

timtay-microsoft:
We're planning to do a release in the next few days, and I'll make sure this gets in for it

PR #891, opened 2020-08-12
fix(iot-dev): Fix issue where errors encountered at http message level were handled twice

timtay-microsoft:
This same logic took place at the IotHubTransport layer based on the result of this method call, so no need to include this logic in the HTTP layer

You can see this logic here:
https://github.com/Azure/azure-iot-sdk-java/blob/5867a9311ec7f30102e37106b3a26ec7f769c802/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java#L1026

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsIotHubConnection.java
@@ -151,11 +151,6 @@ public class HttpsIotHubConnection implements IotHubTransportConnection
                 //Codes_SRS_HTTPSIOTHUBCONNECTION_34_067: [If the response from the service is OK or OK_EMPTY, this function shall notify its listener that a message was sent with no exception.]
                 this.listener.onMessageSent(transportMessage, null);
             }
-            else
-            {
-                //Codes_SRS_HTTPSIOTHUBCONNECTION_34_068: [If the response from the service not OK or OK_EMPTY, this function shall notify its listener that a message was with the mapped IotHubServiceException.]
-                this.listener.onMessageSent(transportMessage, IotHubStatusCode.getConnectionStatusException(status, ""));
-            }
 
             return status;
         }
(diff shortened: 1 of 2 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsIotHubConnection.java
```diff
@@ -151,11 +151,6 @@ public IotHubStatusCode sendMessage(Message message) throws TransportException
                 //Codes_SRS_HTTPSIOTHUBCONNECTION_34_067: [If the response from the service is OK or OK_EMPTY, this function shall notify its listener that a message was sent with no exception.]
                 this.listener.onMessageSent(transportMessage, null);
             }
```
vinagesh, line 153: Can you point to where it is handled?

vinagesh: Never mind. Should have read description earlier :)

timtay-microsoft: Sure, I can add a note that non-200 codes have their exceptions handled in the transport layer

vinagesh: That would be good

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsIotHubConnection.java
```diff
@@ -173,11 +173,9 @@ public IotHubStatusCode sendMessage(Message message) throws TransportException
...
-            else
-            {
-                //Codes_SRS_HTTPSIOTHUBCONNECTION_34_068: [If the response from the service not OK or OK_EMPTY, this function shall notify its listener that a message was with the mapped IotHubServiceException.]
-                this.listener.onMessageSent(transportMessage, IotHubStatusCode.getConnectionStatusException(status, ""));
-            }
+
+            // Status codes other than 200 and 204 have their errors handled in the IotHubTransport layer once this method returns,
+            // so there is no need to call "this.listener.onMessageSent(transportMessage, someException)" from this layer.
```
bikamani, line 178: so was it throwing exception at both layers?

timtay-microsoft: Currently, the transport layer basically gets notified that the sent message failed twice. The first handling of the failed message works as expected, but the second handling fails because the message is already removed from the expected list and has already had its callback executed.

PR #892, opened 2020-08-12
fix(e2e): Fix timing issue in provisioning e2e tests

timtay-microsoft:
There is some propagation delay between when a device is provisioned to a hub and when it is queryable. This commit refactors the test to just query for the device until it can be found, rather than just query once and hope the device was ready to be queried.

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/setup/ProvisioningCommon.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/setup/ProvisioningCommon.java
@@ -44,6 +44,7 @@ import java.util.*;
 import static com.microsoft.azure.sdk.iot.provisioning.device.ProvisioningDeviceClientStatus.PROVISIONING_DEVICE_STATUS_ASSIGNED;
 import static junit.framework.TestCase.fail;
 import static org.junit.Assert.*;
+import static tests.integration.com.microsoft.azure.sdk.iot.iothub.twin.QueryTwinTests.QUERY_TIMEOUT_MILLISECONDS;
 
 public class ProvisioningCommon extends IntegrationTest
 {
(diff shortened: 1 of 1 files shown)


abhipsaMisra:
What kind of delay does this address? Is this to ensure that created the device identity is available on the hub after the Register() call has been completed on the dps device client instance?

timtay-microsoft:
> What kind of delay does this address? Is this to ensure that created the device identity is available on the hub after the Register() call has been completed on the dps device client instance?

It addresses the delay between when a device is registered on a hub through provisioning, and when the device is queryable through this query API

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/setup/ProvisioningCommon.java
```diff
@@ -448,11 +449,25 @@ private void assertProvisionedDeviceWorks(String iothubUri, String deviceId) thr
...
-    {
+    protected void assertProvisionedDeviceCapabilitiesAreExpected(DeviceCapabilities expectedDeviceCapabilities, String provisionedHubConnectionString) throws IOException, IotHubException, InterruptedException {
         DeviceTwin deviceTwin = DeviceTwin.createFromConnectionString(provisionedHubConnectionString);
-        Query query = deviceTwin.queryTwin("SELECT * FROM devices WHERE deviceId = '" + testInstance.provisionedDeviceId +"'");
-        assertTrue(CorrelationDetailsLoggingAssert.buildExceptionMessageDpsIndividualOrGroup("Provisioned device " + testInstance.provisionedDeviceId + "not found in expected hub", getHostName(provisioningServiceConnectionString), testInstance.groupId, testInstance.registrationId), deviceTwin.hasNextDeviceTwin(query));
+
+        boolean deviceFoundInCorrectHub = false;
+        Query query = null;
+        long startTime = System.currentTimeMillis();
+        while (!deviceFoundInCorrectHub)
```
bikamani, line 458: Doesn't this add delay of more than 4 minutes our tests where we keep querying to see if device is provisioned?

bikamani:
> > What kind of delay does this address? Is this to ensure that created the device identity is available on the hub after the Register() call has been completed on the dps device client instance?
> 
> It addresses the delay between when a device is registered on a hub through provisioning, and when the device is queryable through this query API

How often do we see this issue? Also why is the QUERY_TIMEOUT_MILLISECONDS defined as 4 minutes? Does it take that long?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/setup/ProvisioningCommon.java
timtay-microsoft: Only if the device never becomes queryable in this hub, which should never happen

timtay-microsoft:
> How often do we see this issue? Also why is the QUERY_TIMEOUT_MILLISECONDS defined as 4 minutes? Does it take that long?

I have been unable to do a canary merge for several days because of this issue, so it is pretty common. It shouldn't take 4 minutes, but it doesn't hurt to be generous with a timeout like this. The query tests used to have a 1 minute timeout for a similar scenario, but I had to increase it to 4 minutes because 1 minute wasn't always enough. It is usually enough, but we want 100% pass rate.

Issue #910, opened 2020-09-08
[Bug Report] Device Client API completeFileUploadAsync is not an async function


timtay-microsoft:
Need to deprecate that API and add a new one called "completeFileUpload" to make this less confusing

AB#8214907

timtay-microsoft:
#911

timtay-microsoft:
This issue has been fixed as of device client version 1.26.0, so I'm closing this issue

az-iot-builder-01:
@timtay-microsoft, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #911, opened 2020-09-09
feat(iot-dev): Fix file upload complete notification API

timtay-microsoft:
completeFileUploadAsync does not behave asynchronously, so the name was a mistake. Deprecating that API and creating a new one without the Async suffix

#910


abhipsaMisra (approved):
Is this required for the .NET client as well?

vinagesh:
Looks good. Should we patch the LTS ?

vinagesh (approved):

timtay-microsoft:
> Is this required for the .NET client as well?

No, because it is actually async in .NET

This java function returns syncrhonously, so it should never have been post-fixed with "async"

timtay-microsoft:
> Looks good. Should we patch the LTS ?

It isn't really a critical bug fix or a security fix, so I don't see a reason to

vinagesh:
> > Looks good. Should we patch the LTS ?
> 
> It isn't really a critical bug fix or a security fix, so I don't see a reason to

When do we finally delete the deprecated functions?

timtay-microsoft:
> > > Looks good. Should we patch the LTS ?
> > 
> > 
> > It isn't really a critical bug fix or a security fix, so I don't see a reason to
> 
> When do we finally delete the deprecated functions?

Never, probably

Issue #900, opened 2020-08-21
[Bug Report] DeviceClient.open() using AMQPS might throw NullPointerException. 


EinarBergseth:
# Context

- **OS and version used:** Linux version 4.4.187 (build@a78bfeb6277f) (gcc version 7.4.0 (Buildroot 2019.02.2-svn3578) )
- **Java runtime used:** Java(TM) SE Embedded Runtime Environment (build 1.8.0_251-b08, profile compact2, headless)
- **SDK version used:** iot-device-client 1.25.0

## Description of the issue
DeviceClient.open() using AMQPS might throw NullPointerException (if authenticationSessionOpen Timed Out). In this case the function com.microsoft.azure.sdk.iot.device.transport.amqps.AmqpsIotHubConnection.closeAsync() is called where the variable this.connection might be null. See the provided stack trace below.

I don't know how to reproduce this error. 

## Console log of the issue
java.lang.NullPointerException
        at com.microsoft.azure.sdk.iot.device.transport.amqps.AmqpsIotHubConnection.closeAsync(AmqpsIotHubConnection.java:712)
        at com.microsoft.azure.sdk.iot.device.transport.amqps.AmqpsIotHubConnection.close(AmqpsIotHubConnection.java:175)
        at com.microsoft.azure.sdk.iot.device.transport.amqps.AmqpsIotHubConnection.closeConnectionWithException(AmqpsIotHubConnection.java:690)
        at com.microsoft.azure.sdk.iot.device.transport.amqps.AmqpsIotHubConnection.open(AmqpsIotHubConnection.java:143)
        at com.microsoft.azure.sdk.iot.device.transport.IotHubTransport.openConnection(IotHubTransport.java:785)
        at com.microsoft.azure.sdk.iot.device.transport.IotHubTransport.open(IotHubTransport.java:325)
        at com.microsoft.azure.sdk.iot.device.DeviceIO.open(DeviceIO.java:156)
        at com.microsoft.azure.sdk.iot.device.InternalClient.open(InternalClient.java:146)
        at com.microsoft.azure.sdk.iot.device.DeviceClient.open(DeviceClient.java:395)
        at e2u.gateway.commandservice.CommandServiceConnectionImpl.open(CommandServiceConnectionImpl.java:40)
        at e2u.gateway.commandservice.CommandServiceImpl.openConnection(CommandServiceImpl.java:68)
[...]

PR #915, opened 2020-09-15
fix(iot-dev): Fix possible null pointer issue in amqp layer

timtay-microsoft:
Issue #900 has an example of our AmqpsExceptionTranslator throwing a null pointer when the passed in errorCode is null. This PR adds a null check to that argument to ensure a null pointer isn't thrown there

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsExceptionTranslator.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsExceptionTranslator.java
@@ -14,6 +14,13 @@ public class AmqpsExceptionTranslator
 {
     static TransportException convertFromAmqpException(ErrorCondition error)
     {
+        if (error == null)
+        {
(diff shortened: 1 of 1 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsExceptionTranslator.java
```diff
@@ -14,6 +14,13 @@
 {
     static TransportException convertFromAmqpException(ErrorCondition error)
     {
+        if (error == null)
+        {
+            TransportException t = new TransportException("An unknown transport exception occurred");
+            t.setRetryable(true);
```
abhipsaMisra, line 20: shouldn't we mark an unknown error code to not be retryable?

timtay-microsoft: That isn't the current behavior. Down below, if the error code doesn't match any known code, we return the same exception

abhipsaMisra (approved):

Issue #914, opened 2020-09-11
[Bug Report] Expected URL that uses protocol HTTPS but received one that uses protocol 'http'


fnbk:
The Device Client API `HttpsConnection` Function only supports HTTPS connections and throws an exception when HTTP connections are used. 

https://github.com/Azure/azure-iot-sdk-java/blob/11f1ea95ff47838b9fe42b106ebce3d28e2743cd/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsConnection.java#L60-L99

# Background

We are using the Device Client API in a module that connects to the iotedge daemon to read module-twin settings. 

Based on the edge runtime (edge or edge-on-k8s) a different URL will be provided by default:

**edge**
```
IOTEDGE_WORKLOADURI=unix:///var/run/iotedge/workload.sock
```

**edge-on-k8s**
```
IOTEDGE_WORKLOADURI=http://localhost:35001/
```

So in the regular edge environment (moby-engine) there are no issues, but in the kubernetes environment while a sidecar proxy is used to create connections to the edge daemon, the default configuration (`IOTEDGE_WORKLOADURI=http://localhost:35001/`) leads to the following error:

```
08:28:37,503 WARN  [com.iqs.cq.kc.twinconfig.TwinConfigurationProvider] (ServerService Thread Pool -- 71) post init, twinMonitor=null
08:28:37,522 INFO  [com.microsoft.azure.sdk.iot.device.ModuleClient] (ServerService Thread Pool -- 71) Creating module client from environment with protocol MQTT...
08:28:37,522 INFO  [com.microsoft.azure.sdk.iot.device.ModuleClient] (ServerService Thread Pool -- 71) No connection string was configured for this module, so it will get its credentials from the edgelet
08:28:37,577 INFO  [stdout] (ServerService Thread Pool -- 71) Exception
08:28:37,577 ERROR [stderr] (ServerService Thread Pool -- 71) java.lang.IllegalArgumentException: Expected URL that uses protocol HTTPS but received one that uses protocol 'http'.
08:28:37,577 ERROR [stderr] (ServerService Thread Pool -- 71)
[...]

PR #921, opened 2020-09-18
fix(iot-dev): fix issue where edgelet hsm requests did not support http

timtay-microsoft:
Some edgelet uris are http requests to localhosts, so the SDK needs to support that.

issue #914

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsRequest.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsRequest.java
@@ -115,6 +115,7 @@ public class HttpsRequest
     /**
      * Executes the HTTPS request.
      *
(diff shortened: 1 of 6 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsRequest.java
```diff
@@ -93,13 +93,41 @@ public HttpsRequest(URL url, HttpsMethod method, byte[] body, String userAgentSt
...
+
+    /**
+     * Executes the HTTPS request.
+     *
+     * @return an HTTPS response.
+     *
+     * @throws TransportException if the connection could not be established, or the
+     * input/output streams could not be accessed.
+     */
+    private HttpsResponse send(boolean isHttps) throws TransportException
```
azabbasi, line 123: you added javadocs so might as well add @param isHttps

azabbasi (approved):

PR #934, opened 2020-09-24
Refactored auth code to reuse what is there in track1 repo

vinagesh:
# Checklist

# Description of the problem

# Description of the solution


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/DigitalTwinAsyncClient.java
```diff
@@ -44,18 +45,18 @@
      * @param connectionString The IoTHub connection string
      */
     @Builder(builderMethodName = "buildFromConnectionString", builderClassName = "FromConnectionStringBuilder")
-    public DigitalTwinAsyncClient(@NonNull String connectionString) {
-        ServiceConnectionString serviceConnectionString = ServiceConnectionStringParser.parseConnectionString(connectionString);
-        SasTokenProvider sasTokenProvider = serviceConnectionString.createSasTokenProvider();
-        String httpsEndpoint = serviceConnectionString.getHttpsEndpoint();
+    public DigitalTwinAsyncClient(@NonNull String connectionString) throws IOException {
+        IotHubConnectionString iotHubConnectionString = IotHubConnectionStringBuilder.createConnectionString(connectionString);
+        IotHubServiceSasToken iotHubServiceSasToken = new IotHubServiceSasToken(iotHubConnectionString);
```
abhipsaMisra, line 50: do we get token caching and refresh?

vinagesh: Good point. I will check.

vinagesh: If you have context about this, let me know.

vinagesh: Do you know if we had that in the copied code?

vinagesh: Currently we have an expiry of 1 year for the SAS token so this implementation should be sufficient. We should revisit the logic of this in a separate task. I don't think maintaining code separately for DigitalTwinClient is the best option.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/DigitalTwinAsyncClient.java
```diff
@@ -44,18 +45,18 @@
...
+        String httpsEndpoint = "https://" + iotHubConnectionString.getHostName();
```
drwill-ms, line 51: is there not a way in Java to create a well formed URI? String concatenation makes me feel uneasy.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/DigitalTwinAsyncClient.java
drwill-ms: I'm curious about this too. Can we talk about it in parking lot or in a meeting?

vinagesh: yes sure. I wanted to discuss with Abhipsa and the pnp team before bringing it to standup.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/DigitalTwinAsyncClient.java
vinagesh: I don't see it done other ways. I will try to google some more. But would be great to know from others who code in java.

PR #935, opened 2020-09-25
Refactoring code for API version

bikamani:
Refactor API Version changes


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/TransportUtils.java
```diff
@@ -9,9 +9,6 @@
 {
     public static String IOTHUB_API_VERSION = "2020-09-30";
 
-    // The preview API version support has been added to enable support for PnP features. This will be removed once the service goes GA.
-    public static String IOTHUB_API_VERSION_PREVIEW = "2020-09-30";
```
bikamani, line 13: removing since we do not need this anymore

PR #936, opened 2020-09-25
Minor fixes to the samples

azabbasi:
Fix a few minor issues with the samples.


service/iot-service-samples/device-manager-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ModuleManagerSample.java
```diff
@@ -46,6 +53,33 @@ public static void main(String[] args) throws Exception
         System.out.println("Shutting down sample...");
     }
 
+    private static void AddDevice() throws Exception
```
azabbasi, line 56: The sample assumes that the device exists. This makes sure that even if the device doesn't exist the sample runs end to end.

PR #944, opened 2020-09-29
Throwing correct exceptions

vinagesh:
# Checklist

# Description of the problem

# Description of the solution


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/helpers/Tools.java
```diff
@@ -44,7 +45,7 @@
             return Observable.just(result);
         }
         catch (JsonProcessingException e) {
-            return Observable.error(e);
+            return Observable.error(new IotHubException("Failed to parse the resonse"));
```
bikamani, line 48: Is this what's decided to throw as an error to user?

bikamani (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/helpers/Tools.java
vinagesh: yes this is what we discussed.

PR #946, opened 2020-09-29
Moving Pnp helpers to sample

bikamani:
Moving PnP helpers to device samples

### DIFF
--- a/device/iot-device-samples/pnp-device-sample/temperature-controller-device-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/device/PnpHelpers.java
+++ b/device/iot-device-samples/pnp-device-sample/temperature-controller-device-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/device/PnpHelpers.java
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-package com.microsoft.azure.sdk.iot.device.plugandplay;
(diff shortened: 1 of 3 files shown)


device/iot-device-samples/pnp-device-sample/temperature-controller-device-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/device/PnpHelpers.java
```diff
@@ -15,10 +15,11 @@
 
 import static java.util.Collections.singleton;
 import static java.util.Collections.singletonMap;
+
 /*
  A helper class for formatting command requests and properties as per plug and play convention.
 */
-public class PnpHelper {
+public class PnpHelpers {
```
drwill-ms, line 22: Let's call it PnpConvention since you have it in the samples.* package.

drwill-ms (approved):
We'll also need to add batch telemetry/property support, but that can be done in another PR, and possibly after GA.

barustum (approved):

device/iot-device-samples/pnp-device-sample/temperature-controller-device-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/device/PnpHelpers.java
bikamani: I have addressed it. Thanks

PR #951, opened 2020-09-29
Exclude packages for javadocs

bikamani:
Exclude packages for javadocs


service/iot-service-client/pom.xml
```diff
@@ -135,6 +135,11 @@
                     </execution>
                 </executions>
                 <configuration>
+                    <excludePackageNames>
+                        com.microsoft.azure.sdk.iot.service.digitaltwin.generated:
```
bikamani, line 139: For some reason it shows messed up here but looks fine in my editor.

azabbasi (approved):

PR #962, opened 2020-10-07
feat(samples): Add Pnp Convention unit tests

bikamani:
### Changes include
- Add unit tests for Pnp Helper.
- Thermostat and Temperature controller should be modules of sample projects and not service.

### DIFF
--- a/device/iot-device-samples/pnp-device-sample/temperature-controller-device-sample/pom.xml
+++ b/device/iot-device-samples/pnp-device-sample/temperature-controller-device-sample/pom.xml
@@ -23,6 +23,12 @@
             <version>2.11.2</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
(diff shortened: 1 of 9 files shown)


device/iot-device-samples/pnp-device-sample/temperature-controller-device-sample/pom.xml
```diff
@@ -23,6 +23,12 @@
             <version>2.11.2</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
```
barustum, line 29: why do we need to add junit for the samples project?

barustum (approved):

service/pom.xml
```diff
@@ -20,9 +20,6 @@
     <modules>
         <module>iot-service-client</module>
         <module>iot-service-samples</module>
-        <module>iot-service-samples/digitaltwin-service-samples</module>
```
bikamani, line 23: Note: samples shouldn't be included as modules here but in service-samples.

PR #968, opened 2020-10-13
Update preview API version to GA

bikamani:
Update preview API version to GA


abhipsaMisra (approved):
nit: we are adding a small note on the .NET client explaining that even though this is a preview library, we are using a GA version here because of the way service has managed the versioning (preview features are available under a GA service API version, and are restricted based on the region it is deployed under).
This will help address the confusion as to why we are using a GA service API version in a preview library, and also call out that this is not accidental :)

vinagesh (approved):

drwill-ms (approved):

bikamani:
> nit: we are adding a small note on the .NET client explaining that even though this is a preview library, we are using a GA version here because of the way service has managed the versioning (preview features are available under a GA service API version, and are restricted based on the region it is deployed under).
> This will help address the confusion as to why we are using a GA service API version in a preview library, and also call out that this is not accidental :)

It's definitely good to add a note, I'll add this in my next PR

PR #985, opened 2020-10-27
doc(service): Add headers for getter and settter of model Id

bikamani:
Add headers for getter and settter of model Id

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
@@ -647,6 +647,7 @@ public class DeviceTwinDevice
     /**
      * Getter to get model Id
      * @return The DTDL model Id of the device
+     * The value will be null for a non-pnp device.
      * The value will be null for a pnp device until the device connects and registers with the model Id.
      */
     public String getModelId() {
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
@@ -645,17 +645,17 @@ public class DeviceTwinDevice
     }
 
     /**
-     * Getter to get model Id
+     * Getter for a model Id
      * @return The DTDL model Id of the device
      * The value will be null for a non-pnp device.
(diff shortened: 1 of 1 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
```diff
@@ -644,10 +644,21 @@ private TwinCollection setToMap(Set<Pair> set)
         return map;
     }
 
+    /**
+     * Getter to get model Id
```
drwill-ms, line 648: Is this kind of language typical? "Getter to get" and "Setter to set" seems redundant.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
```diff
@@ -644,10 +644,21 @@ private TwinCollection setToMap(Set<Pair> set)
...
+     * @return The DTDL model Id of the device
+     * The value will be null for a non-pnp device.
+     * The value will be null for a pnp device until the device connects and registers with the model Id.
```
drwill-ms, line 651: with **a** model Id.

drwill-ms (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
bikamani: Yes that's the language I have seen for other get and set method.

bikamani: addressed this.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
bikamani: done

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
timtay-microsoft: Then the other places should be updated. This isn't a good description. "Gets the model Id" is much better

PR #987, opened 2020-11-04
refactor(iot-dev, e2e): Fix log statement, add throttle resiliency to tests

timtay-microsoft:
Also add bulk add functionality to e2e tests to help with upcoming multiplexing tests since we don't have this feature in our registry manager yet, and may never add it

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ExportImportDevicesParser.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ExportImportDevicesParser.java
@@ -44,11 +44,9 @@ public class ExportImportDevicesParser {
             throw new IllegalArgumentException("The provided json cannot be null or empty");
         }
 
-        ExportImportDevicesParser exportImportDevicesParser;
         try
         {
-            exportImportDevicesParser = gson.fromJson(json, ExportImportDevicesParser.class);
-            this.exportImportDevices = exportImportDevicesParser.getExportImportDevices();
+            this.exportImportDevices = gson.fromJson(json, ExportImportDevicesParser.class).getExportImportDevices();
         }
         catch (JsonSyntaxException e)
         {
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ExportImportDevicesParser.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ExportImportDevicesParser.java
@@ -22,8 +22,7 @@ public class ExportImportDevicesParser {
      */
     public String toJson()
     {
-        String json = gson.toJson(exportImportDevices);
-        return json;
+        return gson.toJson(exportImportDevices);
     }
 
     /**
@@ -36,7 +35,7 @@ public class ExportImportDevicesParser {
 
     /**
      * Constructor for an ExportImportDevicesParser that is built using the provided json
-     * @param json the json string to build the ExportImportDevicesParser out of
+     * @param json the json string used to build the ExportImportDevicesParser
      */
     public ExportImportDevicesParser(String json)
     {
@@ -45,17 +44,16 @@ public class ExportImportDevicesParser {
             throw new IllegalArgumentException("The provided json cannot be null or empty");
         }
 
-        ExportImportDevicesParser exportImportDevicesParser = null;
+        ExportImportDevicesParser exportImportDevicesParser;
         try
         {
             exportImportDevicesParser = gson.fromJson(json, ExportImportDevicesParser.class);
+            this.exportImportDevices = exportImportDevicesParser.getExportImportDevices();
         }
         catch (JsonSyntaxException e)
         {
             throw new IllegalArgumentException("The provided json could not be parsed");
         }
-
-        this.exportImportDevices = exportImportDevicesParser.getExportImportDevices();
     }
 
     public final Iterable<ExportImportDeviceParser> getExportImportDevices() {
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ExportImportDevicesParser.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ExportImportDevicesParser.java
@@ -1,0 +1,68 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package tests.integration.com.microsoft.azure.sdk.iot.helpers;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
+import com.google.gson.annotations.SerializedName;
+import com.microsoft.azure.sdk.iot.deps.serializer.ExportImportDeviceParser;
+
+import java.util.ArrayList;
+
+public class ExportImportDevicesParser {
+    private static final String DEVICES = "devices";
+    @SerializedName(DEVICES)
+    Iterable<ExportImportDeviceParser> exportImportDevices;
+    private transient static Gson gson = new Gson();
+
+    /**
+     * Converts this into json and returns it
+     * @return the json representation of this
+     */
+    public String toJson()
+    {
+        String json = gson.toJson(exportImportDevices);
+        return json;
+    }
+
+    /**
+     * Empty constructor: Used only to keep GSON happy.
+     */
+    public ExportImportDevicesParser()
+    {
+        exportImportDevices = new ArrayList<>();
+    }
+
+    /**
+     * Constructor for an ExportImportDevicesParser that is built using the provided json
+     * @param json the json string to build the ExportImportDevicesParser out of
+     */
+    public ExportImportDevicesParser(String json)
+    {
+        if (json == null || json.isEmpty())
+        {
+            throw new IllegalArgumentException("The provided json cannot be null or empty");
+        }
+
+        ExportImportDevicesParser exportImportDevicesParser = null;
+        try
+        {
+            exportImportDevicesParser = gson.fromJson(json, ExportImportDevicesParser.class);
+        }
+        catch (JsonSyntaxException e)
+        {
+            throw new IllegalArgumentException("The provided json could not be parsed");
+        }
+
+        this.exportImportDevices = exportImportDevicesParser.getExportImportDevices();
+    }
+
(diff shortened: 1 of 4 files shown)


timtay-microsoft:
The additions to the Tools file are in an effort to get the bulk device creation APIs that IoT Hub exposes, but that this SDK doesn't implement. Typically I'd prefer to just add those APIs to our registry manager so that customers can use them, but it would take more API design work than I'd like to spend now. So in the meantime, the code will just exist in our e2e test code.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ExportImportDevicesParser.java
```diff
@@ -0,0 +1,68 @@
...
+    Iterable<ExportImportDeviceParser> exportImportDevices;
+    private transient static Gson gson = new Gson();
+
+    /**
+     * Converts this into json and returns it
+     * @return the json representation of this
+     */
+    public String toJson()
+    {
+        String json = gson.toJson(exportImportDevices);
```
vinagesh, line 25: return gson.toJson(exportImportDevices) ?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ExportImportDevicesParser.java
```diff
@@ -0,0 +1,68 @@
...
+     * Empty constructor: Used only to keep GSON happy.
+     */
+    public ExportImportDevicesParser()
+    {
+        exportImportDevices = new ArrayList<>();
+    }
+
+    /**
+     * Constructor for an ExportImportDevicesParser that is built using the provided json
+     * @param json the json string to build the ExportImportDevicesParser out of
```
vinagesh, line 39: minor: "the json string used to build the ExportImportDevicesParser" might sound better.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ExportImportDevicesParser.java
```diff
@@ -0,0 +1,68 @@
...
+        try
+        {
+            exportImportDevicesParser = gson.fromJson(json, ExportImportDevicesParser.class);
+        }
+        catch (JsonSyntaxException e)
+        {
+            throw new IllegalArgumentException("The provided json could not be parsed");
+        }
+
+        this.exportImportDevices = exportImportDevicesParser.getExportImportDevices();
```
vinagesh, line 58: minor: This could be moved below line 51

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ExportImportDevicesParser.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public final Iterable<ExportImportDeviceParser> getExportImportDevices() {
```
vinagesh, line 61: Does final mean we can only set this value once in the constructor? Trying to understand how java works.

timtay-microsoft: The ```final``` modifier simply means that the method cannot be overriden by any class that extends this class. You can set/get this value as many times as you'd like, and whenever you'd like

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -139,6 +153,98 @@ public static Device addDeviceWithRetry(RegistryManager registryManager, Device
...
+
+        List<ExportImportDeviceParser> parsers = new ArrayList<>();
+        for (Device device : devices)
+        {
+            ExportImportDeviceParser exportImportDevice = new ExportImportDeviceParser();
+            exportImportDevice.setId(device.getDeviceId());
+            AuthenticationParser authenticationParser = new AuthenticationParser();
+            authenticationParser.setSymmetricKey(new SymmetricKeyParser(device.getSymmetricKey().getPrimaryKey(), device.getSymmetricKey().getSecondaryKey()));
+            exportImportDevice.setAuthentication(authenticationParser);
+            exportImportDevice.setImportMode("create");
```
vinagesh, line 217: nit: create constant for this

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -139,6 +153,98 @@ public static Device addDeviceWithRetry(RegistryManager registryManager, Device
...
+        List<Device> subIterable = new ArrayList<>();
+        List<Device> devicesClone = new ArrayList<>(); //create a clone of the source list so elements can be removed from it instead
+        devicesClone.addAll(devices);
+        while (devicesClone.size() > 0)
+        {
+            Device device = devicesClone.remove(0);
+            subIterable.add(device);
+
+            // wait until sub list has the Iot Hub limit of 100 devices to add, or until there will be no more devices to add
+            if (subIterable.size() > 99 || devicesClone.size() <= 0)
```
vinagesh, line 169: Should this be if or while? Are we not adding only one to the list or am I missing something?

timtay-microsoft: This if block represents what should happen once a list of 100 devices is ready to be added, or if a list of under 100 devices is ready to add and there are no more devices to add after this list

timtay-microsoft: The purpose here is so that this method can take a list of arbitrary length, and chunk that list into smaller lists of 100 devices or fewer so that we can use the bulk add service API which has a limit of 100 devices per request

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ExportImportDevicesParser.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param json the json string used to build the ExportImportDevicesParser
+     */
+    public ExportImportDevicesParser(String json)
+    {
+        if (json == null || json.isEmpty())
+        {
+            throw new IllegalArgumentException("The provided json cannot be null or empty");
+        }
+
+        ExportImportDevicesParser exportImportDevicesParser;
```
drwill-ms, line 47: Why is this declared outside of the try/catch?

timtay-microsoft: This can be simplified, sure

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -139,6 +153,99 @@ public static Device addDeviceWithRetry(RegistryManager registryManager, Device
...
```
drwill-ms, line 169: comparison seems numerically backwards (checking the large number before the small number), but maybe that's just me

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -139,6 +153,99 @@ public static Device addDeviceWithRetry(RegistryManager registryManager, Device
...
+            {
+                long startTime = System.currentTimeMillis();
+                while (System.currentTimeMillis() - startTime < RETRY_TIMEOUT_ON_NETWORK_FAILURE_MILLISECONDS)
+                {
+                    try
+                    {
+                        addDevices(subIterable, connectionString);
+                        break;
+                    }
+                    catch (UnknownHostException | SocketException | SocketTimeoutException e)
```
drwill-ms, line 179: Well that is something interesting that Java can do

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/rules/ThrottleResistantTestRule.java
```diff
@@ -49,6 +50,19 @@ public void evaluate() throws Throwable {
                     log.warn("Thottling detected in test {}, waiting for {} milliseconds and then re-running the test", description.getMethodName(), THROTTLING_RETRY_DELAY_MILLISECONDS, e);
                     Thread.sleep(THROTTLING_RETRY_DELAY_MILLISECONDS);
                 }
+                catch (Exception e)
+                {
+                    if (e.getMessage() != null && e.getMessage().toLowerCase().contains(CONNECTION_REFUSED))
+                    {
+                        log.warn("Thottling detected in test {}, waiting for {} milliseconds and then re-running the test", description.getMethodName(), THROTTLING_RETRY_DELAY_MILLISECONDS, e);
```
drwill-ms, line 57: I only see 2 instances of `{}`, but 3 parameters. Is the exception parameter a special one?

timtay-microsoft: The exception parameter is indeed special here. The log.warn (and log.error) will handle this exception without needing a ```{}```

drwill-ms (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
timtay-microsoft: I like it, personally

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
timtay-microsoft: I'm not sure I follow which half of the if statement you mean here

Issue #989, opened 2020-11-10
[Bug Report] Incorrect signature generation for Symmetric Key Device Provisioning


daawnee:
# Context

- **OS and version used:** N/A
- **Java runtime used:** N/A
- **SDK version used:** N/A

## Description of the issue
When constructing the SAS token, the token URL is URL encoded *before* it is signed.
Thus, the resulting signature included in the SharedAccessSignature is incorrect.

See this line: [RegisterTask.java#L154](https://github.com/Azure/azure-iot-sdk-java/blob/283b0098949015413fcfa84fbf0096b183476042/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/RegisterTask.java#L154)

Expected: `0neXXXXXXXX/registrations/DEVICE_ID`
Actual: `0neXXXXXXXX%2Fregistrations%2FDEVICE_ID`

The actual message to be signed is compiled at [line 162](https://github.com/Azure/azure-iot-sdk-java/blob/283b0098949015413fcfa84fbf0096b183476042/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/RegisterTask.java#L162), where you can observe that it takes the _actual_ URL-encoded version as a basis (`0neXXXXXXXX%2Fregistrations%2FDEVICE_KEY`) instead of the expected, non-URL-encoded version (`0neXXXXXXXX/registrations/DEVICE_KEY`), which then gets HMAC signed in [line 172](https://github.com/Azure/azure-iot-sdk-java/blob/283b0098949015413fcfa84fbf0096b183476042/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/RegisterTask.java#L172), resulting in an incorrect signature.

## Code sample exhibiting the issue
N/A

## Console log of the issue
N/A

AB#8770376

timtay-microsoft:
That does seem incorrect. How does this fix look to you? https://github.com/Azure/azure-iot-sdk-java/pull/990

timtay-microsoft:
Out of curiosity, were there characters in your registrationId got messed up from this? I'd be interested in adding an e2e test around this.

daawnee:
> #990

Seems all right.

daawnee:
> Out of curiosity, were there characters in your registrationId got messed up from this? I'd be interested in adding an e2e test around this.

[...]

PR #990, opened 2020-11-12
fix(prov-dev): Fix issue where sas token url was url encoded before being signed

timtay-microsoft:
the old flow was url encode -> sign -> url encode again

the correct flow is simply sign -> url encode

fixes issue #989


abhipsaMisra (approved):
Given that we had a unit test that expected the string to be signed to already be url encoded, any idea why we might have designed it that way?

timtay-microsoft:
> Given that we had a unit test that expected the string to be signed to already be url encoded, any idea why we might have designed it that way?

Not a clue. It was definitely being url encoded twice, which seems accidental to me

PR #992, opened 2020-11-13
Update samples to include reconnection logic

azabbasi:
image


device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/TransportClientTest.java
```diff
@@ -320,7 +320,7 @@ public void registerDeviceClientThrowsDeviceClientNull()
         TransportClient transportClient = new TransportClient(iotHubClientProtocol);
 
         // act
-        Deencapsulation.invoke(transportClient, "registerDeviceClient", (DeviceClient)null);
```
timtay-microsoft, line 323: You don't need to make this change, my branch should have this fix already

azabbasi: I know ! I have rebased onto your branch before submitting the PR, no idea why it shows up in the PR, if your branch already has this change it should just ignore it in the PR ... strange

timtay-microsoft: I'd pull from my branch again. I've done some force pushes after squashing these kinds of commits

device/iot-device-samples/multiplexing-sample/pom.xml
```diff
@@ -9,21 +9,48 @@
...
+            <version>${slf4j-version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
+            <version>${slf4j-version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.microsoft.azure.sdk.iot</groupId>
+            <artifactId>iot-e2e-common</artifactId>
```
timtay-microsoft, line 43: Is this dependency needed? I don't see anything imported from here

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexingSample.java
```diff
@@ -196,22 +128,34 @@ public static void main(String[] args)
...
+        multiplexClientManager.closeClient();
         System.out.println("Successfully closed the multiplexed connection");
+
+        System.out.println("Shutting down...");
+    }
+
+    private static int acknowledgedSentMessages = 0;
+    private static class TelemetryAcknowledgedEventCallback implements IotHubEventCallback
+    {
+        public void execute(IotHubStatusCode status, Object context)
```
barustum, line 154: Since this callback has a status, I am assuming it could be successful or a failure. Does it make sense to only increment the acknowledgedSentMessages only when it's a success? Otherwise, we might be acknowledging a message received, but in reality, it was unsuccessful.

barustum: Just noticed that we log the status on line 157. So might be good enough.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,161 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+
+import static com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeReason.NO_NETWORK;
+import static com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeReason.RETRY_EXPIRED;
+import static com.microsoft.azure.sdk.iot.device.transport.IotHubConnectionStatus.DISCONNECTED;
+import static com.microsoft.azure.sdk.iot.device.transport.IotHubConnectionStatus.DISCONNECTED_RETRYING;
+
+@Slf4j
```
barustum, line 18: curious to know what that is

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,161 @@
+package samples.com.microsoft.azure.sdk.iot;
+
+import com.microsoft.azure.sdk.iot.device.DeviceTwin.Pair;
```
timtay-microsoft, line 3: Java has a native Pair object that would be fine here. This Pair class in our library was a bad case of reinventing the wheel, so we shouldn't encourage its use

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexingSample.java
```diff
@@ -24,78 +19,9 @@
     // confirm when a device client is connected after registering it to an active multiplexed connection since the .registerDeviceClients(...)
     // call behaves asynchronously when the multiplexing client is already open. Similarly, this callback is used to track
     // when a device client is closed when unregistering it from an active connection.
-    public static class MultiplexedDeviceConnectionStatusChangeTracker implements IotHubConnectionStatusChangeCallback
-    {
-        private boolean isOpen = false;
-        private boolean isDisconnectedRetrying = false;
+    // In this sample we are using MultiplexClientManager to implement the reconnection
```
timtay-microsoft, line 22: Delete the comments above as they are mostly out of date

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeCallback;
+import com.microsoft.azure.sdk.iot.device.MultiplexingClient;
+import lombok.experimental.Delegate;
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+
+@Slf4j
+public class MultiplexClientManager extends ClientManagerBase {
```
timtay-microsoft, line 11: It's probably worthwhile to add a header to this class that explains why it is worth using, and what it does

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
```diff
@@ -0,0 +1,47 @@
...
+
+    MultiplexClientManager(MultiplexingClient multiplexingClient) {
+        this.connectionStatus = ConnectionStatus.DISCONNECTED;
+        this.client = multiplexingClient;
+        this.client.registerConnectionStatusChangeCallback(this, this);
+    }
+
+    @Override
+    public void openClient() throws IOException {
+        this.client.open();
```
timtay-microsoft, line 31: Why is this being overriden if it just delegates the call? Here and for close

vsts/StartEmulator.sh
```diff
@@ -3,7 +3,7 @@
 echo 'Listing available android sdks for installation'
 $ANDROID_HOME/tools/bin/sdkmanager --list | grep system-images
 
-emulatorImage='system-images;android-28;google_apis;x86_64'
```
timtay-microsoft, line 6: This change should also disappear after rebasing

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public void closeClient() throws IOException {
+        this.client.close();
+    }
+
+    @Override
+    public String getClientId() {
+        return  "Multiplexing Client";
```
timtay-microsoft, line 41: Maybe this should be a value that is set at MultiplexingClientManager constructor time so that users can differentiate between multiple MultiplexingClientManager instances

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback callback, Object callbackContext) {
+        if (callback != null) {
+            this.suppliedConnectionStatusChangeCallback = new Pair<>(callback, callbackContext);
+        } else {
+            this.suppliedConnectionStatusChangeCallback = null;
+        }
+    }
+
+    public void doConnect() throws IOException {
```
timtay-microsoft, line 42: I'm not sure I follow why doConnect() and open() both exist. Is there some benefit to splitting them up like this?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,161 @@
...
+                    }
+                }
+            }
+
+            try {
+                log.debug("Sleeping for 10 secs before attempting another open()");
+                Thread.sleep(SLEEP_TIME_BEFORE_RECONNECTING_IN_SECONDS * 1000);
+            }
+            catch (InterruptedException ex) {
+                throw new RuntimeException("InterruptedException in thread sleep: ", ex);
```
timtay-microsoft, line 70: I'd just let the method signature declare InterruptedException as a possibility. Any function that waits typically declares it as a possibility. At the very least, the error message here could be something like "Interrupted while waiting between attempting to open the client" to add some extra context

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,161 @@
...
+    @Override
+    public void execute(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable, Object callbackContext) {
+        Pair<IotHubConnectionStatusChangeCallback, Object> suppliedCallbackPair = this.suppliedConnectionStatusChangeCallback;
+        if (throwable == null)
+        {
+            System.out.println("CONNECTION STATUS UPDATE FOR CLIENT: " + getClientId() + " - " + status + ", " + statusChangeReason);
+        }
+        else
+        {
+            System.out.println("CONNECTION STATUS UPDATE FOR CLIENT: " + getClientId() + " - " + status + ", " + statusChangeReason + ", " + throwable.getMessage());
```
timtay-microsoft, line 109: Might be worthwhile to just call throwable.printStackTrace instead of appending the throwable.getMessage(). getMessage() does not include the full stack trace, so some information is lost

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,161 @@
...
+            suppliedCallbackPair.getKey().execute(status, statusChangeReason, throwable, suppliedCallbackPair.getValue());
+        }
+    }
+
+    // This handles the state where the DeviceClient reports that OperationTimeout has expired, and stops retrying; even though the applied RetryPolicy is still valid.
+    protected boolean shouldDeviceReconnect(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable) {
+        return (status == DISCONNECTED && statusChangeReason == RETRY_EXPIRED && throwable instanceof DeviceOperationTimeoutException);
+    }
+
+    public void handleRecoverableDisconnection() {
```
timtay-microsoft, line 128: The level of nesting within this function is a bit hard to follow. Consider splitting out the runnable into a separate method

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
timtay-microsoft: Unlike .NET, Java doesn't have a property on the exception itself that contains the stacktrace, so this is the best way to print it out when using System.out

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
barustum: I believe because we try to connect in the case of handling a recoverable disconnection; handleRecoverableDisconnection()

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
azabbasi: we are using the same base type for multiplexing client and device client, they are in charge of implementing how close work on their own corresponding clients

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
barustum, line 11: Since this class is in the samples folder, I am guessing this class (as well as the other managers) are intended as helper classes for the samples, correct?
It would be nice to have class documentation on what this class is, how it's intended to be used, and what benefits one would gain from using such helper classes.

barustum: Since this class is in the samples folder, I am guessing this class (as well as the other managers) are intended as helper classes for the samples, correct?
It would be nice to also mention what benefits one would gain from using such helper classes.

Question: if such helper classes are useful, why can't they be promoted to product code?

PR #993, opened 2020-11-16
Multiplexing samples: Device Client Manager and Multiplexing client manager

azabbasi:
With this change, `DeviceClientManager `objects will be wrapped around device clients to manage the reconnection logic. 
`DeviceClientManager `and `MultiplexingClientManager `both extend `ClientManagerBase `that hold the core reconnection logic.
All API calls to the `DeviceClientManager `and `MultiplexingClientManager `will be delegated to the client they are wrapping.

### DIFF
--- a/device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
+++ b/device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
@@ -1,6 +1,5 @@
 package samples.com.microsoft.azure.sdk.iot;
 
-import com.microsoft.azure.sdk.iot.device.DeviceTwin.Pair;
 import com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeCallback;
 import com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeReason;
 import com.microsoft.azure.sdk.iot.device.exceptions.DeviceOperationTimeoutException;
@@ -10,100 +9,83 @@ import lombok.extern.slf4j.Slf4j;
 
 import java.io.IOException;
 
-import static com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeReason.NO_NETWORK;
 import static com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeReason.RETRY_EXPIRED;
 import static com.microsoft.azure.sdk.iot.device.transport.IotHubConnectionStatus.DISCONNECTED;
-import static com.microsoft.azure.sdk.iot.device.transport.IotHubConnectionStatus.DISCONNECTED_RETRYING;
 
 /**
  * This is the base class for DeviceClientManager and MultiplexingClientManager
- * The shared logic for reconnection and handling status change call back exists here.
+ * The shared logic for reconnection and handling status change callback exists here.
  */
 @Slf4j
-public abstract class ClientManagerBase implements IotHubConnectionStatusChangeCallback {
-
-    protected enum ConnectionStatus {
-        DISCONNECTED, CONNECTING, CONNECTED
-    }
-    
-    protected static final Object lock = new Object();
+public abstract class ClientManagerBase implements IotHubConnectionStatusChangeCallback, ConnectionStatusTracker
+{
     protected static final int SLEEP_TIME_BEFORE_RECONNECTING_IN_SECONDS = 2;
-    protected ConnectionStatus connectionStatus;
-    protected Pair<IotHubConnectionStatusChangeCallback, Object> suppliedConnectionStatusChangeCallback;
+
+    // Initialize the connection status as DISCONNECTED
+    protected ConnectionStatus lastKnownConnectionStatus = ConnectionStatus.DISCONNECTED;
+
+    protected ConnectionStatusTracker dependencyConnectionStatusTracker;
 
     public abstract void openClient() throws IOException;
     public abstract void closeClient() throws IOException;
     public abstract String getClientId();
 
-    public void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback callback, Object callbackContext) {
-        if (callback != null) {
-            this.suppliedConnectionStatusChangeCallback = new Pair<>(callback, callbackContext);
-        } else {
-            this.suppliedConnectionStatusChangeCallback = null;
-        }
+    // Since the client manager is in charge of handling the connection status callback, this method is a no-op
+    public void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback callback, Object callbackContext)
+    {
+        return;
     }
 
-    public void doConnect() throws IOException {
-        // Device client does not have retry on the initial open() call. Will need to be re-opened by the calling application
-        while (connectionStatus == ConnectionStatus.CONNECTING) {
-            synchronized (lock) {
-                if(connectionStatus == ConnectionStatus.CONNECTING) {
-                    try {
-                        log.debug("Opening the device client instance " + getClientId() + " ...");
-                        openClient();
-                        connectionStatus = ConnectionStatus.CONNECTED;
-                        break;
-                    }
-                    catch (Exception ex) {
-                        if (ex.getCause() instanceof TransportException && ((TransportException) ex.getCause()).isRetryable()) {
-                            log.warn("Transport exception thrown while opening DeviceClient instance " + getClientId() + ", retrying: ", ex);
-                        } else {
-                            log.error("Non-retryable exception thrown while opening DeviceClient instance " + getClientId() + ": ", ex);
-                            connectionStatus = ConnectionStatus.DISCONNECTED;
-                            throw ex;
-                        }
-                    }
-                }
-            }
-
-            try {
-                log.debug("Sleeping for " + SLEEP_TIME_BEFORE_RECONNECTING_IN_SECONDS + " secs before attempting another open()");
-                Thread.sleep(SLEEP_TIME_BEFORE_RECONNECTING_IN_SECONDS * 1000);
-            }
-            catch (InterruptedException ex) {
-                throw new RuntimeException("Interrupted while waiting between attempting to open the client: ", ex);
-            }
-        }
-    }
-
-    public void closeNow() {
-        synchronized (lock) {
-            try {
+    public void closeNow()
+    {
+        synchronized (lastKnownConnectionStatus)
+        {
+            try
+            {
                 log.debug("Closing the client instance " + getClientId() + " ...");
                 closeClient();
             }
-            catch (IOException e) {
-                log.error("Exception thrown while closing client instance " + getClientId() + " : ", e);
-            } finally {
-                connectionStatus = ConnectionStatus.DISCONNECTED;
+            catch (IOException ex)
+            {
+                log.error("Exception thrown while closing client instance " + getClientId() + " : ", ex);
+            }
+            finally
+            {
+                // Once the connection is closed, set the Status to DISCONNECTED
+                lastKnownConnectionStatus = ConnectionStatus.DISCONNECTED;
             }
         }
     }
 
-    public void open() throws IOException {
-        synchronized (lock) {
-            if(connectionStatus == ConnectionStatus.DISCONNECTED) {
-                connectionStatus = ConnectionStatus.CONNECTING;
-            } else {
+    // When client manager is being opened it first makes sure the client is in a DISCONNECTED state
+    // If the client is in CONNECTING or CONNECTED state, Open will be no-op.
+    public void open() throws IOException
+    {
+        synchronized (lastKnownConnectionStatus)
+        {
+            // Do not attempt to CONNECT if the connection status is not DISCONNECTED. This ensure that only one process is going to attempt to connect
+            if (lastKnownConnectionStatus != ConnectionStatus.DISCONNECTED)
+            {
                 return;
             }
+            else
+            {
+                // Set the connection status to CONNECTING
+                lastKnownConnectionStatus = ConnectionStatus.CONNECTING;
+            }
         }
-        doConnect();
+
+        establishConnection();
     }
 
     @Override
-    public void execute(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable, Object callbackContext) {
-        Pair<IotHubConnectionStatusChangeCallback, Object> suppliedCallbackPair = this.suppliedConnectionStatusChangeCallback;
+    public ConnectionStatus getConnectionStatus(){
+        return this.lastKnownConnectionStatus;
+    }
+
+    @Override
+    public void execute(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable, Object callbackContext)
+    {
         if (throwable == null)
         {
             log.info("CONNECTION STATUS UPDATE FOR CLIENT: " + getClientId() + " - " + status + ", " + statusChangeReason);
@@ -114,52 +96,122 @@ public abstract class ClientManagerBase implements IotHubConnectionStatusChangeC
             throwable.printStackTrace();
         }
 
-        if (shouldDeviceReconnect(status, statusChangeReason, throwable)) {
-            if (suppliedCallbackPair != null) {
-                suppliedCallbackPair.getKey().execute(DISCONNECTED_RETRYING, NO_NETWORK, throwable, suppliedCallbackPair.getValue());
-            }
-
+        if (shouldDeviceReconnect(status, statusChangeReason, throwable))
+        {
             handleRecoverableDisconnection();
-        } else if (suppliedCallbackPair != null) {
-            suppliedCallbackPair.getKey().execute(status, statusChangeReason, throwable, suppliedCallbackPair.getValue());
         }
     }
 
-    // This handles the state where the DeviceClient reports that OperationTimeout has expired, and stops retrying; even though the applied RetryPolicy is still valid.
-    protected boolean shouldDeviceReconnect(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable) {
+    // This detects the state where the DeviceClient reports that OperationTimeout has expired, and stops retrying; even though the applied RetryPolicy is still valid.
+    // The logic to identify whether or not the connection should be established lives in this method.
+    // The client will automatically retry to establish the connection if the error is retryable
+    protected boolean shouldDeviceReconnect(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable)
+    {
         return (status == DISCONNECTED && statusChangeReason == RETRY_EXPIRED && throwable instanceof DeviceOperationTimeoutException);
     }
 
+    // We only expect the connection status to be CONNECTED by the time we enter this state.
     public void handleRecoverableDisconnection() {
-        synchronized (lock) {
-            if (connectionStatus == ConnectionStatus.CONNECTED) {
-                new Thread(new Runnable() {
-                    @Override
-                    public void run() {
-                        log.debug("Attempting reconnect for client: " + getClientId() + " ...");
-                        synchronized (lock) {
-                            if (connectionStatus == ConnectionStatus.CONNECTED) {
-                                try {
-                                    closeClient();
-                                } catch (Exception e) {
-                                    log.warn("Client " + getClientId() + " closeNow failed.", e);
-                                } finally {
-                                    connectionStatus = ConnectionStatus.CONNECTING;
-                                }
-                            } else {
-                                log.debug("Client `" + getClientId() + "` is currently connecting, or already connected; skipping...");
-                                return;
+        // If the lastKnownConnectionStatus is not in a CONNECTED state it can mean two things:
+        // 1: the status is CONNECTING, in which case there is nothing to be done.
+        // 2: the status is DISCONNECTED, in which case connection cannot be re-established.
+        if (lastKnownConnectionStatus == ConnectionStatus.CONNECTED)
+        {
+            new Thread(new Runnable() {
+                @Override
+                public void run() {
+                    log.debug("Attempting reconnect for client: " + getClientId() + " ...");
+                    synchronized (lastKnownConnectionStatus)
+                    {
+                        if (lastKnownConnectionStatus == ConnectionStatus.CONNECTED)
+                        {
+                            try
+                            {
+                                closeClient();
+                            }
+                            catch (Exception ex)
+                            {
+                                log.warn("Client " + getClientId() + " closeNow failed.", ex);
+                            }
+                            finally
+                            {
+                                lastKnownConnectionStatus = ConnectionStatus.CONNECTING;
                             }
                         }
-                        try {
-                            doConnect();
-                        } catch (IOException e) {
-                            log.error("Exception thrown while opening client instance: " + getClientId(), e);
+                        else
+                        {
+                            log.debug("Client `" + getClientId() + "` is currently connecting; skipping...");
+                            return;
                         }
                     }
-                }).start();
-            } else {
-                connectionStatus = ConnectionStatus.DISCONNECTED;
+
+                    // The client is now closed and the connection status is CONNECTING. Connection can be established now.
+                    try
+                    {
+                        establishConnection();
+                    }
+                    catch (IOException ex)
+                    {
+                        log.error("Exception thrown while opening client instance: " + getClientId(), ex);
+                    }
+                }
+            }).start();
+        }
+    }
+
+    public void establishConnection() throws IOException
+    {
+        // Device client does not have retry on the initial open() call. Will need to be re-opened by the calling application
+        // Lock the lastKnownConnectionStus so no other process will be able to change it while the client manager is attempting to open the connection.
+        synchronized (lastKnownConnectionStatus)
+        {
+            while (lastKnownConnectionStatus == ConnectionStatus.CONNECTING)
+            {
+                // If the client has dependencies to another client (in this case it could be the multiplexing client) we have to wait to make sure the
+                // dependent connection is established first.
+                if (dependencyConnectionStatusTracker != null && dependencyConnectionStatusTracker.getConnectionStatus() != ConnectionStatus.CONNECTED)
+                {
+                    try
+                    {
+                        log.info("Waiting for the dependent connection to be established before attempting to open the connection");
+                        Thread.sleep(SLEEP_TIME_BEFORE_RECONNECTING_IN_SECONDS * 1000);
+                    }
+                    catch (InterruptedException ex)
+                    {
+                        throw new RuntimeException("Interrupted while waiting between attempting to open the client: ", ex);
+                    }
+                }
+
+                try
+                {
+                    log.debug("Opening the client instance " + getClientId() + " ...");
+                    openClient();
+                    lastKnownConnectionStatus = ConnectionStatus.CONNECTED;
+                    break;
+                }
+                catch (Exception ex)
+                {
+                    if (ex.getCause() instanceof TransportException && ((TransportException) ex.getCause()).isRetryable())
+                    {
+                        log.warn("Transport exception thrown while opening client instance " + getClientId() + ", retrying: ", ex);
+                    }
+                    else
+                    {
+                        log.error("Non-retryable exception thrown while opening client instance " + getClientId() + ": ", ex);
+                        lastKnownConnectionStatus = ConnectionStatus.DISCONNECTED;
+                        throw ex;
+                    }
+                }
+
+                try
+                {
+                    log.debug("Sleeping for " + SLEEP_TIME_BEFORE_RECONNECTING_IN_SECONDS + " secs before attempting another open()");
+                    Thread.sleep(SLEEP_TIME_BEFORE_RECONNECTING_IN_SECONDS * 1000);
+                }
+                catch (InterruptedException ex)
+                {
+                    throw new RuntimeException("Interrupted while waiting between attempting to open the client: ", ex);
+                }
             }
         }
     }
--- a/device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
+++ b/device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
@@ -24,6 +24,7 @@ public abstract class ClientManagerBase implements IotHubConnectionStatusChangeC
     // Initialize the connection status as DISCONNECTED
     protected ConnectionStatus lastKnownConnectionStatus = ConnectionStatus.DISCONNECTED;
 
+    // Tracks connection status of the dependency connection (multiplexing client connection status for example in this case for DeviceClientManager)
     protected ConnectionStatusTracker dependencyConnectionStatusTracker;
 
     public abstract void openClient() throws IOException;
--- a/device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
+++ b/device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
@@ -1,0 +1,166 @@
+package samples.com.microsoft.azure.sdk.iot;
+
+import com.microsoft.azure.sdk.iot.device.DeviceTwin.Pair;
+import com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeCallback;
+import com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeReason;
+import com.microsoft.azure.sdk.iot.device.exceptions.DeviceOperationTimeoutException;
+import com.microsoft.azure.sdk.iot.device.exceptions.TransportException;
+import com.microsoft.azure.sdk.iot.device.transport.IotHubConnectionStatus;
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+
+import static com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeReason.NO_NETWORK;
+import static com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeReason.RETRY_EXPIRED;
+import static com.microsoft.azure.sdk.iot.device.transport.IotHubConnectionStatus.DISCONNECTED;
+import static com.microsoft.azure.sdk.iot.device.transport.IotHubConnectionStatus.DISCONNECTED_RETRYING;
+
+/**
+ * This is the base class for DeviceClientManager and MultiplexingClientManager
+ * The shared logic for reconnection and handling status change call back exists here.
+ */
+@Slf4j
+public abstract class ClientManagerBase implements IotHubConnectionStatusChangeCallback {
+
+    protected enum ConnectionStatus {
+        DISCONNECTED, CONNECTING, CONNECTED
+    }
+
+    protected static final Object lock = new Object();
+    protected static final int SLEEP_TIME_BEFORE_RECONNECTING_IN_SECONDS = 2;
+    protected ConnectionStatus connectionStatus;
+    protected Pair<IotHubConnectionStatusChangeCallback, Object> suppliedConnectionStatusChangeCallback;
+
+    public abstract void openClient() throws IOException;
+    public abstract void closeClient() throws IOException;
+    public abstract String getClientId();
+
+    public void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback callback, Object callbackContext) {
+        if (callback != null) {
+            this.suppliedConnectionStatusChangeCallback = new Pair<>(callback, callbackContext);
+        } else {
+            this.suppliedConnectionStatusChangeCallback = null;
+        }
+    }
+
+    public void doConnect() throws IOException {
+        // Device client does not have retry on the initial open() call. Will need to be re-opened by the calling application
+        while (connectionStatus == ConnectionStatus.CONNECTING) {
+            synchronized (lock) {
+                if(connectionStatus == ConnectionStatus.CONNECTING) {
+                    try {
+                        log.debug("Opening the device client instance " + getClientId() + " ...");
+                        openClient();
+                        connectionStatus = ConnectionStatus.CONNECTED;
+                        break;
+                    }
+                    catch (Exception ex) {
+                        if (ex.getCause() instanceof TransportException && ((TransportException) ex.getCause()).isRetryable()) {
+                            log.warn("Transport exception thrown while opening DeviceClient instance " + getClientId() + ", retrying: ", ex);
+                        } else {
+                            log.error("Non-retryable exception thrown while opening DeviceClient instance " + getClientId() + ": ", ex);
+                            connectionStatus = ConnectionStatus.DISCONNECTED;
+                            throw ex;
+                        }
+                    }
+                }
+            }
+
+            try {
+                log.debug("Sleeping for 10 secs before attempting another open()");
+                Thread.sleep(SLEEP_TIME_BEFORE_RECONNECTING_IN_SECONDS * 1000);
+            }
+            catch (InterruptedException ex) {
+                throw new RuntimeException("Interrupted while waiting between attempting to open the client: ", ex);
+            }
+        }
+    }
+
+    public void closeNow() {
+        synchronized (lock) {
+            try {
+                log.debug("Closing the client instance " + getClientId() + " ...");
+                closeClient();
+            }
+            catch (IOException e) {
+                log.error("Exception thrown while closing client instance " + getClientId() + " : ", e);
+            } finally {
+                connectionStatus = ConnectionStatus.DISCONNECTED;
+            }
+        }
+    }
+
+    public void open() throws IOException {
+        synchronized (lock) {
+            if(connectionStatus == ConnectionStatus.DISCONNECTED) {
+                connectionStatus = ConnectionStatus.CONNECTING;
+            } else {
+                return;
+            }
+        }
+        doConnect();
+    }
+
+    @Override
+    public void execute(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable, Object callbackContext) {
+        Pair<IotHubConnectionStatusChangeCallback, Object> suppliedCallbackPair = this.suppliedConnectionStatusChangeCallback;
+        if (throwable == null)
+        {
+            System.out.println("CONNECTION STATUS UPDATE FOR CLIENT: " + getClientId() + " - " + status + ", " + statusChangeReason);
+        }
+        else
+        {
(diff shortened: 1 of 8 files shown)


device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
+package samples.com.microsoft.azure.sdk.iot;
+
+import com.microsoft.azure.sdk.iot.device.DeviceTwin.Pair;
```
azabbasi, line 3: @timtay-microsoft  , you rightfully mentioned that we shouldn't be using this implementation and there is an out of the box Pair class in java. 
I found javafx.util.Pair implementation ... is that the one you had in mind?

bikamani:
> _No description provided._

Can you add description for better context? :)

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
timtay-microsoft: My mistake on this one. I thought there was a pair class within Java's collection namespace, but apparently there isn't. This one should be fine, though, sure.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+                            log.error("Non-retryable exception thrown while opening DeviceClient instance " + getClientId() + ": ", ex);
+                            connectionStatus = ConnectionStatus.DISCONNECTED;
+                            throw ex;
+                        }
+                    }
+                }
+            }
+
+            try {
+                log.debug("Sleeping for 10 secs before attempting another open()");
```
timtay-microsoft, line 70: Don't hardcode "10" into this message. Use the value from the variable below, in case we ever change its value

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+        }
+        doConnect();
+    }
+
+    @Override
+    public void execute(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable, Object callbackContext) {
+        Pair<IotHubConnectionStatusChangeCallback, Object> suppliedCallbackPair = this.suppliedConnectionStatusChangeCallback;
+        if (throwable == null)
+        {
+            System.out.println("CONNECTION STATUS UPDATE FOR CLIENT: " + getClientId() + " - " + status + ", " + statusChangeReason);
```
timtay-microsoft, line 109: Since we are using log.debug everywhere else, can you switch these from System.out to log.info or log.debug?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
```diff
@@ -0,0 +1,54 @@
...
+        void close();
+        void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback callback, Object callbackContext);
+    }
+
+    // The methods defined in the interface DeviceClientNonDelegatedFunction will be called on MultiplexingClientManager, and not on MultiplexingClient.
+    @Delegate(excludes = DeviceClientNonDelegatedFunction.class)
+    private final MultiplexingClient client;
+    private final String multiplexClientId;
+
+    MultiplexClientManager(MultiplexingClient multiplexingClient, String multiplexClientId) {
```
timtay-microsoft, line 29: Add a header here if only to clarify that the multiplexClientId is user-defined, and isn't some Id like deviceId where they get it from the portal.

timtay-microsoft (approved):
Just a few small comments, but otherwise looks good!

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
timtay-microsoft: For reference, there is an interface defined for a class like Pair in Map: https://docs.oracle.com/javase/8/docs/api/java/util/Map.Entry.html

But the implementation is private, so we have no access to it

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+    protected static final int SLEEP_TIME_BEFORE_RECONNECTING_IN_SECONDS = 2;
+    protected ConnectionStatus connectionStatus;
+    protected Pair<IotHubConnectionStatusChangeCallback, Object> suppliedConnectionStatusChangeCallback;
+
+    public abstract void openClient() throws IOException;
+    public abstract void closeClient() throws IOException;
+    public abstract String getClientId();
+
+    public void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback callback, Object callbackContext) {
+        if (callback != null) {
```
drwill-ms, line 39: I'd suggest swapping this condition to handle the positive case first

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+        } else {
+            this.suppliedConnectionStatusChangeCallback = null;
+        }
+    }
+
+    public void doConnect() throws IOException {
+        // Device client does not have retry on the initial open() call. Will need to be re-opened by the calling application
+        while (connectionStatus == ConnectionStatus.CONNECTING) {
+            synchronized (lock) {
+                if(connectionStatus == ConnectionStatus.CONNECTING) {
```
drwill-ms, line 50: space after `if`

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+                    try {
+                        log.debug("Opening the device client instance " + getClientId() + " ...");
+                        openClient();
+                        connectionStatus = ConnectionStatus.CONNECTED;
```
drwill-ms, line 54: Is `openClient()` synchronous, such that after it completes we can be sure we're connected?

Shouldn't we wait for a callback that says we're connected to change this internally tracked state to connected?

azabbasi: Yes, it is synchronous.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
drwill-ms, line 50: Suggest changing this to != connecting, and return, so the remaining code doesn't have to be tabbed in 1 more

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+    public void closeNow() {
+        synchronized (lock) {
+            try {
+                log.debug("Closing the client instance " + getClientId() + " ...");
+                closeClient();
+            }
+            catch (IOException e) {
+                log.error("Exception thrown while closing client instance " + getClientId() + " : ", e);
+            } finally {
+                connectionStatus = ConnectionStatus.DISCONNECTED;
```
drwill-ms, line 88: Why is locking not necessary here when changing this state but is elsewhere?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+            }
+        }
+        doConnect();
+    }
+
+    @Override
+    public void execute(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable, Object callbackContext) {
+        Pair<IotHubConnectionStatusChangeCallback, Object> suppliedCallbackPair = this.suppliedConnectionStatusChangeCallback;
+        if (throwable == null)
+        {
```
drwill-ms, line 108: very inconsistent brace usage in this file

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: The finally statement is also in the synchronized block. Am I missing something?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: fixing.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+@Slf4j
+public abstract class ClientManagerBase implements IotHubConnectionStatusChangeCallback {
+
+    protected enum ConnectionStatus {
+        DISCONNECTED, CONNECTING, CONNECTED
+    }
+    
+    protected static final Object lock = new Object();
+    protected static final int SLEEP_TIME_BEFORE_RECONNECTING_IN_SECONDS = 2;
+    protected ConnectionStatus connectionStatus;
```
drwill-ms, line 31: lastKnownConnectionStatus?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+                suppliedCallbackPair.getKey().execute(DISCONNECTED_RETRYING, NO_NETWORK, throwable, suppliedCallbackPair.getValue());
+            }
+
+            handleRecoverableDisconnection();
+        } else if (suppliedCallbackPair != null) {
+            suppliedCallbackPair.getKey().execute(status, statusChangeReason, throwable, suppliedCallbackPair.getValue());
+        }
+    }
+
+    // This handles the state where the DeviceClient reports that OperationTimeout has expired, and stops retrying; even though the applied RetryPolicy is still valid.
```
drwill-ms, line 128: It doesn't handle that state. It detects that state, right?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
+package samples.com.microsoft.azure.sdk.iot;
```
drwill-ms, line 1: Given this is a sample and that the task is complex, it is going to need way more code comments.

I'd like to see high level approach, and then for each method how it fits into the larger picture.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+                        connectionStatus = ConnectionStatus.CONNECTED;
+                        break;
+                    }
+                    catch (Exception ex) {
+                        if (ex.getCause() instanceof TransportException && ((TransportException) ex.getCause()).isRetryable()) {
+                            log.warn("Transport exception thrown while opening DeviceClient instance " + getClientId() + ", retrying: ", ex);
+                        } else {
+                            log.error("Non-retryable exception thrown while opening DeviceClient instance " + getClientId() + ": ", ex);
+                            connectionStatus = ConnectionStatus.DISCONNECTED;
+                            throw ex;
```
drwill-ms, line 63: Does Java share this antipattern with C#? That is, `throw ex;` vs `throw;`?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+                log.error("Exception thrown while closing client instance " + getClientId() + " : ", e);
+            } finally {
+                connectionStatus = ConnectionStatus.DISCONNECTED;
+            }
+        }
+    }
+
+    public void open() throws IOException {
+        synchronized (lock) {
+            if(connectionStatus == ConnectionStatus.DISCONNECTED) {
```
drwill-ms, line 95: Again, I think this would read easier if it was formatted as:

```Java
synchronized (lock) {
    if (connectionStatus != ConnectionStatus.DISCONNECTED) {
        return;
    }

    connectionStatus = ConnectionStatus.CONNECTING;
}

doConnect();
```

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
drwill-ms: What do you think about using the callback to change the cached state vs. setting it here?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+import java.io.IOException;
+
+import static com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeReason.NO_NETWORK;
+import static com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeReason.RETRY_EXPIRED;
+import static com.microsoft.azure.sdk.iot.device.transport.IotHubConnectionStatus.DISCONNECTED;
+import static com.microsoft.azure.sdk.iot.device.transport.IotHubConnectionStatus.DISCONNECTED_RETRYING;
+
+/**
+ * This is the base class for DeviceClientManager and MultiplexingClientManager
+ * The shared logic for reconnection and handling status change call back exists here.
```
drwill-ms, line 20: callback is 1 word

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexingSample.java
```diff
@@ -193,22 +122,34 @@ public static void main(String[] args)
...
         // This call will close all multiplexed device client instances as well
-        multiplexingClient.close();
+        multiplexClientManager.closeClient();
         System.out.println("Successfully closed the multiplexed connection");
+
+        System.out.println("Shutting down...");
+    }
+
+    private static int acknowledgedSentMessages = 0;
+    private static class TelemetryAcknowledgedEventCallback implements IotHubEventCallback
```
drwill-ms, line 146: Perhaps this belongs on the client wrapper classes we have?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Creates an instance of the MultiplexClientManager
+     *
+     * @param multiplexingClient the multiplexing client
+     * @param multiplexClientId user defined Id for the multiplexing client.
+     */
+    MultiplexClientManager(MultiplexingClient multiplexingClient, String multiplexClientId) {
+        this.connectionStatus = ConnectionStatus.DISCONNECTED;
+        this.client = multiplexingClient;
+        this.multiplexClientId = multiplexClientId;
```
drwill-ms, line 38: Where is this used?

drwill-ms: Just for logging, I guess?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+                    public void run() {
+                        log.debug("Attempting reconnect for client: " + getClientId() + " ...");
+                        synchronized (lock) {
+                            if (connectionStatus == ConnectionStatus.CONNECTED) {
+                                try {
+                                    closeClient();
+                                } catch (Exception e) {
+                                    log.warn("Client " + getClientId() + " closeNow failed.", e);
+                                } finally {
+                                    connectionStatus = ConnectionStatus.CONNECTING;
```
drwill-ms, line 147: Why wouldn't this be set to DISCONNECTED?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    public void closeNow() {
+        synchronized (lock) {
+            try {
+                log.debug("Closing the client instance " + getClientId() + " ...");
+                closeClient();
+            }
+            catch (IOException e) {
```
drwill-ms, line 85: Any reason not to use the same exception variable name throughout? We have `ex` and `e` in some places.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+                if(connectionStatus == ConnectionStatus.CONNECTING) {
+                    try {
+                        log.debug("Opening the device client instance " + getClientId() + " ...");
+                        openClient();
+                        connectionStatus = ConnectionStatus.CONNECTED;
+                        break;
+                    }
+                    catch (Exception ex) {
+                        if (ex.getCause() instanceof TransportException && ((TransportException) ex.getCause()).isRetryable()) {
+                            log.warn("Transport exception thrown while opening DeviceClient instance " + getClientId() + ", retrying: ", ex);
```
drwill-ms, line 59: The class comment says this is the base class for device and multiplexing manager clients, but this comment mentions the device client.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
```diff
@@ -0,0 +1,60 @@
...
+        this.client.registerConnectionStatusChangeCallback(this, this);
```
drwill-ms, line 39: How might/should the multiplexing connection status callback calls (where the status is disconnected_retrying) affect the device client manager statuses? They may still think they are connected, but they aren't. Or are they getting the callback too, and if so, what action are they taking?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,166 @@
...
+        }
+        else
+        {
+            log.info("CONNECTION STATUS UPDATE FOR CLIENT: " + getClientId() + " - " + status + ", " + statusChangeReason + ", " + throwable.getMessage());
+            throwable.printStackTrace();
+        }
+
+        if (shouldDeviceReconnect(status, statusChangeReason, throwable)) {
+            if (suppliedCallbackPair != null) {
+                suppliedCallbackPair.getKey().execute(DISCONNECTED_RETRYING, NO_NETWORK, throwable, suppliedCallbackPair.getValue());
```
drwill-ms, line 119: What's the difference between `getClientId()` and `suppliedCallbackPair.getValue()`? Are they both the device Id?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public void closeClient() throws IOException {
+        this.client.close();
+    }
+
+    @Override
+    public String getClientId() {
+        return  multiplexClientId;
```
drwill-ms, line 54: 2 spaces after return => 1

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public MultiplexingClient getMultiplexClient(){
```
drwill-ms, line 57: space before brace

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: good point. will add a lot more comments.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
drwill-ms: I wasn't looking broadly enough to see it is in a larger lock. Do we want this entire operation locked?

I thought the lock was just for conditionally acting on and changing the state of the cached last known connection status

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: no, you always have to throw an exception in the throw statement.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
azabbasi: correct.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: my bad, fixing.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: in this case, yes, they are both deviceId/multiplexing client Id

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: The DISCONNECTED status represents a final disconnected (and not going to retry) status. Connecting is basically saying, we are trying to open this connection and it's not yet connected.

drwill-ms: That might be evident to a sample reader, such as myself, if the enum values had comments describing what they are and when they are used

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
drwill-ms: Perhaps that makes me wonder why the multiplexing client doesn't have a way to uniquely identify it.

Anytime we have to write code in our samples (especially each time), I think that is a sign we are missing something in our client library. Do you agree?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
drwill-ms: Seems could be a source of confusion for a sample reader when we use both in the same function.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
azabbasi: well, the question is how would you actually uniquely identify the multiplexing client? we either have to assign a random identifier to it as a metadata upon initialization, or take it as an input form the user. I don't know how much value there is in providing extra metadata to identify the multiplexing client ...

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
azabbasi: let's chat about this in the live review together.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
drwill-ms: The multiplexing client doesn't need to be identified to the service, but it would be for local use, as you've done with logging. I'd make it an optional, read-only property on the multiplexing client. If not provided, we could initialize it with a GUID - maybe not.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexingSample.java
azabbasi: I tend to agree that Device can have a wrapper class that implements the TelemetryAck but I am not sure if this logic belongs to the same class. it's especially challenging because we can't implement two separate interfaces in JAVA. Let me do a quick implementation spike and see how it looks like.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: I like that

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: This block is updating the connectionStatus and opening the client (in which if there is no exceptions, means it's successful) ... so I think having it in the synchronized block is justified .
Did I understand your comment correctly?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: You mean in the `execute` method instead of `doConnect` ? sure, we can certainly do that, I am curious to know why you think that's a better approach? 
as I mentioned, if `openClient` doesn't throw, it means it successfully established the connection. also, the "connection status change callback" is basically invoked every time there is a change, and that doesn't exactly (or "accurately") maps to the states we are maintaining in this class.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: Comments added

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,220 @@
...
+
+    // Tracks connection status of the dependency connection (multiplexing client connection status for example in this case for DeviceClientManager)
+    protected ConnectionStatusTracker dependencyConnectionStatusTracker;
+
+    public abstract void openClient() throws IOException;
+    public abstract void closeClient() throws IOException;
+    public abstract String getClientId();
+
+    // Since the client manager is in charge of handling the connection status callback, this method is a no-op
+    public void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback callback, Object callbackContext)
```
drwill-ms, line 35: In .NET we'd make this abstract to ensure the inheriting class implemented it. Does Java have an analog?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,220 @@
...
+    {
+        return;
+    }
+
+    public void closeNow()
```
drwill-ms, line 40: Why `closeNow()` instead of just `close()`?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,220 @@
...
+    public void open() throws IOException
+    {
+        synchronized (lastKnownConnectionStatus)
+        {
+            // Do not attempt to CONNECT if the connection status is not DISCONNECTED. This ensure that only one process is going to attempt to connect
+            if (lastKnownConnectionStatus != ConnectionStatus.DISCONNECTED)
+            {
+                return;
+            }
+            else
```
drwill-ms, line 72: As previous block returns, `else` is not necessary

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,220 @@
...
+            {
+                // Set the connection status to CONNECTING
+                lastKnownConnectionStatus = ConnectionStatus.CONNECTING;
+            }
+        }
+
+        establishConnection();
```
drwill-ms, line 79: Same, why not just `connect()`?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    @Override
+    public ConnectionStatus getConnectionStatus(){
+        return this.lastKnownConnectionStatus;
+    }
+
+    @Override
+    public void execute(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable, Object callbackContext)
```
drwill-ms, line 88: Let's avoid the verb execute. It doesn't pass PoliCheck.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,220 @@
...
+            handleRecoverableDisconnection();
+        }
+    }
+
+    // This detects the state where the DeviceClient reports that OperationTimeout has expired, and stops retrying; even though the applied RetryPolicy is still valid.
+    // The logic to identify whether or not the connection should be established lives in this method.
+    // The client will automatically retry to establish the connection if the error is retryable
+    protected boolean shouldDeviceReconnect(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable)
+    {
+        return (status == DISCONNECTED && statusChangeReason == RETRY_EXPIRED && throwable instanceof DeviceOperationTimeoutException);
```
drwill-ms, line 111: For readability, consider putting each condition on a newline:

```Java
return status == DISCONNECTED
    && statusChangeReason == RETRY_EXPIRED
    && throwable instanceof DeviceOperationTimeoutException;
```

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,220 @@
...
+    {
+        // Device client does not have retry on the initial open() call. Will need to be re-opened by the calling application
+        // Lock the lastKnownConnectionStus so no other process will be able to change it while the client manager is attempting to open the connection.
+        synchronized (lastKnownConnectionStatus)
+        {
+            while (lastKnownConnectionStatus == ConnectionStatus.CONNECTING)
+            {
+                // If the client has dependencies to another client (in this case it could be the multiplexing client) we have to wait to make sure the
+                // dependent connection is established first.
+                if (dependencyConnectionStatusTracker != null && dependencyConnectionStatusTracker.getConnectionStatus() == ConnectionStatus.CONNECTING)
```
drwill-ms, line 173: Same idea as above, splitting conditions on multiple lines; it will also reduce line length here which is getting a bit long

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ConnectionStatusTracker.java
```diff
@@ -0,0 +1,6 @@
+package samples.com.microsoft.azure.sdk.iot;
+
+public interface ConnectionStatusTracker
```
drwill-ms, line 3: comment

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java
```diff
@@ -0,0 +1,58 @@
...
+import com.microsoft.azure.sdk.iot.device.DeviceClient;
+import com.microsoft.azure.sdk.iot.device.IotHubConnectionStatusChangeCallback;
+import lombok.experimental.Delegate;
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+
+/**
+ * This class is in charge of handling reconnection logic and registering callbacks for connection status changes.
+ * It will delegate all other calls other than `Open`, `Close` and registerConnectionStatusChangeCallbaack to the inner client (DeviceClient)
```
drwill-ms, line 12: baack

drwill-ms, line 12: I don't know if markdown works for Java doc comments, but you have the first two methods enclosed in backticks, but not the third. We should decorate them consistently.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: I don't think we can avoid it, it's in our SDK interface and we have to implement it :(

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
```diff
@@ -0,0 +1,60 @@
...
+    private interface DeviceClientNonDelegatedFunction
+    {
+        void open();
+        void close();
+        void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback callback, Object callbackContext);
+    }
+
+    // The methods defined in the interface DeviceClientNonDelegatedFunction will be called on MultiplexingClientManager, and not on MultiplexingClient.
+    @Delegate(excludes = DeviceClientNonDelegatedFunction.class)
+    private final MultiplexingClient client;
```
drwill-ms, line 27: Lots of "clients" to distinguish. Perhaps call this `sdkClient`?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
drwill-ms: Because getting confirmation from the SDK client that we've connected is more affirmative than asserting it should be connected.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
drwill-ms: Which SDK? The legacy DeviceClient or the new MultiplexingClient?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: There is the same concept in java, I didn't make it abstract because there is nothing to override, we don't want the subclasses to do anything with this, we just want to make sure the call is delegated to the ClientManager and it doesn't do anything. we can discuss it in more depth if you like.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: The suppliedCallbackPair is removed. We no longer have this issue.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
azabbasi: named them each after what they represent. `deviceClient `and `multiplexingClient `respectively in both `DeviceClientManager `and `MultiplexClientManager `

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
azabbasi: That is an excellent question.
DeviceClient has two different method signatures, close() and closeNow(). close() has been deprecated 

client managers for both DeviceClient and MultiplexingClient are setup such that any call that you make to them they will delegate to the inner client except for open, close and registerConnectionStatusCallback

Now there are 2 different method signatures on the DeviceClient for closing the connection, and we want to make sure we delegate those calls to the client manager and it knows how to override both of those methods.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexClientManager.java
azabbasi: I think at this stage since this is a sample code and it's only used for logging purposes taking it as a parameter in the constructor is reasonable. but I do agree that if at some point users start managing multiple multiplexingClients in their code, there might be a good argument that these clients should be easily identifiable which requires the core SDK logic to change (which is an addition and should be fine)

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexingSample.java
azabbasi: I think that would complicate the logic in client managers unnecessarily. Let's discuss this further if you feel really strongly about this.

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -0,0 +1,255 @@
...
+
+    /**
+     * This detects the state where the DeviceClient reports that OperationTimeout has expired, and sto
+     * The logic to identify whether or not the connection should be established lives in this method.
+     * The client will automatically retry to establish the connection if the error is retryable
+     */
+    protected boolean shouldClientReconnect(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable)
+    {
+        return status == DISCONNECTED
+                && statusChangeReason == RETRY_EXPIRED
```
drwill-ms, line 138: Why is this indented twice?

drwill-ms (approved):

PR #1000, opened 2020-11-18
Update readme with Baltimore cert update info

sharmasejal:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -156,18 +156,7 @@ Contains libraries that enable interactions with the IoT Hub service to perform 
 
 ## Certificates -  Important to know
 
-The Azure IoT Hub certificates presented during TLS negotiation shall be always validated using the appropriate root CA certificate(s).
-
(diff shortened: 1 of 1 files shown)


abhipsaMisra (approved):

timtay-microsoft (changes requested):
This isn't quite how a user should look at our certs. Our device client and module client constructors have optional parameters for taking custom SSLContexts. The readme should have something more like "Unless a custom SSLContext is provided, device clients and module clients will use the public certs for IoT Hub during the SSL handshake"

timtay-microsoft:
The Java SDK doesn't look at the certs registered on the machine running the SDK, as far as I know. We just have a hardcoded public cert for IoT Hub embedded in our SDK that gets loaded. There are other options for adding extra trusted certs as well, but nothing automatic for the customer like this commit suggests.

abhipsaMisra:
You can dismiss my approval @sharmasejal , as expected Java requires a different set of instructions!

timtay-microsoft (approved):

PR #1002, opened 2020-11-19
feat(iot-dev): Add client constructors that take sas token provider interface

timtay-microsoft:
Currently, users who have symmetric key authenticated devices have to provide the whole symmetric key to our SDK. Some customers have asked for the ability to just provide the SAS tokens instead so that they can store the symmetric key somewhere else, which is a reasonable ask. This PR add support for that scenario.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
@@ -144,6 +144,22 @@ public final class DeviceClientConfig
         this.productInfo = new ProductInfo();
     }
 
+
+    public DeviceClientConfig(String hostName, SasTokenProvider sasTokenProvider, ClientOptions clientOptions, String deviceId, String moduleId)
+    {
+        this.authenticationProvider = new IotHubSasTokenProvidedAuthenticationProvider(
+                hostName,
+                deviceId,
+                moduleId,
+                sasTokenProvider,
+                clientOptions != null ? clientOptions.sslContext : null);
+
+        this.useWebsocket = false;
+        this.productInfo = new ProductInfo();
+
+        log.debug("Device configured to use SAS token provided authentication provider");
+    }
+
     /**
      * Constructor for device configs that use x509 authentication
      *
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
@@ -147,12 +147,9 @@ public final class DeviceClientConfig
 
     public DeviceClientConfig(String hostName, SasTokenProvider sasTokenProvider, ClientOptions clientOptions, String deviceId, String moduleId)
     {
-        this.authenticationProvider = new IotHubSasTokenProvidedAuthenticationProvider(
-                hostName,
-                deviceId,
-                moduleId,
-                sasTokenProvider,
-                clientOptions != null ? clientOptions.sslContext : null);
+        SSLContext sslContext = clientOptions != null ? clientOptions.sslContext : null;
+        this.authenticationProvider =
+                new IotHubSasTokenProvidedAuthenticationProvider(hostName, deviceId, moduleId, sasTokenProvider, sslContext);
 
         this.useWebsocket = false;
         this.productInfo = new ProductInfo();
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
@@ -128,6 +128,29 @@ public class InternalClient
         this.deviceIO = new DeviceIO(this.config, sendPeriodMillis, receivePeriodMillis);
     }
 
+    InternalClient(String hostName, String deviceId, String moduleId, SasTokenProvider sasTokenProvider, IotHubClientProtocol protocol, ClientOptions clientOptions, long sendPeriodMillis, long receivePeriodMillis)
+    {
+        if (hostName == null)
+        {
+            throw new IllegalArgumentException("Host name cannot be null");
+        }
+
+        if (protocol == null)
+        {
+            throw new IllegalArgumentException("Protocol cannot be null.");
+        }
+
+        this.config = new DeviceClientConfig(hostName, sasTokenProvider, clientOptions, deviceId, moduleId);
+        this.config.setProtocol(protocol);
+        if (clientOptions != null)
+        {
+            this.config.modelId = clientOptions.getModelId();
+        }
+
+        this.deviceIO = new DeviceIO(this.config, sendPeriodMillis, receivePeriodMillis);
+    }
+
+
     //unused
     InternalClient()
     {
@@ -415,7 +438,7 @@ public class InternalClient
      *
      *	    - <b>SetReceiveInterval</b> - this option is applicable to all protocols
      *	      in case of HTTPS protocol, this option acts the same as {@code SetMinimumPollingInterval}
-     *	      in case of MQTT and AMQP protocols, this option specifies the interval in millisecods
+     *	      in case of MQTT and AMQP protocols, this option specifies the interval in milliseconds
      *	      between spawning a thread that dequeues a message from the SDK's queue of received messages.
      *
      *	    - <b>SetCertificatePath</b> - this option is applicable only
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
@@ -150,7 +150,6 @@ public class InternalClient
         this.deviceIO = new DeviceIO(this.config, sendPeriodMillis, receivePeriodMillis);
     }
 
-
     //unused
     InternalClient()
     {
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
@@ -161,7 +161,7 @@ public class InternalClient
 
     public void open() throws IOException
     {
-        if (this.config.getAuthenticationType() == DeviceClientConfig.AuthType.SAS_TOKEN && this.config.getSasTokenAuthentication().isRenewalNecessary())
+        if (this.config.getAuthenticationType() == DeviceClientConfig.AuthType.SAS_TOKEN && this.config.getSasTokenAuthentication().isAuthenticationProviderRenewalNecessary())
         {
             //Codes_SRS_INTERNALCLIENT_34_044: [If the SAS token has expired before this call, throw a Security Exception]
             throw new SecurityException("Your SasToken is expired");
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/SasTokenProvider.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/SasTokenProvider.java
@@ -1,0 +1,34 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.azure.sdk.iot.device;
+
+/**
+ * Interface for allowing users to control SAS token generation. To see an example of how SAS tokens can be generated
+ * from device connection strings, see {@link com.microsoft.azure.sdk.iot.device.auth.IotHubSasToken}.
+ * @see <a href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#example">This document</a> for more details on sas tokens.
+ */
+public interface SasTokenProvider
+{
+    /**
+     * Returns a SAS token to be used for authentication.
+     * <p>
+     * This function will be called each time the client library needs a SAS token. This will follow different patterns
+     * based on which protocol your client object is using.
+     * <p>
+     * For HTTPS, this function will be called for each HTTPS request made (for instance, once per telemetry send), and does not need
+     * to be a unique token each time. Because of that, users may simply cache and return the same SAS token repeatedly here.
+     * Alternatively, users can generate very short-lived SAS tokens since they will only need to be valid for a relative short period of time.
+     * The user of this API is responsible for tracking when to renew the SAS token based on how long the previous token was valid for.
+     * <p>
+     * For AMQPS/AMQPS_WS, this function will be called once when first opening the connection, and then will be called again
+     * at some point prior to the previous SAS token's expiry time in order to proactively renew the connection's authentication.
+     * This proactive renewal takes place at around 85% of the previous SAS token's lifespan.
+     * <p>
+     * For MQTT/MQTT_WS, this function will be called once when first opening the connection, and again each time the previous
+     * SAS token has expired and the client closes and re-opens the connection. MQTT/MQTT_WS does not currently support
+     * proactive token renewal.
+     * @return a SAS token to be used for authentication.
+     */
+    public char[] getSasToken();
+}
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
@@ -76,10 +76,13 @@ public abstract class IotHubSasTokenAuthenticationProvider extends IotHubAuthent
     }
 
     /**
-     * Returns true if the saved sas token needs to be manually renewed by the user
-     * @return true if the saved sas token needs to be manually renewed by the user
+     * Returns true if the this authentication provider is no longer valid. If true, users will need to create a new
+     * DeviceClient instance to get a new authentication provider. The most common case for this is if the user
+     * provides a SAS token, but no symmetric key, and that SAS token has expired. At that point, the user's client
+     * won't be able to authenticate anymore.
+     * @return true if the this authentication provider is no longer valid. False otherwise
      */
-    public boolean isRenewalNecessary()
+    public boolean isAuthenticationProviderRenewalNecessary()
     {
         //Codes_SRS_IOTHUBSASTOKENAUTHENTICATION_34_017: [If the saved sas token has expired, this function shall return true.]
         return (this.sasToken != null && this.sasToken.isExpired());
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
@@ -34,7 +34,7 @@ public abstract class IotHubSasTokenAuthenticationProvider extends IotHubAuthent
     protected IotHubSasToken sasToken;
 
     public abstract boolean canRefreshToken();
-    public abstract String getRenewedSasToken(boolean proactivelyRenew, boolean forceRenewal) throws IOException, TransportException;
+    public abstract char[] getSasToken() throws IOException, TransportException;
 
     public IotHubSasTokenAuthenticationProvider(String hostname, String gatewayHostname, String deviceId, String moduleId)
     {
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenHardwareAuthenticationProvider.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenHardwareAuthenticationProvider.java
@@ -120,7 +120,7 @@ public class IotHubSasTokenHardwareAuthenticationProvider extends IotHubSasToken
      * @return always returns false as the hardware authentication mechanism will never need to be updated with a new key or token
      */
     @Override
-    public boolean isRenewalNecessary()
+    public boolean isAuthenticationProviderRenewalNecessary()
     {
         //Hardware will always be able to generate new sas tokens
         //Codes_SRS_IOTHUBSASTOKENHARDWAREAUTHENTICATION_34_012: [This function shall return false.]
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenHardwareAuthenticationProvider.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenHardwareAuthenticationProvider.java
@@ -65,25 +65,14 @@ public class IotHubSasTokenHardwareAuthenticationProvider extends IotHubSasToken
     /**
      * Getter for SasToken. If the saved token has expired, this method shall renew it if possible
      *
-     * @param proactivelyRenew if true, this method will generate a fresh sas token even if the previously saved token
-     *                                 has not expired yet as long as the current token has lived beyond its buffer.
-     *                                 Use this for pre-emptively renewing sas tokens.
-
      * @throws IOException if generating the sas token from the TPM fails
      * @return The value of SasToken
      */
-    public String getRenewedSasToken(boolean proactivelyRenew, boolean forceRenewal) throws IOException
+    public char[] getSasToken() throws IOException
     {
-        if (this.shouldRefreshToken(proactivelyRenew) || forceRenewal)
-        {
-            //Codes_SRS_IOTHUBSASTOKENHARDWAREAUTHENTICATION_34_035: [If the saved sas token has expired and there is a security provider, the saved sas token shall be refreshed with a new token from the security provider.]
-            //Codes_SRS_IOTHUBSASTOKENHARDWAREAUTHENTICATION_34_036: [If the saved sas token has not expired and there is a security provider, but the sas token should be proactively renewed, the saved sas token shall be refreshed with a new token from the security provider.]
-            String sasTokenString = this.generateSasTokenSignatureFromSecurityProvider(this.tokenValidSecs);
-            this.sasToken = new IotHubSasToken(this.hostname, this.deviceId, null, sasTokenString, this.moduleId, 0);
-        }
-
-        //Codes_SRS_IOTHUBSASTOKENHARDWAREAUTHENTICATION_34_005: [This function shall return the saved sas token.]
-        return this.sasToken.toString();
+        String sasTokenString = this.generateSasTokenSignatureFromSecurityProvider(this.tokenValidSecs);
+        this.sasToken = new IotHubSasToken(this.hostname, this.deviceId, null, sasTokenString, this.moduleId, 0);
+        return this.sasToken.toString().toCharArray();
     }
 
     @Override
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
@@ -66,7 +66,7 @@ public class IotHubSasTokenProvidedAuthenticationProvider extends IotHubSasToken
 
         // Assuming that the token's life "starts" now for the sake of figuring out when it needs to be renewed. Users
         // could theoretically give us a SAS token that started a while ago, but since we have no way of figuring that out,
-        // we will conservatively just renew at 85% of the remaining time on the token, rather than 85% of the time the token
+        // we will conservatively just renew at timeBufferPercentage% of the remaining time on the token, rather than timeBufferPercentage% of the time the token
         // has existed for.
         long tokenValidSeconds = expiryTimeSeconds - (System.currentTimeMillis() / 1000);
 
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
@@ -48,6 +48,7 @@ public class IotHubSasTokenProvidedAuthenticationProvider extends IotHubSasToken
     @Override
     public boolean canRefreshToken()
     {
+        // User is always capable of providing a new SAS token when using this authentication provider.
         return true;
     }
 
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
@@ -5,10 +5,16 @@
 
 package com.microsoft.azure.sdk.iot.device.auth;
 
+import com.microsoft.azure.sdk.iot.device.ClientOptions;
+import com.microsoft.azure.sdk.iot.device.IotHubClientProtocol;
 import com.microsoft.azure.sdk.iot.device.SasTokenProvider;
 
 import javax.net.ssl.SSLContext;
 
+/**
+ * {@link IotHubSasTokenAuthenticationProvider} implementation where the tokens are provided by an instance of {@link SasTokenProvider}.
+ * This is used in cases like when the user creates a device client with {@link com.microsoft.azure.sdk.iot.device.DeviceClient#DeviceClient(String, String, SasTokenProvider, IotHubClientProtocol, ClientOptions)}
+ */
 public class IotHubSasTokenProvidedAuthenticationProvider extends IotHubSasTokenAuthenticationProvider
 {
     SasTokenProvider sasTokenProvider;
@@ -19,7 +25,7 @@ public class IotHubSasTokenProvidedAuthenticationProvider extends IotHubSasToken
 
         if (sasTokenProvider == null)
         {
-            throw new IllegalArgumentException("sas token provider cannot be null");
+            throw new IllegalArgumentException("SAS token provider cannot be null");
         }
 
         this.sasTokenProvider = sasTokenProvider;
@@ -34,7 +40,7 @@ public class IotHubSasTokenProvidedAuthenticationProvider extends IotHubSasToken
     @Override
     public void setTokenValidSecs(long tokenValidSecs)
     {
-        throw new UnsupportedOperationException("Cannot configure sas token time to live when custom sas token provider is in use");
+        throw new UnsupportedOperationException("Cannot configure SAS token time to live when custom sas token provider is in use");
     }
 
     @Override
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
@@ -32,7 +32,7 @@ public class IotHubSasTokenProvidedAuthenticationProvider extends IotHubSasToken
     }
 
     @Override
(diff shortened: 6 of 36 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/SasTokenProvider.java
```diff
@@ -0,0 +1,34 @@
...
+     * <p>
+     * This function will be called each time the client library needs a SAS token. This will follow different patterns
+     * based on which protocol your client object is using.
+     * <p>
+     * For HTTPS, this function will be called for each HTTPS request made (for instance, once per telemetry send), and does not need
+     * to be a unique token each time. Because of that, users may simply cache and return the same SAS token repeatedly here.
+     * Alternatively, users can generate very short-lived SAS tokens since they will only need to be valid for a relative short period of time.
+     * The user of this API is responsible for tracking when to renew the SAS token based on how long the previous token was valid for.
+     * <p>
+     * For AMQPS/AMQPS_WS, this function will be called once when first opening the connection, and then will be called again
```
barustum, line 24: Is AMQPS a thing? or should this be AMQP?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
```diff
@@ -144,6 +144,22 @@ public DeviceClientConfig(IotHubAuthenticationProvider authenticationProvider) t
...
 
+
+    public DeviceClientConfig(String hostName, SasTokenProvider sasTokenProvider, ClientOptions clientOptions, String deviceId, String moduleId)
+    {
+        this.authenticationProvider = new IotHubSasTokenProvidedAuthenticationProvider(
+                hostName,
+                deviceId,
+                moduleId,
+                sasTokenProvider,
+                clientOptions != null ? clientOptions.sslContext : null);
```
drwill-ms, line 155: How come no tricks? `clientOptions?.sslContext

:(

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenHardwareAuthenticationProvider.java
```diff
@@ -65,25 +65,14 @@ public IotHubSasTokenHardwareAuthenticationProvider(String hostname, String gate
...
      * Getter for SasToken. If the saved token has expired, this method shall renew it if possible
      *
-     * @param proactivelyRenew if true, this method will generate a fresh sas token even if the previously saved token
-     *                                 has not expired yet as long as the current token has lived beyond its buffer.
-     *                                 Use this for pre-emptively renewing sas tokens.
-
      * @throws IOException if generating the sas token from the TPM fails
      * @return The value of SasToken
      */
-    public String getRenewedSasToken(boolean proactivelyRenew, boolean forceRenewal) throws IOException
```
drwill-ms, line 75: Is this not customer public (e.g. breaking change)?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
```diff
@@ -0,0 +1,68 @@
...
+{
+    SasTokenProvider sasTokenProvider;
+    char[] lastSasToken;
+
+    public IotHubSasTokenProvidedAuthenticationProvider(String hostName, String deviceId, String moduleId, SasTokenProvider sasTokenProvider, SSLContext sslContext) {
+        super(hostName, null, deviceId, moduleId, sslContext);
+
+        if (sasTokenProvider == null)
+        {
+            throw new IllegalArgumentException("sas token provider cannot be null");
```
drwill-ms, line 22: `SAS` as it is an acronym

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        this.sasTokenProvider = sasTokenProvider;
+    }
+
+    @Override
+    public boolean isRenewalNecessary()
+    {
+        return false;
```
drwill-ms, line 31: Why is this `false`? If it can be refreshed, wouldn't renewal be necessary?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
```diff
@@ -0,0 +1,68 @@
...
+*  Licensed under the MIT license. See LICENSE file in the project root for full license information.
+*/
+
+package com.microsoft.azure.sdk.iot.device.auth;
+
+import com.microsoft.azure.sdk.iot.device.SasTokenProvider;
+
+import javax.net.ssl.SSLContext;
+
+public class IotHubSasTokenProvidedAuthenticationProvider extends IotHubSasTokenAuthenticationProvider
```
drwill-ms, line 12: Meaningful class comment

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Override
+    public int getMillisecondsBeforeProactiveRenewal()
+    {
+        // Seconds since UNIX epoch when this sas token will expire
+        long expiryTimeSeconds = IotHubSasToken.getExpiryTimeFromToken(new String(lastSasToken));
+
+        // Assuming that the token's life "starts" now for the sake of figuring out when it needs to be renewed. Users
+        // could theoretically give us a SAS token that started a while ago, but since we have no way of figuring that out,
+        // we will conservatively just renew at 85% of the remaining time on the token, rather than 85% of the time the token
```
drwill-ms, line 61: I don't see any code referring to 85%

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/SasTokenProvider.java
timtay-microsoft: "AMQP" is to "AMQPS" what "HTTP" is to "HTTPS". Typically we say AMQPS because we do SSL handshakes when opening AMQP connections to secure them

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
timtay-microsoft: Because Java :(

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenHardwareAuthenticationProvider.java
timtay-microsoft: This is not customer public, no

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
timtay-microsoft: This function is more about "Does the customer need to create a new client because they created this client with a SAS token rather than a symmetric key, and that SAS token has expired"

timtay-microsoft: Since the user is providing all the SAS tokens, this function should never return true.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
timtay-microsoft: That sits in a class that this class extends

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
drwill-ms: Or maybe `canRefreshToken()` should be false then?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/SasTokenProvider.java
drwill-ms: This doesn't apply to MQTT then?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
drwill-ms: Seems all the talk about 85% then is irrelevant there. That responsibility is delegated to the base class. If it changes in the base class, this comment would be wrong, so best to just abstract it and say, the base class schedules early renewal via `this.timeBufferPercentage`, and we'll adhere to that in this impl.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
drwill-ms: Oh, this isn't the impl for when a SAS-token provided? I'm confused.

drwill-ms (changes requested):
Can we get a sample of how a customer should use it?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
timtay-microsoft: Let me just rename this method to help with the confusion.

timtay-microsoft:
@drwill-ms on the topic of adding a ```force``` flag to the ```getSasToken()``` callback to notify the user that they need to refresh their token:

From my investigation so far, the SDK itself can't reliably say that a given SAS token authenticated request/connection got a 401 because it was expired, so I'm hesitant to add that flag. In cases where the symmetric keys of the device get cycled, our SDK could typically figure out if the SAS token was not expired if it still had a lot of time left to live, but there would always be room for error on that since we are comparing the client's clock to the server's clock. Because of that, I believe there would be cases where we'd call ```getSasToken(forceRenewal=false)``` and the user would just give the cached token again.

Since the SDK really doesn't have enough information in the 401 error message, I'd rather we just have the same callback for getting all SAS tokens and let the user be responsible for choosing when to renew vs when to cache.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/SasTokenProvider.java
timtay-microsoft: You'd think it would, but I don't think "MQTTS" is a thing, terminology-wise. Our MQTT connections always do an SSL handshake though, just like HTTPS and AMQPS/AMQPS_WS.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
timtay-microsoft: Good idea

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
```diff
@@ -0,0 +1,76 @@
...
+package com.microsoft.azure.sdk.iot.device.auth;
+
+import com.microsoft.azure.sdk.iot.device.ClientOptions;
+import com.microsoft.azure.sdk.iot.device.IotHubClientProtocol;
+import com.microsoft.azure.sdk.iot.device.SasTokenProvider;
+
+import javax.net.ssl.SSLContext;
+
+/**
+ * {@link IotHubSasTokenAuthenticationProvider} implementation where the tokens are provided by an instance of {@link SasTokenProvider}.
```
drwill-ms, line 15: `a custom instance of`?

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
```diff
@@ -0,0 +1,370 @@
...
+    private static final int D2C_MESSAGE_TIMEOUT = 2000; // 2 seconds
+    private static List<String> failedMessageListOnClose = new ArrayList(); // List of messages that failed on close
+
+    /**
+     * Helper class for turning symmetric keys into SAS tokens. It also provides some helpful functions around
+     * if this token should be renewed.
+     */
+    protected static class SasToken
+    {
+        private static final String RAW_SIGNATURE_FORMAT = "%s\n%s";
```
drwill-ms, line 38: I wonder if these should be constants in the SDK so someone can just refer to them instead of having to know or refer to your sample.

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
```diff
@@ -0,0 +1,370 @@
...
+                // Base64 encode the HMAC encrypted byte[]
+                byte[] base64EncodedHmacEncryptedSignature = Base64.encodeBase64(hmacEncryptedSignature);
+
+                // Convert byte[] of base64 encoded and HMAC encrypted bits to a UTF-8 String
+                String utf8Sig = new String(base64EncodedHmacEncryptedSignature, SIGNATURE_CHARSET);
+
+                // URL encode the string
+                String urlEncodedSignature = URLEncoder.encode(utf8Sig, SIGNATURE_CHARSET.name());
+
+                this.sasToken = String.format(SHARED_ACCESS_SIGNATURE_FORMAT,
```
drwill-ms, line 85: should first params be on the next line?

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
```diff
@@ -0,0 +1,370 @@
...
+                        ResourceURIFieldKey,
```
drwill-ms, line 86: These are tabbed in twice

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
```diff
@@ -0,0 +1,370 @@
...
+            Mac hMacSha256 = Mac.getInstance(hmacSha256);
+            hMacSha256.init(secretKey);
+            encryptedSig = hMacSha256.doFinal(signature);
+
+            return encryptedSig;
+        }
+
+        private String buildScope(String hostName, String deviceId)
+        {
+            return String.format("%s/devices/%s", hostName, deviceId);
```
drwill-ms, line 137: Why is this format not with the class constants?

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
```diff
@@ -0,0 +1,370 @@
...
+
+    protected static class EventCallback implements IotHubEventCallback
+    {
+        public void execute(IotHubStatusCode status, Object context)
+        {
+            Message msg = (Message) context;
+
+            System.out.println("IoT Hub responded to message "+ msg.getMessageId()  + " with status " + status.name());
+
+            if (status==IotHubStatusCode.MESSAGE_CANCELLED_ONCLOSE)
```
drwill-ms, line 198: Spaces around `==`

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -138,7 +161,7 @@
 
     public void open() throws IOException
     {
-        if (this.config.getAuthenticationType() == DeviceClientConfig.AuthType.SAS_TOKEN && this.config.getSasTokenAuthentication().isRenewalNecessary())
+        if (this.config.getAuthenticationType() == DeviceClientConfig.AuthType.SAS_TOKEN && this.config.getSasTokenAuthentication().isAuthenticationProviderRenewalNecessary())
```
abhipsaMisra, line 164: nit: I am a bit confused by the naming of this - we aren't renewing the authentication provider itself, right?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/SasTokenProvider.java
```diff
@@ -0,0 +1,34 @@
...
+     * at some point prior to the previous SAS token's expiry time in order to proactively renew the connection's authentication.
+     * This proactive renewal takes place at around 85% of the previous SAS token's lifespan.
```
abhipsaMisra, line 26: If I understand coreectly, for a user implementing this interface, they would need to implement their own renewal logic, right? This statement is only for our library's implementation of `SasTokenProvider`, which incidentally contains the 85% lifetime renewal logic?

Correct me if I am wrong, this is what I understand - 
From an interface perspective, our library calls `getSasToken()` whenever it requires a new SAS token (at it's simplest implementation - every time for Http, on CONNECT for Mqtt, while opening cbs link for Amqp).
For a user of this interface, you can then add any caching or renewal logic as you require in your implementation of `getSasToken()`, while the sdk can add additional calls to `getSasToken()` on reconnection, with an amqp refresh timer, etc.

If this is correct, then we should instead state that any renewal or caching logic is the ownership of whoever implements `SasTokenProvider` interface.

abhipsaMisra: Wait, I see a lot of authentication provider classes in our library; would you mind listing their hierarchy? That will help understand which method (in the auth provider interface) our transport layer calls into.

abhipsaMisra: If `SasTokenProvider` is not what our transport layer calls into, then my assumption above might not be correct.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
```diff
@@ -76,10 +76,13 @@ public void setTokenValidSecs(long tokenValidSecs)
     }
 
     /**
-     * Returns true if the saved sas token needs to be manually renewed by the user
-     * @return true if the saved sas token needs to be manually renewed by the user
+     * Returns true if the this authentication provider is no longer valid. If true, users will need to create a new
+     * DeviceClient instance to get a new authentication provider. The most common case for this is if the user
+     * provides a SAS token, but no symmetric key, and that SAS token has expired. At that point, the user's client
+     * won't be able to authenticate anymore.
+     * @return true if the this authentication provider is no longer valid. False otherwise
```
abhipsaMisra, line 83: ok, this answers a question that I asked above - why do we want to expose something like this? Wouldn't it be easier to simply state that if you create an authentication provider using a shared access signature, then you cannot renew tokens and your client will be unusable once your token expires.

abhipsaMisra:
I'll sync up with you offline Tim, I think I am a bit unclear on how all these token providers relate to each other (since there seem to be some providers between what the transport layer calls into, and what the user implements).

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
timtay-microsoft: We kind of are. ```isAuthenticationProviderRenewalNecessary``` is only true when the user provides a single SAS token, and it expires, and we can't refresh it because we don't have the device key

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
timtay-microsoft: It's tempting, but we don't have a great place to put it. None of our "public" classes are directly tied to our connection string/sas token logic

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/SasTokenProvider.java
timtay-microsoft: I'd need a whiteboard to illustrate these authentication providers and their hierarchies. Just focusing on the SAS side of the tree, IotHubSasTokenAuthenticationProvider is the "root" class for all symmetric key based authentication mechanisms. From there, there are several implementations depending on if the device key is present, or if the TPM is the provider, or if the new SAS token provider interface is provided.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
timtay-microsoft: We don't expose this. These AuthenticationProvider classes are strictly internal. This particular function is used only in a few special cases where we throw an exception. I'd love to get rid of it, but that would be a breaking change.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
azabbasi: So are they really renewing the authentication provider or just a SAS token? the name is really confusing to me.

azabbasi: so does the SasTokenProvider control the auth keys to generate the SAS token?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/SasTokenProvider.java
azabbasi: I once attempted to draw the hierarchy by just looking at the code and I failed on the first few attempts ... it's a confusing logic.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
```diff
@@ -34,7 +34,7 @@
     protected IotHubSasToken sasToken;
 
     public abstract boolean canRefreshToken();
-    public abstract String getRenewedSasToken(boolean proactivelyRenew, boolean forceRenewal) throws IOException, TransportException;
+    public abstract char[] getSasToken() throws IOException, TransportException;
```
azabbasi, line 37: This class is a public class yes? can we just change the method ?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
azabbasi: I am  confused here again, can't they hold the logic to provide a new SAS token in their implementation of the SasTokenProvider?

azabbasi: that way they would still be able to keep using the DeviceClient.

azabbasi: it's just not a great experience if I have to keep creating a device client every time my SAS expires.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenHardwareAuthenticationProvider.java
```diff
@@ -65,25 +65,14 @@ public IotHubSasTokenHardwareAuthenticationProvider(String hostname, String gate
     /**
      * Getter for SasToken. If the saved token has expired, this method shall renew it if possible
      *
-     * @param proactivelyRenew if true, this method will generate a fresh sas token even if the previously saved token
```
azabbasi, line 68: so if we are repurposing this method entirely, where does this logic go? the proactive renewal approach that is.

azabbasi: did we never use this?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/SasTokenProvider.java
timtay-microsoft: > This statement is only for our library's implementation of SasTokenProvider, which incidentally contains the 85% lifetime renewal logic?

Our SDK will still ask for a new SAS token once the previous token has 15% of it's life left, but the SDK doesn't mandate anything about that token. It's not recommended, but customers could just return the same SAS token at that point, for instance.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
timtay-microsoft: No, these AuthenticationProvider class are not part of our public API surface. You can't construct a device client with one of these things

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
```diff
@@ -0,0 +1,371 @@
...
+public class CustomSasTokenProviderSample
+{
+    private static final int D2C_MESSAGE_TIMEOUT = 2000; // 2 seconds
+    private static List<String> failedMessageListOnClose = new ArrayList(); // List of messages that failed on close
+
+    /**
+     * Helper class for turning symmetric keys into SAS tokens. It also provides some helpful functions around
+     * if this token should be renewed.
+     */
+    protected static class SasToken
```
azabbasi, line 36: I think it's important to make the distinction that this is a helper class by including that in the name. 
`SampleSasTokenHelper `? or something of that nature.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/SasTokenProviderImpl.java
```diff
@@ -0,0 +1,45 @@
...
+import com.microsoft.azure.sdk.iot.device.IotHubConnectionString;
+import com.microsoft.azure.sdk.iot.device.SasTokenProvider;
+import com.microsoft.azure.sdk.iot.device.auth.IotHubSasToken;
+
+import java.net.URISyntaxException;
+
+/**
+ * Basic implementation of the SasTokenProvider interface, for test purposes.
+ */
+public class SasTokenProviderImpl implements SasTokenProvider
```
azabbasi, line 15: I assume this implementation works out of the box if they provide the connection string. Is that correct?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenHardwareAuthenticationProvider.java
timtay-microsoft: We only ever called this method with the same flags set, (true and true), so we can just get rid of the other cases

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/SasTokenProviderImpl.java
timtay-microsoft: It does, but this lives in our test code. I'd prefer users pick up the sample code's implementation rather than this one.

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
timtay-microsoft: This class is functional enough to replace the similar classes we have defined in our SDK, so I kind of like the name as is.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenAuthenticationProvider.java
timtay-microsoft: > I am confused here again, can't they hold the logic to provide a new SAS token in their implementation of the SasTokenProvider?

They can. They do not need to create a new device client each time their SAS token expires. This particular method only raises issues if the user creates a device client instance through a old constructor that allows for a single SAS token to be provided. In the case that the user uses that old constructor, and their SAS token expires, then this method returns true, signaling to the Transport layer to fire a notification to the user.

For users of the new constructor that takes the SasTokenProvider interface, they will never need to create a new device client because they will always be capable of providing new SAS tokens.

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
drwill-ms: It could be a standalone constants class

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
azabbasi: then maybe we should include this as the default implementation of our own that ships along with the interface?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
timtay-microsoft: I'd just ignore this function, to be honest. It only is relevant in very niche cases. For users who use the new DeviceClient constructor in this PR, they will never enter into this if block.

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
drwill-ms: But it isn't a SasToken, so the name is misleading, don't you think?

timtay-microsoft: Sure then, I can rename this to ```SasTokenHelper```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
```diff
@@ -144,6 +144,22 @@ public DeviceClientConfig(IotHubAuthenticationProvider authenticationProvider) t
         this.productInfo = new ProductInfo();
     }
 
+
+    public DeviceClientConfig(String hostName, SasTokenProvider sasTokenProvider, ClientOptions clientOptions, String deviceId, String moduleId)
+    {
+        this.authenticationProvider = new IotHubSasTokenProvidedAuthenticationProvider(
+                hostName,
```
drwill-ms, line 151: these are tabbed in twice

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -128,6 +128,29 @@
...
+        this.config = new DeviceClientConfig(hostName, sasTokenProvider, clientOptions, deviceId, moduleId);
+        this.config.setProtocol(protocol);
+        if (clientOptions != null)
+        {
+            this.config.modelId = clientOptions.getModelId();
+        }
+
+        this.deviceIO = new DeviceIO(this.config, sendPeriodMillis, receivePeriodMillis);
+    }
+
```
drwill-ms, line 152: there are 2 newlines here, when I think 1 is expected

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenProvidedAuthenticationProvider.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public void setTokenValidSecs(long tokenValidSecs)
+    {
+        throw new UnsupportedOperationException("Cannot configure SAS token time to live when custom SAS token provider is in use");
+    }
+
+    @Override
+    public boolean canRefreshToken()
```
drwill-ms, line 49: As this is overridden and specified for this impl, I think it warrants a comment as to why it is `true`.

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
```diff
@@ -0,0 +1,371 @@
...
+
+                // Convert byte[] of base64 encoded and HMAC encrypted bits to a UTF-8 String
+                String utf8Sig = new String(base64EncodedHmacEncryptedSignature, SIGNATURE_CHARSET);
+
+                // URL encode the string
+                String urlEncodedSignature = URLEncoder.encode(utf8Sig, SIGNATURE_CHARSET.name());
+
+                this.sasToken = String.format(
+                    SHARED_ACCESS_SIGNATURE_FORMAT,
+                        RESOURCE_URI_FIELD_KEY,
```
drwill-ms, line 89: What's with the weird tabbing going on here where some params are further in than others?

drwill-ms (approved):

device/iot-device-samples/custom-sas-token-provider-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/CustomSasTokenProviderSample.java
timtay-microsoft: I'll fix it

PR #1004, opened 2020-11-20
fix(documentation): Adding conditional docs generation

vinagesh:
# Checklist

# Description of the problem
As we have created a new preview branch, all packages have the same version as master and not the latest preview version. This leads to generation of incorrect docs as we have only released a new device-client preview. To ensure that the gh-pages are only updated for the corresponding package, I have added a conditional generation which can be set in the pipeline. This way, we can just generate the required documentation and keep the others ones as is.

This change will go to preview when I merge master to preview.

Pipeline now has the following options. They will be used once this PR is checked in.

image

### DIFF
--- a/vsts/generate_javadocs_branch.ps1
+++ b/vsts/generate_javadocs_branch.ps1
@@ -15,7 +15,16 @@ function TestLastExitCode {
     }
 }
 
-function CreateJavadocReleaseBranch($GitHubName, $GitHubEmail, $Sources, $FolderName) {
+function CreateJavadocReleaseBranch(
+    $GitHubName,
+    $GitHubEmail,
+    $Sources,
+    $FolderName,
+    $UpdateServiceClientDocs,
+    $UpdateDeviceClientDocs,
+    $UpdateProvisioningClientDocs,
+    $UpdateDepsDocs) {
+
     if ($([string]::IsNullOrWhiteSpace($GitHubName) -eq $true)) {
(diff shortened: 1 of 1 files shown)


vsts/generate_javadocs_branch.ps1
```diff
@@ -104,10 +114,26 @@ function CreateJavadocReleaseBranch($GitHubName, $GitHubEmail, $Sources, $Folder
...
+    {
+        Move-Item -Force -Path ..\device -Destination ..\$FolderName\device
+    }
+
+    if($UpdateServiceClientDocs)
+    {
+        Move-Item -Force -Path ..\service -Destination ..\$FolderName\service
+    }
+
+    if($UpdateProvisioningClientDocs)
```
timtay-microsoft, line 133: There are several clients within the provisioning folder, so this could be more specific. We have provisioning device client, provisioning service client, security provider, etc.

timtay-microsoft (approved):

vsts/generate_javadocs_branch.ps1
timtay-microsoft: Probably not a dealbreaker though

vinagesh: Makes sense. I'm hoping we won't release those in preview :P There will be too many variables to deal with so I did not add them all. If we really need it, we can add it. Let me know if that sounds reasonable.

PR #1006, opened 2020-11-20
fix(script): Fix gh-pages generation script

vinagesh:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/vsts/generate_javadocs_branch.ps1
+++ b/vsts/generate_javadocs_branch.ps1
@@ -115,22 +115,22 @@ function CreateJavadocReleaseBranch(
     # Move the generated content to the correct folder. The folder will be different for master and preview branches.
     New-Item -Path ../$FolderName -ItemType Directory
 
-    if($UpdateDepsDocs)
+    if ($UpdateDepsDocs -eq "True")
     {
         Move-Item -Force -Path ..\deps -Destination ..\$FolderName\deps
     }
 
-    if($UpdateDeviceClientDocs)
+    if ($UpdateDeviceClientDocs -eq "True")
     {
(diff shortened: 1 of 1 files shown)


vsts/generate_javadocs_branch.ps1
```diff
@@ -115,22 +115,22 @@ function CreateJavadocReleaseBranch(
     # Move the generated content to the correct folder. The folder will be different for master and preview branches.
     New-Item -Path ../$FolderName -ItemType Directory
 
-    if($UpdateDepsDocs)
+    if ($UpdateDepsDocs -eq "True")
```
vinagesh, line 118: This seems to be input from the pipeline as a string and not bool

timtay-microsoft (approved):

vsts/generate_javadocs_branch.ps1
barustum: nit and I am ok if we don't to it: maybe do a toLowerString (or whatever the PowerShell equivalence) in case the backend break us by sending us true instead of True.

vinagesh: This is sent from the pipeline and I think it will be nice to have if we enter it incorrectly. Will add it. My only laziness is about the wait for pipeline to pass but that is not a good excuse :)

vinagesh: Btw I checked and -eq is case insensitive so we are good. you need to do -ceq for a case sensitive comparision on powershell.

azabbasi: I think we can even check for both cases with -or ... so if the pipeline ever fixed their issue, we won't break.

PR #1010, opened 2020-12-01
feat(iot-dev): Fix multiplexing client not reporting unauthorized exceptions

timtay-microsoft:
Previously, if a device client registered to the multiplexing client had an out of date connection string, the register call would block infinitely. Now the method will throw an exception instead.

This also adds an optional timeout to the registration/unregistration APIs to allow users to have shorter or longer timeouts than our new default of 1 minute.

I've added new Exception types to help give users all the information about which devices failed to authenticate. One new exception comes with a map of deviceId->authentication exception so that users know which keys need refreshing. 

As a part of this work, I made it so that partially successful opens/registrations are possible. If you register a mix of device clients with and without up to date credentials, the open call will throw, but the connection will open for all the devices that had the proper credentials. Similarly, when registering bulk devices onto an active connection, the call will throw if any have incorrect credentials, but the other clients will still be registered successfully to the connection. This was done so that users don't have to question which devices need to be re-registered.

Note that this also changes the thrown exceptions of most of the multiplexing client APIs. This is a breaking change, but is fine since this feature is in preview.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -164,14 +164,36 @@ void open() throws IOException
...
-    void registerMultiplexedDeviceClient(List<DeviceClientConfig> configs) throws InterruptedException
+    // Functionally the same as "open()", but without wrapping any thrown TransportException into an IOException
+    void multiplexingClientOpen() throws TransportException
     {
-        this.transport.registerMultiplexedDeviceClient(configs);
+        try
+        {
+            open();
+        }
+        catch (IOException e)
```
timtay-microsoft, line 174: We did this silly thing in the DeviceClient to work around the fact that we can't throw TransportException's directly in methods like ```deviceClient.open()``` becuase the open API existed before the TransportException did. To get around it, we just nested the meaningful exception into an IOException. I don't want to do the same thing for the multiplexing client though, so this code un-nests the exception when possible.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/MultiplexingClientDeviceRegistrationFailedException.java
```diff
@@ -0,0 +1,85 @@
...
+    public MultiplexingClientDeviceRegistrationFailedException()
+    {
+        super();
+    }
+
+    /**
+     * Construct a new MultiplexingClientDeviceRegistrationFailedException with no nested exception but with an error message.
+     * @param message The top level message for this exception.
+     */
+    public MultiplexingClientDeviceRegistrationFailedException(String message)
```
timtay-microsoft, line 29: I should maybe rename this to ```MultiplexingClientDeviceRegistrationAuthenticationFailedException``` to be more specific, since it is only ever thrown when a device's authentication fails

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -164,14 +164,36 @@ void open() throws IOException
         }
     }
 
-    void registerMultiplexedDeviceClient(List<DeviceClientConfig> configs) throws InterruptedException
+    // Functionally the same as "open()", but without wrapping any thrown TransportException into an IOException
+    void multiplexingClientOpen() throws TransportException
```
drwill-ms, line 168: Would feel more natural to me if the verb were first: `openMultiplexingClient`. Thoughts?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
drwill-ms: Add that as a code comment, then please

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -251,18 +273,25 @@ public void close() throws IOException
...
-     * Completes all current outstanding requests and closes the IoT Hub client.
-     * Must be called to terminate the background thread that is sending data to
-     * IoT Hub. After {@code close()} is called, the IoT Hub client is no longer
-     *  usable. If the client is already closed, the function shall do nothing.
-     *
-     * @throws IOException if the connection to an IoT Hub cannot be closed.
-     */
-    public void multiplexClose() throws IOException
+    // Functionally the same as "close()", but without wrapping any thrown TransportException into an IOException
+    public void multiplexClose() throws TransportException
```
drwill-ms, line 277: Same: `closeMultiplexingClient`

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -119,6 +124,8 @@ else if (receivePeriod == 0) //default builder value for this option, signals th
         this.deviceIO = new DeviceIO(hostName, protocol, sslContext, proxySettings, sendPeriod, receivePeriod);
     }
 
+    private static final String OPEN_ERROR_MESSAGE = "Failed to open the multiplexed connection";
```
drwill-ms, line 127: Naming consistency: multiplexing

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -719,19 +731,47 @@ public void registerMultiplexedDeviceClient(List<DeviceClientConfig> configs) th
...
+                while (deviceIsNotConnected && deviceRegistrationException == null)
                 {
                     Thread.sleep(100);
+
+                    deviceIsNotConnected = deviceConnectionStates.get(deviceId) != IotHubConnectionStatus.CONNECTED;
+                    deviceRegistrationException = multiplexingDeviceRegistrationFailures.remove(deviceId);
+                    boolean operationHasTimedOut = System.currentTimeMillis() >= timeoutTime;
+                    if (operationHasTimedOut)
+                    {
+                        throw new MultiplexingClientDeviceRegistrationFailedException("Timed out waiting for all device registrations to finish");
```
drwill-ms, line 752: puncuation?

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexingSample.java
```diff
@@ -102,16 +115,57 @@ public static void main(String[] args)
...
+        }
+        catch (MultiplexingClientDeviceRegistrationFailedException e)
+        {
+            // When registering device clients to an active multiplexed connection, one to all of the devices may fail
+            // to register if they have out-of-date or otherwise incorrect connection strings, for instance. The thrown exception
+            // here contains a map of deviceId -> registration failure so that you can tell which devices failed to register,
+            // and why each device failed to register.
+            System.out.println("Encountered an exception while registering devices to the active multiplexed connection: " + e.getMessage());
+            Map<String, Exception> registrationExceptions = e.getRegistrationExceptions();
+            for (String deviceId : registrationExceptions.keySet()) {
```
drwill-ms, line 148: inconsistent brace usage

drwill-ms (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -164,14 +164,36 @@ void open() throws IOException
...
+            open();
+        }
+        catch (IOException e)
+        {
+            Throwable cause = e.getCause();
+            if (cause != null && cause instanceof TransportException)
+            {
+                throw (TransportException) cause;
+            }
+            else
```
azabbasi, line 181: it seems like we don't need the else since the if statement throws.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -127,14 +134,44 @@ else if (receivePeriod == 0) //default builder value for this option, signals th
      * <p>
      * If this client is already open, then this method will do nothing.
      * <p>
-     * @throws IOException If any IO errors occur while opening the multiplexed connection.
+     * @throws MultiplexingClientException If any IO or authentication errors occur while opening the multiplexed connection.
+     * @throws MultiplexingClientDeviceRegistrationFailedException If one or many of the registered devices failed to authenticate.
```
azabbasi, line 137: `MultiplexingClientDeviceRegistrationFailedException` seems a bit long. I think we don't need the word `fail` in there.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: This doesn't open the multiplexing client though. It is the open method used exclusively by the multiplexing client

timtay-microsoft: I'll rename this to something like "openWithoutWrappingException" instead, so that this isn't necessarily limited to multiplexing clients though

drwill-ms: I don't know how someone would know that.

timtay-microsoft: And I'll do the same for the similar close method

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
azabbasi: that doesn't really make it any shorter, but still ...

azabbasi:
> As a part of this work, I made it so that partially successful opens/registrations are possible. If you register a mix of device clients with and without up to date credentials, the open call will throw, but the connection will open for all the devices that had the proper credentials. Similarly, when registering bulk devices onto an active connection, the call will throw if any have incorrect credentials, but the other clients will still be registered successfully to the connection. This was done so that users don't have to question which devices need to be re-registered.


This is interesting. I think this behavior is desirable, connect as many as you can and give me a list of `<deviceId, exception>` pairs for those that failed ... 
Do you think there is a scenario where someone would want an all or nothing sort of `open` functionality?

azabbasi:
> > As a part of this work, I made it so that partially successful opens/registrations are possible. If you register a mix of device clients with and without up to date credentials, the open call will throw, but the connection will open for all the devices that had the proper credentials. Similarly, when registering bulk devices onto an active connection, the call will throw if any have incorrect credentials, but the other clients will still be registered successfully to the connection. This was done so that users don't have to question which devices need to be re-registered.
> 
> This is interesting. I think this behavior is desirable, connect as many as you can and give me a list of `<deviceId, exception>` pairs for those that failed ...
> Do you think there is a scenario where someone would want an all or nothing sort of `open` functionality?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -127,14 +133,44 @@ else if (receivePeriod == 0) //default builder value for this option, signals th
...
```
azabbasi, line 137: could there be any other reason for the device registration to fail? is it only authentication failures that would cause a `MultiplexingClientDeviceRegistrationFailedException` to be thrwon?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -127,14 +133,44 @@ else if (receivePeriod == 0) //default builder value for this option, signals th
...
     {
         synchronized (this.operationLock)
         {
             log.info("Opening multiplexing client");
-            this.deviceIO.open();
+            try
+            {
+                this.deviceIO.openWithoutWrappingException();
+            }
+            catch (TransportException e)
```
azabbasi, line 150: correct me if I am wrong, but `TransportException` is not in the `MultiplexingClientException` hierarchy 
From what I can tell, `MultiplexingClientException` extends Exception and `MultiplexingClientDeviceRegistrationFailedException` extends `MultiplexingClientException` so line 157 would not be true ever ! no?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -200,13 +244,100 @@ public void close() throws IOException
...
+    public void registerDeviceClient(DeviceClient deviceClient) throws InterruptedException, MultiplexingClientException {
+        this.registerDeviceClient(deviceClient, DEFAULT_REGISTRATION_TIMEOUT_MILLISECONDS);
+    }
+
+    /**
+     * Register a device client to this multiplexing client. This method may be called before or after opening the
+     * multiplexed connection.
+     * <p>
+     * Users should use {@link #registerDeviceClients(Iterable)} for registering multiple devices as it has some
+     * performance improvements over repeatedly calling this method. This method blocks on each registration, whereas
```
azabbasi, line 262: `repeatedly calling this method for individual device registrations`

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
abhipsaMisra, line 137: Q - both of these exception types point towards authentication related errors, what is the difference in the scenarios where they are thrown?
MultiplexingClientException = while initial open vs MultiplexingClientDeviceRegistrationFailedException = registered but subsequently keys expired?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -200,13 +244,100 @@ public void close() throws IOException
...
+     * The registered device clients must belong to the same IoT Hub as all previously registered device clients.
+     * <p>
+     * If any of these device clients are already registered to this multiplexing client, then then this method will
+     * not do anything to that particular device client. All other provided device clients will still be registered though.
+     * <p>
+     * @throws InterruptedException If the thread gets interrupted while waiting for the registration to succeed. This
+     * will never be thrown if the multiplexing client is not open yet.
+     * @throws MultiplexingClientDeviceRegistrationFailedException If one or more devices failed to register. Details for each failure can be found
+     * in this exception. Any devices not found in the map of registration exceptions provided by this exception have registered successfully.
+     * @throws com.microsoft.azure.sdk.iot.device.exceptions.MultiplexingClientDeviceRegistrationTimeoutException If this operation takes longer than the default timeout allows.
```
azabbasi, line 334: does `com.microsoft.azure.sdk.iot.device.exceptions.MultiplexingClientDeviceRegistrationTimeoutException` need to be fully qualified?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -238,11 +369,22 @@ public void registerDeviceClient(DeviceClient deviceClient) throws InterruptedEx
...
+     * @param timeoutMilliseconds How long (in milliseconds) to let this operation wait for all registrations to complete. If this threshold is passed, a {@link com.microsoft.azure.sdk.iot.device.exceptions.MultiplexingClientDeviceRegistrationTimeoutException} is thrown.
      */
-    public void registerDeviceClients(Iterable<DeviceClient> deviceClients) throws InterruptedException {
+    public void registerDeviceClients(Iterable<DeviceClient> deviceClients, long timeoutMilliseconds) throws InterruptedException, MultiplexingClientException
+    {
         Objects.requireNonNull(deviceClients);
 
+        if (timeoutMilliseconds <= 0)
+        {
+            throw new IllegalArgumentException("Cannot set a device registration timeout of less than 0 milliseconds");
```
azabbasi, line 385: less or equal? can it be 0?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -339,13 +481,69 @@ public void registerDeviceClients(Iterable<DeviceClient> deviceClients) throws I
...
     {
         Objects.requireNonNull(deviceClient);
         List<DeviceClient> clientList = new ArrayList<>();
         clientList.add(deviceClient);
-        unregisterDeviceClients(clientList);
+        unregisterDeviceClients(clientList, timeoutMilliseconds);
+    }
+
+    /**
+     * Unregister device clients from this multiplexing client. This method may be called before or after opening the
```
azabbasi, line 526: nit: maybe add the word `multiple` at the beginning ...
`Unregister one or many device clients ...`

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/TransportClient.java
```diff
@@ -110,12 +114,16 @@ public void open() throws IllegalStateException, IOException
...
-                this.deviceIO.registerMultiplexedDeviceClient(configList);
+                this.deviceIO.registerMultiplexedDeviceClient(configList, DEFAULT_REGISTRATION_TIMEOUT_MILLISECONDS);
             }
             catch (InterruptedException e)
             {
                 throw new IOException("Interrupted while registering device clients to the multiplexed connection", e);
             }
+            catch (MultiplexingClientException e)
+            {
+                throw new IOException("Failed to register one or more device clients to the multiplexed connection", e);
```
azabbasi, line 125: can you elaborate why you change the exception type here? I am sure you have documented it somewhere in this PR but I am having a hard time following :D

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/MultiplexingClientDeviceRegistrationFailedException.java
azabbasi: I think that is a good change. you probably don't even need the word `failed` in there as Exception kind of gives that away

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/MultiplexingClientDeviceRegistrationFailedException.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public Map<String, Exception> getRegistrationExceptions() {
+        return registrationExceptions;
+    }
+
+    /**
+     * Sets the full registrations exception map. This will overwrite any previously saved mappings.
+     * @param registrationExceptions the new full registrations exception map.
+     */
+    public void setRegistrationExceptionsMap(Map<String, Exception> registrationExceptions)
```
azabbasi, line 81: does this need to be public? who would be able to set it ? I mean whoever it is, should probably use addRegistrationException

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MultiplexingSample.java
```diff
@@ -102,19 +115,60 @@ public static void main(String[] args)
...
         // If instead each client was registered separately through multiplexingClient.registerDeviceClient(), it would
         // take a longer time since it would block on each registration completing, rather than block on all registrations completing
         System.out.println("Registering " + multiplexedDeviceCount + " clients to the multiplexing client...");
-        multiplexClientManager.registerDeviceClients(multiplexedDeviceClients.values());
-
+        try
+        {
+            multiplexClientManager.registerDeviceClients(multiplexedDeviceClients.values());
+        }
+        catch (MultiplexingClientDeviceRegistrationFailedException e)
```
azabbasi, line 140: Consider delegating this logic to the `MultiplexClientManager`.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -127,14 +134,44 @@ else if (receivePeriod == 0) //default builder value for this option, signals th
...
+                this.deviceIO.multiplexingClientOpen();
+            }
+            catch (TransportException e)
+            {
+                // AMQP layer will throw a TransportException with a nested MultiplexingClientException in some cases. In these
+                // cases, the wrapping of the MultiplexingClientException is strictly to avoid adding more thrown exception types, and
+                // no extra context is added into the TransportException layer. Because of that, this layer will strip out the TransportException
+                // if the TransportException's cause was a MultiplexingClientException.
+                Throwable cause = e.getCause();
+                if (cause != null && cause instanceof MultiplexingClientDeviceRegistrationFailedException)
```
abhipsaMisra, line 157: MultiplexingClientDeviceRegistrationFailedException doesn't seem to be an instance of TransportException, it looks like it was inherited from Exception directly. Missed, maybe?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -127,14 +134,44 @@ else if (receivePeriod == 0) //default builder value for this option, signals th
...
+                {
+                    MultiplexingClientDeviceRegistrationFailedException newException =
+                            new MultiplexingClientDeviceRegistrationFailedException(OPEN_ERROR_MESSAGE, cause);
```
abhipsaMisra, line 160: we are creating a new MultiplexingClientDeviceRegistrationFailedException with an existing MultiplexingClientDeviceRegistrationFailedException? I don't think I get what is happening here, could you explain this bit?

timtay-microsoft:
> This is interesting. I think this behavior is desirable, connect as many as you can and give me a list of <deviceId, exception> pairs for those that failed ...
Do you think there is a scenario where someone would want an all or nothing sort of open functionality?

I don't see a strong use-case for an "all-or-nothing" type approach as it would be very expensive to open a 1000-device connection only to tear it all down if one device has an invalid key. The now deprecated TransportClient offers something like that though, if customers wanted it.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -238,11 +330,62 @@ public void registerDeviceClient(DeviceClient deviceClient) throws InterruptedEx
...
+     * @throws com.microsoft.azure.sdk.iot.device.exceptions.MultiplexingClientDeviceRegistrationTimeoutException If this operation takes longer than the provided timeout allows.
+     * @throws MultiplexingClientException If any other Exception is thrown, it will be nested into this exception.
+     * @param deviceClients The device clients to associate with this multiplexing client.
+     * @param timeoutMilliseconds How long (in milliseconds) to let this operation wait for all registrations to complete. If this threshold is passed, a {@link com.microsoft.azure.sdk.iot.device.exceptions.MultiplexingClientDeviceRegistrationTimeoutException} is thrown.
+     */
+    public void registerDeviceClients(Iterable<DeviceClient> deviceClients, long timeoutMilliseconds) throws InterruptedException, MultiplexingClientException
+    {
         Objects.requireNonNull(deviceClients);
 
+        if (timeoutMilliseconds <= 0)
```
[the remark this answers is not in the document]
timtay-microsoft: Java 7 and below don't support unsigned ints/longs unfortunately. I love the idea, though!

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: I think I will rename this particular exception to ```MultiplexingClientDeviceRegistrationAuthenticationException``` and if there are other cases when the SDK needs to throw, I can add a new Exception type to cover them. Right now, this exception is only thrown in authentication failure cases.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: Line 157 is checking the nested exception of this caught exception. I did add one special case where the lower layer throws a ```TransportException``` with a nested ```MultiplexingClientDeviceRegistrationAuthenticationException``` so that we could handle this code path. In particular, it is when you call open after registering your devices.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: ```MultiplexingClientException``` is the super type of ```MultiplexingClientDeviceRegistrationFailedException``` here. Normally I'd only call out that the method throws the super type, since that is all that Javadocs require, but it felt useful to call out ```MultiplexingClientDeviceRegistrationFailedException``` in particular as it can be handled differently from ```MultiplexingClientException```. The more specific exception has a map of deviceId->registration exception that users can read when reacting to such exceptions.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: Strangely, yes. I can't figure out why though

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: If we allowed 0, it would immediately throw a timeout exception, which doesn't seem useful. Using HTTP libraries as a model, setting a timeout to 0 usually means "infinite timeout" but I don't see a good use case for that when you can just set a really long timeout already.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/MultiplexingClientDeviceRegistrationFailedException.java
timtay-microsoft: It is only public because this function is called from the multiplexing client, which sits in a different package from these exceptions. I don't expect customers to call it, but I can't make it private without moving these new exceptions to the base package.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: This case is checking the cause of the thrown exception, not the thrown exception itself. I added one special case in the AMQP layer to throw a TransportException with a nested MultiplexingClientDeviceRegistrationFailedException so that the AMQP layer had a convenient way to report which, if any, devices failed to register.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: I'm going to rework this a little bit. Instead of this design, I'll have a class that extends TransportException but that has the same information as MultiplexingClientDeviceRegistrationFailedException in it. That way this new exception can be caught at this layer instead of catching a cause here.

timtay-microsoft: The important thing that I'm striving for here is that the top level exception should contain that map of deviceId -> registration failure exception rather than just having it in a nested exception.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
azabbasi: ah, I totallly missed the fact that you are checking the inner exception

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
azabbasi: I mean it's a technicality issue, the sentence mentions it cannot be less than 0, the user can argue "it's not, it's 0" ... so maybe 
`Cannot set a device registration timeout of 0 milliseconds or less`

azabbasi (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -146,19 +178,27 @@ public void open() throws IOException
...
-            }
+                for (DeviceClient deviceClient : this.multiplexedDeviceClients.values())
+                {
+                    deviceClient.closeFileUpload();
+                }
 
-            this.deviceIO.multiplexClose();
+                this.deviceIO.closeWithoutWrappingException();
+            }
+            catch (TransportException | IOException e)
```
abhipsaMisra, line 198: Q - if we get an IOException here, should we extract the transport exception from the IOException and throw that within the new MultiplexingClientException?

abhipsaMisra: To avoid the actual exception info being 2 levels lower

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -173,9 +213,61 @@ public void close() throws IOException
      * multiplexed connection.
      * <p>
      * Users should use {@link #registerDeviceClients(Iterable)} for registering multiple devices as it has some
-     * performance improvements over repeatedly calling this method. This method blocks on each registration, whereas
+     * performance improvements over repeatedly calling this method for individual device registrations. This method blocks on each registration, whereas
+     * {@link #registerDeviceClients(Iterable)} blocks on all of the registrations after starting them all asynchronously.
+     * <p>
+     * A device client can be unregistered using {@link #unregisterDeviceClient(DeviceClient)}, {@link #unregisterDeviceClient(DeviceClient, long)},
+     * {@link #unregisterDeviceClients(Iterable)}, or {@link #unregisterDeviceClients(Iterable, long)}. A device client will not be unregistered
+     * automatically if it encounters a non-retryable exception, so users are responsible for unregistering a device client
```
abhipsaMisra, line 221: Q - "A device client will not be unregistered automatically if it encounters a non-retryable exception" - is this statement applicable only during the registration call, or during device operation as well?
i.e. -> encounter a non-retryable exception during registration vs during sending telemetry.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -173,9 +213,61 @@ public void close() throws IOException
...
+     * when they no longer want it in this multiplexing client.
+     * <p>
+     * Up to {@link #MAX_MULTIPLEX_DEVICE_COUNT_AMQPS} devices can be registered on a multiplexed AMQPS connection,
+     * and up to {@link #MAX_MULTIPLEX_DEVICE_COUNT_AMQPS_WS} devices can be registered on a multiplexed AMQPS_WS connection.
+     * <p>
+     * If the multiplexing client is already open, then this device client will automatically
+     * be opened, too. If the multiplexing client is not open yet, then this device client will not be opened until
+     * {@link MultiplexingClient#open()} is called.
```
abhipsaMisra, line 229: 👍

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -173,9 +213,61 @@ public void close() throws IOException
...
+     * <p>
+     * If the multiplexed connection is already open, then this call will add this device client to the
+     * multiplexed connection, and then will block until the registration has been completed.
```
abhipsaMisra, line 232: q - "and then will block until the registration has been completed" -> should this say "and then will block until the device client has been opened" instead?
also, adding an already opened device client to an opened mux connection will simply add it to the hashmap (or whichever collection is used) and will return immediately, correct?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -339,17 +497,49 @@ public void registerDeviceClients(Iterable<DeviceClient> deviceClients) throws I
...
      */
-    public void unregisterDeviceClient(DeviceClient deviceClient) throws InterruptedException
+    public void unregisterDeviceClient(DeviceClient deviceClient) throws InterruptedException, MultiplexingClientException
+    {
+        this.unregisterDeviceClient(deviceClient, DEFAULT_UNREGISTRATION_TIMEOUT_MILLISECONDS);
+    }
+
+    /**
+     * Unregister a device client from this multiplexing client. This method may be called before or after opening the
+     * multiplexed connection.
```
abhipsaMisra, line 511: Q - unregister will do a graceful close of the transport layer, right? So "register -> open -> send telemetry -> unregister -> register -> send telemetry" should be fine?

abhipsaMisra: On that note, what happens if I call deviceClient.close() but don't unregister the client from the mux connection. Will it just hang around in the connection blocking one position? If I call deviceClient.open() after that, will it be usable again?

abhipsaMisra: I see this note below: "It cannot be used in non-multiplexing scenarios", so will deviceclient.close() throw an exception?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/TransportClient.java
abhipsaMisra: probably because of the method signature "public void open() throws IllegalStateException, IOException"

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/MultiplexingClientDeviceRegistrationTimeoutException.java
```diff
@@ -0,0 +1,46 @@
...
+    {
+        super(message);
+    }
+
+    /**
+     * Construct a new MultiplexingClientDeviceRegistrationTimeoutException with a nested exception and an error message.
+     * @param message The top level message for this exception.
+     * @param cause The nested exception.
+     */
+    public MultiplexingClientDeviceRegistrationTimeoutException(String message, Throwable cause)
```
abhipsaMisra, line 33: q - how detailed is this message/ throwable? does it give any info about which device actually timed out (while opening/ closing)?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubListener.java
```diff
@@ -56,4 +56,13 @@
      * @param deviceId the Id of the device that the session belongs to
      */
     void onMultiplexedDeviceSessionLost(Throwable e, String connectionId, String deviceId);
+
+    /**
+     * Callback to be fired when the multiplexed connection fails to register a device session.
+     * @param connectionId the Id of the connection, used to identify which of possible many reconnection attempts
```
abhipsaMisra, line 62: 👍

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -337,6 +339,15 @@ public void onMultiplexedDeviceSessionLost(Throwable e, String connectionId, Str
         }
     }
 
+    @Override
+    public void onMultiplexedDeviceSessionRegistrationFailed(String connectionId, String deviceId, Exception e)
+    {
+        if (connectionId != null && connectionId.equals(this.iotHubTransportConnection.getConnectionId()))
```
abhipsaMisra, line 345: Q - "connectionId.equals(this.iotHubTransportConnection.getConnectionId())" is this to avoid the scenario where by the time this callback was fired the client had already initiated a new reconnection attempt?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: The IOException caught here is only thrown from the closing of the FileUpload code. That code doesn't have nested TransportExceptions, so it should be fine to not try to un-nest this particular exception

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: It is applicable at all times. The only time when a device becomes unregistered is when the user calls multiplexingClient.unregister()

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: You can't register a device client that has already been opened, whether it was opened on a different multiplexing client, or just opened as a non-multiplexing client connection.

> should this say "and then will block until the device client has been opened"

I'm deliberately trying to use the phrase "register" rather than "open" in this context since the API itself is called "**register**DeviceClient"

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: > Q - unregister will do a graceful close of the transport layer, right? So "register -> open -> send telemetry -> unregister -> register -> send telemetry" should be fine?

Unregister doesn't close the multiplexing connection, it simply removes the device session associated with that device client. So that flow you outlined will work, yes

timtay-microsoft: > On that note, what happens if I call deviceClient.close() but don't unregister the client from the mux connection. Will it just hang around in the connection blocking one position? If I call deviceClient.open() after that, will it be usable again?

I have a special UnsupportedOperationException to prevent users from calling open/close on the device client itself when the client is multiplexed. The only way to open and close a multiplexed client is by registering it/unregistering it

> I see this note below: "It cannot be used in non-multiplexing scenarios", so will deviceclient.close() throw an exception?

Yep. That way users can't mix up if their client is multiplexed or not

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/MultiplexingClientDeviceRegistrationTimeoutException.java
timtay-microsoft: The message will be a hardcoded "Timed out waiting for all device unregistrations to finish." and there will not be a nested throwable, at least for now. The only reason I provide overloaded constructors that take throwables is because that is standard for exceptions.

> does it give any info about which device actually timed out 

No, because this timeout is for all of the device registrations, not per device registration. If a user calls the singular multiplexingClient.registerDeviceClient(), then they will know which client timed out. For the bulk APIs, though, I expect users will just call the API again with the same set of device clients to be registered. The multiplexing client allows for you to "re-register" a client or a set of clients without throwing, after all.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
timtay-microsoft: Yeah, the IotHubTransport layer may spawn a new AmqpsIotHubConnection instance during retry logic. This connectionId prevents us from logging/retrying errors from old connections that we've moved on from

abhipsaMisra (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: Oh, good catch, I misunderstood what you were getting at

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
abhipsaMisra: > You can't register a device client that has already been opened, whether it was opened on a different multiplexing client, or **just opened as a non-multiplexing client connection**.

Is this documented?

PR #1012, opened 2020-12-03
fix(iot-dev): Fix possible NPE in AMQP layer

timtay-microsoft:
Sending a get twin request while the subscribe to desired properties request was in progress seemed to cause reliable NPE in transport clients that start twin for each of their devices. By waiting for the in progress subscriptions to finish before sending actual messages over these links, the NPE goes away. The NPE itself comes from within the AMQP library, so my best guess is that there is a race condition of some sort that we would hit previously

Also simplify some code in our executor service cleanup logic to match what we checked into preview earlier.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -740,35 +740,13 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
         }
     }
 
-    private void executorServicesCleanup() throws TransportException
+    private void executorServicesCleanup()
     {
         if (this.executorService != null)
         {
             log.trace("Shutdown of executor service has started");
-            this.executorService.shutdown();
-            try
-            {
-                // Wait a while for existing tasks to terminate
-                if (!this.executorService.awaitTermination(MAX_WAIT_TO_TERMINATE_EXECUTOR, TimeUnit.SECONDS))
-                {
-                    this.executorService.shutdownNow(); // Cancel currently executing tasks
-                    // Wait a while for tasks to respond to being cancelled
-                    if (!this.executorService.awaitTermination(MAX_WAIT_TO_TERMINATE_EXECUTOR, TimeUnit.SECONDS))
-                    {
-                        log.trace("Pool did not terminate");
-                    }
-                }
-
-                this.executorService = null;
-            }
-            catch (InterruptedException e)
-            {
-                log.warn("Interrupted while cleaning up executor services", e);
-                // (Re-)Cancel if current thread also interrupted
-                this.executorService.shutdownNow();
-                this.executorService = null;
-                throw new TransportException("Waited too long for the connection to close.", e);
-            }
+            this.executorService.shutdownNow();
+            this.executorService = null;
             log.trace("Shutdown of executor service completed");
         }
     }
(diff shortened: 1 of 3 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -740,35 +740,13 @@ else if (this.connection.getLocalState() == EndpointState.CLOSED && this.connect
...
             log.trace("Shutdown of executor service has started");
-            this.executorService.shutdown();
-            try
-            {
-                // Wait a while for existing tasks to terminate
-                if (!this.executorService.awaitTermination(MAX_WAIT_TO_TERMINATE_EXECUTOR, TimeUnit.SECONDS))
-                {
-                    this.executorService.shutdownNow(); // Cancel currently executing tasks
-                    // Wait a while for tasks to respond to being cancelled
-                    if (!this.executorService.awaitTermination(MAX_WAIT_TO_TERMINATE_EXECUTOR, TimeUnit.SECONDS))
```
azabbasi, line 756: it looks like we were trying really hard to make sure the executorService gets terminated before nulling it out. 
What are the consequences of not waiting for these anymore?

timtay-microsoft: Nothing really. This executor was used to manage the one worker thread that we extend to proton-j to run the AMQP connection. When this function is called, we are deliberately closing everything, including this thread pool.

timtay-microsoft: For reference, this method is only called when opening the connection fails or when we have finished closing a successful AMQP connection. In both cases, there is nothing left that we need to be careful about cleaning up in the AMQP library.

azabbasi (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -355,6 +355,26 @@ else if (subscriptionType == DEVICE_OPERATION_TWIN_SUBSCRIBE_DESIRED_PROPERTIES_
                         || senderLinkHandler instanceof AmqpsTwinSenderLinkHandler && messageType == DEVICE_TWIN
                         || senderLinkHandler instanceof AmqpsMethodsSenderLinkHandler && messageType == DEVICE_METHODS)
                 {
+                    if (messageType == DEVICE_TWIN)
+                    {
+                        if (explicitInProgressTwinSubscriptionMessage != null)
+                        {
+                            // Don't send any twin messages while a twin subscription is in progress. Wait until the subscription
+                            // has been acknowledged by the service before sending it.
```
abhipsaMisra, line 363: Q - from where will this message be retried?

abhipsaMisra (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
timtay-microsoft: The function that calls this checks to see if the message send was successful. If it wasn't, it requeues the message. that's why we return ```false``` here to mimic a sending failure

PR #1015, opened 2020-12-07
fix(iot-dev): Fix cleanup logic in AMQP layer when interrupted during open

timtay-microsoft:
exception should be retryable, and the catch block should close the reactor and shut down the executor service, not just shut down the executor service. Close() covers this well so a call to it from this catch block works well

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -158,8 +158,10 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
             }
             catch (InterruptedException e)
             {
-                executorServicesCleanup();
-                throw new TransportException("Interrupted while waiting for links to open for AMQP connection", e);
+                this.close();
+                TransportException interruptedTransportException = new TransportException("Interrupted while waiting for links to open for AMQP connection", e);
+                interruptedTransportException.setRetryable(true);
+                throw interruptedTransportException;
             }
         }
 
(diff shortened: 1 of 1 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -158,8 +158,10 @@ public void open(Queue<DeviceClientConfig> deviceClientConfigs) throws Transport
             }
             catch (InterruptedException e)
             {
-                executorServicesCleanup();
```
barustum, line 161: I am assuming we only want to clean up on close?

barustum (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
timtay-microsoft: The ```close``` method already calls ```executorServicesCleanup```, so I'm just delaying it until after the reactor has stopped

drwill-ms (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -181,6 +183,7 @@ public void close() throws TransportException
         }
         catch (InterruptedException e)
         {
+            this.executorServicesCleanup();
             throw new TransportException("Interrupted while closing proton reactor", e);
```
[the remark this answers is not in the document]
timtay-microsoft: It's like we are nuclear physicists or something!

PR #1016, opened 2020-12-07
[Preview] Fix multiplexingClient not allowing re-registered clients to subscribe to twin/methods

timtay-microsoft:
If a client subscribes to twin and method, then unregisters from its multiplexing client, and is later re-registered, the default behavior will be to not preserve those subscriptions. We can add an optional parameter to enable this sort of behavior later, if users are interested.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
@@ -1081,7 +1081,8 @@ public class InternalClient
 
     /**
      * Returns if this client is or ever was registered to a {@link MultiplexingClient} instance. Device clients that were
-     * cannot be used in non-multiplexed connections.
+     * cannot be used in non-multiplexed connections. Device clients that aren't registered to any multiplexing client
+     * will still return true.
      * @return true if this client is or ever was registered to a {@link MultiplexingClient} instance, false otherwise.
      */
     public boolean isMultiplexed()
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
@@ -649,6 +649,20 @@ public class InternalClient
         }
     }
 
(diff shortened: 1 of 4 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -627,6 +627,20 @@ void startTwinInternal(IotHubEventCallback twinStatusCallback, Object twinStatus
...
+    void markTwinAsUnsubscribed()
+    {
+        this.twin = null;
+    }
+
+    // only used by the MultiplexingClient class to signal to this client that it needs to re-register methods
+    // callbacks
+    void markMethodsAsUnsubscribed()
+    {
+        this.method = null;
```
timtay-microsoft, line 641: Elsewhere in this class, this field is assigned when the user wants to subscribe to methods. When not null, this layer assumes that it is already subscribed. So setting this value to null again allows users to subscribe to methods again later if they re-register

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -601,6 +625,12 @@ public void unregisterDeviceClients(Iterable<DeviceClient> deviceClients, long t
                 log.info("Unregistering device {} from multiplexing client", deviceClientToUnregister.getConfig().getDeviceId());
                 this.multiplexedDeviceClients.remove(deviceClientToUnregister.getConfig().getDeviceId());
                 deviceClientToUnregister.setDeviceIO(null);
+
+                // Clear the device client's subscriptions after it has been unregistered from the multiplexing client
+                // We may make this optional at some point in the future if users want to be able to preserve subscriptions.
+                deviceClientToUnregister.markTwinAsUnsubscribed();
+                deviceClientToUnregister.markMethodsAsUnsubscribed();
+                deviceClientToUnregister.setMessageCallback(null, null);
```
timtay-microsoft, line 633: MessageCallback allows for us and users to set the callback to null, but methods and twin do not, hence the need for these new markTwin/MethodAsUnsubscribed methods

drwill-ms (approved):
Might want to refer to the new public method in the sample

PR #1019, opened 2020-12-08
fix(iot-dev): Fix issue where unregistering a multiplexed device that isn't registered loops infinitely

timtay-microsoft:
This loop happens in the AMQP worker thread, so users wouldn't notice this, but it is still a bug.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -952,41 +952,41 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
             {
                 log.trace("Removing session handler for device {}", amqpsSessionHandler.getDeviceId());
                 this.sessionHandlers.remove(amqpsSessionHandler);
+
+                // Need to find the sas token renewal handler that is tied to this device
+                AmqpsSasTokenRenewalHandler sasTokenRenewalHandlerToRemove = null;
+                for (AmqpsSasTokenRenewalHandler existingSasTokenRenewalHandler : this.sasTokenRenewalHandlers)
+                {
+                    if (existingSasTokenRenewalHandler.amqpsSessionHandler.getDeviceId().equals(configToUnregister.getDeviceId()))
+                    {
+                        sasTokenRenewalHandlerToRemove = existingSasTokenRenewalHandler;
+
+                        // Stop the sas token renewal handler from sending any more authentication messages on behalf of this device
+                        log.trace("Closing sas token renewal handler for device {}", configToUnregister.getDeviceId());
+                        sasTokenRenewalHandlerToRemove.close();
+                        break;
+                    }
+                }
+
+                if (sasTokenRenewalHandlerToRemove != null)
+                {
+                    this.sasTokenRenewalHandlers.remove(sasTokenRenewalHandlerToRemove);
+                }
+
+                this.reconnectionsScheduled.remove(configToUnregister.getDeviceId());
+
+                log.debug("Closing device session for multiplexed device {}", configToUnregister.getDeviceId());
+                amqpsSessionHandler.closeSession();
+
+                configsUnregisteredSuccessfully.add(configToUnregister);
+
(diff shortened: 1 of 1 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -952,41 +952,41 @@ private void checkForNewlyUnregisteredMultiplexedClientsToStop()
             {
                 log.trace("Removing session handler for device {}", amqpsSessionHandler.getDeviceId());
                 this.sessionHandlers.remove(amqpsSessionHandler);
-            }
-            else
-            {
-                log.warn("Attempted to remove device session for device {} from multiplexed connection, but device was not currently registered.", configToUnregister.getDeviceId());
-                return;
```
timtay-microsoft, line 959: The issue here was that this code forgot to add this config to the set of "successfully" unregistered configs so that the end of the function could remove it from the queue of configs to unregister. It can't be removed at this point since that would be a concurrent modification exception since this function loops over that queue.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -952,41 +952,42 @@ private void checkForNewlyUnregisteredMultiplexedClientsToStop()
...
-            configsUnregisteredSuccessfully.add(configToUnregister);
+                configsUnregisteredSuccessfully.add(configToUnregister);
 
-            configToUnregister = configsToUnregisterIterator.hasNext() ? configsToUnregisterIterator.next() : null;
+                configToUnregister = configsToUnregisterIterator.hasNext() ? configsToUnregisterIterator.next() : null;
+            }
+            else
+            {
+                log.warn("Attempted to remove device session for device {} from multiplexed connection, but device was not currently registered.", configToUnregister.getDeviceId());
+                configsUnregisteredSuccessfully.add(configToUnregister);
```
barustum, line 988: this line and the one after happen regardless what the code flow is. Maybe move outside the if/else.

timtay-microsoft: Good catch, will do!

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -952,41 +952,40 @@ private void checkForNewlyUnregisteredMultiplexedClientsToStop()
...
+            }
+            else
+            {
+                log.warn("Attempted to remove device session for device {} from multiplexed connection, but device was not currently registered.", configToUnregister.getDeviceId());
+            }
 
             configsUnregisteredSuccessfully.add(configToUnregister);
-
             configToUnregister = configsToUnregisterIterator.hasNext() ? configsToUnregisterIterator.next() : null;
+
```
barustum, line 988: extra space before closing brace

barustum (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -952,41 +952,40 @@ private void checkForNewlyUnregisteredMultiplexedClientsToStop()
...
 
-            this.reconnectionsScheduled.remove(configToUnregister.getDeviceId());
+                this.reconnectionsScheduled.remove(configToUnregister.getDeviceId());
 
-            log.debug("Closing device session for multiplexed device {}", configToUnregister.getDeviceId());
-            amqpsSessionHandler.closeSession();
+                log.debug("Closing device session for multiplexed device {}", configToUnregister.getDeviceId());
+                amqpsSessionHandler.closeSession();
+            }
+            else
```
drwill-ms, line 981: consider handling the error/smaller case first, so both blocks appear closer to the if condition. Otherwise, by the time a reader gets down here they've lost the context of what `else` means.

drwill-ms (approved):

Issue #555, opened 2019-07-15
Add java 12 support


maxtar:
- **OS and version used:** Windows 10

- **Java runtime used:** OpenJdk 12.0.1.

- **SDK version used:** OpenJdk 12.0.1.

# Description of the issue:
When I run test under java 12 I receiving exception described bellow. With java 8 all works fine.

# Console log of the issue:
```
Could not open the connection
java.io.IOException: Could not open the connection
	at com.microsoft.azure.sdk.iot.device.DeviceIO.open(DeviceIO.java:164)
	at com.microsoft.azure.sdk.iot.device.InternalClient.open(InternalClient.java:130)
	at com.microsoft.azure.sdk.iot.device.DeviceClient.open(DeviceClient.java:317)
...
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:500)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:481)
	at org.junit.platform.commons.util.ReflectionUtils.newInstance(ReflectionUtils.java:453)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:62)
	at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.invokeTestClassConstructor(ClassTestDescriptor.java:345)
	at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.instantiateTestClass(ClassTestDescriptor.java:290)
	at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.instantiateTestClass(ClassTestDescriptor.java:281)
	at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.instantiateAndPostProcessTestInstance(ClassTestDescriptor.java:269)
	at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.lambda$testInstancesProvider$2(ClassTestDescriptor.java:259)
[...]

PR #1020, opened 2020-12-09
Fixed https://github.com/Azure/azure-iot-sdk-java/issues/555 by chang…

maxtar:
…ed keystore type.

[555](https://github.com/Azure/azure-iot-sdk-java/issues/555)

# Description of the problem

In java 9 default keystore container was changed to pkcs12 from pks.

# Description of the solution

Changed creating keystore to pks type.

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
@@ -29,8 +29,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.UUID;
 
-public class IotHubSSLContext
-{
+public class IotHubSSLContext {
     private SSLContext sslContext = null;
 
     private static final String SSL_CONTEXT_INSTANCE = "TLSv1.2";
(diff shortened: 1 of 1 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
```diff
@@ -202,7 +188,7 @@ private void generateSSLContextWithKeys(String publicKeyCertificateString, Strin
         char[] temporaryPassword = generateTemporaryPassword();
 
         //Codes_SRS_IOTHUBSSLCONTEXT_34_020: [The constructor shall create a keystore containing the public key certificate and the private key.]
-        KeyStore keystore = KeyStore.getInstance(KeyStore.getDefaultType());
+        KeyStore keystore = KeyStore.getInstance("jks");
```
jamdavi, line 191: We shouldn't hard code anyting here but instead rely on the `keystore.type` property. See the documentation for an explanation:

>public static final String getDefaultType​()
Returns the default keystore type as specified by the keystore.type security property, or the string "jks" (acronym for "Java keystore") if no such property exists.
>
>The default keystore type can be used by applications that do not want to use a hard-coded keystore type when calling one of the getInstance methods, and want to provide a default keystore type in case a user does not specify its own.
>
>The default keystore type can be changed by setting the value of the keystore.type security property to the desired keystore type.
>
>Returns:
>the default keystore type as specified by the keystore.type security property, or the string "jks" if no such property exists.
>See Also:
>security properties


https://docs.oracle.com/javase/9/docs/api/index.html?java/security/KeyStore.html

jamdavi (changes requested):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
maxtar: Ok. Will change our properties. Thanx.

jamdavi:
@maxtar Thanks for taking the time to investigate your issue, I think the only change you requested was the keystore.type. I'll keep this open for a couple of days to make sure there isn't a need for any other code changes and close it by Friday 12/18/2020 if I don't hear back.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
```diff
@@ -43,31 +42,29 @@
...
      * Creates a SSLContext for the IotHub.
      *
-     * @throws KeyStoreException  if no Provider supports a KeyStoreSpi implementation for the specified type or
-     *                            if the keystore has not been initialized,
-     *                            or the given alias already exists and does not identify an entry containing a trusted certificate,
-     *                            or this operation fails for some other reason.
-     * @throws KeyManagementException As per https://docs.oracle.com/javase/7/docs/api/java/security/KeyManagementException.html
-     * @throws IOException If the certificate provided was null or invalid
-     * @throws CertificateException As per https://docs.oracle.com/javase/7/docs/api/java/security/cert/CertificateException.html
+     * @throws KeyStoreException        if no Provider supports a KeyStoreSpi implementation for the specified type or
```
bikamani, line 45: extra whitespaces

maxtar:
No need. Problem resolved by security file

PR #1021, opened 2020-12-09
fix(iot-dev): fix issue where AMQP layer sometimes throws "uncorrelated channel" NPE

timtay-microsoft:
Closing a session locally before the session was opened remotely causes a NPE to throw from proton with error message "uncorrelated channel: X", so this will fix our session handlers to not close locally before the session has been opened remotely

The relevant proton code can be seen [here](https://github.com/apache/qpid-proton-j/blob/fcdc70521186d626490657823289bd952121f0ce/proton-j/src/main/java/org/apache/qpid/proton/engine/impl/TransportImpl.java#L1213).

Note that the proton function in question is a handler function for "session begin" frames sent by the service

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
@@ -107,6 +107,7 @@ public class AmqpsSessionHandler extends BaseHandler implements AmqpsLinkStateCa
             {
                 // This flag signals to this session handler to close the session once the service has opened it remotely
                 // see above for more details on why.
+                log.trace("Session handler was closed but the service has not opened the session remotely yet, so the session will be closed once that happens.");
                 this.sessionHandlerClosedBeforeRemoteSessionOpened = true;
             }
         }
@@ -130,6 +131,9 @@ public class AmqpsSessionHandler extends BaseHandler implements AmqpsLinkStateCa
         this.sessionOpenedRemotely = true;
         if (this.sessionHandlerClosedBeforeRemoteSessionOpened)
         {
+            // If the session handler was closed earlier, before this session opened remotely, then now is the soonest
+            // that the session itself can safely be closed.
+            log.trace("Closing an out of date session now that the service has opened the session remotely.");
             this.session.close();
         }
         else if (this.deviceClientConfig.getAuthenticationType() == DeviceClientConfig.AuthType.X509_CERTIFICATE)
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
@@ -127,7 +127,7 @@ public class AmqpsSessionHandler extends BaseHandler implements AmqpsLinkStateCa
     public void onSessionRemoteOpen(Event e)
     {
         log.trace("Device session opened remotely for device {}", this.getDeviceId());
-
+        this.sessionOpenedRemotely = true;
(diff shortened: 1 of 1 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -80,13 +80,35 @@ public void setSession(Session session)
...
+        this.sessionOpenedRemotely = false;
+        this.sessionHandlerClosedBeforeRemoteSessionOpened = false;
     }
 
     public void closeSession()
     {
         if (this.session != null)
         {
-            this.session.close();
+            if (this.sessionOpenedRemotely)
```
jamdavi, line 91: Is this ever being set somewhere or is this for the future?

timtay-microsoft: Good catch, somehow this didn't get committed to this PR. I've added it being set to true on the onSessionRemoteOpen event now

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -105,7 +127,12 @@ public void onSessionFinal(Event e)
     public void onSessionRemoteOpen(Event e)
     {
         log.trace("Device session opened remotely for device {}", this.getDeviceId());
-        if (this.deviceClientConfig.getAuthenticationType() == DeviceClientConfig.AuthType.X509_CERTIFICATE)
+        this.sessionOpenedRemotely = true;
+        if (this.sessionHandlerClosedBeforeRemoteSessionOpened)
+        {
+            this.session.close();
```
jamdavi, line 133: Since we're intentionally closing this session but not opening a new one in this method we should definitely log it as an error/warning. 

Do we want session.close() to throw here ?

timtay-microsoft: The connection handler that manages 1 to many of these session handlers is generally in charge of managing which sessions are opened and closed. In this particular case, there is no need to open a new session from this session handler since the connection handler deliberately closed this session handler.

The expected flow here during retry logic is to have the connection handler create a new session handler for the device that is reconnecting, and if that session handler doesn't open in time, then we close it, wait for a few seconds, and then open a new session handler to try again

timtay-microsoft: In essence, the session handler is disposable, just like the amqp session itself is here

timtay-microsoft: > Do we want session.close() to throw here ?

This is expected behavior to respond to the service acknowledging the beginning of a session that we no longer track, so no

jamdavi: Got it, so what ever is calling these onXXXX methods due to the event should have context of the state and this just fired out of order due to some unforeseen circumstance.

Still probably worth logging that this condition happened either here or in the closeSession event above

jamdavi (approved):
Looks good to me. Would like to see the additional logging so we know we got into this situation

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
timtay-microsoft: Good idea, I can add some extra logs for this

PR #1036, opened 2020-12-31
fix: Java code style warnings

bikamani:
This PR is to fix code style warnings plus few others.

code style warnings:

image

misc:

- [Error handling] Catch block may ignore exception (5)
- [Logging] Number of placeholders does not match number of arguments in logging call (1)

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Socks5SocketFactory.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Socks5SocketFactory.java
@@ -194,8 +194,8 @@ public class Socks5SocketFactory extends SocketFactory
             {
                 try {
                     mLocalIP = InetAddress.getByName(mLocalHost);
-                } catch (UnknownHostException e) {
-                    e.printStackTrace();
+                } catch (UnknownHostException e)
+                {
                 }
             }
             return mLocalIP;
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Socks5SocketFactory.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Socks5SocketFactory.java
@@ -183,7 +183,6 @@ public class Socks5SocketFactory extends SocketFactory
             return isLocal() ? super.getLocalPort() : mProxied.mPort;
         }
 
-        @SneakyThrows
         @Override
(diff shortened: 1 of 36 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Socks5SocketFactory.java
```diff
@@ -181,6 +183,7 @@ public int getLocalPort() {
             return isLocal() ? super.getLocalPort() : mProxied.mPort;
         }
 
+        @SneakyThrows
```
timtay-microsoft, line 186: I don't think this is the correct idea. This changes the behavior of the method to throw the exception rather than just return null when an UnknownHostException is thrown. I'd recommend logging the exception rather than doing a sneaky throws

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Socks5SocketFactory.java
```diff
@@ -243,13 +240,7 @@ public InetAddress getLocalAddress()
...
                 if ((type != ATYP_DOMAINNAME))
                 {
-                    try
-                    {
-                        mIp = InetAddress.getByName(mHost);
-                    }
-                    catch (final UnknownHostException uh_ex)
-                    {
-                    }
+                    mIp = InetAddress.getByName(mHost);
```
timtay-microsoft, line 243: Same comment here as well

timtay-microsoft (approved):
Address the one comment, then this looks good to me

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Socks5SocketFactory.java
bikamani: I am not sure if just logging will get rid of this warning. I will check that.

bikamani: Does work. I have updated to log it instead.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Socks5SocketFactory.java
bikamani: updated

PR #1038, opened 2021-01-04
fix(iot-dev): Fix multiplexing client so that it unwinds previous bulk registrations if one client isn't supported

timtay-microsoft:
Previously, the local state of which devices were registered already would be incorrect upon encountering an error like this if the error wasn't triggered by the first device client in the list to register.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
@@ -417,12 +417,6 @@ public class MultiplexingClient
         {
             List<DeviceClientConfig> deviceClientConfigsToRegister = new ArrayList<>();
 
-            // When registering in bulk, some clients may cause this method to throw (for example if the client uses MQTT),
-            // so it is important to be able to unwind the bulk operation when this happens. This list
-            // contains the Ids of the clients that were ready to be registered before encountering an error like that
-            // so that they can be removed from the local state.
-            List<String> partiallyRegisteredClientIds = new ArrayList<>();
-
             for (DeviceClient deviceClientToRegister : deviceClients)
             {
                 DeviceClientConfig configToAdd = deviceClientToRegister.getConfig();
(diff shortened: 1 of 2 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -459,7 +460,7 @@ public void registerDeviceClients(Iterable<DeviceClient> deviceClients, long tim
 
                 if (deviceClientToRegister.getDeviceIO() != null && deviceClientToRegister.getDeviceIO().isOpen() && !deviceClientToRegister.isMultiplexed)
                 {
-                    throw new IllegalStateException("Cannot register a device client to a multiplexed connection when the device client was already opened.");
```
timtay-microsoft, line 462: All the other exceptions like this one threw UnsupportedOperationException, so it seemed appropriate for me to change this one to match

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -475,7 +476,6 @@ public void registerDeviceClients(Iterable<DeviceClient> deviceClients, long tim
                 }
                 else
                 {
-                    this.multiplexedDeviceClients.put(deviceClientToRegister.getConfig().getDeviceId(), deviceClientToRegister);
```
timtay-microsoft, line 478: Putting the client in the map of multiplexed devices here was premature since the registration hasn't actually happened yet. The client has only been validated to ensure that it has the expected protocol, hostname, etc.

drwill-ms (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
```diff
@@ -1397,6 +1399,107 @@ public void registerDevicesWithIncorrectCredentialsBeforeOpenThrowsOnOpen() thro
...
+            this.testInstance.multiplexingClient.registerDeviceClients(this.testInstance.deviceClientArray);
+        }
+        catch (UnsupportedOperationException e)
+        {
+            expectedExceptionThrown = true;
+        }
+
+        assertTrue("Expected bulk registration to throw an UnsupportedOperationException", expectedExceptionThrown);
+
+        // Because one of the devices in the bulk registration was MQTT, and therefore unsupported, none of the other clients
```
abhipsaMisra, line 1494: nit: you probably forgot to update this - I see you're calling this method from other test scenarios as well (x509 device, device already opened etc.), so the desc shouldn't say "mqtt" specifically.

abhipsaMisra (approved):

azabbasi (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -487,6 +487,13 @@ public void registerDeviceClients(Iterable<DeviceClient> deviceClients, long tim
             }
 
             this.deviceIO.registerMultiplexedDeviceClient(deviceClientConfigsToRegister, timeoutMilliseconds);
+
+            for (DeviceClient successfullyRegisteredClient : deviceClients)
+            {
+                // Only update the local state map once the register call has succeeded
+                String deviceIdToRegister = successfullyRegisteredClient.getConfig().getDeviceId();
```
bikamani, line 494: Nit: Is it better to call this registeredDeviceId?

PR #1039, opened 2021-01-05
Bump dependency versions

timtay-microsoft:
Guava is only used in our e2e tests, but we may as well bump its version.

Apache commons is a bit more interesting. We use it for base64 and base32 encoding in some places, but we also have a local implementation that we use in other places.

While bumping the apache commons version, I figured it was time to consolidate so that all places in our library use the same base64 library instead of mixing it like this

Also incorporated the changes from #848 since they are tangentially related. See that PR for more comments.


deps/pom.xml
```diff
@@ -43,15 +43,6 @@
     </properties>
 
     <dependencies>
-        <dependency>
-            <groupId>commons-codec</groupId>
-            <artifactId>commons-codec</artifactId>
-            <!--
-            Note! Do not increase the version of this dependency
-            Android devices are unable to use versions of this library higher than version 1.2
```
timtay-microsoft, line 51: I'm still running some tests around Android support, but this change doesn't appear to break anything despite this comment I left a while back

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/util/Base64.java
```diff
@@ -1,324 +0,0 @@
-// Copyright (c) Microsoft. All rights reserved.
```
timtay-microsoft, line 1: We don't need our own implementation of base64 encoding/decoding. Apache commons will handle that for us

device/iot-device-client/pom.xml
```diff
@@ -97,11 +97,6 @@
             <artifactId>gson</artifactId>
             <version>2.8.0</version>
         </dependency>
-        <dependency>
-            <groupId>commons-codec</groupId>
-            <artifactId>commons-codec</artifactId>
-            <version>1.10</version>
-        </dependency>
```
timtay-microsoft, line 104: device client will get this commons dependency from our deps package, so no need to redeclare it here as well

barustum (approved):

provisioning/security/dice-provider-emulator/pom.xml
```diff
@@ -65,7 +65,7 @@
         <dependency>
             <groupId>commons-codec</groupId>
             <artifactId>commons-codec</artifactId>
-            <version>1.5</version>
```
abhipsaMisra, line 68: that is quite a bump!

abhipsaMisra (approved):

service/iot-service-client/pom.xml
```diff
@@ -95,6 +91,12 @@
             <groupId>com.microsoft.azure</groupId>
             <artifactId>azure-storage</artifactId>
             <version>4.0.0</version>
+            <exclusions>
+                <exclusion>
```
bikamani, line 95: what is this exclusion for?

bikamani (approved):

service/iot-service-client/pom.xml
timtay-microsoft: this Azure Storage version comes with a dependency on the excluded library so that we can control the version that our library uses

PR #1040, opened 2021-01-05
Clean up code warnings

bikamani:
This PR is based on intellij's "Code Cleanup" functionality.

After code cleanup - 
image

Before
image


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/TwinMetadata.java
```diff
@@ -60,23 +60,9 @@ protected synchronized boolean update(String lastUpdated, Integer lastUpdatedVer
...
-        else if(this.lastUpdatedVersion.equals(lastUpdatedVersion))
-        {
-            updated = false;
-        }
-        else
-        {
-            updated = true;
+            updated = lastUpdatedVersion != null;
         }
+        else updated = !this.lastUpdatedVersion.equals(lastUpdatedVersion);
```
timtay-microsoft, line 65: No one line else blocks. Please add braces

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/util/Tools.java
```diff
@@ -215,8 +215,8 @@ public static void appendJsonAttribute(StringBuilder strBuilder, String name, St
             // Codes_SRS_SDK_JAVA_TOOLS_12_023: [The function shall append the input StringBuilder string using the following format: “name”:”value” if isQuoted is false]
             strBuilder.append("\"");
 
-            if (Tools.isNullOrEmpty(name))
-                strBuilder.append("");
+            if (Tools.isNullOrEmpty(name)) {
+            }
```
timtay-microsoft, line 219: This if block doesn't do anything when the condition is met, so you can get rid of it

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketUpgrade.java
```diff
@@ -314,14 +314,7 @@ public Boolean validateUpgradeReply(byte[] responseBytes)
...
 
-        if ((isStatusLineOk) && (isUpgradeHeaderOk) && (isConnectionHeaderOk) && (isProtocolHeaderOk) && (isAcceptHeaderOk))
-        {
-            return true;
-        }
-        else
-        {
-            return false;
-        }
+        return (isStatusLineOk) && (isUpgradeHeaderOk) && (isConnectionHeaderOk) && (isProtocolHeaderOk) && (isAcceptHeaderOk);
```
timtay-microsoft, line 317: I don't think we need the parenthesis here, right?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -170,7 +170,7 @@ public void onMessageSent(Message message, Throwable e)
                 //Codes_SRS_IOTHUBTRANSPORT_34_005: [If there was a packet in the inProgressPackets queue tied to the
                 // provided message, and the provided throwable is null, this function shall set the status of that
                 // packet to OK_EMPTY and add it to the callbacks queue.]
-                this.log.trace("Message was sent by this client, adding it to callbacks queue with OK_EMPTY ({})", message);
+                IotHubTransport.log.trace("Message was sent by this client, adding it to callbacks queue with OK_EMPTY ({})", message);
```
timtay-microsoft, line 173: I'm not sure I understand why we are making this particular change. Does the related warning say why to call into the log statically rather than through ```this```?

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/Tools.java
```diff
@@ -215,8 +215,8 @@ public static void appendJsonAttribute(StringBuilder strBuilder, String name, St
             // Codes_SRS_SERVICE_SDK_JAVA_TOOLS_12_023: [The function shall append the input StringBuilder string using the following format: "name":"value" if isQuoted is false]
             strBuilder.append("\"");
 
-            if (Tools.isNullOrEmpty(name))
-                strBuilder.append("");
+            if (Tools.isNullOrEmpty(name)) {
+            }
```
timtay-microsoft, line 219: This if block has no contents, so it should be removed

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ExportImportDevice.java
```diff
@@ -240,12 +240,8 @@ else if (!Tools.areEqual(this.getStatus(), otherExportImportDevice.getStatus()))
             {
                 return false;
             }
-            else if (!Tools.areEqual(this.getImportMode(), otherExportImportDevice.getImportMode()))
-            {
-                return false;
-            }
+            else return Tools.areEqual(this.getImportMode(), otherExportImportDevice.getImportMode());
```
timtay-microsoft, line 243: Don't use one line else statements

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/Tools.java
```diff
@@ -215,8 +215,8 @@ public static void appendJsonAttribute(StringBuilder strBuilder, String name, St
             // Codes_SRS_SERVICE_SDK_JAVA_TOOLS_12_023: [The function shall append the input StringBuilder string using the following format: "name":"value" if isQuoted is false]
             strBuilder.append("\"");
 
-            if (Tools.isNullOrEmpty(name))
-                strBuilder.append("");
+            if (Tools.isNullOrEmpty(name)) {
```
timtay-microsoft, line 218: This if block has no contents, so it should be removed

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
bikamani: I also thought about this one in particular if it is needed or not. I do not know answer to whether this is solving a warning, not sure how to figure that out other than running code cleanup twice. Are you concerned for readability here?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketUpgrade.java
bikamani: correct! I'll remove these

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
bikamani: yes it resolves this warning:
image

bikamani: removed this instead.

jamdavi:
Here is a comparison of what changed before and after the auto cleanup. I ran this against the same data I extracted last week. Note, I remove the test files and samples from my analysis.
```
                                                   [Code maturity] Deprecated API usage Before: 31    After: 38
                                [Code style issues] Standard Charset object can be used Before: 7     After: 0
                                   [Code style issues] Unnecessary conversion to String Before: 7     After: 0
                                              [Code style issues] Unnecessary semicolon Before: 6     After: 0
                                                    [Compiler issues] Unchecked warning Before: 49    After: 47
                                 [Control flow issues] 'if' statement with common parts Before: 2     After: 3
                                         [Control flow issues] Redundant 'if' statement Before: 10    After: 1
                   [Declaration redundancy] Access static member via instance reference Before: 106   After: 0
                                      [Error handling] Catch block may ignore exception Before: 2     After: 1
                                                                [Imports] Unused import Before: 112   After: 116
                                         [Java 7] Explicit type can be replaced with <> Before: 11    After: 9
                                    [Java 8] Anonymous type can be replaced with lambda Before: 17    After: 16
                                                       [No grouping] unused declaration Before: 987   After: 849
                      [Verbose or redundant code constructs] Redundant String operation Before: 10    After: 5
                             [Verbose or redundant code constructs] Redundant type cast Before: 9     After: 0
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransportMessage.java
```diff
@@ -202,14 +202,11 @@ public boolean isMessageAckNeeded(IotHubClientProtocol protocol)
...
-                    || this.operationType == DeviceOperations.DEVICE_OPERATION_METHOD_SUBSCRIBE_REQUEST
-                    || this.operationType == DeviceOperations.DEVICE_OPERATION_TWIN_UNSUBSCRIBE_DESIRED_PROPERTIES_REQUEST)
-            {
-                //This is a SUBSCRIBE action in MQTT which we currently treat synchronously, so there is no message ack
-                // unlike other send operations
-                return false;
-            }
+            //This is a SUBSCRIBE action in MQTT which we currently treat synchronously, so there is no message ack
+            // unlike other send operations
+            return this.operationType != DeviceOperations.DEVICE_OPERATION_TWIN_SUBSCRIBE_DESIRED_PROPERTIES_REQUEST
```
vinagesh, line 207: Why not just keep the same logical OR? Is this recommended practice in java?
return x || y ||z vs return !x && !y && !z

vinagesh (approved):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/TwinProperty.java
```diff
@@ -96,10 +96,7 @@ protected Boolean addProperty(String key, Object value, Integer propertyVersion)
         {
             /* Codes_SRS_TWINPARSER_21_078: [If any `value` is null, the updateDesiredProperty shall delete it from the collection and report on Json.] */
             /* Codes_SRS_TWINPARSER_21_084: [If any `value` is null, the updateReportedProperty shall delete it from the collection and report on Json.] */
-            if(property.containsKey(key))
-            {
-                property.remove(key);
-            }
+            property.remove(key);
```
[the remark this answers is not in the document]
timtay-microsoft: In java, it is a no-op, not an exception, so it should be safe

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/RegisterManagerParser.java
```diff
@@ -114,10 +114,7 @@ protected boolean setDeviceId(String deviceId) throws IllegalArgumentException
         if((this.deviceId == null) || (deviceId == null) || (!this.deviceId.equals(deviceId)))
         {
             this.deviceId = deviceId;
-            if(this.deviceId != null)
-            {
-                return true;
-            }
+            return this.deviceId != null;
```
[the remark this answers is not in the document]
bikamani: That wouldn't work for the case where deviceId value is not null but different than this.deviceId, right? There are probably other ways to make it one line but I think it will less readable.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/util/Base64.java
```diff
@@ -0,0 +1,324 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package com.microsoft.azure.sdk.iot.deps.util;
+
+public final class Base64
```
timtay-microsoft, line 6: I got rid of this code yesterday since we should be using apache commons for our base64 encoding/decoding. Don't re-add this class back into our codebase

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/TwinProperty.java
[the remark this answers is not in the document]
bikamani: Line 70, 76 takes care of that case

timtay-microsoft (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransportMessage.java
bikamani: That's a very good question, I don't the answer. will do research and get back to you.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/util/Base64.java
bikamani: I merged the changes from master, weird it's showing up. Let me check what happened.

PR #1043, opened 2021-01-06
refactor(deps): Remove old AMQP websocket implementation

timtay-microsoft:
We use a published library for AMQPS_WS that we share with the event hubs team, so we don't need this local implementation


vinagesh (approved):

azabbasi (approved):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpsConnection.java
```diff
@@ -5,8 +5,8 @@
 
 package com.microsoft.azure.sdk.iot.deps.transport.amqp;
 
+import com.microsoft.azure.proton.transport.ws.impl.WebSocketImpl;
```
bikamani, line 8: Just wondering, why did we move to this and who maintains this codebase?

timtay-microsoft: This Proton-J websocket implementation now only lives [here](https://github.com/Azure/qpid-proton-j-extensions) and we share ownership with the eventhubs SDK team. The motivation behind this was that both our teams needed a websocket implementation that works with proton, so it made sense for us to both use the same implementation.

timtay-microsoft:
Closing this PR for now. For some reason, the service client code doesn't work with the websocket library, so it will just have to continue using our local implementation

Issue #1027, opened 2020-12-21
SSL Context in service Client 


gianni-di-marco:
It will be usefull to change SSL Context in ServiceClient (like the feature in DeviceClient) to connect to iot hub in a private azure stack appliance.

in AmqpConnectionHandler.java it   SSLContext is initialized  with iotHub ca certificates.
As far as I can see there is no way to change it.
 

```java:
private SslDomain makeDomain(SslDomain.Mode mode)
    {
        SslDomain domain = Proton.sslDomain();

        try
        {
            // Need the base trusted certs for IotHub in our ssl context. IotHubSSLContext handles that
            domain.setSslContext(new IotHubSSLContext().getSSLContext());
        }
        catch (Exception e)
        {
            this.savedException = e;
        }

        domain.init(mode);

        return domain;
    }
```
AB#9010062

timtay-microsoft:
I like this idea a lot. I'll try to get this feature added soon.

timtay-microsoft:
Closing this issue because service client version 1.27.0 now has this feature

az-iot-builder-01:
@gianni-di-marco, @timtay-microsoft, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

mojitodaiquiri:
Hi, there,
it does't work for me, I just had a quick look and seems that the sslcontext option does't go deep through to the message sender....
in ServiceClient the variable is initialize like that
       // Codes_SRS_SERVICE_SDK_JAVA_SERVICECLIENT_12_007: [The constructor shall create a new instance of AmqpSend object]
        this.amqpMessageSender = new AmqpSend(hostName, userName, sasToken, this.iotHubServiceClientProtocol, options.getProxyOptions());

with no ssl context.

PR #1044, opened 2021-01-06
feat(iot-service): Add custom SSLContext option to service client construction

timtay-microsoft:
This enables users to load their preferred public certificates which is necessary for supporting private Hub environments

 #1027 for reference

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
@@ -8,6 +8,7 @@ package com.microsoft.azure.sdk.iot.service;
 import com.microsoft.azure.sdk.iot.service.transport.amqps.AmqpReceive;
 import lombok.extern.slf4j.Slf4j;
 
+import javax.net.ssl.SSLContext;
 import java.io.IOException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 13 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
```diff
@@ -88,6 +89,23 @@ public FeedbackReceiver(String hostName, String userName, String sasToken, IotHu
...
+     * Constructor to verify initialization parameters
+     * Create instance of AmqpReceive
+     *
+     * @param hostName The iot hub host name
+     * @param userName The iot hub user name
+     * @param sasToken The iot hub SAS token for the given device
+     * @param iotHubServiceClientProtocol protocol to be used
+     * @param proxyOptions the proxy options to tunnel through, if a proxy should be used.
+     * @param sslContext the SSL context to use during the TLS handshake when opening the connection. If null, a default
+     *                   SSL context will be generated.
```
abhipsaMisra, line 106: nit: should we specify here that the default ssl context generated trusts IoT hub's public root certs only?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClientOptions.java
```diff
@@ -14,4 +16,12 @@
      */
     @Getter
     private ProxyOptions proxyOptions;
+
+
+    /**
+     * The SSL context to use when opening the AMQPS/AMQPS_WS connections. If not set, this library will generate the default
+     * SSL context that trusts the IoT Hub public certificates.
```
abhipsaMisra, line 23: ah, I see that you've mentioned this here :)
maybe we should add this to the public API javadoc as well

abhipsaMisra (approved):

PR #1047, opened 2021-01-11
Fixes [Verbose or redundant code constructs] 'StringBuilder' can be replaced with 'String' (31)

jamdavi:
Fixes the [Verbose or redundant code constructs] 'StringBuilder' can be replaced with 'String' warning.


barustum (approved):

barustum:
Curious to know why the recommendation to replace stringbuilder with string. Probably an optimization?

jamdavi:
> Curious to know why the recommendation to replace stringbuilder with string. Probably an optimization?

Yes, per the docs it checks for SB being used for a constant number of concats.

PR #1053, opened 2021-01-11
[JUnit] simplify assertions

timtay-microsoft:


deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/Helpers.java
```diff
@@ -102,11 +101,11 @@ else if(actualValue == null)
                 {
                     if(message == null)
                     {
-                        assertTrue("Expected key:" + key + " does not exist in Actual Map", false);
+                        fail("Expected key:" + key + " does not exist in Actual Map");
```
abhipsaMisra, line 104: nice, I've never used a `fail` assertion before

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/Helpers.java
```diff
@@ -119,23 +118,23 @@ else if(expectedValue instanceof Map)
...
+                            fail(message);
                         }
                     }
                 }
                 else if(expectedValue instanceof ArrayList)
                 {
                     if(message == null)
                     {
-                        assertTrue("Map failed on " + key + ": " + actualValue + " != " + expectedValue, actualValue.toString().equals(expectedValue.toString()));
+                        Assert.assertEquals("Map failed on " + key + ": " + actualValue + " != " + expectedValue, actualValue.toString(), expectedValue.toString());
```
abhipsaMisra, line 133: do we need to state the class qualifier here?

timtay-microsoft: Nope. Both objects here are strings, so the assertEquals(string, string) overload is called

abhipsaMisra (approved):

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/Helpers.java
abhipsaMisra: No, what I meant is, why do we need to call `Assert.assertEquals` (and not `assertEquals` directly).

PR #1055, opened 2021-01-12
[Probable Bugs] Unused Assignments

timtay-microsoft:
Nothing appears to actually be a bug, but this will get rid of the relevant warnings


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubConnectionString.java
```diff
@@ -41,12 +41,12 @@
     private static final Charset CONNECTION_STRING_CHARSET = StandardCharsets.UTF_8;
 
     private String hostName = null;
```
barustum, line 43: any idea why it didn't complain about the other fields, like hostName and deviceId?

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTest.java
```diff
@@ -305,9 +305,8 @@ public void manyExtendsOfAbstractClassDoesNotChangeConfig() throws TransportExce
     @Test(expected = IllegalArgumentException.class)
     public void constructorThrowsExceptionIfConnectionIsNotInitialised() throws TransportException
     {
-        Mqtt mockMqtt = null;
         //act
-        mockMqtt = new MqttMessaging(null, CLIENT_ID, mockedIotHubListener, null, "", "", false, new HashMap<Integer, Message>());
+        Mqtt mockMqtt = new MqttMessaging(null, CLIENT_ID, mockedIotHubListener, null, "", "", false, new HashMap<Integer, Message>());
```
barustum, line 309: can we skip the assignment all together, since the variable is not used?

timtay-microsoft: Good idea

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubConnectionString.java
timtay-microsoft: Not sure, but I can certainly make the same change to them as well

barustum (approved):

PR #1056, opened 2021-01-12
[Probable Bugs] Throwable not thrown + [Probable Bug] Suspicious collections method calls

timtay-microsoft:
These were all instances in unit test code where we simulated methods throwing these methods. The tests work as expected, so these are all just false positives


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsCbsReceiverLinkHandler.java
```diff
@@ -70,7 +70,7 @@ private void handleCBSResponseMessage(Receiver receiver)
             if (amqpsMessage.getApplicationProperties() != null && amqpsMessage.getProperties() != null)
             {
                 Properties properties = amqpsMessage.getProperties();
-                Object correlationId = properties.getCorrelationId();
+                UUID correlationId = (UUID) properties.getCorrelationId();
```
timtay-microsoft, line 73: Proton's message's correlationId's are of type Object, but we only ever assign them to be instances of UUIDs, so this is a safe cast to do

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/transport/IotHubTransportTest.java
```diff
@@ -29,6 +29,7 @@
 /**
  * Unit tests for IotHubTransportPacket.
  */
+@SuppressWarnings("ThrowableNotThrown")
```
abhipsaMisra, line 32: Does the warning suggest that the unit test should be throwing the declared exceptions?

timtay-microsoft: The warning is complaining because of how jmockit expects us to make a mocked method return by throwing an exception. In our tests, we do something like

```
someMockInstance.thisMethodShouldThrow();
result = new IOException();
```

The warning sees that we create an exception instance, but never throw it and complains. This is the expected way for jmockit to make mocked methods throw though, so I just suppressed these warnings

abhipsaMisra (approved):

PR #1057, opened 2021-01-12
[Probable Bugs] multiple warnings fixed

timtay-microsoft:
See each commit message for more details, but this PR fixes several Probable Bugs warnings


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MessageProperty.java
```diff
@@ -39,7 +39,6 @@
         reservedPropertyNames.add("iothub-operation");
         reservedPropertyNames.add("iothub-partition-key");
         reservedPropertyNames.add("iothub-ack");
-        reservedPropertyNames.add("iothub-connection-device-id");
```
timtay-microsoft, line 42: This value was being added to the set twice, so I removed one instance

iot-e2e-tests/edge-e2e/src/main/java/glue/ModuleGlue.java
```diff
@@ -151,7 +151,7 @@ public void connect(String transportType, String connectionString, Certificate c
             ModuleClient client = new ModuleClient(connectionString, protocol);
 
             String cert = caCertificate.getCert();
-            if (cert != null && cert != "")
+            if (cert != null && cert.isEmpty())
```
barustum, line 154: shouldn't this be !cert.isEmpty()?

abhipsaMisra: same

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/transport/IotHubTransportTest.java
```diff
@@ -29,6 +29,7 @@
 /**
  * Unit tests for IotHubTransportPacket.
  */
+@SuppressWarnings("ThrowableNotThrown")
```
abhipsaMisra, line 32: is this because we annotate the test with the expected exception?

iot-e2e-tests/edge-e2e/src/main/java/glue/ModuleGlue.java
timtay-microsoft: Good catch, I have fixed this now

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/transport/IotHubTransportTest.java
timtay-microsoft: I know that I explained this in another PR after you made this comment, so in case someone else sees this comment and is still confused:

> The warning is complaining because of how jmockit expects us to make a mocked method return by throwing an exception. In our tests, we do something like

>```
>someMockInstance.thisMethodShouldThrow();
>result = new IOException();
>```

> The warning sees that we create an exception instance, but never throw it and complains. This is the expected way for > jmockit to make mocked methods throw though, so I just suppressed these warnings

barustum (approved):

abhipsaMisra (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/ProxiedSSLSocket.java
```diff
@@ -96,7 +96,7 @@ private void doTunnelHandshake(Socket tunnel, String host, int port) throws IOEx
         if (this.proxyUsername != null && this.proxyPassword != null)
         {
             String base64EncodedCredentials = new String(Base64.encodeBase64(String.format("%s:%s", this.proxyUsername, new String(this.proxyPassword)).getBytes(byteEncoding)));
-            proxyConnectMessage += String.format("Proxy-Authorization: Basic %s\r\n", base64EncodedCredentials, TransportUtils.USER_AGENT_STRING);
+            proxyConnectMessage += String.format("Proxy-Authorization: Basic %s\r\nUser-Agent: %s\r\n", base64EncodedCredentials, TransportUtils.USER_AGENT_STRING);
```
azabbasi, line 99: how did you find this? what is the impact?

azabbasi (approved):
just one question. looks good

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/ProxiedSSLSocket.java
timtay-microsoft: The warning advised me that the string.formal call had more parameters than "%s" instances, so I noticed we were missing the above

timtay-microsoft: No meaningful impact. Proxy connect requests don't require a user agent string. It's mostly just used for logging purposes on the user's proxy

PR #1058, opened 2021-01-12
[Performance] fix all performance warnings

timtay-microsoft:
Nothing too exciting that users would notice, from what I can tell


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/http/HttpResponse.java
```diff
@@ -149,16 +149,15 @@ protected static String canonicalizeFieldName(String field)
 
     protected static String flattenValuesList(List<String> values)
     {
-        String valuesStr = "";
+        StringBuilder valuesStr = new StringBuilder();
         for (String value : values)
         {
-            valuesStr += value + ",";
+            valuesStr.append(value).append(",");
```
abhipsaMisra, line 155: I am sure this would be appreciated.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
```diff
@@ -196,7 +196,7 @@ private void generateSSLContextWithKeys(String publicKeyCertificateString, Strin
         Key privateKey = IotHubSSLContext.parsePrivateKey(privateKeyString);
         Collection<X509Certificate> certChain = IotHubSSLContext.parsePublicKeyCertificate(publicKeyCertificateString);
 
-        X509Certificate[] certs = certChain.toArray(new X509Certificate[certChain.size()]);
+        X509Certificate[] certs = certChain.toArray(new X509Certificate[0]);
```
abhipsaMisra, line 199: is this always 0?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketUpgrade.java
```diff
@@ -234,7 +234,7 @@ public String createUpgradeRequest()
         {
             for (Map.Entry<String, String> entry : _additionalHeaders.entrySet())
             {
-                stringBuilder.append(entry.getKey() + ": " + entry.getValue()).append(_endOfLine);
+                stringBuilder.append(entry.getKey()).append(": ").append(entry.getValue()).append(_endOfLine);
```
abhipsaMisra, line 237: more string additions in a loop! 😮

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
timtay-microsoft: The newly created array will not be size 0, but for the sake of converting a collection to an array, this is the recommended way, yes

timtay-microsoft: "There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).
In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.
This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs)."

abhipsaMisra (approved):

PR #1060, opened 2021-01-12
Update java-api-reference

jlorich:
Updated java-api-reference to be the correct javadoc url since it currently 404s.  However, I do wonder if we should be referencing https://docs.microsoft.com/en-us/java/api/com.microsoft.azure.sdk.iot.device?view=azure-java-stable instead of the JavaDoc.  We seem to have both and use them inconsistently page to page.


timtay-microsoft:
@sindhu-nagesh for awareness since she last looked at our github pages, but this looks good to me

vinagesh:
Looks good to me too!

vinagesh (approved):

timtay-microsoft (approved):

PR #1061, opened 2021-01-12
[Threading Issues] fix threading issues warnings

timtay-microsoft:
A few notable cases in our production code that certainly should be fixed


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/TwinProperty.java
```diff
@@ -29,7 +29,7 @@
     private static final String LAST_UPDATE_TAG = "$lastUpdated";
     private static final String LAST_UPDATE_VERSION_TAG = "$lastUpdatedVersion";
 
-    private Object lock = new Object();
```
timtay-microsoft, line 32: All locks must be final

device/iot-device-samples/multiplexing-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ClientManagerBase.java
```diff
@@ -63,7 +64,7 @@ public void close()
 
     public void closeNow()
     {
-        synchronized (lastKnownConnectionStatus)
```
timtay-microsoft, line 66: Not generally a good idea to synchronize on non-final objects. Since this state is supposed to change over time, it makes more sense to add a lock to synchronize on instead

device/iot-device-samples/send-serialized-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendSerializedEvent.java
```diff
@@ -82,47 +112,46 @@ public void execute(IotHubStatusCode status, Object context)
         {
             System.out.println("IoT Hub responded to message with status " + status.name());
 
-            if (context != null) {
-                synchronized (context) {
-                    context.notify();
```
timtay-microsoft, line 87: Wait and notify pattern doesn't fit well with the context style callback approach our library has. Countdownlatch is a bit easier to use here

device/iot-device-samples/send-serialized-event/src/main/java/samples/com/microsoft/azure/sdk/iot/SendSerializedEvent.java
```diff
@@ -143,20 +173,20 @@ public static void main(String[] args)
...
-                Object lockobj = new Object();
+                CountDownLatch messageSentLatch = new CountDownLatch(1);
                 EventCallback callback = new EventCallback();
-                client.sendEventAsync(msg, callback, lockobj);
+                client.sendEventAsync(msg, callback, messageSentLatch);
 
-                synchronized (lockobj) {
-                    lockobj.wait();
-                }
+                messageSentLatch.await();
```
abhipsaMisra, line 183: the perils of using callbacks and still awaiting on them!

abhipsaMisra (approved):

PR #1064, opened 2021-01-13
[Declaration redundancy] Access static member via instance reference

abhipsaMisra:


abhipsaMisra:
This is all in unit-test code

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContextTest.java
```diff
@@ -67,11 +67,11 @@ private void generateSSLContextExpectations() throws NoSuchAlgorithmException, K
         new NonStrictExpectations()
         {
             {
-                mockedSSLContext.getInstance(anyString);
+                SSLContext.getInstance(anyString);
```
bikamani, line 70: shouldn't you be deleting Mocked instances? Is it referenced elsewhere?

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpConnectionTest.java
```diff
@@ -97,7 +97,7 @@ public void AmqpsConnectionConstructorSucceeds() throws IOException
         new NonStrictExpectations()
         {
             {
-                mockedProton.reactor((ReactorOptions)any, (Handler)any);
+                Proton.reactor((ReactorOptions)any, (Handler)any);
```
bikamani, line 100: same comment as above here and other places.

provisioning/provisioning-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/provisioning/service/auth/ProvisioningServiceSasTokenTest.java
```diff
@@ -66,12 +66,12 @@ public void constructorSucceeded() throws Exception
             final SecretKeySpec secretKeySpec;
             Mac mac;
```
bikamani, line 67: what is Mac btw? :)

provisioning/provisioning-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/provisioning/service/auth/ProvisioningServiceSasTokenTest.java
```diff
@@ -66,12 +66,12 @@ public void constructorSucceeded() throws Exception
...
             {
```
bikamani, line 68: Is this extra braces?

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContextTest.java
abhipsaMisra: the mocked ssl context is used in our tests (that is the class under test). Instance references to static members don't make sense, and that is what has been removed.

abhipsaMisra: The ssl context is not directly tested, but IotHubSslContext class, which holds a reference to ssl context. The mocked instance is used for setting expectations etc.

provisioning/provisioning-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/provisioning/service/auth/ProvisioningServiceSasTokenTest.java
abhipsaMisra: https://docs.oracle.com/javase/7/docs/api/javax/crypto/Mac.html

provisioning/provisioning-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/provisioning/service/auth/ProvisioningServiceSasTokenTest.java
abhipsaMisra: it looks like an anonymous block, possibly to limit variable scope

abhipsaMisra: https://stackoverflow.com/questions/1563030/anonymous-code-blocks-in-java

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpConnectionTest.java
abhipsaMisra: the mocked instances might be used for other assertions in these unit tests. the only instances where we were calling static methods on these instances have been reverted.

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContextTest.java
bikamani: Is it complaining about static method getInstance?

bikamani: Does line 71 make any sense after this change?

abhipsaMisra: > Is it complaining about static method getInstance?

Yes, it does, since invoking a static method on an instance does not make any sense.

> Does line 71 make any sense after this change?

You are right, it doesn't seem to help with the test expectation. Let me see if Mockit does anything special here, in that case my change will need to be replaced by a suppression instead.

provisioning/provisioning-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/provisioning/service/auth/ProvisioningServiceSasTokenTest.java
bikamani: Do you see its usage here?

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContextTest.java
bikamani: you may want to check same with other places. If this is redundant piece of code, we should refactor it.

abhipsaMisra: I actually take that back, what we are doing here is: call SSLContext.getInstance() -> which should return an SSL context instance, which we assign to as the mocked instance in L71. 
So yes, it does make sense.

azabbasi (approved):

provisioning/provisioning-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/provisioning/service/auth/ProvisioningServiceSasTokenTest.java
abhipsaMisra: I don't see any issue in the current test and refactoring the entire test doesn't seem worthwhile to me.

deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContextTest.java
bikamani: I am not sure how that's being verified though. Look at constructorCreatesSSLContext test for reference

provisioning/provisioning-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/provisioning/service/auth/ProvisioningServiceSasTokenTest.java
bikamani: It's ok if you don't want to refactor this part. It caught my attention as you modified the piece of code inside this block.

PR #1066, opened 2021-01-15
[Java 8] Anonymous type can be replaced with lambda

jamdavi:
[Java 8] Anonymous type can be replaced with lambda


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/generated/implementation/DigitalTwinsImpl.java
```diff
@@ -131,15 +126,12 @@ public Object call(ServiceResponseWithHeaders<Object, DigitalTwinGetDigitalTwinH
         }
         final String apiVersion = "2020-09-30";
         return service.getDigitalTwin(id, apiVersion)
-            .flatMap(new Func1<Response<ResponseBody>, Observable<ServiceResponseWithHeaders<Object, DigitalTwinGetDigitalTwinHeaders>>>() {
```
barustum, line 134: Other folks in the team who are more familiar with this code can confirm, but isn't this generated code? Can we change it, or do we change it elsewhere and re-generate?

device/iot-device-samples/pnp-device-sample/temperature-controller-device-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/device/TemperatureController.java
```diff
@@ -178,23 +178,19 @@ public static void main(String[] args) throws IOException, URISyntaxException, P
         maxTemperature.put(THERMOSTAT_1, 0.0d);
         maxTemperature.put(THERMOSTAT_2, 0.0d);
 
-        new Thread(new Runnable() {
-            @SneakyThrows({InterruptedException.class, IOException.class})
```
abhipsaMisra, line 182: q - were these "throws" declarations redundant, given that our `main` already declares that it throws these two exceptions?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/generated/implementation/DigitalTwinsImpl.java
timtay-microsoft: Is there a way to disable analyzing generated code for warnings?

abhipsaMisra: Yeah, we wouldn't want to modify generated code. If there isn't a way to disable analyzing for specific file paths, we could add a suppression at class level.

jamdavi: Agreed, reverted.

timtay-microsoft (approved):

device/iot-device-samples/pnp-device-sample/temperature-controller-device-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/device/TemperatureController.java
jamdavi: Actually these are required. The build breaks if these are left. Will need to suppress.

PR #1069, opened 2021-01-19
[Numeric Issues]

timtay-microsoft:
Simplify some arithmetic to satisfy the warnings inspector

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpsConnection.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpsConnection.java
@@ -26,7 +26,7 @@ import java.util.concurrent.*;
 @Slf4j
 public class AmqpsConnection extends ErrorLoggingBaseHandlerWithCleanup
 {
-    private static final int MAX_WAIT_TO_OPEN_CLOSE_CONNECTION = 1*60*1000; // 1 minute timeout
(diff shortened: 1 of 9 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpsConnection.java
```diff
@@ -26,7 +26,7 @@
 @Slf4j
 public class AmqpsConnection extends ErrorLoggingBaseHandlerWithCleanup
 {
-    private static final int MAX_WAIT_TO_OPEN_CLOSE_CONNECTION = 1*60*1000; // 1 minute timeout
+    private static final int MAX_WAIT_TO_OPEN_CLOSE_CONNECTION = 60 * 1000; // 1 minute timeout
```
vinagesh, line 29: Not sure if renaming would be out of scope but other than ERROR_INJECTION_MESSAGE_EFFECT_TIMEOUT_MILLISECONDS, none of the other names include the units (milliseconds)

vinagesh (approved):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpsConnection.java
timtay-microsoft: A bit out of scope, but I'd also like to fix that later

PR #1071, opened 2021-01-20
[Java 7 warnings]

timtay-microsoft:
Collapses some multiple catch clauses that all did the same thing, and replace a try/finally with a try with resources

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
@@ -319,9 +319,8 @@ public class IotHubSSLContext
             Security.addProvider(new BouncyCastleProvider());
 
             CertificateFactory certFactory = CertificateFactory.getInstance("X.509");
-            final PemReader publicKeyCertificateReader = new PemReader(new StringReader(publicKeyCertificateString));
 
-            try
+            try (PemReader publicKeyCertificateReader = new PemReader(new StringReader(publicKeyCertificateString)))
             {
                 PemObject possiblePublicKeyCertificate;
                 while (((possiblePublicKeyCertificate = publicKeyCertificateReader.readPemObject()) != null))
@@ -345,10 +344,6 @@ public class IotHubSSLContext
                         break;
                     }
                 }
-            }
-            finally
-            {
-                publicKeyCertificateReader.close();
             }
 
             return certChain;
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/SignatureHelper.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/SignatureHelper.java
@@ -3,6 +3,8 @@
 
 package com.microsoft.azure.sdk.iot.device.auth;
 
+import lombok.extern.slf4j.Slf4j;
+
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
 import java.io.UnsupportedEncodingException;
@@ -16,6 +18,7 @@ import static org.apache.commons.codec.binary.Base64.decodeBase64;
 import static org.apache.commons.codec.binary.Base64.encodeBase64;
 
 /** Builds the authorization signature as a composition of functions. */
+@Slf4j
 public final class SignatureHelper
 {
     /**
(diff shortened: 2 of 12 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/SignatureHelper.java
```diff
@@ -80,14 +80,10 @@
             hMacSha256.init(secretKey);
             encryptedSig = hMacSha256.doFinal(sig);
         }
-        catch (NoSuchAlgorithmException e)
+        catch (NoSuchAlgorithmException | InvalidKeyException e)
         {
             // should never happen, since the algorithm is hard-coded.
```
vinagesh, line 85: Add both reasons in comment here?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
```diff
@@ -319,9 +319,8 @@ private static Key parsePrivateKey(String privateKeyString) throws CertificateEx
             Security.addProvider(new BouncyCastleProvider());
 
             CertificateFactory certFactory = CertificateFactory.getInstance("X.509");
-            final PemReader publicKeyCertificateReader = new PemReader(new StringReader(publicKeyCertificateString));
 
-            try
+            try (PemReader publicKeyCertificateReader = new PemReader(new StringReader(publicKeyCertificateString)))
```
vinagesh, line 323: Does the close happen automatically due to this new syntax?

vinagesh (approved):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
timtay-microsoft: Yep, this is kind of like a using block in .NET

abhipsaMisra: Interesting

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/SignatureHelper.java
```diff
@@ -80,13 +80,9 @@
...
         }
-        catch (NoSuchAlgorithmException e)
+        catch (NoSuchAlgorithmException | InvalidKeyException e)
         {
-            // should never happen, since the algorithm is hard-coded.
-        }
-        catch (InvalidKeyException e)
-        {
-            // should never happen, since the input key type is hard-coded.
+            // should never happen, since the algorithm and input key are hard-coded.
```
abhipsaMisra, line 85: might be worth adding a comment stating that the only reason we catch and swallow this is to avoid having to declare these exceptions in the method signature.
also, should it say "will never happen", definitively

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/SignatureHelper.java
```diff
@@ -80,13 +80,9 @@
             hMacSha256.init(secretKey);
             encryptedSig = hMacSha256.doFinal(sig);
         }
-        catch (NoSuchAlgorithmException e)
+        catch (NoSuchAlgorithmException | InvalidKeyException e)
         {
-            // should never happen, since the algorithm is hard-coded.
-        }
-        catch (InvalidKeyException e)
```
abhipsaMisra, line 87: can we say for sure that InvalidKeyException won't happen? should we throw incase it does? (since algorithm is hard-coded, NoSuchAlgorithmException might not happen; but it seems to me that InvalidKeyException might still be thrown).

abhipsaMisra (approved):
only question regarding swallowing a possible InvalidKeyException in the signature helper

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/SignatureHelper.java
timtay-microsoft: This is kind of a public API, so I'd rather not start throwing this exception. I can certainly add a log statement to help with debugging though

PR #1073, opened 2021-01-20
Fix(e2e) fix e2e test

timtay-microsoft:
Test threw an indexOutOfBounds exception because the list that it tried to get the hostname from was always empty, so the test has been refactored to get the hostname from a different list of clients that is never empty instead

Without this change, the test always fails. The only reason it got like this was because we don't run this test at the gate since it is a 15 minute test

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/TokenRenewalTests.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/TokenRenewalTests.java
@@ -107,10 +107,11 @@ public class TokenRenewalTests extends IntegrationTest
     public void tokenRenewalWorks() throws Exception
     {
         List<InternalClient> clients = createClientsToTest();
+        String hostname = clients.get(0).getConfig().getIotHubHostname();
         ArrayList<DeviceClient> amqpMultiplexedClients = new ArrayList<>();
-        MultiplexingClient amqpMultiplexingClient = createMultiplexedClientToTest(AMQPS, amqpMultiplexedClients);
+        MultiplexingClient amqpMultiplexingClient = createMultiplexedClientToTest(AMQPS, amqpMultiplexedClients, hostname);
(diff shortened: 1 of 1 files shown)


barustum (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/TokenRenewalTests.java
```diff
@@ -253,17 +254,16 @@ else if (protocol == AMQPS || protocol == AMQPS_WS)
         return clients;
     }
 
-    private MultiplexingClient createMultiplexedClientToTest(IotHubClientProtocol protocol, List<DeviceClient> createdClients) throws IotHubException, IOException, URISyntaxException, MultiplexingClientException, InterruptedException {
-        String hostName = createdClients.get(0).getConfig().getIotHubHostname();
-        MultiplexingClient multiplexingClient = new MultiplexingClient(hostName, protocol);
+    private MultiplexingClient createMultiplexedClientToTest(IotHubClientProtocol protocol, List<DeviceClient> clientsToCreate, String hostname) throws IotHubException, IOException, URISyntaxException, MultiplexingClientException, InterruptedException {
+        MultiplexingClient multiplexingClient = new MultiplexingClient(hostname, protocol);
```
bikamani, line 258: I wonder why hostname gets null in case of device client and not for internal client list?

bikamani (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/TokenRenewalTests.java
timtay-microsoft: "clientsToCreate" is by design empty when this method is called. The method's job is to add clients to the list, not to read from the list. Because of that, it didn't make sense to try to get the hostname from the empty list

abhipsaMisra (approved):

PR #1075, opened 2021-01-20
[Declaration redundancy] Actual method parameter is the same constant - tests

abhipsaMisra:
The warning has been suppressed in test classes, since the flagged params were being referenced in helper methods

### DIFF
--- a/deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/serializer/MethodParserTest.java
+++ b/deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/serializer/MethodParserTest.java
@@ -500,7 +500,8 @@ public class MethodParserTest
         methodParser.toJsonElement();
     }
 
-    private static final TestMethod createMethodRequestWithTimeout(
+    @SuppressWarnings("SameParameterValue") // Since this is a helper method, the params can be passed any value.
+    private static TestMethod createMethodRequestWithTimeout(
             String methodName,
             long responseTimeOut,
             long connectTimeout,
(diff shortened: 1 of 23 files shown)


deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/serializer/MethodParserTest.java
```diff
@@ -579,7 +583,7 @@ private static final TestMethod createMethodRequestWithConnectTimeout(
         return testMethod;
     }
 
-    private static final TestMethod createMethodResponse(Integer status, String payload, boolean wrapPayloadAsString, Object expectedPayload)
```
abhipsaMisra, line 582: no point in declaring private static methods as final, they cannot be overridden anyway

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/DeviceTwinCommon.java
```diff
@@ -512,7 +513,7 @@ public void tearDownNewDeviceAndModule()
         }
     }
 
-    protected void readReportedPropertiesAndVerify(DeviceState deviceState, String startsWithKey, String startsWithValue, int expectedReportedPropCount) throws IOException, IotHubException, InterruptedException
```
abhipsaMisra, line 515: the property key is a constant, so replaced it inline

timtay-microsoft (approved):

PR #1078, opened 2021-01-21
[Memory] Inner class may be 'static'

abhipsaMisra:
Code inspection details:
An inner class may be static if it doesn't reference its enclosing instance.
A static inner class does not keep an implicit reference to its enclosing instance. This prevents a common cause of memory leaks and uses less memory per instance of the class.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUpload.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUpload.java
@@ -27,7 +27,7 @@ public final class FileUpload
     private final HttpsTransportManager httpsTransportManager;
     private final ScheduledExecutorService taskScheduler;
     private final FileUploadStatusCallBack fileUploadStatusCallBack;
-    private static Queue<FileUploadInProgress> fileUploadInProgressesSet;
+    static Queue<FileUploadInProgress> fileUploadInProgressesSet;
 
     /**
      * CONSTRUCTOR
@@ -112,31 +112,6 @@ public final class FileUpload
         FileUploadTask fileUploadTask = new FileUploadTask(blobName, inputStream, streamLength, httpsTransportManager, fileUploadStatusCallBack, newUpload);
 
         newUpload.setTask(taskScheduler.submit(fileUploadTask));
-    }
-
-    private static final class FileUploadStatusCallBack implements IotHubEventCallback
-    {
-        @Override
-        public synchronized void execute(IotHubStatusCode status, Object context)
-        {
-            if(context instanceof FileUploadInProgress)
-            {
-                FileUploadInProgress uploadInProgress = (FileUploadInProgress) context;
-                uploadInProgress.triggerCallback(status);
-                try
-                {
-                    fileUploadInProgressesSet.remove(context);
-                }
-                catch (ClassCastException | NullPointerException | UnsupportedOperationException e)
-                {
-                    log.error("FileUploadStatusCallBack received callback for unknown FileUpload", e);
-                }
-            }
-            else
-            {
-                log.error("FileUploadStatusCallBack received callback for unknown FileUpload");
-            }
-        }
     }
 
     /**
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUpload.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUpload.java
@@ -130,7 +130,7 @@ public final class FileUpload
         newUpload.setTask(taskScheduler.submit(fileUploadTask));
     }
 
-    private final class FileUploadStatusCallBack implements IotHubEventCallback
+    private static final class FileUploadStatusCallBack implements IotHubEventCallback
     {
         @Override
         public synchronized void execute(IotHubStatusCode status, Object context)
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUpload.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUpload.java
@@ -27,7 +27,7 @@ public final class FileUpload
     private final HttpsTransportManager httpsTransportManager;
     private final ScheduledExecutorService taskScheduler;
     private final FileUploadStatusCallBack fileUploadStatusCallBack;
-    private final Queue<FileUploadInProgress> fileUploadInProgressesSet;
+    private static Queue<FileUploadInProgress> fileUploadInProgressesSet;
 
     /**
      * CONSTRUCTOR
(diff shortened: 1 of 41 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1232,7 +1232,7 @@ private TransportException singleReconnectAttempt()
     /**
      * Task for adding a packet back to the waiting queue. Used for delaying message retry
      */
-    public class MessageRetryRunnable implements Runnable
+    public static class MessageRetryRunnable implements Runnable
```
abhipsaMisra, line 1235: although this class is public, there is no scenario in which users would initialize an instance of this class and pass it to the sdk

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/HubTierConnectionTests.java
```diff
@@ -147,7 +147,7 @@ public void SetProxyIfApplicable()
         }
     }
 
-    public class HubTierConnectionTestInstance
+    public static class HubTierConnectionTestInstance
```
abhipsaMisra, line 150: test class

provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/parser/X509RegistrationResultParser.java
```diff
@@ -24,7 +24,7 @@
      * Class representing X509CertificateInfo
      * https://docs.microsoft.com/en-us/rest/api/iot-dps/RuntimeRegistration/RegisterDevice#definitions_x509certificateinfo
      */
-    public class X509CertificateInfo
+    public static class X509CertificateInfo
```
abhipsaMisra, line 27: @timtay-microsoft is this something users initialize?

timtay-microsoft: Nope. We initialize it internally for DPS, but users never make an instance of it

timtay-microsoft (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUpload.java
```diff
@@ -130,7 +130,7 @@ public synchronized void uploadToBlobAsync(
         newUpload.setTask(taskScheduler.submit(fileUploadTask));
     }
 
-    private final class FileUploadStatusCallBack implements IotHubEventCallback
+    private static final class FileUploadStatusCallBack implements IotHubEventCallback
```
abhipsaMisra, line 133: this class is mocked here and below: https://github.com/Azure/azure-iot-sdk-java/blob/master/device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTest.java#L361
which now fails (since it is now a private static inner class, and mockit is unable to mock it): 
"java.lang.IllegalArgumentException: FileUploadStatusCallBack is not an inner class"

Since these tests rely on performing an action on the mocked callback, I don't see a way to refactor it without mocking the said callback.

Suggestions are welcome.

timtay-microsoft: It may be easier to just take this inner class and put it in it's own file. If it is in its own class file, it wouldn't have to be static. That should make it easier to mock, right?

abhipsaMisra: That makes sense, sure. Let me give that a shot 👍🏽

drwill-ms: Without knowing too much about this specific test, I see our choices as:

1. Make the base class accessible to the test (or the world, if necessary).
2. Drop the test.
3. Implement what was in the base class in the test class.

Which of those aren't actual options? What other options do you see?

abhipsaMisra: The inner class under question here is an implementation of our public IoT hub event callback interface.
1. I don't really see a value in making the class public, just because it is only ever invoked by our sdk internally, and users wouldn't gain anything from having access to it. Updating the inner class to not be enclosed within a parent class is an option.
Having a public interface and a private implementation is useful when mocking classes for unit tests; however, with this being a private inner class makes it complicated (plus mockit support). Looking up online, it looks like powermock (another mocking framework) does have support for this feature though.
2. Last resort.
3. I am not sure what this means - implement the sourcecode in the test class?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/ProxiedSSLSocket.java
```diff
@@ -162,6 +162,7 @@ private void doTunnelHandshake(Socket tunnel, String host, int port) throws IOEx
     }
 
     @RequiredArgsConstructor
+    static
```
bikamani, line 165: Is it just the git showing this on new line?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/twin/DesiredPropertiesTests.java
```diff
@@ -347,6 +347,7 @@ public void testUpdateDesiredProperties() throws IOException, InterruptedExcepti
     }
 
     @AllArgsConstructor
+    static
```
bikamani, line 350: same here?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUpload.java
timtay-microsoft: For 3, this inner class is actually used in our production code, so it can't just be moved to test code

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/ProxiedSSLSocket.java
abhipsaMisra: it's not not git, it actually was on a new line! :D

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUpload.java
abhipsaMisra: So moving to its own class isn't that straightforward, since these classes rely on shared private variables

drwill-ms: I don't mean moved to test code, just copied. That said, it sounds like Tim's suggestion is best anyway.

PR #1079, opened 2021-01-21
[Compiler Error] Unchecked warnings

timtay-microsoft:
These warnings were about casting types, using Collection types without a type (List vs List<String>).

In addition, I removed an unused, non-public class "Template" that served no purpose, and fixed some spacing issues


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinCollection.java
```diff
@@ -112,7 +114,6 @@
      */
     public TwinCollection()
     {
-        /* SRS_TWIN_COLLECTION_21_001: [The constructor shall create a new instance of the super class.] */
```
jamdavi, line 115: Sweeping comment for this type of removal. Are these comments no longer important to the SDK?

jamdavi (approved):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinCollection.java
timtay-microsoft: Yeah, we've stopped using these lines to correlate from production code to unit tests

timtay-microsoft: Since they are often out of date, I've been deleting them over time so they stop cluttering up the code

PR #1080, opened 2021-01-21
[Class Structure] Field can be local Warnings

bikamani:
This PR is to address a category of "field can be local" warnings

To do: Update xml comments for setForceUpdate


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketImpl.java
```diff
@@ -21,7 +21,6 @@
 
 public class WebSocketImpl implements WebSocket, TransportLayer
 {
-    private final int _maxFrameSize = (4 * 1024) + (16 * WebSocketHeader.MED_HEADER_LENGTH_MASKED);
     private boolean _tail_closed = false;
```
vinagesh, line 24: Just curious what is the warning if something is global like this?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/SymmetricKey.java
```diff
@@ -34,6 +29,7 @@ public SymmetricKey()
     {
         try
         {
+            String encryptionMethod = "AES";
```
vinagesh, line 32: I observe a pattern where we have moved constants to local variables. Is that a warning because it has not been used more than once?

vinagesh (approved):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketImpl.java
bikamani: This particular one was used only in one place-  warning is "field can be local"

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/SymmetricKey.java
bikamani: It's both. Warning because of the usage only in one place.

timtay-microsoft: Would making the field be "private static final" also fix this warning? I also see this as opposite to how we typically handle these magic strings

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/RawTwinQuery.java
```diff
@@ -16,8 +16,6 @@
 public class RawTwinQuery
 {
     private IotHubConnectionString iotHubConnectionString = null;
-    private final long USE_DEFAULT_TIMEOUT = 0;
-    private final int DEFAULT_PAGE_SIZE = 100;
```
timtay-microsoft, line 20: Same question here. Wouldn't marking these as static resolve the issue, too?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/QueryRequestParser.java
```diff
@@ -13,9 +13,6 @@
 public class QueryRequestParser
 {
     private static final String QUERY_TAG = "query";
-    @Expose(deserialize = false)
-    @SerializedName(QUERY_TAG)
-    private String query = null;
```
timtay-microsoft, line 18: This change seems wrong to me. This is a json field declaration that is used for serialization/deserialization. I don't think you can make this a local field

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/SymmetricKey.java
bikamani: Good q: I don't know, will try it out and see. Is it more of a preference thing though?
I also thought about how we follow it currently however I saw this being used only once and made the change accordingly.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/QueryRequestParser.java
bikamani: This is not being referenced anywhere though!

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/Configuration.java
```diff
@@ -297,7 +292,9 @@ public void setForceUpdate(Boolean forceUpdate) throws IllegalArgumentException
             throw new IllegalArgumentException("forceUpdate cannot be null");
         }
 
-        this.forceUpdate = forceUpdate;
+        /**
+         * Flip-flop helper for sending a forced update
+         */
```
abhipsaMisra, line 297: this comment is unattached to any field :D
it looks like unused field "forceUpdate" has been removed, this comment should be removed as well

abhipsaMisra: also, this setter is no longer setting any field, that doesn't look right

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/SymmetricKey.java
abhipsaMisra: +1 to making this "private static final", it would be better to have those constant defined at class level

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/QueryRequestParser.java
timtay-microsoft: It is accessed via reflection when converting to and from json though, so this isn't safe to remove.

timtay-microsoft: I'd just suppress the warning on this instance

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/Configuration.java
bikamani: pretty sure removed the comment but not sure why it showed up after merge.
Wil lremove it

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/SymmetricKey.java
bikamani: This is only used at one place and since the warning. I don't prefer suppressing the warning here unless I am missing any other benefits of having it defined at class level. Otherwise I'll leave this as it is.

bikamani: private static works, thanks @timtay-microsoft

timtay-microsoft (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/BaseDevice.java
```diff
@@ -355,8 +350,6 @@ public void setForceUpdate(Boolean forceUpdate) throws IllegalArgumentException
         {
             throw new IllegalArgumentException();
         }
-
-        this.forceUpdate = forceUpdate;
```
abhipsaMisra, line 359: q - setForceUpdate now doesn't do anything other than throw an exception is the passed arg is null. Couldn't we just remove this method altogether?

abhipsaMisra: or is validation its only purpose? in that case, since we cannot rename a public method, we should probably add a comment explaining that.

bikamani: In this context, addressing the warning for forceUpdate which is not referenced/used anywhere else. For the implementation. I did my research but haven't found why this was implemented in the first place!! I will address adding comment in follow up PR.

PR #1081, opened 2021-01-21
[Control flow issues] if statements optimization

abhipsaMisra:
- Inspection that can extract common parts from branches of if statement.
- Reports if statements which can be simplified to a single assignment, return or assert statement.

### DIFF
--- a/device/iot-device-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java
+++ b/device/iot-device-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java
@@ -29,14 +29,7 @@ public class DeviceTwinSample
         @Override
         public void execute(IotHubStatusCode status, Object context)
         {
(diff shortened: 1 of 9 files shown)


device/iot-device-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java
```diff
@@ -29,14 +29,7 @@
...
         {
-            if((status == IotHubStatusCode.OK) || (status == IotHubStatusCode.OK_EMPTY))
-            {
-                Succeed.set(true);
-            }
-            else
-            {
-                Succeed.set(false);
-            }
+            Succeed.set((status == IotHubStatusCode.OK) || (status == IotHubStatusCode.OK_EMPTY));
```
vinagesh, line 32: This logic has been repeated. Could be extracted to a common helper class with something like IsIotHubSucceeded(IotHubStatusCode status).

vinagesh (approved):

timtay-microsoft (approved):

PR #1084, opened 2021-01-22
[Probable bugs] Constant conditions and exceptions

timtay-microsoft:
Removing unnecessary null checks, suppressing warnings about false positive possible null pointer cases, and fixing some incorrectly qualified json fields


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketHandlerImpl.java
```diff
@@ -214,7 +214,7 @@ else if (payloadLength == WebSocketHeader.PAYLOAD_EXTENDED_16)
                     retVal = WebSocketMessageType.WEB_SOCKET_MESSAGE_TYPE_HEADER_CHUNK;
                 }
             }
-            else if (payloadLength == WebSocketHeader.PAYLOAD_EXTENDED_64)
```
timtay-microsoft, line 217: statement was always true

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketHandlerImpl.java
```diff
@@ -245,10 +245,6 @@ else if(opcode == WebSocketHeader.OPCODE_CONTINUATION)
                 {
                     retVal = WebSocketMessageType.WEB_SOCKET_MESSAGE_TYPE_CHUNK;
                 }
-                else
-                {
-                    retVal = WebSocketMessageType.WEB_SOCKET_MESSAGE_TYPE_UNKNOWN;
```
timtay-microsoft, line 250: value was also assign before this as the default value

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceTwin/Device.java
```diff
@@ -58,47 +49,31 @@ public void setReportedProp(Property reportedProp)
...
-            for (Iterator repProperty = reportedProp.iterator(); repProperty.hasNext();)
-            {
-                repProperty.next();
-                repProperty.remove();
-            }
+            repProperty.next();
+            repProperty.remove();
         }
 
-        if (desiredProp != null)
```
timtay-microsoft, line 95: desiredProp is never null here

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceTwin/Device.java
```diff
@@ -58,47 +49,31 @@ public void setReportedProp(Property reportedProp)
...
-        **Codes_SRS_DEVICE_25_008: [**This method shall add the parameters to the map even if callback and object pair are null**]**
-         */
+
         this.desiredProp.put(desiredProp, new Pair<>(desiredPropCallBack, desiredPropCallBackContext));
     }
 
     public void clean()
     {
-        //Codes_SRS_DEVICE_34_009: [The method shall remove all the reported and desired properties set by the user so far.]
-        if (reportedProp != null)
```
timtay-microsoft, line 86: reportedProp is never null here

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceTwin/DeviceMethod.java
```diff
@@ -133,11 +133,7 @@ public IotHubMessageResult execute(Message message, Object callbackContext)
         @Override
         public void execute(IotHubStatusCode responseStatus, Object callbackContext)
         {
-            if (deviceMethodStatusCallback != null)
```
timtay-microsoft, line 136: never null

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasToken.java
```diff
@@ -168,7 +168,7 @@ else if(this.signature != null && this.expiryTime != 0L && this.scope!= null)
         }
         else
         {
-            return null;
```
timtay-microsoft, line 171: This should never be hit, so instead of returning null which would cause NPEs everywhere, throw an exception instead

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTask.java
```diff
@@ -144,7 +145,6 @@ public void run()
         catch (StorageException | IOException | IllegalArgumentException | URISyntaxException e)
         {
             log.error("File upload failed to upload the stream to the blob", e);
-            fileUploadCompletionNotification = new FileUploadCompletionNotification(sasUriResponse.getCorrelationId(), false, -1, "Failed to upload to storage.");
```
timtay-microsoft, line 147: Moving this default value assignment to earlier so that the finally block always has a non-null instance to use

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -227,7 +227,7 @@ public void onMessageSent(Message message, String deviceId, Throwable e)
                 }
             }
         }
-        else if (message != null)
```
timtay-microsoft, line 230: never null

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1592,8 +1602,13 @@ private void addToCallbackQueue(IotHubTransportPacket packet)
 
     private DeviceClientConfig getDefaultConfig()
     {
-        Iterator<DeviceClientConfig> configsIterator = this.deviceClientConfigs.values().iterator();
-        return configsIterator.hasNext() ? configsIterator.next() : null;
+        for (DeviceClientConfig config : this.deviceClientConfigs.values())
+        {
+            return config;
```
timtay-microsoft, line 1607: deliberately returning the first config. This is just a bit easier to read, I believe

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -80,12 +80,7 @@ public MqttIotHubConnection(DeviceClientConfig config) throws IllegalArgumentExc
             {
                 throw new IllegalArgumentException("deviceID cannot be null or empty.");
             }
-            if (config.getIotHubName() == null || config.getIotHubName().length() == 0)
```
timtay-microsoft, line 83: This same check is done a few lines above

device/iot-device-samples/transportclient-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/TransportClientSample.java
```diff
@@ -250,21 +250,13 @@ public static void main(String[] args)
             protocol = IotHubClientProtocol.AMQPS_WS;
         }
 
-        String pathToCertificate1 = null;
```
timtay-microsoft, line 253: These were never assigned, so I removed them

provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/contract/http/ContractAPIHttp.java
```diff
@@ -126,7 +126,7 @@ private HttpRequest prepareRequest(
         request.setHeaderField(USER_AGENT, USER_AGENT_VALUE);
         request.setHeaderField(ACCEPT, ACCEPT_VALUE);
         request.setHeaderField(CONTENT_TYPE, ACCEPT_VALUE + "; " + ACCEPT_CHARSET);
-        request.setHeaderField(CONTENT_LENGTH, payload != null ? String.valueOf(payload.length) : "0");
```
timtay-microsoft, line 129: Payload is always non null at this point since there is a null check above

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/DeviceRegistrationState.java
```diff
@@ -52,14 +52,16 @@
...
-    private final String createdDateTimeUtc = null;
+    @SuppressWarnings("unused") // used by reflection during json serialization/deserialization
+    private String createdDateTimeUtc;
     private transient Date createdDateTimeUtcDate;
 
     // last update date and time
     private static final String LAST_UPDATED_DATETIME_UTC_TAG = "lastUpdatedDateTimeUtc";
     @Expose
     @SerializedName(LAST_UPDATED_DATETIME_UTC_TAG)
-    private final String lastUpdatedDateTimeUtc = null;
```
timtay-microsoft, line 62: creating a final string that will always be null doesn't make much sense, so I removed the final qualifier so that it can be json serialized/deserialized with meaningful data.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobsResponseParser.java
```diff
@@ -20,30 +20,29 @@
  */
 public class JobsResponseParser
 {
-    @Expose(deserialize = false)
-    private static final String DATEFORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'";
-    @Expose(deserialize = false)
-    private static final String TIMEZONE = "UTC";
```
abhipsaMisra, line 26: unused?

abhipsaMisra (approved):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobsResponseParser.java
timtay-microsoft: Yep, unused. Even weirder is that they were exposed for json serialization

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/JobsResponseParser.java
```diff
@@ -20,30 +20,29 @@
...
     // Condition for device query to get devices to execute the job on
     private static final String QUERYCONDITION_TAG = "queryCondition";
     @SerializedName(QUERYCONDITION_TAG)
+    @SuppressWarnings("unused") // used by reflection during json serialization/deserialization
     private String queryCondition;
 
     // Scheduled job start time in UTC.
     private static final String CREATETIME_TAG = "createdTime";
     @SerializedName(CREATETIME_TAG)
+    @SuppressWarnings("unused") // used by reflection during json serialization/deserialization
```
bikamani, line 39: Just curious how did you validate this is unused?

timtay-microsoft: There's some logic below where we attempt to create a date time object from this string, if it was deserialized into

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasToken.java
barustum: Given that this is a public method, could this introduce a breaking change if a customer is using this method?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1548,12 +1555,15 @@ private boolean hasOperationTimedOut(long startTime)
     {
         if (startTime == 0)
         {
-            //Codes_SRS_IOTHUBTRANSPORT_34_077: [If the provided start time is 0, this function shall return false.]
             return false;
         }
 
-        //Codes_SRS_IOTHUBTRANSPORT_34_044: [This function shall return if the provided start time was long enough ago
-        // that it has passed the device operation timeout threshold.]
+        if (this.getDefaultConfig() == null)
```
barustum, line 1561: nit: maybe bundle with the previous if statement since the outcome is the same.

barustum (approved):
Just left one comment I am curious about:
https://github.com/Azure/azure-iot-sdk-java/pull/1084#discussion_r562939579

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasToken.java
timtay-microsoft: I'd chalk this class up as not public considering we never ask a user to instantiate it directly. At most, people may use this as a utility class of sorts, and I don't expect them to hit this code path even in that case

PR #1085, opened 2021-01-22
Fix maven Pom warnings

timtay-microsoft:
Removes duplicate groupId designations (when not specified, groupId is inherited), and suppresses a warning about a duplicate dependency declaration


device/iot-device-samples/pom.xml
```diff
@@ -40,6 +40,11 @@
         <module>custom-sas-token-provider-sample</module>
     </modules>
     <dependencies>
+        <!--suppress MavenDuplicateDependenciesInspection -->
+        <!--
+        This duplicate dependency declaration warning can be ignored as one child Pom.xml duplicates this dependency declaration
+        demonstrate how to exclude in order to demonstrate excluding bouncycastle dependency from the device client library.
```
abhipsaMisra, line 46: I think there is a typo here: "...duplicates this dependency declaration ~demonstrate how to exclude~ in order to demonstrate excluding bouncycastle dependency"

abhipsaMisra (approved):
So we now have the group Id defined at the parent level of each project, and everyone else under it inherits it, correct?

timtay-microsoft:
> So we now have the group Id defined at the parent level of each project, and everyone else under it inherits it, correct?

That's the idea, yep!

PR #1086, opened 2021-01-22
Fix warning about trailing whitespace in .properties files

timtay-microsoft:


abhipsaMisra (approved):
Such a complicated PR to review!

PR #1091, opened 2021-01-25
Misc warnings fix

abhipsaMisra:
This PR contains small warnings fixes that are probably a result of other warning fixes that we've done - eg. we've removed some used code block which made some import statement unnecessary.

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/TwinTags.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/TwinTags.java
@@ -35,7 +35,7 @@ public class TwinTags
     protected synchronized JsonElement update(Map<String, Object> tagsMap) throws IllegalArgumentException
     {
         JsonElement innerDiff = updateFromMap(tagsMap, tags);
-        if((innerDiff == null) || (innerDiff.toString().equals("{}")))
+        if(innerDiff.toString().equals("{}"))
         {
             return null;
         }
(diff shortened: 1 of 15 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/TwinTags.java
```diff
@@ -35,7 +35,7 @@ protected TwinTags()
     protected synchronized JsonElement update(Map<String, Object> tagsMap) throws IllegalArgumentException
     {
         JsonElement innerDiff = updateFromMap(tagsMap, tags);
-        if((innerDiff == null) || (innerDiff.toString().equals("{}")))
```
abhipsaMisra, line 38: updateFromMap never returns a null JsonElement (it returns a new or modified JsonObject)

provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/parser/ProvisioningErrorParser.java
```diff
@@ -25,6 +25,7 @@
     private String message;
 
     private static final String INFO = "info";
+    @SuppressWarnings("MismatchedQueryAndUpdateOfCollection") // "info" is populated by reflection during json deserialization
```
abhipsaMisra, line 28: this warning complained that info is never updated but only queried.

abhipsaMisra: Since this is populated during deserialization I cannot make this final

barustum (approved):

bikamani (approved):

PR #1092, opened 2021-01-25
Fix or suppress warnings for statements with empty body

abhipsaMisra:


timtay-microsoft (approved):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpReactor.java
```diff
@@ -32,7 +32,11 @@ public void run() throws HandlerException
         this.reactor.setTimeout(REACTOR_TIMEOUT);
         this.reactor.start();
 
-        while(this.reactor.process()){}
+        //noinspection StatementWithEmptyBody
+        while(this.reactor.process())
+        {
+            // The empty while loop is to ensure that reactor thread runs as long as it has messages to process
```
bikamani, line 38: I didn't quite understand, what is the resolution of warning here?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpResponseVerification.java
```diff
@@ -6,53 +6,41 @@
...
+            return;
         }
-        else if (state.getClass().equals(Accepted.class))
-        {
 
-        }
-        else if (state.getClass().equals(Rejected.class))
-        {
-            Rejected rejectedState = (Rejected)state;
+        if (state.getClass().equals(Rejected.class)) {
```
bikamani, line 31: Is this else if --> if change for some warning here?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpReactor.java
abhipsaMisra: the warning has been suppressed by using the following statement:
`//noinspection StatementWithEmptyBody`

abhipsaMisra: that is because we do want an empty while block here, to simply wait until all processing has been completed (adding some kind of log statement to circumvent the warning wouldn't be the best solution here, since we'll end up with a lot of log statements and also cause cpu overhead)

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpResponseVerification.java
abhipsaMisra: the warning was again that all these else if blocks contained empty body. 

The reason I chose to modify the if-else checks is to separate 2 cases
 -> (state == null) => assign an exception and return
-> statue is not null => perform subsequent condition checks

previously we have all the checks being performed in a single linear flow.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpReactor.java
bikamani: Agreed.

PR #1096, opened 2021-01-27
feat(iot-service): Add cbs auth support to ServiceClient, add TokenCredential constructor

timtay-microsoft:
Still need to add some javadocs and tests, but this is a good start

Opening this PR mostly to get early feedback on the new APIs, not because the code is ready to merge into master yet.


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/CbsAuthorizationType.java
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.azure.sdk.iot.deps.transport.amqp;
+
+/**
+ * An enumeration of supported authorization methods for AMQP connections that use Claims Based Security links.
+ */
+public enum CbsAuthorizationType
```
timtay-microsoft, line 9: Intentionally putting this in the deps package in case we need to re-use it for device client scenarios

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/CbsAuthorizationType.java
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Authorize with CBS through a shared access signature.
+     */
+    SHARED_ACCESS_SIGNATURE("azure-devices.net:sastoken");
```
timtay-microsoft, line 14: Currently the only token auth type Hub supports, but this is where we'd add an enum for RBAC style auth tokens later

drwill-ms (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/IotHubServiceSasToken.java
```diff
@@ -20,7 +20,7 @@
  */
 public final class IotHubServiceSasToken
 {
-    long TOKEN_VALID_SECS = 365*24*60*60;
```
timtay-microsoft, line 23: I want to call this out as a change that I believe isn't a breaking change, but I'd like to hear some opinions on this.

With CBS auth and proactive token renewal, customers who were used to having connection stay open for a long time (more than an hour) will still see the same behavior even though the SDK will be proactively renewing their connection with shorter lived tokens rather than opening the connection with one really long-lived SAS token

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
```diff
@@ -116,12 +153,11 @@ public void onConnectionBound(Event event)
                 ((TransportInternal)transport).addTransportLayer(webSocket);
             }
 
-            Sasl sasl = transport.sasl();
-            sasl.plain(this.userName, this.sasToken);
+            transport.sasl().setMechanisms("ANONYMOUS");
```
timtay-microsoft, line 156: Note that this does not mean we aren't authenticating. It just means we have a custom authorization system in place with the CBS tokens, rather than using a more AMQP standard method like "Plain"

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
```diff
@@ -125,6 +127,21 @@ public FeedbackReceiver(String hostName, String userName, String sasToken, IotHu
         // Codes_SRS_SERVICE_SDK_JAVA_FEEDBACKRECEIVER_12_003: [The constructor shall create a new instance of AmqpReceive object]
         this.amqpReceive = new AmqpReceive(hostName, userName, sasToken, iotHubServiceClientProtocol, proxyOptions, sslContext);
     }
+
+    public FeedbackReceiver(String hostName, TokenCredential authenticationTokenProvider, CbsAuthorizationType authorizationType, IotHubServiceClientProtocol iotHubServiceClientProtocol, ProxyOptions proxyOptions, SSLContext sslContext)
```
barustum, line 131: Add method docs.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -65,11 +75,41 @@ public static ServiceClient createFromConnectionString(String connectionString,
...
+            IotHubServiceSasToken sasToken = new IotHubServiceSasToken(iotHubConnectionString);
+            OffsetDateTime sasTokenExpiryOffsetDateTime = OffsetDateTime.ofInstant(Instant.ofEpochMilli(sasToken.getExpiryTimeMillis()), ZoneId.systemDefault());
+            AccessToken accessToken = new AccessToken(sasToken.toString(), sasTokenExpiryOffsetDateTime);
+            return Mono.just(accessToken);
+        };
+
+        return new ServiceClient(iotHubConnectionString.hostName, authenticationTokenProvider, CbsAuthorizationType.SHARED_ACCESS_SIGNATURE, iotHubServiceClientProtocol, options);
+    }
+
+    public static ServiceClient createFromTokenCredential(String hostName, TokenCredential authenticationTokenProvider, CbsAuthorizationType authorizationType, IotHubServiceClientProtocol iotHubServiceClientProtocol) throws IOException
```
barustum: method docs since public.

Also for the methods below

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
```diff
@@ -5,26 +5,32 @@
...
+import org.apache.qpid.proton.engine.Transport;
 import org.apache.qpid.proton.engine.impl.TransportInternal;
 import org.apache.qpid.proton.reactor.Reactor;
 
 import javax.net.ssl.SSLContext;
 import java.io.IOException;
 
 @Slf4j
-public class AmqpConnectionHandler extends ErrorLoggingBaseHandlerWithCleanup
+public abstract class AmqpConnectionHandler extends ErrorLoggingBaseHandlerWithCleanup implements CbsSessionStateCallback
```
barustum, line 33: Is this considered a breaking change? since users who instantiate a AmqpConnectionHandler can no longer do that?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/IotHubServiceSasToken.java
drwill-ms: Seems fine to me. Can you make your best case as to why it would not be fine?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/CbsAuthorizationType.java
abhipsaMisra: Doesn't the amqp connection for device client currently use cbs links?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/IotHubServiceSasToken.java
timtay-microsoft: If the cbs roken renewal logic breaks for any reason, users will see earlier disconnections than expected. I don't see it as a very likely issue, though. I plan on adding some token renewal tests for covering that scenario

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpConnectionHandler.java
timtay-microsoft: Users don't instantiate this class, so I don't see this as a breaking change, no. We create it internally and don't expose any instances to the user.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
abhipsaMisra, line 131: will TokenCredential and CbsAuthorizationType both exist for a workflow? won't TokenCredential always be for authType = rbac?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/IotHubServiceSasToken.java
abhipsaMisra: lgtm

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
timtay-microsoft: TokenCredential will allow for tokens of Symmetric keys (what we currently support) and for tokens of RBAC keys (what we will support). The important thing here is that the user tells us what type of keys their TokenCredential will generate. When we send the CBS auth message, we have to forward both the token and the type of that token so that the service knows how to interpret the token

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/IotHubServiceSasToken.java
drwill-ms: I'd say that is a good thing.

timtay-microsoft:
Closing this PR since I'm going to break this feature up into a few smaller PRs instead

PR #1098, opened 2021-01-29
Force Update methods only performs validation

bikamani:
Refactoring code, deprecating Apis taking forceUpdate as seems appropriate.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
```diff
@@ -60,41 +60,11 @@
      */
     private Boolean httpBatchSerializeAsString;
 
-    /**
-     * [Optional] Used when batching on HTTP Default: UTF-8.
-     */
```
bikamani, line 65: removing unused fields

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/BaseDevice.java
```diff
@@ -340,10 +340,12 @@ public long getCloudToDeviceMessageCount()
 
     /**
      * Setter for force update boolean
+     * @deprecated  This method currently only validates forceUpdate parameter
      *
      * @param forceUpdate - Boolean controlling if the update should be forced or not
      * @throws IllegalArgumentException if the provided argument is null
      */
+    @Deprecated
```
bikamani, line 348: Deprecating this API and others using this since it doesn't serve the purpose as per the documentation. There is nowhere in SDK we do anything special to force update. 

Since this is public, can not remove. Decided to deprecate and add a comment that explains what is expected using this.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/Configuration.java
```diff
@@ -281,10 +281,12 @@ public String getEtag()
 
     /**
      * Setter for force update boolean
+     * @deprecated  This method currently only validates forceUpdate parameter
      *
      * @param forceUpdate - Boolean controlling if the update should be forced or not
      * @throws IllegalArgumentException if the provided argument is null
      */
+    @Deprecated
```
bikamani, line 289: same as above

service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/RegistryManagerTest.java
```diff
@@ -675,57 +675,6 @@ public void updateDevice_good_case() throws Exception
...
-                device.setForceUpdate(false);
-                mockHttpRequest.setHeaderField("If-Match", "*");
-            }
-        };
-    }
-
-    // Tests_SRS_SERVICE_SDK_JAVA_REGISTRYMANAGER_12_034: [The function shall throw IllegalArgumentException if the input device is null]
-    // Assert
-    @Test (expected = IllegalArgumentException.class)
-    public void updateDeviceForce_input_null() throws Exception
```
bikamani, line 687: Removing these since we have similar tests for API not taking forceUpdate parameter.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -419,13 +419,15 @@ public Device updateDevice(Device device) throws IOException, IotHubException
 
     /**
      * Update device with forceUpdate input parameter
+     * @deprecated Please use {@link #updateDevice(Device)}
```
timtay-microsoft, line 422: I'd add an explanation here. Something like "the forceUpdate argument does nothing so this method will always behave the same as {link updateDevice(Device)}"

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -492,13 +493,15 @@ public Device updateDevice(Device device, Boolean forceUpdate) throws IOExceptio
 
     /**
      * Async wrapper for forced updateDevice() operation
+     * @deprecated Please use {@link #updateDeviceAsync(Device)}
```
timtay-microsoft, line 496: Same comment here

timtay-microsoft (approved):
Just missing an explanation in code for some of these deprecations. Otherwise looks good

abhipsaMisra (approved):

azabbasi (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
bikamani: yeah sounds good, will add that

PR #1102, opened 2021-02-02
[Code Maturity] Deprecated member is still used 

bikamani:
Most of these are referenced and used, I have added comments where it seemed necessary.
Suppressing these warnings in this PR

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
@@ -160,7 +160,7 @@ public class Message
         initialize();
 
         this.body = body.getBytes(DEFAULT_IOTHUB_MESSAGE_CHARSET);
-        this.setContentType(DEFAULT_IOTHUB_MESSAGE_CHARSET.name());
+        this.setContentTypeFinal(DEFAULT_IOTHUB_MESSAGE_CHARSET.name());
     }
 
     
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/BaseDevice.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/BaseDevice.java
@@ -345,6 +345,8 @@ public class BaseDevice
      * @param forceUpdate - Boolean controlling if the update should be forced or not
      * @throws IllegalArgumentException if the provided argument is null
      */
+    // Supressing this warning since this method is used by public APIs in Registry Manager
+    @SuppressWarnings("DeprecatedIsStillUsed")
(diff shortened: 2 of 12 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
```diff
@@ -160,7 +160,7 @@ public Message(String body)
         initialize();
 
         this.body = body.getBytes(DEFAULT_IOTHUB_MESSAGE_CHARSET);
-        this.setContentType(DEFAULT_IOTHUB_MESSAGE_CHARSET.name());
+        this.setContentTypeFinal(DEFAULT_IOTHUB_MESSAGE_CHARSET.name());
```
bikamani, line 163: setContentType is deprecated method, this seemed to be a fine change. Let me know otherwise.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/BaseDevice.java
```diff
@@ -345,6 +345,8 @@ public long getCloudToDeviceMessageCount()
      * @param forceUpdate - Boolean controlling if the update should be forced or not
      * @throws IllegalArgumentException if the provided argument is null
      */
+    // Supressing this warning since this method is used by public APIs in Registry Manager
+    @SuppressWarnings("DeprecatedIsStillUsed")
```
azabbasi, line 349: Are those APIs also deprecated?

azabbasi: I guess what I am asking is why is the property deprecated if the methods that use it are not ! (if that is the case)

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/Message.java
timtay-microsoft: Yep, this is a safe change

timtay-microsoft (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/BaseDevice.java
bikamani: method itself is deprecated here.

bikamani: Does that answer your question?

bikamani: Talked offline.

PR #1103, opened 2021-02-02
Add logging to both retry policies and the RetryDecision class

jamdavi:
Mirroring the effort done in [Add a log statement to indicate a NoRetry retry policy was enabled](https://github.com/Azure/azure-iot-sdk-csharp/pull/1763) for the csharp SDK.

The ask from the CSS team is as below:
In certain situations customers raise support issues stating that the client does not recover on disconnection, only to discover on inspection of customer code that they have explicitly disabled the inbuilt retry policy in the client.
In order to avoid long delays in identifying the issue, the CSS team wants us to log a statement in case the retry policy is disabled.

This PR adds logging to the following classes `NoRetry`, `ExponentialBackoffWithJitter`, and `RetryDecision` to indicate the creation of the policy or the current decision.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/NoRetry.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/NoRetry.java
@@ -17,7 +17,7 @@ import lombok.extern.slf4j.Slf4j;
 public class NoRetry implements RetryPolicy
 {
     public NoRetry() {
-        log.info("NOTE: A new instance of {} has been created, the client will not perform any retries on disconnect.", NoRetry.class.getName());
+        log.info("NOTE: A new instance of NoRetry has been created, the client will not perform any retries on disconnect.");
     }
     /**
      * Always says to not retry.
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/NoRetry.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/NoRetry.java
@@ -17,7 +17,7 @@ import lombok.extern.slf4j.Slf4j;
 public class NoRetry implements RetryPolicy
 {
     public NoRetry() {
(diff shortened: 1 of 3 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/NoRetry.java
```diff
@@ -8,12 +8,17 @@
...
+import lombok.extern.slf4j.Slf4j;
 
 /**
   * Represents a retry policy that performs no retries.
   */
+@Slf4j
 public class NoRetry implements RetryPolicy
 {
+    public NoRetry() {
+        log.info("NOTE: A new instance of {} has been created, the client will not perform any retries on disconnect.", NoRetry.class.getName());
```
timtay-microsoft, line 20: Since the name of these classes shouldn't change, you should hard code the name rather than get it through reflection each time

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/RetryDecision.java
```diff
@@ -26,6 +29,13 @@ public RetryDecision(boolean shouldRetry, long duration)
         // Codes_SRS_RETRYDECISION_28_001: [The constructor shall save the duration and getRetryDecision]
         this.duration = duration;
         this.shouldRetry = shouldRetry;
+
+        // When a retry policy specifies the descision we should log what the new timing was, this will help with backoff debugging
+        if (!shouldRetry){
+            log.info("NOTE: A new instance of {} has been created with retry disabled, the client will not perform any retries.", RetryDecision.class.getName());
+        } else {
+            log.info("NOTE: A new instance of {} has been created with retry enabled, the client will perform retries every {} milliseconds.", RetryDecision.class.getName(), duration);
```
timtay-microsoft, line 37: > the client will perform retries every {} milliseconds.

This retry decision is only for one retry, to is should be "the client will retry after {} milliseconds"

timtay-microsoft (approved):

abhipsaMisra (approved):

barustum (approved):

PR #1112, opened 2021-02-05
feat(iot-service): Add AzureSasCredential and TokenCredential support to RegistryManager

timtay-microsoft:


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -59,15 +66,20 @@ public RegistryManager()
                 .httpConnectTimeout(RegistryManagerOptions.DEFAULT_HTTP_CONNECT_TIMEOUT_MS)
                 .httpReadTimeout(RegistryManagerOptions.DEFAULT_HTTP_READ_TIMEOUT_MS)
                 .build();
+        hostName = ""; // only so that hostName can be declared final
```
drwill-ms, line 69: Is `final` the same as `readonly` in C#?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -128,7 +237,6 @@ public void open()
      */
     public void close()
     {
-        // Codes_SRS_SERVICE_SDK_JAVA_REGISTRYMANAGER_34_087: [The function shall tell this object's executor service to shutdown]
```
drwill-ms, line 131: Why don't we like these comments?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
timtay-microsoft: No, it just means that the field will always be assigned a value at constructor time. This constructor is a bit of a special case because we never intended for it to exist. Because of that, assigning an empty hostname is fine here

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
timtay-microsoft: They clutter up the code base a lot

timtay-microsoft: Also because many of them are out of date. We haven't followed this convention of commenting in years.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -59,15 +66,20 @@ public RegistryManager()
...
     /**
      * Static constructor to create instance from connection string
      *
      * @param connectionString The iot hub connection string
      * @return The instance of RegistryManager
-     * @throws IOException This exception is thrown if the object creation failed
+     * @deprecated because this method declares a thrown IOException even though it never throws an IOException. Users
+     * are recommended to use {@link #RegistryManager(String)} instead since it does not declare this exception even
+     * though it constructs the same RegistryManager.
+     * @throws IOException This exception is never thrown.
```
bikamani, line 80: So the reason why we are deprecating this is because it never throws an exception as noted in documentation?

barustum (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
barustum: I believe Tim investigated, and an IOException is never thrown. Maybe it did at some point, but that's no longer the case.

timtay-microsoft: Yep. I created a non-static constructor that takes the same arguments which will replace this static constructor

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -99,19 +143,84 @@ public static RegistryManager createFromConnectionString(String connectionString
...
+
+    /**
+     * Create a new RegistryManager instance.
+     *
+     * @param hostName The hostname of your IoT Hub instance (For instance, "your-iot-hub.azure-devices.net")
+     * @param azureSasCredential The SAS token provider that will be used for authentication.
+     * @param options The connection options to use when connecting to the service.
+     * @return The instance of RegistryManager
+     */
+    public RegistryManager(String hostName, AzureSasCredential azureSasCredential, RegistryManagerOptions options)
```
vinagesh, line 211: So many overloads in java :D

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -1056,30 +1064,22 @@ public JobProperties getJob(String jobId) throws IllegalArgumentException, IOExc
...
         }
 
-        // Codes_SRS_SERVICE_SDK_JAVA_REGISTRYMANAGER_28_002: [The function shall deserialize the given module object to Json string]
         String moduleJson = module.toDeviceParser().toJson();
 
-        // Codes_SRS_SERVICE_SDK_JAVA_REGISTRYMANAGER_28_003: [The function shall get the URL for the module]
-        URL url = iotHubConnectionString.getUrlModule(module.getDeviceId(), module.getId());
-        // Codes_SRS_SERVICE_SDK_JAVA_REGISTRYMANAGER_28_004: [The function shall create a new SAS token for the module]
-        String sasTokenString = new IotHubServiceSasToken(this.iotHubConnectionString).toString();
+        URL url = IotHubConnectionString.getUrlModule(this.hostName, module.getDeviceId(), module.getId());
```
vinagesh, line 1074: This looks like repeated code pattern. Can we extract to a method like CreatHttpRequest()?

vinagesh: Scratch that. It is not as repeated as I thought. Different Http requests and urls.

vinagesh: However, it could all be done in a single line. But leave it to your preference.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -1638,11 +1571,29 @@ private HttpRequest CreateRequest(URL url, HttpMethod method, byte[] payload, St
...
     }
+
+    private String getAuthenticationToken()
+    {
+        // Three different constructor types for this class, and each type provides either a TokenCredential implementation,
+        // an AzureSasCredential instance, or just the connection string. The sas token can be retrieved from the non-null
+        // one of the three options.
+        if (this.credential != null)
+        {
+            return this.credential.getToken(new TokenRequestContext()).block().getToken();
```
vinagesh, line 1590: For this case, you will have to append - "bearer" to the token value and add the expiry of the token.
https://github.com/Azure/azure-sdk-for-net/blob/8970711b5c82c9fac3f7135529ab57de5a294f08/sdk/core/Azure.Core/src/Pipeline/BearerTokenAuthenticationPolicy.cs#L218
It is also worth sending an email to the service team to confirm if this is what they expect. I have not gotten this far in code so had not sent an email.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
azabbasi: isn't that what readonly means in c#?
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/readonly

drwill-ms: Got it. I think the comment could be updated to explain that a bit more.

`// we don't use hostname in this ctor, so assign it an empty value to satisfy the field being marked final`

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
drwill-ms: You think they offer no value to future readers?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
timtay-microsoft: Haha, sadly

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
timtay-microsoft: The particular URL to use changes per method, but I can factor out the next line since every request should have an authentication token

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -99,19 +143,84 @@ public static RegistryManager createFromConnectionString(String connectionString
             throw new IllegalArgumentException("RegistryManagerOptions cannot be null for this constructor");
         }
 
-        // Codes_SRS_SERVICE_SDK_JAVA_REGISTRYMANAGER_12_002: [The constructor shall create an IotHubConnectionString object from the given connection string]
-        IotHubConnectionString iotHubConnectionString = IotHubConnectionStringBuilder.createConnectionString(connectionString);
+        this.iotHubConnectionString =
+                IotHubConnectionStringBuilder.createIotHubConnectionString(connectionString);
```
bikamani, line 147: Nit: I see some extra spaces here

timtay-microsoft: Intellij leaves two tabs when indenting like this, so I don't see them as extra

vinagesh (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
drwill-ms: See, I'm not alone. Why does IntelliJ get to set a new standard? :p

PR #1114, opened 2021-02-05
feat(iot-service): Add TokenCredential and AzureSasCredential support to digital twin clients

timtay-microsoft:


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/DigitalTwinAsyncClient.java
```diff
@@ -39,30 +45,88 @@
...
-    adapter.serializer().registerModule(stringModule);
-    RestClient simpleRestClient = new RestClient.Builder()
-        .withBaseUrl(httpsEndpoint)
-        .withCredentials(new ServiceClientCredentialsProvider(sasTokenProvider))
-        .withResponseBuilderFactory(new ServiceResponseBuilder.Factory())
-        .withSerializerAdapter(adapter)
-        .build();
-
-    IotHubGatewayServiceAPIsImpl protocolLayerClient = new IotHubGatewayServiceAPIsImpl(simpleRestClient);
-    _protocolLayer = new DigitalTwinsImpl(simpleRestClient.retrofit(), protocolLayerClient);
```
timtay-microsoft, line 59: The indentation of this code was a bit off

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/authentication/ServiceClientBearerTokenCredentialProvider.java
```diff
@@ -0,0 +1,36 @@
...
+import lombok.NonNull;
+import okhttp3.Interceptor;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+
+/**
+ * Implementation of {@link ServiceClientCredentials} that provides RBAC based bearer tokens.
+ */
+@AllArgsConstructor
+public class ServiceClientBearerTokenCredentialProvider implements ServiceClientCredentials {
```
timtay-microsoft, line 17: It's worth noting that we have a class ```ServiceClientCredentialsProvider``` that looks about the same, but that uses a SasTokenProvider rather than a BearerTokenProvider to get the authentication tokens.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/authentication/BearerTokenProvider.java
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package com.microsoft.azure.sdk.iot.service.digitaltwin.authentication;
+
+/**
+ * Interface for providing a Bearer authentication token for use in RBAC based authentication.
+ */
+public interface BearerTokenProvider {
```
timtay-microsoft, line 9: This new interface is basically the same as the existing ```SasTokenProvider``` but the token retrieval method in this interface specifies the token is a bearer token rather than a SAS token

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/DigitalTwinAsyncClient.java
```diff
@@ -39,30 +45,88 @@
...
+     */
+    public DigitalTwinAsyncClient(String hostName, TokenCredential credential) {
+        final SimpleModule stringModule = new SimpleModule("String Serializer");
+        stringModule.addSerializer(new DigitalTwinStringSerializer(String.class, objectMapper));
+        BearerTokenProvider bearerTokenProvider = () -> credential.getToken(new TokenRequestContext()).block().getToken();
+
+        JacksonAdapter adapter = new JacksonAdapter();
+        adapter.serializer().registerModule(stringModule);
+        RestClient simpleRestClient = new RestClient.Builder()
+                .withBaseUrl("https://" + hostName) //hostname is only "my-iot-hub.azure-devices.net" so we need to add "https://"
```
azabbasi, line 90: is this indented correctly? seems to be two tabs

azabbasi: same below

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/DigitalTwinAsyncClient.java
```diff
@@ -39,30 +45,88 @@
...
+            .build();
+
+        IotHubGatewayServiceAPIsImpl protocolLayerClient = new IotHubGatewayServiceAPIsImpl(simpleRestClient);
+        _protocolLayer = new DigitalTwinsImpl(simpleRestClient.retrofit(), protocolLayerClient);
+    }
+
+    /**
+     * Creates an implementation instance of {@link DigitalTwins} that is used to invoke the Digital Twin features
+     *
+     * @param hostName The hostname of your IoT Hub instance (For instance, "your-iot-hub.azure-devices.net")
```
azabbasi, line 77: does it have to have the `https://` prefix in there? I know that ADT client doesn't like it when the prefix is missing.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/authentication/BearerTokenProvider.java
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package com.microsoft.azure.sdk.iot.service.digitaltwin.authentication;
+
+/**
+ * Interface for providing a Bearer authentication token for use in RBAC based authentication.
```
azabbasi, line 7: nit: `Bearer Authentication Token` or `bearer authentication token`

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/DigitalTwinAsyncClient.java
timtay-microsoft: In ADT land, we ask for the "endpoint", not the hostname, which is why we require the "https://" prefix in ADT. For these clients, I opted to ask for the hostname instead, for consistency with the other service clients in this SDK. Since the HTTP pipeline we use in the digital twin clients still needs an endpoint rather than a hostname, I've made it so the SDK pre-pends the hostname with "https://" in order to turn the given hostname into an endpoint

azabbasi (approved):

PR #1116, opened 2021-02-08
feat(iot-service): Add TokenCredential and AzureSasCredential support to JobClient

timtay-microsoft:


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/Job.java
```diff
@@ -22,59 +23,27 @@
     private String jobId;
     private JobClient jobClient;
 
-    /**
-     * CONSTRUCTOR
-     *
-     * @param connectionString is the IoTHub connection string.
-     * @throws IOException This exception is thrown if the object creation failed
-     */
-    Job(String connectionString) throws IOException
```
timtay-microsoft, line 31: This class is not the actual JobClient, for reference. There is a class called "JobClient" that is the more interesting one.

timtay-microsoft: This class is mostly used as a return type from existing APIs like "twinClient.scheduleTwinJob(...)"

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobClient.java
```diff
@@ -25,37 +32,122 @@
  */
 public class JobClient
 {
-    private final static long USE_DEFAULT_TIMEOUT = 0L;
-    private final static long MAX_TIMEOUT = Integer.MAX_VALUE - 24000;
```
timtay-microsoft, line 29: Unused fields

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceMethod.java
```diff
@@ -348,7 +348,19 @@ public Job scheduleDeviceMethod(
             throw new IllegalArgumentException("negative maxExecutionTimeInSeconds");
         }
 
-        Job job = new Job(this.getAuthenticationToken());
+        Job job;
+        if (this.credential != null)
+        {
+            job = new Job(this.hostName, this.credential);
```
azabbasi, line 354: is this the TokenCredential type? can we name it as such?

timtay-microsoft: From previous discussions with Sindhu and David, we agreed to call the TokenCredential argument "credential" to match the naming within the azure identity library itself

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobClient.java
```diff
@@ -25,37 +32,122 @@
...
     private final static Integer DEFAULT_PAGE_SIZE = 100;
 
     private final static byte[] EMPTY_JSON = "{}".getBytes();
 
-    private IotHubConnectionString iotHubConnectionString = null;
+    private String hostName;
+    private TokenCredential credential;
```
azabbasi, line 40: just for clarity, let's call it `tokenCredential`

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobClient.java
```diff
@@ -25,37 +32,122 @@
...
+    }
 
-        /* Codes_SRS_JOBCLIENT_21_003: [The constructor shall create a new DeviceMethod instance and return it.] */
-        JobClient jobClient = new JobClient();
+    /**
+     * Create a new JobClient instance.
+     *
+     * @param hostName The hostname of your IoT Hub instance (For instance, "your-iot-hub.azure-devices.net")
+     * @param credential The custom {@link TokenCredential} that will provide authentication tokens to
+     *                                    this library when they are needed. The provided tokens must be Json Web Tokens.
```
azabbasi, line 96: is this normal to have so much space before the next line text begins?

timtay-microsoft: It is not, I'll fix it

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobClient.java
```diff
@@ -434,21 +580,52 @@ public synchronized JobResult getNextJob(Query query) throws IOException, IotHub
...
         Query jobResponseQuery = new Query(pageSize, QueryType.JOB_RESPONSE);
 
-        //Codes_SRS_JOBCLIENT_25_045: [The queryDeviceJob shall send a query request on the query object using Query URL, HTTP GET method and wait for the response by calling sendQueryRequest.]
         String jobTypeString = (jobType == null) ? null : jobType.toString();
         String jobStatusString = (jobStatus == null) ? null : jobStatus.toString();
-        jobResponseQuery.sendQueryRequest(iotHubConnectionString, iotHubConnectionString.getUrlQuery(jobTypeString, jobStatusString), HttpMethod.GET, MAX_TIMEOUT);
+        ProxyOptions proxyOptions = options.getProxyOptions();
+        Proxy proxy = proxyOptions != null ? proxyOptions.getProxy() : null;
+
+        if (this.credential != null)
```
azabbasi, line 603: curious, can this logic live in the Query class instead?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobClientOptions.java
```diff
@@ -0,0 +1,42 @@
...
+    private final ProxyOptions proxyOptions;
+
+    /**
+     * The http read timeout to a specified timeout, in milliseconds. A non-zero value specifies the timeout when reading from
+     * Input stream after a connection is established to a resource. If the timeout expires before there is data available
+     * for read, a java.net.SocketTimeoutException is raised. A timeout of zero is interpreted as an infinite timeout.
+     * By default, this value is {@link #DEFAULT_HTTP_READ_TIMEOUT_MS}. Must be a non-negative value.
+     */
+    @Getter
+    private final int httpReadTimeout;
```
azabbasi, line 32: Do we have the option of using `Duration` or `Period` instead? I am not sure if they are supported in Java 7 though ... I like using Timespan in .NET instead of having an integer with no unit indication ... I am just thinking out loud

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobClient.java
azabbasi: resolved by another comment thread.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobClientOptions.java
timtay-microsoft: Thankfully the service client has always been Java 8+. Despite that, the HTTP library that we use takes these values as ints, so it feels appropriate for us to ask for them as ints in order to avoid any conversion problems.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobClient.java
timtay-microsoft: By passing in all three of the possible authentication providers and letting the query class figure it out? Let me try it out and see what that saves us here

azabbasi: yes, I saw the same sort of logic a couple of times in this PR.

azabbasi: just something to think about, it might not actually pan out.

azabbasi (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/Job.java
```diff
@@ -22,59 +23,27 @@
...
+    Job(String connectionString)
     {
         if ((connectionString == null) || connectionString.isEmpty())
         {
-            /* Codes_SRS_JOB_21_001: [The constructor shall throw IllegalArgumentException if the input string is null or empty.] */
             throw new IllegalArgumentException("Connection string cannot be null or empty");
         }
 
-        /* Codes_SRS_JOB_21_003: [If no jobId is provided, the constructor shall generate a unique jobId to identify the Job in the Iothub.] */
```
vinagesh, line 39: This makes it sound like jobId should be a input parameter too. But looks like in all Job constructors, we don't take it as input. Do we need that as an option for customers to send?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/Job.java
```diff
@@ -87,35 +56,32 @@
...
+        DeviceTwinDevice updateTwin,
+        Date startTimeUtc,
+        long maxExecutionTimeInSeconds)
+        throws IOException, IotHubException
     {
-        /* Codes_SRS_JOB_21_006: [If the updateTwin is null, the scheduleUpdateTwin shall throws IllegalArgumentException.] */
-        if(updateTwin == null)
+        if (updateTwin == null)
         {
             throw new IllegalArgumentException("null updateTwin");
```
vinagesh, line 68: I know this already exists but I happened to notice the language on all these IllegalArgumentException is not a full sentence like elsewhere. Does it sound reasonable to update it?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/Job.java
```diff
@@ -133,38 +99,43 @@ void scheduleUpdateTwin(String queryCondition,
...
+        Object payload,
+        Date startTimeUtc,
+        long maxExecutionTimeInSeconds)
+        throws IOException, IotHubException
     {
-        /* Codes_SRS_JOB_21_012: [If the methodName is null or empty, the scheduleDeviceMethod shall throws IllegalArgumentException.] */
-        if((methodName == null) || methodName.isEmpty())
+        if (Tools.isNullOrEmpty(methodName))
         {
             throw new IllegalArgumentException("null updateTwin");
```
vinagesh, line 114: same language observation here and a few exceptions in the lines below.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobClient.java
```diff
@@ -25,37 +32,122 @@
...
+    /**
+     * Constructor to create instance from connection string
+     *
+     * @param connectionString The iot hub connection string
+     * @param options The connection options to use when connecting to the service.
+     * @return The instance of JobClient
+     */
+    public JobClient(String connectionString, JobClientOptions options)
+    {
+        this.iotHubConnectionString = IotHubConnectionStringBuilder.createIotHubConnectionString(connectionString);
```
vinagesh, line 86: null/empty check for connectionstring and options?

vinagesh (approved):

vinagesh:
LGTM. We just need to update the logic of returning the bearer token in the format confirmed by the service team.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/serviceclient/JobClientTests.java
```diff
@@ -123,14 +133,14 @@ private JobResult queryJobResponseResult(String jobId, JobType jobType, JobStatu
...
             jobResult = jobClient.getNextJob(query);
-            if(jobResult.getJobId().equals(jobId) &&
-                    (jobResult.getJobType() == jobType) &&
-                    (jobResult.getJobStatus() == jobStatus))
+            if (jobResult.getJobId().equals(jobId) &&
+                (jobResult.getJobType() == jobType) &&
+                (jobResult.getJobStatus() == jobStatus))
             {
-                System.out.println("Iothub confirmed " + jobId + " " + jobStatus + " for " + jobType);
+                log.info("Iothub confirmed {} {} for ", jobId, jobStatus, jobType);
```
barustum, line 143: missing a pair of curly braces at the end of this string.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/serviceclient/JobClientTests.java
```diff
@@ -253,16 +262,16 @@ public void scheduleUpdateTwinSucceed() throws IOException, IotHubException, Int
...
 
         // asserts for the service side.
-        if(jobExceptions.size() != 0)
+        if (jobExceptions.size() != 0)
         {
-            for (Map.Entry<String, Exception> jobException: jobExceptions.entrySet())
+            for (Map.Entry<String, Exception> jobException : jobExceptions.entrySet())
             {
-                System.out.println(jobException.getKey() + " throws " + jobException.getValue().getMessage());
+                log.error("{} threw", jobException.getKey(), jobException.getValue());
```
barustum, line 269: missing a pair of curly braces at the end of this string.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/serviceclient/JobClientTests.java
barustum: ... or are these parameters for log.info?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/Job.java
timtay-microsoft: Customers don't create instances of ```Job```, so we never need to take it as an input. We only create a value for it from a GUID so that when we schedule a twin update or a method invocation, we have a way to track that job by Id in later calls to jobClient.getJob(...).

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/Job.java
timtay-microsoft: Sure, now is a good a time as any

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobClient.java
timtay-microsoft: This will help simplify our code, so sure I can do this

barustum (approved):

PR #1117, opened 2021-02-10
refactor(samples): Add RBAC and AzureSasCredential samples for service clients

timtay-microsoft:


service/iot-service-samples/azure-sas-credential-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/AzureSasCredentialSample.java
```diff
@@ -0,0 +1,354 @@
...
+     * will make this token only usable when creating/updating/deleting device identities.
+     * For more examples, see <a href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#use-security-tokens-from-service-components">Using security tokens from service components</a>
+     * @param key The shared access key for your IoT Hub.
+     * @return The shared access signature. This value can be used as is to build a {@link AzureSasCredential} instance
+     * @throws UnsupportedEncodingException If UTF_8 is not a supported encoding on your device.
+     * @throws NoSuchAlgorithmException If HmacSHA256 algorithm isn't found.
+     * @throws InvalidKeyException If initializing the HmacSHA256 SHA fails.
+     * @see <a href="docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security">Control access to IoT Hub</a>
+     */
+    @SuppressWarnings("unused")
```
timtay-microsoft, line 114: This method is for reference purposes, because people who are interested in this constructor will also be interested in learning how to create a shared access signature

service/iot-service-samples/azure-sas-credential-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/AzureSasCredentialSample.java
```diff
@@ -0,0 +1,359 @@
...
+        runTwinClientSample(iotHubHostName, credential, newDeviceId);
+
+        runServiceClientSample(iotHubHostName, credential, newDeviceId);
+
+        // The AzureSasCredential object supports updating the shared access signature which is important to do
+        // once the previous signature has expired, or is close to expiring. This sample assumes that no renewal will
+        // be necessary since the sample completes in a few seconds. After updating the credential object, any
+        // clients that were constructed with that credential will start using the updated signature
+
+        //String updatedSignature = "some new shared access signature";
```
barustum, line 88: did you mean to leave these commented snippets?

timtay-microsoft: Yes. This sample doesn't require the user to provide a second shared access signature, but I did want to leave this code in to draw the reader's attention to this feature of AzureSasCredential

service/iot-service-samples/azure-sas-credential-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/AzureSasCredentialSample.java
```diff
@@ -0,0 +1,362 @@
...
+        // without giving them the connection string by giving them shared access signatures instead.
+
+        // This sample includes a helper function that shows how to build shared access signatures below.
+        // Search for "generateSharedAccessSignature"
+        //
+        // More details on how to build this shared access signature can be found here:
+        // https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#security-token-structure
+        AzureSasCredential credential = new AzureSasCredential(sharedAccessSignature);
+
+        // "my-azure-iot-hub.azure-devices.net" for example
```
vinagesh, line 77: These examples in comments are great :)

abhipsaMisra (approved):
lgtm

service/iot-service-samples/azure-sas-credential-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/AzureSasCredentialSample.java
vinagesh: If you want to draw attention to it, I suggest adding some comments. It is not very clear otherwise. However, I thing what might be more beneficial is showing how they can update it. In this case the renewal is left to the customer.

service/iot-service-samples/azure-sas-credential-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/AzureSasCredentialSample.java
vinagesh: Can you add that as a comment in code?

service/iot-service-samples/azure-sas-credential-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/AzureSasCredentialSample.java
```diff
@@ -0,0 +1,362 @@
...
+            }
+        }
+        catch (IotHubException | IOException e)
+        {
+            log.error("Failed to query the jobs for your IoT Hub", e);
+            System.exit(-1);
+        }
+    }
+
+    private static void runDeviceMethodClientSample(String iotHubHostName, AzureSasCredential credential, String deviceId)
```
vinagesh, line 324: I know we want to show usage of all clients here. But I feel the samples are really long this way. We should brainstorm about this a little bit. 1) Do we really need to call all clients as the only thing we want to show is how to construct and pass these credentials 2) If we want to add samples for all clients, should we make it as an option in the existing samples. That way customers can see there are two ways to authenticate based on the APIs they want to use rather than come to a sample specific to token auth.

service/iot-service-samples/azure-sas-credential-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/AzureSasCredentialSample.java
timtay-microsoft: I did add that comment in the javadocs for this method

service/iot-service-samples/azure-sas-credential-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/AzureSasCredentialSample.java
timtay-microsoft: I agree that these samples are a bit big when all they really need to show off is the constructors. I'm not sure I like making the existing samples configurable to use two different authentication types though. That would be a lot of work, and I don't think it would help the samples stay focused. 

We could potentially split the samples I created in this PR into 1 sample per client to keep them brief. I feel there would be a lot of duplicate code to make this happen, so I'm a bit hesitant to do that.

We could also just remove the getTwin/invokeMethod/createDevice calls so that the sample simply calls the new constructors. I like that style for users who are just reading the samples, but when we run the sample, there would be no validation that the user actually provided the correct credentials until we actually make a service API call like getTwin/invokeMethod/etc.

vinagesh: I agree we need to show some API call. What if you skip all clients and just show the sample with one? I feel very lengthy samples get confusing and is usually not recommended.

service/iot-service-samples/azure-sas-credential-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/AzureSasCredentialSample.java
vinagesh: They are! I somehow did not see them.

Issue #1032, opened 2020-12-23
Android - Provisioning service fails when minify enabled.


devzeze:
# Description of the issue:
Its possible to build the app with latest iot-device-client and provisioning service with minify enabled however when running provisioning it fails at runtime.
Note that device client is able to register.

# Code sample exhibiting the issue:

    plugins {
        id 'com.android.application'
    }
    android {
        compileSdkVersion 30
        buildToolsVersion "30.0.0"

    defaultConfig {
        applicationId "com.demo.iottwin"
        minSdkVersion 26
        targetSdkVersion 30
        versionCode 1
        versionName "1.0"

        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"

        multiDexEnabled true
    }

    buildTypes {
        debug {
            minifyEnabled true
            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
        }
        release {
            minifyEnabled true
            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
        }
    }
    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }

    dependencies {

    implementation 'androidx.appcompat:appcompat:1.2.0'
    implementation 'com.google.android.material:material:1.2.1'
    implementation 'androidx.constraintlayout:constraintlayout:2.0.4'
    testImplementation 'junit:junit:4.+'
    androidTestImplementation 'androidx.test.ext:junit:1.1.2'
    androidTestImplementation 'androidx.test.espresso:espresso-core:3.3.0'

    //  IoT
    implementation('com.microsoft.azure.sdk.iot:iot-device-client:1.28.0') {
        exclude module: 'azure-storage'
    }
    implementation 'com.microsoft.azure.sdk.iot.provisioning:provisioning-device-client:1.8.5'
    implementation 'com.microsoft.azure.sdk.iot.provisioning.security:x509-provider:1.1.4'
}

# Console log of the issue:

    2020-12-23 13:56:04.031 17439-17439/com.demo.iottwin E/iotdemo: iotregistering IotHubConnectionStatus CONNECTED
    2020-12-23 13:56:04.941 17439-17439/com.demo.iottwin E/iotdemo: iotprovisioning provisioningDeviceClientStatus: PROVISIONING_DEVICE_STATUS_ERROR
[...]

PR #1119, opened 2021-02-11
Add proguard to the android gate

timtay-microsoft:
[Proguard](https://www.guardsquare.com/en/products/proguard) is an android specific minifying/obfuscation tool that some users apply to their code. These changes will make it so that our android e2e test runs will be done on minified versions of our SDK so that users don't run into issues like #1032


device/iot-device-samples/android-sample/app/proguard-rules.pro
```diff
@@ -15,3 +15,26 @@
...
+#-keep class com.microsoft.azure.sdk.iot.service.ImportMode { *; }
+#-keep class com.microsoft.azure.sdk.iot.service.AuthenticationMechanism { *; }
+
+# These are the recommended exclusions when using the Device Provisioning Service device client SDK.
+# They are commented out because this sample only uses the IoT Hub device client SDK
+#-keep class com.microsoft.azure.sdk.iot.provisioning.device.internal.parser.* { *; }
+
+# These are the recommended exclusions when using the Device Provisioning Service service client SDK.
+# They are commented out because this sample only uses the IoT Hub device client SDK
+#-keep class com.microsoft.azure.sdk.iot.provisioning.service.configs.* { *; }
```
timtay-microsoft, line 40: These "-keep class x.y.z" lines just mean that proguard should deliberately **not** minify the class or package in question

device/iot-device-samples/android-sample/build.gradle
```diff
@@ -6,11 +6,15 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.4'
+        classpath 'com.android.tools.build:gradle:4.1.2'
```
timtay-microsoft, line 9: Out gradle versions were a bit out of date, so now seems like an appropriate time to bump them

device/iot-device-samples/android-sample/app/proguard-rules.pro
```diff
@@ -15,3 +15,26 @@
 #-keepclassmembers class fqcn.of.javascript.interface.for.webview {
 #   public *;
 #}
+
+# JSON parser classes don't mix well with minification because JSON parser classes use reflection
+# based on the name of the field to create the JSON, and minification changes those names. As a result
+# all the message payload serializing and deserializing breaks. These lines explicitly exclude
+# the packages where we keep our JSON parsing classes from minification to avoid this issue
```
abhipsaMisra, line 22: Interesting find!

device/iot-device-samples/android-sample/app/proguard-rules.pro
```diff
@@ -15,3 +15,26 @@
...
+-keep class com.microsoft.azure.sdk.iot.deps.serializer.* { *; }
+-keep class com.microsoft.azure.sdk.iot.deps.twin.* { *; }
+-keep class com.microsoft.azure.sdk.iot.device.edge.MethodRequest { *; }
+-keep class com.microsoft.azure.sdk.iot.device.hsm.parser.ErrorResponse { *; }
+-keep class com.microsoft.azure.sdk.iot.device.hsm.parser.SignRequest { *; }
+
+# These are the recommended exclusions when using the IoT Hub service client SDK. They are commented
+# out because this sample only uses the IoT Hub device client SDK
+#-keep class com.microsoft.azure.sdk.iot.service.ImportMode { *; }
```
abhipsaMisra, line 31: This is very useful info, but is the device client sample the right place for this? Do we have service client/ provisioning etc android samples where we could place these lines instead?

abhipsaMisra (approved):
cool!

device/iot-device-samples/android-sample/app/proguard-rules.pro
timtay-microsoft: We don't have an android sample for service client or for our provisioning clients, so this is the best place for now

PR #1120, opened 2021-02-12
feat(iot-service): Add DigitalTwinClientOptions support to digital twin clients

timtay-microsoft:
same as other options objects; HTTP read and connect timeouts, and proxy support.

I've seen our nightly builds often fail on http read timeouts on digital twin tests, so this feature should help our tests be more resilient, too

I'm adding this to the RBAC preview code since that is where we are already adding a bunch of digital twin constructors.

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
@@ -179,6 +179,30 @@ public class DigitalTwinClientTests extends IntegrationTest
         assertEquals(responseWithHeaders.body().getMetadata().getModelId(), E2ETestConstants.THERMOSTAT_MODEL_ID);
     }
 
+    @Test(expected = IllegalArgumentException.class)
+    @StandardTierHubOnlyTest
+    public void digitalTwinConstructorThrowsForNegativeConnectTimeout() {
+        // arrange
+        DigitalTwinClientOptions clientOptions =
+            DigitalTwinClientOptions.builder()
+                .httpConnectTimeout(-1)
+                .build();
+
+        digitalTwinClient = new DigitalTwinClient(IOTHUB_CONNECTION_STRING, clientOptions);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    @StandardTierHubOnlyTest
+    public void digitalTwinConstructorThrowsForNegativeReadTimeout() {
+        // arrange
+        DigitalTwinClientOptions clientOptions =
+            DigitalTwinClientOptions.builder()
+                .httpReadTimeout(-1)
+                .build();
+
+        digitalTwinClient = new DigitalTwinClient(IOTHUB_CONNECTION_STRING, clientOptions);
+    }
+
     @Test
     @StandardTierHubOnlyTest
     public void getDigitalTwinWithAzureSasCredential() {
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
@@ -9,10 +9,13 @@ import com.microsoft.azure.sdk.iot.device.DeviceTwin.*;
 import com.microsoft.azure.sdk.iot.service.Device;
 import com.microsoft.azure.sdk.iot.service.IotHubConnectionString;
 import com.microsoft.azure.sdk.iot.service.IotHubConnectionStringBuilder;
+import com.microsoft.azure.sdk.iot.service.ProxyOptions;
 import com.microsoft.azure.sdk.iot.service.RegistryManager;
+import com.microsoft.azure.sdk.iot.service.RegistryManagerOptions;
 import com.microsoft.azure.sdk.iot.service.auth.AuthenticationType;
 import com.microsoft.azure.sdk.iot.service.auth.IotHubServiceSasToken;
 import com.microsoft.azure.sdk.iot.service.digitaltwin.DigitalTwinClient;
+import com.microsoft.azure.sdk.iot.service.digitaltwin.DigitalTwinClientOptions;
 import com.microsoft.azure.sdk.iot.service.digitaltwin.UpdateOperationUtility;
 import com.microsoft.azure.sdk.iot.service.digitaltwin.customized.DigitalTwinGetHeaders;
 import com.microsoft.azure.sdk.iot.service.digitaltwin.models.*;
@@ -24,6 +27,8 @@ import org.junit.*;
 import org.junit.rules.Timeout;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
+import org.littleshoot.proxy.HttpProxyServer;
+import org.littleshoot.proxy.impl.DefaultHttpProxyServer;
 import tests.integration.com.microsoft.azure.sdk.iot.digitaltwin.helpers.E2ETestConstants;
 import tests.integration.com.microsoft.azure.sdk.iot.helpers.IntegrationTest;
 import tests.integration.com.microsoft.azure.sdk.iot.helpers.Tools;
@@ -31,6 +36,8 @@ import tests.integration.com.microsoft.azure.sdk.iot.helpers.annotations.Digital
 import tests.integration.com.microsoft.azure.sdk.iot.helpers.annotations.StandardTierHubOnlyTest;
 
 import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.net.Proxy;
 import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
 import java.time.ZoneOffset;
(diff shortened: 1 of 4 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/DigitalTwinAsyncClient.java
```diff
@@ -60,7 +82,18 @@ public DigitalTwinAsyncClient(String connectionString) {
...
+        Proxy proxy = null;
+        if (proxyOptions != null)
+        {
+            proxy = proxyOptions.getProxy();
+        }
+
         RestClient simpleRestClient = new RestClient.Builder()
+            .withConnectionTimeout(options.getHttpConnectTimeout(), TimeUnit.MILLISECONDS)
+            .withReadTimeout(options.getHttpReadTimeout(), TimeUnit.MILLISECONDS)
+            .withProxy(proxy) // assigning a null proxy here just means no proxy will be used
```
abhipsaMisra, line 96: nice, glad this works!

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
```diff
@@ -54,6 +61,9 @@
     private DeviceClient deviceClient;
     private DigitalTwinClient digitalTwinClient = null;
     private static final String DEVICE_ID_PREFIX = "DigitalTwinServiceClientTests_";
+    protected static HttpProxyServer proxyServer;
+    protected static String testProxyHostname = "127.0.0.1";
+    protected static int testProxyPort = 8769;
```
abhipsaMisra, line 66: worth adding a note about our proxy server dependency that requires us to use this particular port

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/DigitalTwinClientOptions.java
```diff
@@ -0,0 +1,43 @@
...
+     * The options that specify what proxy to tunnel through. If null, no proxy will be used
+     */
+    @Getter
+    private final ProxyOptions proxyOptions;
+
+    /**
+     * The http read timeout to a specified timeout, in milliseconds. A non-zero value specifies the timeout when reading from
+     * Input stream after a connection is established to a resource. If the timeout expires before there is data available
+     * for read, a java.net.SocketTimeoutException is raised. A timeout of zero is interpreted as an infinite timeout.
+     * By default, this value is {@link #DEFAULT_HTTP_READ_TIMEOUT_MS}. Must be a non-negative value.
```
abhipsaMisra, line 30: does the http client used by our generated client do this validation? (timeout should be >= 0)

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
```diff
@@ -108,12 +128,51 @@ public static void cleanUpAfterClass()
...
+    @Test
+    @StandardTierHubOnlyTest
+    public void getDigitalTwinWithProxy() {
+        // arrange
+        Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(testProxyHostname, testProxyPort));
+        ProxyOptions proxyOptions = new ProxyOptions(proxy);
+        DigitalTwinClientOptions clientOptions =
+            DigitalTwinClientOptions.builder()
+                .proxyOptions(proxyOptions)
+                .httpReadTimeout(0)
```
abhipsaMisra, line 167: any test to verify that these timeout values are honored?

abhipsaMisra (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
timtay-microsoft: I'm not sure I follow what you mean. The port number here is pretty arbitrary

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/DigitalTwinClientOptions.java
timtay-microsoft: Yep, the rest client throws an IllegalArgumentException for values below 0

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
abhipsaMisra: I didn't know that the port no here is arbitrary. For .NET, the port no used is dictated by the proxy server configuration used by the test pipeline; which is 8888 for tinyproxy and 3xxx (I don't remember exactly) for squid. Since these are configured to be the listening port for the test proxy server, that is what we need to supply to the client for our e2e tests.
I thought that java followed a similar process, and hence thought that the port no is relevant.

How is the test proxy server configured for java?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/DigitalTwinClientOptions.java
```diff
@@ -0,0 +1,43 @@
...
+import lombok.Getter;
+
+/**
+ * Configurable options for all digital twin client operations. This options bundle is used by both
+ * {@link DigitalTwinClient} and {@link DigitalTwinAsyncClient}.
+ */
+@Builder
+public class DigitalTwinClientOptions
+{
+    protected static final Integer DEFAULT_HTTP_READ_TIMEOUT_MS = 24000; // 24 seconds
```
bikamani, line 17: How are we deciding this number?

timtay-microsoft: These are the defaults we had for earlier options objects. These values were based on some similar hardcoded timeout values that have been in place for a very long time, but were definitely arbitrary.

The idea behind these defaults was so that customers who don't use our options objects will still see roughly the same behavior from the SDK. They will still get read and connect timeouts like the old client used to give

vinagesh (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
timtay-microsoft: Oh, in java we spin up the proxy programmatically in the e2e test class itself. In doing so, we can choose an arbitrary port for it to listen on. The variable ```testProxyPort``` here is referenced directly when creating the proxy, and when creating the service client's proxy settings. Because of that, the only real limitation here is that the port has to be a different port from any other e2e test class because we can't open two proxies on one port at the same time, and these tests run in parallel with other test classes

timtay-microsoft: For instance: https://github.com/Azure/azure-iot-sdk-java/blob/41b952292a67f88270cd414eb30dd620b48ebd75/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/SendMessagesCommon.java#L188

abhipsaMisra: Got it, that explains! It assume it is a start-the-proxy-server-instance-if-it-isn't-running-already kind of flow?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
timtay-microsoft: It's tricky to test this since we can't just make an HTTP call and hope that it will time out in even 1 millisecond. I can add some tests though that ensure that a negative timeout value throws, since that should show that the rest client is getting these values

abhipsaMisra: That's fair, we don't need to test their http client, just that they acknowledge our custom values :)

bikamani (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
timtay-microsoft: > Got it, that explains! It assume it is a start-the-proxy-server-instance-if-it-isn't-running-already kind of flow?

That's the idea, yeah

PR #1122, opened 2021-02-17
refactor(e2e): Add token renewal tests for AzureSasCredential constructors

timtay-microsoft:


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
```diff
@@ -225,6 +225,36 @@ public void getDigitalTwinWithAzureSasCredential() {
...
+
+        AzureSasCredential sasCredential = new AzureSasCredential(serviceSasToken.toString());
+        digitalTwinClient = new DigitalTwinClient(iotHubConnectionStringObj.getHostName(), sasCredential);
+
+        // get a digital twin before the first shared access signature expires
+        // don't care about the return value, just checking that the request isn't unauthorized
+        digitalTwinClient.getDigitalTwin(deviceId, BasicDigitalTwin.class);
+
+        // wait so that the previous shared access signature expires
+        Thread.sleep(2 * tokenLifespanMilliseconds);
```
abhipsaMisra, line 247: how about a getTwin -> unauth -> getTwin test, asserting that the token actually did expire

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
```diff
@@ -225,6 +225,36 @@ public void getDigitalTwinWithAzureSasCredential() {
...
+
+        // Renew the expired shared access signature
+        serviceSasToken = new IotHubServiceSasToken(iotHubConnectionStringObj);
```
abhipsaMisra, line 250: q - does this constructor's javadoc call out that the token generated will have a default ttl (1 hr?)?

timtay-microsoft: It does not, so I'll add it

timtay-microsoft:
```java
    /**
     * Constructor. Generates a SAS token that grants access to an IoT Hub for
     * the default amount of time of 1 hour.
     *
     * @param iotHubConnectionString Connection string object containing the connection parameters
     */
```

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
timtay-microsoft: Good idea, I've reworked the tests a bit to deliberately cause a 401 error

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/exceptions/IotHubExceptionManager.java
```diff
@@ -130,4 +130,64 @@ else if (responseStatus > 300)
...
         // Codes_SRS_SERVICE_SDK_JAVA_IOTHUBEXCEPTIONMANAGER_12_012: [The function shall return without exception if the response status equal or less than 300]
     }
+
+    /**
+     * Return a new exception instance that best matches the given HTTP status code and description
+     * @param responseStatus The HTTP status code (404, 500, etc.)
+     * @param description The HTTP response body
+     * @return a new exception instance that best matches the given HTTP status code
+     */
+    public static IotHubException mapException(int responseStatus, String description)
```
timtay-microsoft, line 140: I needed to add this function because the only other error code to exception mapping logic takes in an httpResponse as the argument. For the AMQP CBS code path, there is no httpResponse object to give it

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/CbsSessionHandler.java
```diff
@@ -131,7 +132,7 @@ public DeliveryState handleAuthenticationResponseMessage(int status, String desc
         }
         else
         {
-            IotHubException e = new IotHubException(status, description);
+            IotHubException e = IotHubExceptionManager.mapException(status, description);
```
timtay-microsoft, line 135: This was a small bug in my earlier work where I didn't actually map to the more specific IotHubException subtype. Now this will throw an IotHubUnauthorizedException for instance.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
```diff
@@ -225,6 +235,56 @@ public void getDigitalTwinWithAzureSasCredential() {
...
+        AzureSasCredential sasCredential = new AzureSasCredential(serviceSasToken.toString());
+        digitalTwinClient = new DigitalTwinClient(iotHubConnectionStringObj.getHostName(), sasCredential);
+
+        // get a digital twin with a valid SAS token in the AzureSasCredential instance
+        // don't care about the return value, just checking that the request isn't unauthorized
+        digitalTwinClient.getDigitalTwin(deviceId, BasicDigitalTwin.class);
+
+        // deliberately expire the SAS token to provoke a 401 to ensure that the digital twin client is using the shared
+        // access signature that is set here.
+        sasCredential.update(SasTokenTools.makeSasTokenExpired(serviceSasToken.toString()));
```
abhipsaMisra, line 260: wouldn't it be easier to create a token that is valid for 2-3 secs and then have it expire? or are we concerned that networking delays of > 2-3secs might occus and have the initial request time out?

abhipsaMisra (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
timtay-microsoft: Yeah, I tried to just let the previous token expire, but I couldn't get consistent 401's from the service. I figure that this current approach that just sends a wildly out of date token would completely remove the risk for clock skew or other small timing issues.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/SasTokenTools.java
```diff
@@ -0,0 +1,38 @@
...
+    private static final String KEY_VALUE_PAIR_SEPARATOR = "&";
+    private static final String EXPIRY_TIME_KEY = KEY_VALUE_PAIR_SEPARATOR + "se=";
+
+    /**
+     * Takes in a valid SAS token and returns an expired version of that same SAS token. This function works for both
+     * device and service SAS tokens.
+     * @param sasToken a valid SAS token to make expired.
+     * @return an expired SAS token.
+     */
+    public static String makeSasTokenExpired(String sasToken)
```
timtay-microsoft, line 20: Our library's public SAS token classes don't allow you to create a token that expires before the current moment, so I added this function in our test code instead. It really should only be used in test scenarios like token renewal tests

Issue #1035, opened 2020-12-30
Crashing while reconnecting, MqttAsyncClient is null


nazmul32:
# Context

- Android 10
- Azure IoT SDK Version, 1.28.0 

## Description of the issue
Application is crashing while reconnecting in the latest SDK. This happens in SDK and almost frequent sometimes. I cannot remember any user initiated call that can cause this issue. I will be highly glad to have some insight regarding this.

## Console log of the issue
2020-12-30 17:47:57.086 18113-18652/com.test.iothub E/AndroidRuntime: FATAL EXCEPTION: azure-iot-sdk-ReconnectionTask:91126a52-8c8c-4ce3-9cc7-1f7e0ee9d01a
    Process: com.test.iothub, PID: 18113
    java.lang.NullPointerException: Attempt to invoke virtual method 'boolean org.eclipse.paho.client.mqttv3.MqttAsyncClient.isConnected()' on a null object reference
        at com.microsoft.azure.sdk.iot.device.transport.mqtt.Mqtt.subscribe(Mqtt.java:282)
        at com.microsoft.azure.sdk.iot.device.transport.mqtt.MqttMessaging.start(MqttMessaging.java:62)
        at com.microsoft.azure.sdk.iot.device.transport.mqtt.MqttIotHubConnection.open(MqttIotHubConnection.java:201)
        at com.microsoft.azure.sdk.iot.device.transport.IotHubTransport.openConnection(IotHubTransport.java:785)
        at com.microsoft.azure.sdk.iot.device.transport.IotHubTransport.singleReconnectAttempt(IotHubTransport.java:916)
        at com.microsoft.azure.sdk.iot.device.transport.IotHubTransport.reconnect(IotHubTransport.java:865)
        at com.microsoft.azure.sdk.iot.device.transport.IotHubTransport.handleDisconnection(IotHubTransport.java:820)
        at com.microsoft.azure.sdk.iot.device.transport.IotHubTransport.onConnectionLost(IotHubTransport.java:253)
        at com.microsoft.azure.sdk.iot.device.transport.ReconnectionNotifier$1.run(ReconnectionNotifier.java:21)
        at java.lang.Thread.run(Thread.java:919)

bikamani:
@nazmul32 Could you add some more details about the code exhibiting this issue? Is the device getting disconnected often? What does your application code look like to handle connection status changes?

nazmul32:
[...]

Issue #991, opened 2020-11-13
NullPointerException when closing connection - ClientState.quiesce


devzeze:
Several crash reports on Crashlytics, I think you have all information needed here.

# Context

Android (several versions)
com.microsoft.azure.sdk.iot:iot-device-client:1.25.0

## Description of the issue
We are getting some crash reports in Crashlytics.

## Code sample exhibiting the issue
In some scenarios, or closing the app we close conection:

    public void destroy() {
        removeCallbacksAndMessages(null);
        try {
            deviceClientRegistration.closeNow();
        } catch (Exception e) {
            Log.e(TAG, "Exception closing Device Client", e);
        }

    }

## Console log of the issue
Crash Stack Trace

Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void java.util.Vector.clear()' on a null object reference
       at org.eclipse.paho.client.mqttv3.internal.ClientState.quiesce(ClientState.java:1356)
       at org.eclipse.paho.client.mqttv3.internal.ClientComms$DisconnectBG.run(ClientComms.java:775)
       at java.lang.Thread.run(Thread.java:784)

AB#8799730

timtay-microsoft:
Can you provide more context on the android devices in question? We have some known compatibility issues with android API versions below 24 that may be relevant here. How consistently do you see this error?

devzeze:
For a relatively small amount of apps, few hundreds, in last 90 days I had this 6 times, it shows up like every 15 days.

in Crashlytics stats:

67% Samsung.
67% Android 10.
33% Background.

Also, some of those occurred during automated tests with Appium, not sure if it might have influence also.

To sum up its not critical, happen not very often.

devzeze:
Hi,

I made a sum up of the NullPointerException we had in last weeks on our Android app. Info obtained in Crashlytics.


1. ClientState.java line 1356

Fatal Exception: java.lang.NullPointerException
Attempt to invoke virtual method 'void java.util.Vector.clear()' on a null object reference

Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void java.util.Vector.clear()' on a null object reference
       at org.eclipse.paho.client.mqttv3.internal.ClientState.quiesce(ClientState.java:1356)
       at org.eclipse.paho.client.mqttv3.internal.ClientComms$DisconnectBG.run(ClientComms.java:775)
       at java.lang.Thread.run(Thread.java:919)


[...]

PR #1125, opened 2021-02-19
fix(iot-device): fix potential NPE cases within MQTT layer

timtay-microsoft:
Instead of having null checks and null assignments everywhere, now important fields like the mqttAsyncClient are final, set in the constructor, and never set to null.

While I was fixing this, I cleaned up some confusing passing of clients around between the layers so that a given layer doesn't have to null check fields retrieved from a different layer repeatedly

The MqttConnection class largely just passed through calls to the mqttAsyncClient, so I deleted it and moved what little logic it had into more appropriate places within this layer.

This should fix both #991 and #1035


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
```diff
@@ -439,11 +413,6 @@ public void deliveryComplete(IMqttDeliveryToken iMqttDeliveryToken)
         }
     }
 
-    public Pair<String, byte[]> peekMessage()
-    {
-        return this.allReceivedMessages.peek();
-    }
```
timtay-microsoft, line 445: No need for a function this simple, so the places that called this previously now just call this.allReceivedMessages.peek();

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttConnection.java
```diff
@@ -1,245 +0,0 @@
-/*
- *  Copyright (c) Microsoft. All rights reserved.
- *  Licensed under the MIT license. See LICENSE file in the project root for full license information.
- */
-
-package com.microsoft.azure.sdk.iot.device.transport.mqtt;
```
timtay-microsoft, line 6: This class had very little meaningful code in it, since it just delegated every API to the mqttAsyncClient. I moved the important code out of this class and then deleted it

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttDeviceMethod.java
```diff
@@ -211,11 +199,4 @@ private void throwMethodsTransportException(String message) throws TransportExce
         transportException.setIotHubService(TransportException.IotHubService.METHODS);
         throw transportException;
     }
-
-    private void throwMethodsTransportException(Exception e) throws TransportException
```
timtay-microsoft, line 215: unused private function

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -39,51 +39,208 @@
...
+        {
+            sslContext = this.config.getAuthenticationProvider().getSSLContext();
+        }
+        catch (IOException e)
+        {
+            throw new TransportException("Failed to get SSLContext", e);
+        }
+
+        char[] password = null;
+        if (this.config.getAuthenticationType() == DeviceClientConfig.AuthType.SAS_TOKEN)
```
timtay-microsoft, line 99: A lot of this constructor logic was moved up from open time to constructor time so that the deviceMessaging/deviceTwin/deviceMethod fields could be declared final to help avoid NPEs

timtay-microsoft: No IO operations were moved for this, though. Just initialization logic for the mqtt client that this layer uses

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttConnectionTest.java
```diff
@@ -1,547 +0,0 @@
...
-import java.util.Queue;
-import java.util.concurrent.ConcurrentLinkedQueue;
-
-import static org.junit.Assert.*;
-
-/*
-    Unit test for MqttConnection
-    Coverage : 100% method, 100% line
- */
-public class MqttConnectionTest
```
timtay-microsoft, line 26: Since the class that this unit tested is now deleted, we can delete this unit test file

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
```diff
@@ -49,56 +54,57 @@
...
-     * @param listener the listener to be called back upon connection established/lost and upon a message being delivered
+     * @param mqttAsyncClient the connection to use
      * @param messageListener the listener to be called back upon a message arriving
-     * @param connectionId the id of the connection
      * @param deviceId the Id of the device this connection belongs to
-     * @throws IllegalArgumentException if the provided mqttConnection is null
      */
-    public Mqtt(MqttConnection mqttConnection, IotHubListener listener, MqttMessageListener messageListener, String connectionId, Map<Integer, Message> unacknowledgedSentMessages, String deviceId) throws IllegalArgumentException
+    public Mqtt(
+        MqttAsyncClient mqttAsyncClient,
```
drwill-ms, line 89: Does changing these parameters constitute a breaking change? Class and ctor are public, but I know with Java that isn't a sure way to know what is public interface.

timtay-microsoft: Nothing in the MQTT layer is considered part of our public API. Users aren't creating these types

timtay-microsoft: We have added and removed params on these classes many times in the past

drwill-ms (approved):

Issue #1123, opened 2021-02-18
It seems that the device SDK connection not stable under mutiple thread


wuwei3:
# Context

- **OS and version used:**  Windows 10, and Ubuntu 18.04
- **Java runtime used:** 1.8
- **SDK version used:** iot-device-client-1.29.0.jar

## Description of the issue

  I have 10 simulation device, I want to report data (send temperature, humidity, CO2 data to the devices) every 7 mins, I used the Java thread pool to finish it, but it always throwing connection failed exception after 10 or 15 times later, it was ok maybe before 10 times

## Code sample exhibiting the issue

![7](https://user-images.githubusercontent.com/46955295/108308800-5909cd80-71eb-11eb-8efa-200724a5be82.png)

## Console log of the issue
Caused by: com.microsoft.azure.sdk.iot.provisioning.device.internal.exceptions.ProvisioningDeviceConnectionException: Exception opening connection
at com.microsoft.azure.sdk.iot.provisioning.device.internal.contract.mqtt.ContractAPIMqtt.open(ContractAPIMqtt.java:175)
at com.microsoft.azure.sdk.iot.provisioning.device.internal.contract.mqtt.ContractAPIMqtt.authenticateWithProvisioningService(ContractAPIMqtt.java:228)
at com.microsoft.azure.sdk.iot.provisioning.device.internal.task.RegisterTask.authenticateWithSasToken(RegisterTask.java:202)
at com.microsoft.azure.sdk.iot.provisioning.device.internal.task.RegisterTask.authenticateWithDPS(RegisterTask.java:319)
at com.microsoft.azure.sdk.iot.provisioning.device.internal.task.RegisterTask.call(RegisterTask.java:341)
at com.microsoft.azure.sdk.iot.provisioning.device.internal.task.RegisterTask.call(RegisterTask.java:37)
at java.util.concurrent.FutureTask.run(FutureTask.java:266)
at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
... 4 more
Caused by: java.io.IOException: Unable to subscribe on topic : $dps/registrations/res/#
at com.microsoft.azure.sdk.iot.deps.transport.mqtt.MqttConnection.subscribe(MqttConnection.java:236)
at com.microsoft.azure.sdk.iot.provisioning.device.internal.contract.mqtt.ContractAPIMqtt.open(ContractAPIMqtt.java:170)
... 11 more
Caused by: Timed out waiting for a response from the server (32000)
[...]

PR #1126, opened 2021-02-19
fix(deps): Fix MQTT subscribe timeout being 1 second

timtay-microsoft:
1 second timeout is a bit too short to accommodate when the service is a bit slower due to load. 1 minute is a bit more reasonable

#1123 

https://github.com/Azure/azure-iot-sdk-java/issues/1123#issuecomment-782443807


drwill-ms (approved):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/mqtt/MqttConnection.java
```diff
@@ -30,7 +30,7 @@
     private static final int KEEP_ALIVE_INTERVAL = 230;
     private static final int MQTT_VERSION = 4;
     private static final boolean SET_CLEAN_SESSION = false;
-    static final int MAX_WAIT_TIME = 1000;
+    static final int SUBSCRIBE_REQUEST_MAX_WAIT_TIME_MILLISECONDS = 60 * 1000; // 1 minute
```
bikamani, line 33: Do you know what we use in dotNet?

bikamani (approved):

Issue #780, opened 2020-05-17
SecurityProviderTPMHsm.java clears the existing EK


MMaiero:
- **OS and version used:** Raspbian 

- **Java runtime used:** openjdk 1.8.0_252

- **SDK version used:** iot-device-client:1.18.0

# Description of the issue:
I'm experimenting the usage of TPM with the Azure SDK. My setup is composed by a raspberry-pi with an Infineon TPM. Infineon provides the EK certificate and using the tpm2-tools I'm able to generate and persist the EK at address 0x81010001.
When using Azure IoT Edge the key generated is used to connect to the cloud. But, instead if I use the SDK, the SecurityProviderTPMHsm class constructor first removes the existing EK and then generates a new one using parameters that seem to match the official TCG template but that result in a different EK and, obviously, registration ID.
The EK seeds are not changed so, in theory, the generated EK should be the same either using the official TCG tpm2-tools and the Microsoft implementation.

My questions are two: 
- why the SDK evicts the existing EK instead of using it, if available
- why the EK generated by the SDK is not the same as the one generated by the official TPM2-tools and matching the stored Infineon EK certificate? 

# Code sample exhibiting the issue:
N/A

# Console log of the issue:
N/A

# Need Support?
* Have a feature request for SDKs? Please post it on [User Voice](https://feedback.azure.com/forums/321918-azure-iot) to help us prioritize.
* Have a technical question? Ask on [Stack Overflow](https://stackoverflow.com/questions/tagged/azure-iot-hub) with tag “azure-iot-hub”
* Need Support? Every customer with an active Azure subscription has access to support with guaranteed response time.  Consider submitting a ticket and get assistance from Microsoft support team
* Found a bug? Please help us fix it by thoroughly documenting it and filing an issue on GitHub (C, Java, .NET, Node.js, Python).

timtay-microsoft:
This does appear to be a bug. #1128 should provide the fix for this issue.

timtay-microsoft:
tpm provider version 1.1.3 has fixed this issue, so I'm closing this issue

PR #1128, opened 2021-02-22
fix(security-tpm): Fix bug where TPM security providers cleared TPM's SRK and EK at constructor time

timtay-microsoft:
These security providers are expected to use the existing EK and SRK if they are available, and to create them only if it didn't exist previously. There was no reason to delete the existing EK and SRK every time.

The .NET SDK does not clear the SRK or EK at constructor time, for reference

I'm running this by a member of the TSS team since they more or less gave this code to us a while back, and they have a lot more context on this code.

#780 for additional context


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/setup/ProvisioningCommon.java
```diff
@@ -239,7 +239,6 @@ public void tearDown()
 
         if (testInstance != null && testInstance.securityProvider != null && testInstance.securityProvider instanceof SecurityProviderTPMEmulator)
         {
-            testInstance.provisioningServiceClient = null;
```
timtay-microsoft, line 242: Fixing a small test bug as well. No need to null this client when it is still used below for cleanup

provisioning/security/tpm-provider-emulator/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/hsm/SecurityProviderTPMEmulator.java
```diff
@@ -66,8 +68,6 @@ public SecurityProviderTPMEmulator() throws SecurityProviderException
     {
         //SRS_SecurityProviderTPMEmulator_25_001: [ The constructor shall start the local TPM Simulator, clear persistent for EK and SRK if it exist, create persistent primary for EK and SRK. ]
         tpm = TpmFactory.localTpmSimulator();
-        clearPersistent(tpm, EK_PERSISTENT_HANDLE, "EK");
-        clearPersistent(tpm, SRK_PERSISTENT_HANDLE, "SRK");
```
abhipsaMisra, line 70: q - is clearing the ek and srk ever relevant for a simulated tpm device (assuming it is a test device with a programmatically generated ek/ srk)?

provisioning/security/tpm-provider/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/hsm/SecurityProviderTPMHsm.java
```diff
@@ -135,6 +134,8 @@ private TPMT_PUBLIC createPersistentPrimary(Tpm tpm, TPM_HANDLE hPersistent, TPM
             throw new SecurityProviderException("Unexpected failure {" +  rc.name() + "} of TPM2_ReadPublic for {" + primaryRole + "}");
         }
 
+        log.info("Creating {} since the TPM didn't have one already", primaryRole);
```
abhipsaMisra, line 137: Q - Wouldn't a hardware tpm device always come with an embedded ek? Is it even relevant for us to be creating an ek for an hsm device?

provisioning/security/tpm-provider-emulator/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/hsm/SecurityProviderTPMEmulator.java
timtay-microsoft: It likely isn't relevant for TPM simulators. The TPM simulator that we use has no EK or SRK on startup. Whether we attempt to clear them here or not should result in nothing done. With that said, we may as well remove this unnecessary code

provisioning/security/tpm-provider/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/hsm/SecurityProviderTPMHsm.java
timtay-microsoft: In practice, yes, hardware TPMs are expected to have an EK. In theory though, it isn't required. I don't expect most users to hit this code path

timtay-microsoft: To add additional context, TPMs support clearing EKs and replacing them with new EKs. That's why this code technically worked without these changes. It's just an unusual case.

abhipsaMisra: That's true, I wouldn't expect this code path to be exercised with hsm flow; no harm in leaving it here though

abhipsaMisra (approved):

jamdavi (approved):

PR #1129, opened 2021-02-23
fix(iot-device): Fix bug where AMQPS_WS clients couldn't send messages with payloads up to 256 kB

timtay-microsoft:
Before this fix, they could send messages up to 16 kB, but this fix brings the limit in line with the IoT Hub size limits, and makes AMQPS_WS consistent with all other protocols.

For additional context: https://github.com/Azure/azure-iot-sdk-java/issues/742


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/telemetry/SendMessagesTests.java
```diff
@@ -104,15 +104,7 @@ public void sendLargestMessages() throws Exception
     {
         this.testInstance.setup();
 
-        if (this.testInstance.protocol == AMQPS_WS)
```
timtay-microsoft, line 107: No need for a special case here. AMQPS_WS can send the same maximum size messages that the other protocols can.

drwill-ms (approved):

PR #1134, opened 2021-02-26
refactor(iot-service): Add AAD auth token caching for HTTP clients

timtay-microsoft:
AMQP client doesn't need it since it has proactive renewal in place already.

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
@@ -178,7 +178,7 @@ public class RegistryManager
      */
     public RegistryManager(String hostName, TokenCredential credential, RegistryManagerOptions options)
     {
-        Objects.requireNonNull(credential, "credentialCache cannot be null");
+        Objects.requireNonNull(credential, "credential cannot be null");
         Objects.requireNonNull(options, "options cannot be null");
         if (Tools.isNullOrEmpty(hostName))
         {
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
@@ -14,6 +14,7 @@ import com.microsoft.azure.sdk.iot.deps.serializer.DeviceParser;
 import com.microsoft.azure.sdk.iot.deps.serializer.JobPropertiesParser;
 import com.microsoft.azure.sdk.iot.deps.serializer.RegistryStatisticsParser;
 import com.microsoft.azure.sdk.iot.service.auth.IotHubServiceSasToken;
+import com.microsoft.azure.sdk.iot.service.auth.TokenCredentialCache;
 import com.microsoft.azure.sdk.iot.service.exceptions.IotHubException;
 import com.microsoft.azure.sdk.iot.service.exceptions.IotHubExceptionManager;
 import com.microsoft.azure.sdk.iot.service.transport.http.HttpMethod;
@@ -44,7 +45,7 @@ public class RegistryManager
     private static final int EXECUTOR_THREAD_POOL_SIZE = 10;
     private ExecutorService executor;
     private final String hostName;
-    private TokenCredential credential;
+    private TokenCredentialCache credentialCache;
     private AzureSasCredential azureSasCredential;
     private IotHubConnectionString iotHubConnectionString;
 
@@ -177,7 +178,7 @@ public class RegistryManager
      */
     public RegistryManager(String hostName, TokenCredential credential, RegistryManagerOptions options)
     {
-        Objects.requireNonNull(credential, "credential cannot be null");
+        Objects.requireNonNull(credential, "credentialCache cannot be null");
         Objects.requireNonNull(options, "options cannot be null");
         if (Tools.isNullOrEmpty(hostName))
         {
@@ -186,7 +187,7 @@ public class RegistryManager
 
         this.executor = Executors.newFixedThreadPool(EXECUTOR_THREAD_POOL_SIZE);
         this.options = options;
-        this.credential = credential;
+        this.credentialCache = new TokenCredentialCache(credential);
         this.hostName = hostName;
     }
 
@@ -1547,9 +1548,9 @@ public class RegistryManager
         // Three different constructor types for this class, and each type provides either a TokenCredential implementation,
         // an AzureSasCredential instance, or just the connection string. The sas token can be retrieved from the non-null
         // one of the three options.
-        if (this.credential != null)
+        if (this.credentialCache != null)
         {
-            return this.credential.getToken(new TokenRequestContext()).block().getToken();
+            return this.credentialCache.getAccessToken().getToken();
         }
         else if (this.azureSasCredential != null)
         {
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
@@ -1,0 +1,72 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.azure.sdk.iot.service.auth;
+
+import com.azure.core.credential.AccessToken;
+import com.azure.core.credential.TokenCredential;
+import com.azure.core.credential.TokenRequestContext;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Objects;
+
+/**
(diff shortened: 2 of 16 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -177,7 +178,7 @@ public RegistryManager(String hostName, TokenCredential credential)
      */
     public RegistryManager(String hostName, TokenCredential credential, RegistryManagerOptions options)
     {
-        Objects.requireNonNull(credential, "credential cannot be null");
+        Objects.requireNonNull(credential, "credentialCache cannot be null");
```
vinagesh, line 181: This gets thrown back to the user. So it should still read 'credential cannot be null' as they are not aware of our internal implementation.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
```diff
@@ -0,0 +1,72 @@
...
+import java.time.Instant;
+import java.util.Objects;
+
+/**
+ * This class generates AAD authentication tokens from a TokenCredential but caches previous tokens when they aren't near
+ * expiry.
+ */
+public class TokenCredentialCache
+{
+    private final static int MINUTES_BEFORE_PROACTIVE_RENEWAL = 9;
```
vinagesh, line 20: I made it a round figure of 10 on .net :D But 9 is fine too.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
```diff
@@ -0,0 +1,72 @@
...
+        this.tokenCredential = tokenCredential;
+    }
+
+    /**
+     * Get a valid AAD authentication token. This may be the same as a previously returned token if it is not near expiration time yet.
+     * @return a valid AAD authentication token.
+     */
+    public AccessToken getAccessToken()
+    {
+        if (this.accessToken == null)
```
vinagesh, line 41: Can it just be if (this.accessToken == null || isAccessTokenCloseToExpiry(this.accessToken)) ?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
```diff
@@ -0,0 +1,72 @@
...
+     * @return the TokenCredential instance that was set in the constructor.
+     */
+    public TokenCredential getTokenCredential()
+    {
+        return this.tokenCredential;
+    }
+
+    private static boolean isAccessTokenCloseToExpiry(AccessToken accessToken)
+    {
+        Duration remainingTimeToLive = Duration.between(Instant.now(), accessToken.getExpiresAt().toInstant());
```
vinagesh, line 64: Nice notation of between in java :)

vinagesh: Just curious - Is there a concept of utc and non-utc at all or everything is utc ?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
```diff
@@ -0,0 +1,72 @@
...
+        if (remainingTimeToLive.toMinutes() <= MINUTES_BEFORE_PROACTIVE_RENEWAL)
```
vinagesh, line 65: Can it just be 'return remainingTimeToLive.toMinutes() <= MINUTES_BEFORE_PROACTIVE_RENEWAL' ?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public TokenCredentialCache(TokenCredential tokenCredential)
+    {
+        Objects.requireNonNull(tokenCredential, "tokenCredential cannot be null");
+
+        this.tokenCredential = tokenCredential;
+    }
+
+    /**
+     * Get a valid AAD authentication token. This may be the same as a previously returned token if it is not near expiration time yet.
```
vinagesh, line 36: Would be good to add an explanation of what near expiration time is in the implementation here.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
timtay-microsoft: I did a "less than or equal to" operation, so our functions should work the same in including 9 minutes

service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCacheTest.java
```diff
@@ -0,0 +1,173 @@
...
+        AccessToken accessToken2 = cache.getAccessToken();
+
+        assertEquals(mockAccessToken, accessToken);
+        assertEquals(mockAccessToken, accessToken2);
+    }
+
+    @Test
+    public void tokenCredentialProactivelyRenewsToken()
+    {
+        TokenCredentialCache cache = new TokenCredentialCache(mockTokenCredential);
```
vinagesh, line 63: Do you need line 63 to 74? It seems like it is the same as what was tested in the previous method.

service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCacheTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+        // 8 minutes from the current time, should fit within the proactive renewal range
+        final long milliseconds = System.currentTimeMillis() + (8 * 60 * 1000);
+        new Expectations()
+        {
+            {
+                mockAccessToken.getExpiresAt();
+                result = Instant.ofEpochMilli(milliseconds).atOffset(ZoneOffset.UTC);
+
+                mockTokenCredential.getToken((TokenRequestContext) any).block();
```
vinagesh, line 86: I might not be understanding the syntax here well but does this not mean you are setting the value for what should be returned on calling getToken? How will this test the logic? I think you should be testing the logic in isAccessTokenCloseToExpiry which return true or false to be sure your dateTime comparison logic is tested.

vinagesh: The way I read this was:
Setup - If getToken is call, return mockAccessToken
Execute - Call getToken
So there will be no situation that these values will be different.

timtay-microsoft: This is mocking the TokenCredential that powers the class under test. This line means that the test expects the tokenCredential to get a token, and the result of that call is defined by the line after it.

timtay-microsoft: Mocking the AccessToken's ```getExpiresAt``` method is what is really helping to test the ```isAccessTokenCloseToExpiry``` method

service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCacheTest.java
timtay-microsoft: The expectations look the same, but then this test changes the expectations before attempting to get the next token. I do need these lines to mimic the typical first generation of an access token. The next expectations make it so that the renewal logic kicks in based off of the expiry time of the mock access token

vinagesh (approved):

service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCacheTest.java
vinagesh: So how can you turn this into a test that might fail? If you change isAccessTokenCloseToExpiry from <= 9 to >9, will it fail?

vinagesh: Ok I think I finally understand the syntax! Thanks for the explanation :)

Issue #1135, opened 2021-03-02
File Descriptor leak on open() when using AMQPS protocol


mattixtech:
# Context
Calling `open()` on the IoT hub device client leaks file descriptors when unable to connect when the AMQPS protocol is used.

- **OS and version used:** macOS 11.1, Ubuntu Core 18, Ubuntu Core 20
- **Java runtime used:** tested w/ Java 1.8 & Java 11
- **SDK version used:** 1.29.2

## Description of the issue
When attempting to `open()` the IoT device client connection when the network is down, the client leaks file descriptors depending on how many times `open()` is called. This seems to be specific to using the AMQPS protocol. This is an issue when attempting reconnect logic. For example if you continue trying to connect while the network is down until you get a successful connection, the number of FDs grows until the process runs out of FDs.

This might happen when connecting for the first time while the network is down or when processing the connection status callback for a non retryable error that requires a new client to be instantiated and `open()`ed.

If/when the client does connect and enters steady state, the leaked FDs appear to remain permanently leaked and the number allocated to the process does not appear to decrease.

I tested this issue using MQTT and could not reproduce the issue, therefore I suspect it may be specific to AMQPS, I did not test any other protocol combinations.

## Code sample exhibiting the issue
Please see reproduction example here: https://github.com/mattixtech/iot-device-client-bug-demo/blob/master/src/main/java/org/example/Example.java

Simply run this program with a valid connection string without a network connection and observe the number of file descriptors continue to grow unbounded. For example on macOS I can run this program and monitor the number of FDs with the following command:
```
watch "lsof -p <pid> | wc -l"
```

## Console log of the issue
N/A

AB#9435166

timtay-microsoft:
The open call to a device client doesn't have retry logic, so can you clarify the scenario a bit more? Is it that a file descriptor is leaked when you call open and the open isn't successful, or is it leaked during reconnection logic on a previously established connection?

timtay-microsoft:
Ah, I just saw your sample code. So the issue here is just when you try to open a device client that previously failed to open. We'll try to look into this soon. Thanks for filing the issue!

mattixtech:
> The open call to a device client doesn't have retry logic, so can you clarify the scenario a bit more? Is it that a file descriptor is leaked when you call open and the open isn't successful, or is it leaked during reconnection logic on a previously established connection?

Ultimately if there is no network connection, the call to `open()` fails with an exception and the leak happens at this point.

[...]

PR #1136, opened 2021-03-02
fix(iot-dev): Fix bug where AMQP layer doesn't release resources after failing to open

timtay-microsoft:
reactor.free() releases file descriptors that were grabbed by proton reactor. Since the connection failed to open, there is no need to hang on to them.

#1135 for more context

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -1125,6 +1125,8 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
             }
             catch (HandlerException e)
             {
+                iotHubReactor.free();
+                
                 TransportException transportException = new TransportException(e);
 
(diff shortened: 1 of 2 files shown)


abhipsaMisra (approved):
Ah, so this is something like closing the tcp connection underneath on disconnect?

timtay-microsoft:
> Ah, so this is something like closing the tcp connection underneath on disconnect?

From the Javadocs from proton-j:

```java
    /**
     * Frees any resources (such as sockets and selectors) held by the reactor
     * or its children.
     */
```

vinagesh (approved):

mattixtech:
Hey @timtay-microsoft, given that the leak happened even when calling `closeNow()` that seems to imply to me that the close path may also be missing the same logic. Do you think we have an issue there too or will this updated path be exercised in both cases?

mattixtech:
> Hey @timtay-microsoft, given that the leak happened even when calling `closeNow()` that seems to imply to me that the close path may also be missing the same logic. Do you think we have an issue there too or will this updated path be exercised in both cases?

What I mean to say is if we close a connection that was opened fine, will we clean up these reactor resources in that case?

timtay-microsoft:
@mattixtech good point. I'll go ahead and open another PR for that fix as well. I do see that the reactor isn't always free'd there, too.

PR #1137, opened 2021-03-03
fix(iot-dev): Fix bug where AMQP reactor isn't free'd if closing encounters an exception

timtay-microsoft:
continuing on from #1135, this change ensures that even if the AMQP layer encounters an exception while closing manually, that the reactor is always free'd at the end

#1135 for additional context


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -262,23 +262,28 @@ public void open() throws TransportException
     public void close() throws TransportException
     {
         log.debug("Shutting down amqp layer...");
-        closeAsync();
-
         try
         {
-            log.trace("Waiting for reactor to close...");
-            closeReactorLatch.await(MAX_WAIT_TO_CLOSE_CONNECTION, TimeUnit.MILLISECONDS);
+            closeAsync();
```
drwill-ms, line 267: Any code comments worthy to include given there was a bug here earlier? I'm not asserting there is, just asking you to consider. :)

drwill-ms (approved):

Issue #719, opened 2020-02-28
IoT Service Client does not expose functionality for replacing a device twin


mattixtech:
Using iot-service-client version 1.2.0

# Description of the issue:
The iot-service-client currently does not support replacing a device twin. It only has a method for updating the device twin. According to the HTTP API (https://docs.microsoft.com/en-us/rest/api/iothub/service/replacetwin) it appears a replace is done the same as an update but sending HTTP method PUT instead of PATCH.

The Java SDK can be updated easily to add support for this and I have tested it in a fork here: https://github.com/mattixtech/azure-iot-sdk-java/commit/7ad66f1f948946642c24b0403540b02edbb3b162. The replace call is made exactly the same as the existing call to update aside from substituting the HTTP method to PUT.

With thew new method I was able to replace the desired properties of my IoT device entirely overwriting the existing contents which is the expected behaviour.

Adding this functionality to the SDK would make replacing device twins much more convenient since the alternative seems to be to explicitly set all the existing keys that are no longer required to null which requires you know all the keys at the time of replacing the twin.

I'm assuming the intent of the SDK is to have feature parity with the other language SDKs and HTTP API. If that is the case I don't see any reason to not update the Java SDK to support this.

Thanks!!

mattixtech:
This feature will be very handy for our use case. Can I please get some feedback on whether or not we could get this updated in the SDK. I'm hesitant to open a PR because I'm not sure about the documentation requirements/testing needed.

Thanks

prmathur-microsoft:
At the moment we dont have an ETA for this feature. We will add it to backlog.

timtay-microsoft:
I'm looking to finally get this feature in soon. I'm aiming to release it alongside the other issue that you filed.

timtay-microsoft:
Closing this issue as this feature has been released as of Service Client version 1.28.0

PR #1139, opened 2021-03-03
feat(iot-service): Add Twin Client API to replace a twin

timtay-microsoft:
Previously, the client only exposed a PATCH API. #719 shows that customers are interested in having a replace API as well.

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
@@ -133,6 +133,7 @@ public class DeviceTwin
         /*
         **Codes_SRS_DEVICETWIN_25_012: [** The function shall set eTag, tags, desired property map, reported property map on the user device **]**
          */
+        device.setVersion(twinState.getVersion());
         device.setModelId(twinState.getModelId());
         device.setETag(twinState.getETag());
         device.setTags(twinState.getTags());
@@ -216,7 +217,7 @@ public class DeviceTwin
         **Codes_SRS_DEVICETWIN_25_020: [** The function shall verify the response status and throw proper Exception **]**
          */
         Proxy proxy = options.getProxyOptions() != null ? options.getProxyOptions().getProxy() : null;
-        HttpResponse response = DeviceOperations.request(this.iotHubConnectionString, url, HttpMethod.PATCH, twinJson.getBytes(StandardCharsets.UTF_8), String.valueOf(requestId++),options.getHttpConnectTimeout(), options.getHttpReadTimeout(), proxy);
+        DeviceOperations.request(this.iotHubConnectionString, url, HttpMethod.PATCH, twinJson.getBytes(StandardCharsets.UTF_8), String.valueOf(requestId++),options.getHttpConnectTimeout(), options.getHttpReadTimeout(), proxy);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
```diff
@@ -133,6 +133,7 @@ private void getTwinOperation(URL url, DeviceTwinDevice device) throws IotHubExc
         /*
         **Codes_SRS_DEVICETWIN_25_012: [** The function shall set eTag, tags, desired property map, reported property map on the user device **]**
          */
+        device.setVersion(twinState.getVersion());
         device.setModelId(twinState.getModelId());
```
timtay-microsoft, line 137: This was a small bug in getTwin where we didn't set the version properly

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
```diff
@@ -216,7 +217,7 @@ public synchronized void updateTwin(DeviceTwinDevice device) throws IotHubExcept
         **Codes_SRS_DEVICETWIN_25_020: [** The function shall verify the response status and throw proper Exception **]**
          */
         Proxy proxy = options.getProxyOptions() != null ? options.getProxyOptions().getProxy() : null;
-        HttpResponse response = DeviceOperations.request(this.iotHubConnectionString, url, HttpMethod.PATCH, twinJson.getBytes(StandardCharsets.UTF_8), String.valueOf(requestId++),options.getHttpConnectTimeout(), options.getHttpReadTimeout(), proxy);
```
timtay-microsoft, line 219: The variable declaration was unused since this method has no return object

abhipsaMisra: no return object meaning void? how was this even compiling?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
```diff
@@ -259,6 +260,72 @@ public void replaceTags(DeviceTwinDevice device) throws UnsupportedOperationExce
...
+
+        TwinState twinState = new TwinState(device.getTagsMap(), device.getDesiredMap(), null);
+        String twinJson = twinState.toJsonElement().toString();
+
+        Proxy proxy = options.getProxyOptions() != null ? options.getProxyOptions().getProxy() : null;
+
+        HttpResponse httpResponse = DeviceOperations.request(
+            this.iotHubConnectionString,
+            url,
+            HttpMethod.PUT,
```
abhipsaMisra, line 296: I could have just checked the code, but asking here since you would have gone over it recently anyway - is this the same as what we currently have in place, but we simply replace a PATCH operation with a PUT?

abhipsaMisra (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
```diff
@@ -259,6 +260,72 @@ public void replaceTags(DeviceTwinDevice device) throws UnsupportedOperationExce
...
+            twinJson.getBytes(StandardCharsets.UTF_8),
+            String.valueOf(requestId++),
```
abhipsaMisra, line 298: Interesting approach for request Id

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
timtay-microsoft: Sorry, I meant that the HTTP response has no payload. It is just a 204. The method still returns an HTTP response object, we just don't do anything with it

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
timtay-microsoft: Yeah, this is copied from the other methods. I don't care for it that much, to be honest.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
timtay-microsoft: Pretty much, yeah. The PATCH and PUT payload look exactly the same and go to the same URL.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
abhipsaMisra: the tone was lost in my comment above - :neutral_face:

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
timtay-microsoft: One small difference is that this new PUT API does allow for an empty set of desired properties and tags. The PATCH API makes you provide at least one desired property or tag to update, but the PUT allows you to "clear" a twin

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
timtay-microsoft: I may revisit how we handle request Ids at a later time

PR #1147, opened 2021-03-08
fix(iot-device): Fix bug where TransportExceptions thrown in Amqp lay…

timtay-microsoft:
…er open were thrown before cleaning up reactor and threadpool

InterruptedExceptions were behaving correctly, but TransportExceptions were not

The mistake I made here was assuming that the ReactorRunner thread that encounters these exceptions would itself throw a HandlerException and that the runner's catch block would close the reactor after encountering it. In reality, the reactor runner does not throw a HandlerException when the device hits an IOException or similar exception. Instead, the main thread monitoring this reactor thread is the one who throws the exception. Because of that, I had to move the reactor freeing responsibilities to the main thread as well to cover these error cases

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -1143,8 +1143,6 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
             }
             catch (HandlerException e)
             {
-                iotHubReactor.free();
-                
                 TransportException transportException = new TransportException(e);
 
                 // unclassified exceptions are treated as retryable in ProtonJExceptionParser, so they should be treated
@@ -1152,6 +1150,10 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
                 transportException.setRetryable(true);
 
                 this.listener.onConnectionLost(transportException, connectionId);
+            }
+            finally
+            {
+                iotHubReactor.free();
             }
 
             return null;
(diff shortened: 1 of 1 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -244,9 +244,20 @@ public void open() throws TransportException
...
+            {
+                // clean up network resources and thread scheduler before exiting this layer. Subsequent open attempts
+                // will create a new reactor and a new executor service
+                this.reactor.free();
+                this.executorServicesCleanup();
+                throw e;
+            }
             catch (InterruptedException e)
             {
-                this.close();
```
timtay-microsoft, line 249: Close() has some logic that isn't necessary to run when the connection wasn't opened successfully, so instead of calling close, I just moved the necessary logic up to these catch blocks

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -1132,8 +1143,6 @@ public Object call()
             }
             catch (HandlerException e)
             {
-                iotHubReactor.free();
```
timtay-microsoft, line 1135: It turns out that if an IOException is thrown due to no network, this catch block isn't run because the AMQP layer delegates the exception to the main thread instead. To help cover similar cases where the reactor thread encounters but does not throw an exception, it makes more sense for this free() to be in the finally block.

drwill-ms (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -1142,6 +1151,10 @@ public Object call()
 
                 this.listener.onConnectionLost(transportException, connectionId);
             }
+            finally
+            {
+                iotHubReactor.free();
```
drwill-ms, line 1156: Why no `this.` in this case?

timtay-microsoft: No reason, I suppose. We don't really have strict guidelines on if we use them or not in Java

bikamani (approved):

barustum (approved):

PR #1148, opened 2021-03-08
fix(e2e): Add retry logic to e2e fault injection tests for when fault…

timtay-microsoft:
… message wasn't sent successfully

Due to some inconsistent behavior from the service, we assign a message expiry time to our fault injection messages so that we don't continuously send the faulting message over and over. The issue is that if the message expires before it is successfully sent, then the test will have failed to fault the connection, and our recovery logic can't be tested.

This change brings some simple retry logic that will continue to send the fault injection message until it reaches the service. This includes a special case for not treating MESSAGE_EXPIRED as a successful send


timtay-microsoft:
For context, IoT Hub responds differently to error injection messages sent over MQTT vs over AMQP. In the MQTT case, they sever the connection immediately without sending an ack for the error injection message itself. From the client side, we have retry logic built in that will just attempt to send that message again once the connection is established. This would be a pointless loop, so the expiry time on the message protects us from sending the fault injection message over and over. AMQP does send an ack on the message before closing the connection, thankfully

PR #1149, opened 2021-03-09
refactor(iot-dev): Restrict HTTP request support to only localhost requests

timtay-microsoft:
See the comment within the code change for more details

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsConnection.java
@@ -106,8 +106,8 @@ public class HttpsConnection
         {
             // Currently, IoT Edge runtime can dictate that this client library sends an HTTP request, but it will
             // only ever be to localhost. The Edge team is currently working to remove this requirement. Eventually,
-            // they will just use unix domain sockets instead, and once that happens, we can remove support for HTTP
-            // requests entirely.
+            // the edge runtime will just make this client use unix domain sockets instead.
+            // Once that happens, we can remove support for HTTP requests entirely.
             throw new IllegalArgumentException("Cannot do HTTP requests to any host other than localhost");
         }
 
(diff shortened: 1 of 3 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsConnection.java
```diff
@@ -101,6 +101,16 @@ public HttpsConnection(URL url, HttpsMethod method, final ProxySettings proxySet
             throw new IllegalArgumentException(errMsg);
         }
 
+        String host = url.getHost();
+        if (!isHttps && !host.equalsIgnoreCase("localhost"))
+        {
+            // Currently, IoT Edge runtime can dictate that this client library sends an HTTP request, but it will
+            // only ever be to localhost. The Edge team is currently working to remove this requirement. Eventually,
+            // the edge runtime will just make this client use unix domain sockets instead.
+            // Once that happens, we can remove support for HTTP requests entirely.
```
drwill-ms, line 110: Do you want to reference the bug Id we got from the Edge team for tracking?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsConnection.java
```diff
@@ -101,6 +101,16 @@ public HttpsConnection(URL url, HttpsMethod method, final ProxySettings proxySet
...
+            throw new IllegalArgumentException("Cannot do HTTP requests to any host other than localhost");
```
drwill-ms, line 111: Is this worth bringing up as a breaking change with Mark? My guess is there is no way a customer should run into this, therefore it isn't really breaking. Can they?

timtay-microsoft: As long as the edge workload URI always points to localhost, which the edge team confirmed, then this change shouldn't be noticed by anyone. There are no other codepaths that made this SDK send an HTTP request

drwill-ms (approved):

vinagesh (approved):

barustum (approved):

Issue #1154, opened 2021-03-16
[Bug Report] DeviceClient's open() times out if C2D message queue is full


Eugeniu-N:
# Context

- **OS and version used:** Windows 10
- **Java runtime used:** 1.8
- **SDK version used:** maven dependency:
<groupId>com.microsoft.azure.sdk.iot</groupId>
<artifactId>iot-device-client</artifactId>
<version>1.29.3</version>

## Description of the issue
I used the service client on another project and filled my C2D message queue completely (50 messages), obviously when I try to send more I get an exception. Then I try to start the simulated device, the program hangs at `client.open()` and then returns an exception.
I start the simulated device with another connection string, belonging to a device with a non-full C2D message queue and I can connect just fine.

## Code sample exhibiting the issue
```
public class DeviceRead {
	// IOT Hub connection string to read from and send messages to
	private static String connString1 = "xxx";

	// Using the MQTT protocol to connect to IoT Hub
	private static IotHubClientProtocol protocol = IotHubClientProtocol.MQTT;
	private static DeviceClient client;
	
	public static void main(String[] args) throws IOException, URISyntaxException {
		// Connect to the IoT hub.
		client = new DeviceClient(connString1, protocol);
		MessageCallback callback = new AppMessageCallback();
		client.setMessageCallback(callback, null);
		System.out.println("Opening client...");
		client.open();
		
		// Stop the application.
		System.out.println("Press ENTER to exit.");
		System.in.read();
		
		System.out.println("Closing client...");
		client.closeNow();
	}

	private static Logger messageCallbackLogger = Logger.getLogger("AppMessageCallback");
	private static class AppMessageCallback implements MessageCallback {
		public IotHubMessageResult execute(Message msg, Object context) {
			messageCallbackLogger.info(String.format("Received message from hub: %s, with message id: %s", new String(msg.getBytes(), Message.DEFAULT_IOTHUB_MESSAGE_CHARSET), msg.getMessageId()));
			
			return IotHubMessageResult.COMPLETE;
		}
	}
}
```

## Console log of the issue
Note that `com.microsoft.docs.iothub.samples.SimulatedDevice_READ_ACK.main(SimulatedDevice_READ_ACK.java:39)` line number refers to the line `client.open()`
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
[...]

PR #1155, opened 2021-03-17
fix(iot-dev): Fix deadlock within MQTT layer when client opens with message callback set and cloud to device message queue is full

timtay-microsoft:
IoT Hub sends the cloud to device messages as soon as it receives the mqtt subscription on the topic, even before it sends the SUBACK that free's up this client to continue on with its open logic (subscribing to a topic is done in a blocking fashion in this client). By allowing the MQTT layer to receive messages while opening, this deadlock is avoided since we can acknowledge the cloud to device messages and get to the service sending the SUBACK

issue #1154 for additional context

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
@@ -346,35 +346,32 @@ public class MqttIotHubConnection implements IotHubTransportConnection, MqttMess
             return IotHubStatusCode.BAD_FORMAT;
         }
 
-        synchronized (this.mqttConnectionStateLock) // cannot simultaneously open/close the connection and send a message
+        if (this.state == IotHubConnectionStatus.DISCONNECTED)
         {
-            if (this.state == IotHubConnectionStatus.DISCONNECTED)
-            {
-                throw new IllegalStateException("Cannot send event using a closed MQTT connection");
-            }
-
-            IotHubStatusCode result = IotHubStatusCode.OK_EMPTY;
-
-            if (message.getMessageType() == DEVICE_METHODS)
-            {
-                this.deviceMethod.start();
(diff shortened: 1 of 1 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -346,35 +346,32 @@ public IotHubStatusCode sendMessage(Message message) throws TransportException
             return IotHubStatusCode.BAD_FORMAT;
         }
 
-        synchronized (this.mqttConnectionStateLock) // cannot simultaneously open/close the connection and send a message
```
timtay-microsoft, line 349: This layer still uses this lock to prevent simultaneous open/close calls, but doesn't need to prevent simultaneous open calls and message receive handling.

azabbasi: So if I am reading this correctly, this lock only locks the open and close operations and not message handling correct?

timtay-microsoft: Yep. And that is sufficient. Higher layers in this SDK already do some basic locking to prevent sending messages when the client is not open, so it is safe to remove these particular synchronization blocks

timtay-microsoft: In case the terminology is new to you, a synchronized block in Java is equivalent to grabbing and then releasing a semaphore.

```java
public void foo()
{
    this.runSomeThreadSafeOperation();

    synchronized (this.lock)
    {
        this.runSomeThreadUnsafeOperation();
    }
}
```

the method foo can be called by multiple threads without worrying about two threads running "runSomeThreadUnsafeOperation" at the same time

azabbasi (approved):

PR #1158, opened 2021-03-19
refactor(iot-dev): Upgrade azure storage SDK dependency to the latest

timtay-microsoft:
A lot of client objects changed names since we took this old dependency (4 major version bumps total), so a lot of code had to be rewritten to use this latest version

This is a nice change on it's own, but it is largely driven by some compatibility issues that the old azure storage SDK has was the incoming azure identity library AAD auth changes. Since both master and preview would benefit from this change, this code will be merged into master first, and preview will get it when we merge master into preview shortly thereafter

### DIFF
--- a/device/iot-device-client/pom.xml
+++ b/device/iot-device-client/pom.xml
@@ -57,15 +57,15 @@
             <version>${iot-deps-version}</version>
         </dependency>
         <dependency>
-            <groupId>com.microsoft.azure</groupId>
-            <artifactId>azure-storage</artifactId>
-            <version>8.1.0</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>com.fasterxml.jackson.core</groupId>
-                    <artifactId>jackson-databind</artifactId>
-                </exclusion>
-            </exclusions>
+            <groupId>com.azure</groupId>
+            <artifactId>azure-storage-blob</artifactId>
+            <version>12.8.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.11</version>
+            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>org.eclipse.paho</groupId>
--- a/device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTaskTest.java
+++ b/device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTaskTest.java
@@ -1,5 +1,7 @@
 package tests.unit.com.microsoft.azure.sdk.iot.device.fileupload;
 
+import com.azure.storage.blob.BlobClient;
+import com.azure.storage.blob.BlobClientBuilder;
 import com.microsoft.azure.sdk.iot.deps.serializer.FileUploadCompletionNotification;
 import com.microsoft.azure.sdk.iot.deps.serializer.FileUploadSasUriRequest;
 import com.microsoft.azure.sdk.iot.deps.serializer.FileUploadSasUriResponse;
@@ -10,9 +12,6 @@ import com.microsoft.azure.sdk.iot.device.ResponseMessage;
 import com.microsoft.azure.sdk.iot.device.fileupload.FileUploadTask;
 import com.microsoft.azure.sdk.iot.device.transport.IotHubTransportMessage;
 import com.microsoft.azure.sdk.iot.device.transport.https.HttpsTransportManager;
-import com.microsoft.azure.storage.StorageException;
-import com.microsoft.azure.storage.StorageExtendedErrorInformation;
-import com.microsoft.azure.storage.blob.CloudBlockBlob;
 import mockit.Deencapsulation;
 import mockit.Mocked;
 import mockit.NonStrictExpectations;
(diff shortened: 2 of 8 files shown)


device/iot-device-client/pom.xml
```diff
@@ -57,15 +57,15 @@
...
-                    <artifactId>jackson-databind</artifactId>
-                </exclusion>
-            </exclusions>
+            <groupId>com.azure</groupId>
+            <artifactId>azure-storage-blob</artifactId>
+            <version>12.8.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
```
timtay-microsoft, line 66: We used to get this dependency from the older azure storage sdk, and we were using a few classes from it, so I had to add this dependency explicitly

device/iot-device-samples/file-upload-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/FileUploadSample.java
```diff
@@ -142,11 +146,14 @@ private static void uploadFile(DeviceClient client, String baseDirectory, String
 
             try
             {
-                // Note that other versions of the Azure Storage SDK can be used here instead. The latest can be found here:
-                // https://github.com/Azure/azure-sdk-for-java/tree/master/sdk/storage#azure-storage-sdk-client-library-for-java
```
timtay-microsoft, line 146: No need for this note anymore since we are using the latest version. We only added this note previously to encourage users to avoid this older storage SDK version

service/iot-service-client/pom.xml
```diff
@@ -86,18 +86,6 @@
             <version>1.18.8</version>
             <scope>provided</scope>
         </dependency>
-        <dependency>
-            <groupId>com.microsoft.azure</groupId>
-            <artifactId>azure-storage</artifactId>
```
timtay-microsoft, line 91: Even though this was scoped as a test dependency, we weren't actually using this in the service client anywhere, so I removed it

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/serviceclient/ExportImportTests.java
```diff
@@ -320,8 +344,8 @@ private static void runImportJob(List<ExportImportDevice> devices, ImportMode im
         // Creating the Azure storage blob and uploading the serialized string of devices
         InputStream stream = new ByteArrayInputStream(blobToImport);
         String importBlobName = "devices.txt";
-        CloudBlockBlob importBlob = importContainer.getBlockBlobReference(importBlobName);
-        importBlob.deleteIfExists();
+        BlobClient importBlob = importContainer.getBlobClient(importBlobName);
```
azabbasi, line 347: isn't this supposed to be `BlockBlobClient` ?

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTaskTest.java
```diff
@@ -655,89 +660,15 @@ public void runFileUploadResponseParserThrows() throws IOException, IllegalArgum
         Deencapsulation.invoke(fileUploadTask, "run");
     }
 
-    /* Tests_SRS_FILEUPLOADTASK_21_022: [If the upload to blob failed, the run shall create a notification the IoT Hub with `isSuccess` equals false, `statusCode` equals -1.] */
```
drwill-ms, line 658: Can you speak to the removal of this code? Why is it no longer needed?

azabbasi (approved):
Just wondering if one of the `BlobClient`s is supposed to be `BlobBlockClient` or not.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/serviceclient/ExportImportTests.java
drwill-ms: Bob Lawblaw's Block Blob

https://arresteddevelopment.fandom.com/wiki/Bob_Loblaw

drwill-ms (approved):

drwill-ms:
Will this come with any behavior changes? Should we consider a major version rev?

device/iot-device-samples/file-upload-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/FileUploadSample.java
abhipsaMisra: Isn't this a good information to relay anyway, that they are free to use any version of the storage sdk with our clients?

abhipsaMisra (approved):

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTaskTest.java
timtay-microsoft: It was testing if the library threw a certain exception, and the new version of the library doesn't throw it anymore

timtay-microsoft:
> Will this come with any behavior changes? Should we consider a major version rev?

Nope. The behavior is still the same

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/serviceclient/ExportImportTests.java
timtay-microsoft: Good catch, it should be. I'll fix it

PR #1163, opened 2021-03-23
refactor(iot-dev, deps, iot-service): Fix a few miscellaneous warnings

timtay-microsoft:
Unused methods, unused variables, and a few other fixes

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpDeviceOperations.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpDeviceOperations.java
@@ -168,9 +168,4 @@ public class AmqpDeviceOperations
     {
         this.amqpProperties.put(Symbol.getSymbol(key), value);
     }
-
-    protected void clearAmqpLinkProperty()
-    {
-        this.amqpProperties.clear();
-    }
 }
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketImpl.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketImpl.java
@@ -234,7 +234,6 @@ public class WebSocketImpl implements WebSocket, TransportLayer
 
         private void sendToUnderlyingInput()
         {
-            boolean _readComplete =  false;
             switch (_lastType)
             {
                 case WEB_SOCKET_MESSAGE_TYPE_UNKNOWN:
@@ -254,7 +253,6 @@ public class WebSocketImpl implements WebSocket, TransportLayer
 
                     _wsInputBuffer.compact();
                     _wsInputBuffer.flip();
-                    _readComplete = true;
                     break;
                 case WEB_SOCKET_MESSAGE_TYPE_CLOSE:
                     _wsInputBuffer.flip();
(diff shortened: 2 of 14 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/amqp/AmqpDeviceOperations.java
```diff
@@ -168,9 +168,4 @@ protected void addAmqpLinkProperty(String key, String value)
     {
         this.amqpProperties.put(Symbol.getSymbol(key), value);
     }
-
-    protected void clearAmqpLinkProperty()
```
timtay-microsoft, line 172: Unused in our code, and inaccessible to users of this library, so it is safe to remove

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketImpl.java
```diff
@@ -234,7 +234,6 @@ private void readInputBuffer()
 
         private void sendToUnderlyingInput()
         {
-            boolean _readComplete =  false;
```
timtay-microsoft, line 237: This variable is assigned to, but never read

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -41,6 +41,7 @@
  * </p>
  * The client supports HTTPS 1.1 and AMQPS 1.0 transports.
  */
+@SuppressWarnings("deprecation") // A lot of references to deprecated APIs, but all of them are still in use
```
timtay-microsoft, line 44: Moving this suppression to the class level because we have 10+ references to the now deprecated TransportClient and other deprecated fields

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1597,6 +1597,7 @@ private void addToCallbackQueue(IotHubTransportPacket packet)
 
     private DeviceClientConfig getDefaultConfig()
     {
+        //noinspection LoopStatementThatDoesntLoop
```
timtay-microsoft, line 1600: The alternative is to write out the full ```this.deviceClientConfigs.values().iterator().hasNext()``` and so on, and the current implementation seems easier to read

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
jamdavi, line 44: I think I'm okay with this, but does this also suppress warnings when we use depreciated members from other packages?

timtay-microsoft: It would suppress them, yes. That's the one drawback to doing this

jamdavi: I think this is alright. Will this suppress when we use other APIs that are depreciated from 3rd parties?

jamdavi (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
jamdavi: It doesn't seem like we change the called APIs very often. I'm sure compilation or our tests will catch when a third party implementation breaks.

Issue #1164, opened 2021-03-24
[Bug Report] IoT Hub device client MQTT layer does not use a renewed SAS token when reconnecting


timtay-microsoft:
This is a regression that only affects device client versions 1.29.2, 1.29.3, and 1.29.4

When reconnecting after a previous SAS token has expired, the above device client versions will attempt to connect using the old SAS token rather than creating a new one. As a result, reconnection will fail with an Unauthorized exception

AB#9588401

timtay-microsoft:
I'll get the fix for this in ASAP

timtay-microsoft:
Closing this issue as the fix has been released as of device client version 1.30.0

PR #1165, opened 2021-03-24
fix(iot-dev): Fix a (regression) bug where device client's MQTT layer did not use a …

timtay-microsoft:
…renewed SAS token when reconnecting

#1164

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
@@ -100,25 +100,14 @@ public class MqttIotHubConnection implements IotHubTransportConnection, MqttMess
             throw new TransportException("Failed to get SSLContext", e);
         }
 
-        char[] password = null;
         if (this.config.getAuthenticationType() == DeviceClientConfig.AuthType.SAS_TOKEN)
         {
             log.trace("MQTT connection will use sas token based auth");
-            try
-            {
-                password = this.config.getSasTokenAuthentication().getSasToken();
-            }
-            catch (IOException e)
-            {
-                throw new TransportException("Failed to get sas token", e);
-            }
-
             this.webSocketQueryString = NO_CLIENT_CERT_QUERY_STRING;
         }
         else if (this.config.getAuthenticationType() == DeviceClientConfig.AuthType.X509_CERTIFICATE)
         {
             log.trace("MQTT connection will use X509 certificate based auth");
-            password = null;
         }
 
         //URLEncoder follows HTML spec for encoding urls, which includes substituting space characters with '+'
@@ -180,7 +169,6 @@ public class MqttIotHubConnection implements IotHubTransportConnection, MqttMess
             this.serverUri = SSL_PREFIX + host + SSL_PORT_SUFFIX;
         }
 
-        MqttAsyncClient mqttAsyncClient = buildMqttAsyncClient(this.serverUri, clientId);
         MqttConnectOptions connectOptions = new MqttConnectOptions();
         connectOptions.setKeepAliveInterval(KEEP_ALIVE_INTERVAL);
         connectOptions.setCleanSession(SET_CLEAN_SESSION);
@@ -214,17 +202,11 @@ public class MqttIotHubConnection implements IotHubTransportConnection, MqttMess
             connectOptions.setSocketFactory(sslContext.getSocketFactory());
         }
 
-        if (password != null && password.length > 0)
-        {
-            connectOptions.setPassword(password);
-        }
-
         // these variables are shared between the messaging, twin and method subclients
         Map<Integer, Message> unacknowledgedSentMessages = new ConcurrentHashMap<>();
         Queue<Pair<String, byte[]>> receivedMessages = new ConcurrentLinkedQueue<>();
 
         this.deviceMessaging = new MqttMessaging(
-            mqttAsyncClient,
             deviceId,
             this,
             moduleId,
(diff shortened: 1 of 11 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -100,25 +100,14 @@ public MqttIotHubConnection(DeviceClientConfig config) throws IllegalArgumentExc
             throw new TransportException("Failed to get SSLContext", e);
         }
 
-        char[] password = null;
         if (this.config.getAuthenticationType() == DeviceClientConfig.AuthType.SAS_TOKEN)
         {
             log.trace("MQTT connection will use sas token based auth");
-            try
-            {
-                password = this.config.getSasTokenAuthentication().getSasToken();
```
timtay-microsoft, line 109: We can't just set the password once in the constructor since this MqttIotHubConnection object will get opened and closed multiple times. This was the main cause of the bug

timtay-microsoft: I moved this logic to happen at open time instead, so that every time we open a new connection, a new SAS token will be used

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -180,7 +169,6 @@ else if (this.config.getAuthenticationType() == DeviceClientConfig.AuthType.X509
             this.serverUri = SSL_PREFIX + host + SSL_PORT_SUFFIX;
         }
 
-        MqttAsyncClient mqttAsyncClient = buildMqttAsyncClient(this.serverUri, clientId);
```
timtay-microsoft, line 183: MqttAsyncClients are single use, so instead of constructing one at MqttIotHubConnection constructor time, we'll create one each time we open the MqttIotHubConnection

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -273,6 +251,27 @@ public void open() throws TransportException
...
+                }
+            }
+
+            // MqttAsyncClient's are unusable after they have been closed. This logic creates a new client
+            // each time an open is called
+            MqttAsyncClient mqttAsyncClient = buildMqttAsyncClient(this.serverUri, this.clientId);
+            mqttAsyncClient.setCallback(this.deviceMessaging);
+            this.deviceMessaging.setMqttAsyncClient(mqttAsyncClient);
+            this.deviceTwin.setMqttAsyncClient(mqttAsyncClient);
+            this.deviceMethod.setMqttAsyncClient(mqttAsyncClient);
```
timtay-microsoft, line 273: As noted above, we only assign the (single use) mqtt async client at open time, so there is no risk of an mqtt async client calling open again after it was already closed

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/TokenRenewalTests.java
```diff
@@ -102,7 +102,9 @@ public static void stopProxy()
      * @throws Exception if the test fails in any way
      */
     @Test
-    @ContinuousIntegrationTest
+    // While this test is long enough to be considered a continuous integration test, it is the only test that truly validates
+    // that the device client's token renewal logic works. Because of that, it needs to be run at the gate.
+    //@ContinuousIntegrationTest
     public void tokenRenewalWorks() throws Exception
```
timtay-microsoft, line 108: This is a 15 minute test, just FYI. I don't love putting this in the gate, but since this regression slipped by in part because this test isn't at the gate, it feels wise to add it back to the gate

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -273,6 +251,27 @@ public void open() throws TransportException
 
             log.debug("Opening MQTT connection...");
 
+            if (this.config.getSasTokenAuthentication() != null)
+            {
+                try
+                {
+                    log.trace("Setting password for MQTT connection since it is a SAS token authenticated connection");
```
drwill-ms, line 258: Is `open()` called internally each time after a disconnect due to SAS expiration?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/TokenRenewalTests.java
```diff
@@ -102,7 +102,9 @@ public static void stopProxy()
      * @throws Exception if the test fails in any way
      */
     @Test
-    @ContinuousIntegrationTest
+    // While this test is long enough to be considered a continuous integration test, it is the only test that truly validates
+    // that the device client's token renewal logic works. Because of that, it needs to be run at the gate.
+    //@ContinuousIntegrationTest
```
drwill-ms, line 107: Might as well remove this tag then?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
timtay-microsoft: Yes

drwill-ms (approved):

PR #1168, opened 2021-03-30
fix(device): encode ModelId during MQTT connect

drwill-ms:
Same as [this change](https://github.com/Azure/azure-iot-sdk-csharp/commit/666df0e5607fdd9d462cdf4227afc54342bf7dba) in the .NET repo, the device's model Id should be URL encoded before being added to the URI, because it is user-supplied data and could potentially have characters that would break the URI.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
@@ -121,8 +121,8 @@ public class MqttIotHubConnection implements IotHubTransportConnection, MqttMess
             password = null;
         }
 
-        //URLEncoder follows HTML spec for encoding urls, which includes substituting space characters with '+'
-        // We want "%20" for spaces, not '+', however, so replace them manually after utf-8 encoding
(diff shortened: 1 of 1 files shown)


azabbasi (approved):

vinagesh (approved):

barustum (approved):

timtay-microsoft (approved):

bikamani (approved):

timtay-microsoft:
I'm not sure if it is relevant for model ids, but take note that URLEncoder doesn't behave exactly the same way in Java that the .NET encoder does. Above your code is an example where we have a special handling of space characters. If model ids can't have spaces, then what you have should be fine

```java
//URLEncoder follows HTML spec for encoding urls, which includes substituting space characters with '+'
// We want "%20" for spaces, not '+', however, so replace them manually after utf-8 encoding
String userAgentString = this.config.getProductInfo().getUserAgentString();
String clientUserAgentIdentifier;
try
{
    clientUserAgentIdentifier = "DeviceClientType=" + URLEncoder.encode(userAgentString, StandardCharsets.UTF_8.name()).replaceAll("\\+", "%20");
}
catch (UnsupportedEncodingException e)
{
    throw new TransportException("Failed to get URLEncode the user agent string", e);
}
```

drwill-ms:
> If model ids can't have spaces, then what you have should be fine

Shouldn't have spaces, but it is user-provided data so it **could** be anything. I'll update the code.

PR #1172, opened 2021-04-13
fix(ado): Rerun failed tests up to twice if necessary

timtay-microsoft:
This change unblocks the gate while we work to reduce the number of registry operations done in a typical gated test run.

For additional context if you missed why the gate is blocked, IoT Hub has started responding much slower to registry operations as of late, and the load of registry operations imposed by our tests is pretty high, so a lot of tests would just time out when trying to create a device to run a test on. Re-running these timed out tests after all other tests have finished seems to resolve this issue for us.

### DIFF
--- a/iot-e2e-tests/common/pom.xml
+++ b/iot-e2e-tests/common/pom.xml
@@ -139,6 +139,7 @@
                     </execution>
                 </executions>
                 <configuration>
+                    <rerunFailingTestsCount>2</rerunFailingTestsCount>
                     <includes>
                         <!--By default, only test files that end in ___Test.java are recognized as tests. This statement extends that to ___Tests.java files-->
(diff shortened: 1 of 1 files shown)


drwill-ms (approved):

iot-e2e-tests/common/pom.xml
```diff
@@ -139,6 +139,7 @@
                     </execution>
                 </executions>
                 <configuration>
+                    <rerunFailingTestsCount>2</rerunFailingTestsCount>
```
bikamani, line 142: Is this more of a temporary fix while service issue is resolved?

timtay-microsoft: Hopefully, yes.

PR #1174, opened 2021-04-15
refactor(ado): Recycle test identities

timtay-microsoft:
rather than deleting a given test identity at the end of a test, they can now be recycled and reused by a later test. The RECYCLE_TEST_IDENTITIES environment variable determines if the identities are recycled or deleted, and defaults to deleted. Our gates will recycle identities though.

This change is towards the broader effort of reducing registry operations performed by our gates


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/IntegrationTest.java
```diff
@@ -77,6 +77,9 @@ protected void finished(Description description)
     public static boolean runProvisioningTests = Boolean.parseBoolean(Tools.retrieveEnvironmentVariableValue("RUN_PROVISIONING_TESTS", "true"));
     public static boolean runDigitalTwinTests = Boolean.parseBoolean(Tools.retrieveEnvironmentVariableValue("RUN_DIGITAL_TESTS", "true"));
 
+    // Determines if the tear down for a given test should delete the device identity, or recycle it so that another test can use it
+    public static boolean recycleIdentities = Boolean.parseBoolean(Tools.retrieveEnvironmentVariableValue("RECYCLE_TEST_IDENTITIES", "false"));
```
timtay-microsoft, line 81: The "false" here is just the default value. I figure people running test locally would rather the test delete the identity for them

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/FileUploadTests.java
```diff
@@ -256,7 +256,7 @@ public static void stopProxy()
 
     private DeviceClient setUpDeviceClient(IotHubClientProtocol protocol) throws URISyntaxException, InterruptedException, IOException, IotHubException, GeneralSecurityException
     {
-        TestDeviceIdentity testDeviceIdentity = Tools.getTestDevice(iotHubConnectionString, protocol, testInstance.authenticationType);
+        TestDeviceIdentity testDeviceIdentity = Tools.getTestDevice(iotHubConnectionString, protocol, testInstance.authenticationType, false);
```
timtay-microsoft, line 259: All the ", false" additions here just say that the returned test identity can be an identity with modified twin. Most tests don't care if their identity has any desired or reported properties

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/TransportClientTests.java
```diff
@@ -210,21 +210,16 @@ public static void tearDown() throws IOException
         }
     }
 
-    @Before
-    public void setupTestInstance() throws InterruptedException, IotHubException, URISyntaxException, IOException, GeneralSecurityException
-    {
-        testInstance.setup();
```
timtay-microsoft, line 216: Moving this to a per-test basis since at least one test in this suite requires a few test identities with no twin, while the rest are fine with using a modified twin.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/errorinjection/SendMessagesErrInjTests.java
```diff
@@ -296,32 +296,16 @@ public void sendMessagesWithTcpConnectionDropNotifiesUserIfRetryExpires() throws
     private void errorInjectionTestFlowNoDisconnect(Message errorInjectionMessage, IotHubStatusCode expectedStatus, boolean noRetry) throws IOException, IotHubException, URISyntaxException, InterruptedException, ModuleClientException, GeneralSecurityException
     {
         // Arrange
-        Device targetDevice = null;
```
timtay-microsoft, line 299: Missed this in an earlier PR, but this logic is within SendMessagesCommon already, so no need to setup a new test identity here as well

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -137,99 +154,320 @@ public static RegistryManager getRegistyManager(String iotHubConnectionString) t
...
+     * @param protocol The device side protocol for the client to use.
+     * @param authenticationType The device side authentication type for the client to use.
+     * @param needCleanTwin True if the returned device identity needs to not have any pre-existing desired or reported properties, false otherwise.
+     * @return A {@link TestDeviceIdentity} that was either recycled from a previous test when possible, or was just created for this test.
+     * @throws URISyntaxException If the connection string cannot be parsed.
+     * @throws IOException If the registry addition of a device fails.
+     * @throws IotHubException If the registry addition of a device fails.
+     * @throws GeneralSecurityException If creating the x509 certificates for an x509 device fails.
+     */
+    public static TestDeviceIdentity getTestDevice(String iotHubConnectionString, IotHubClientProtocol protocol, AuthenticationType authenticationType, boolean needCleanTwin) throws URISyntaxException, IOException, IotHubException, GeneralSecurityException
```
timtay-microsoft, line 195: This is admittedly a bit of a dense function, but it can be read like:

- if a cached device with twin modifications is available, and the test doesn't require a clean twin, then return that device.
- else if a cached device with clean twin is available, return that device
- else create a new batch of devices. Return one, and cache the rest for future callers of this method

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -137,99 +154,320 @@ public static RegistryManager getRegistyManager(String iotHubConnectionString) t
...
+     * @param protocol The device side protocol for the client to use.
+     * @param authenticationType The device side authentication type for the client to use.
+     * @param needCleanTwin True if the returned module identity needs to not have any pre-existing desired or reported properties, false otherwise.
+     * @return A {@link TestDeviceIdentity} that was either recycled from a previous test when possible, or was just created for this test.
+     * @throws URISyntaxException If the connection string cannot be parsed.
+     * @throws IOException If the registry addition of a module fails.
+     * @throws IotHubException If the registry addition of a module fails.
+     * @throws GeneralSecurityException If creating the x509 certificates for an x509 module fails.
+     */
+    public static TestModuleIdentity getTestModule(String iotHubConnectionString, IotHubClientProtocol protocol, AuthenticationType authenticationType, boolean needCleanTwin) throws URISyntaxException, IOException, IotHubException, InterruptedException, ModuleClientException, GeneralSecurityException
```
timtay-microsoft, line 321: Similar to getTestDevice, except there is no batch add API for modules, so instead of adding 100 at a time, it just creates the identities at the time that they are requested without caching any for future callers. Modules only get cached when a test finished up with them.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -137,99 +154,320 @@ public static RegistryManager getRegistyManager(String iotHubConnectionString) t
...
+            }
         }
     }
 
+    /**
+     * If the environment variable RECYCLE_TEST_IDENTITIES is set to true, then the identity will be recycled. Otherwise it
+     * will be removed from the IoT Hub's registry.
+     * @param testIdentity The test identity to dispose/recycle.
+     */
     public static void disposeTestIdentity(TestIdentity testIdentity, String iotHubConnectionString)
```
timtay-microsoft, line 393: Again, a bit dense, but since there are several queues of device and module identities, this function sorts the given testIdentity into the right queue

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ExportImportDeviceParser.java
```diff
@@ -16,6 +16,11 @@
     @SerializedName(ID_NAME)
     private String id;
 
+    private static final String MODULE_ID_NAME = "moduleId";
+    @Expose
+    @SerializedName(MODULE_ID_NAME)
+    private String moduleId;
```
timtay-microsoft, line 22: Adding this so that the e2e tests can start adding modules in bulk as well

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
vinagesh: With this PR, you have added support for bulk device and module add right? That is what the code looked like below but wanted to double check. That would be worth mentioning in our next release but if it goes in in this PR, we might forget to mention it. Maybe you should add that to the commit message or something.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -137,99 +154,440 @@ public static RegistryManager getRegistyManager(String iotHubConnectionString) t
...
+        }
+    }
+
+    private static void disposeTestDevices(Iterable<Device> devices, String iotHubConnectionString)
+    {
+        List<Device> subList = new ArrayList<>();
+        for (Device device : devices)
+        {
+            // Bulk device removal is limited to, at most, 100 devices per request. Create a list of 100 devices,
+            // delete them, and then create another list of 100 devices until all devices have been deleted
```
vinagesh, line 563: typo - delete them and then ~~create~~delete another

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -137,99 +154,440 @@ public static RegistryManager getRegistyManager(String iotHubConnectionString) t
...
+            }
+        }
+    }
+
+    private static void disposeTestDevices(Iterable<Device> devices, String iotHubConnectionString)
+    {
+        List<Device> subList = new ArrayList<>();
+        for (Device device : devices)
+        {
+            // Bulk device removal is limited to, at most, 100 devices per request. Create a list of 100 devices,
```
vinagesh, line 562: typo ~~create~~ Delete a list of 100 devices

vinagesh: Or maybe delete in batches of 100 devices

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -137,99 +154,440 @@ public static RegistryManager getRegistyManager(String iotHubConnectionString) t
...
+            // delete them, and then create another list of 100 devices until all devices have been deleted
+            subList.add(device);
```
vinagesh, line 564: Nit: More of a personal preference with respect to code clarity -
```
if(subList.size() <= 100)
{
   subList.add(device);
}
else
{
}
```

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -262,6 +620,153 @@ public static Device addDeviceWithRetry(RegistryManager registryManager, Device
...
+            parsers.add(exportImportDevice);
+        }
+
+        ExportImportDevicesParser body = new ExportImportDevicesParser();
+        body.setExportImportDevices(parsers);
+
+        String sasTokenString = new IotHubServiceSasToken(iotHubConnectionString).toString();
+
+        HttpRequest request = new HttpRequest(url, HttpMethod.POST, body.toJson().getBytes());
+        request.setReadTimeoutMillis(IntegrationTest.HTTP_READ_TIMEOUT);
```
vinagesh, line 663: Would be good to extract all this common request stuff to a helper method.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -137,99 +154,440 @@ public static RegistryManager getRegistyManager(String iotHubConnectionString) t
         return registryManager;
     }
 
-    public static TestDeviceIdentity getTestDevice(String iotHubConnectionString, IotHubClientProtocol protocol, AuthenticationType authenticationType) throws URISyntaxException, IOException, IotHubException, GeneralSecurityException, InterruptedException
+    private static Queue<TestDeviceIdentity> testSasDeviceQueue = new ConcurrentLinkedQueue<>();  // queue contains SAS device identities that are ready to be re-used, and that have no twin modifications yet
+    private static Queue<TestDeviceIdentity> testX509DeviceQueue = new ConcurrentLinkedQueue<>(); // queue contains x509 device identities that are ready to be re-used, and that have no twin modifications yet
+    private static Queue<TestDeviceIdentity> testSasDeviceWithTwinQueue = new ConcurrentLinkedQueue<>(); // queue contains SAS device identities that are ready to be re-used, and that have some twin modifications to them already
```
vinagesh, line 159: If the number of tests that need a fresh twin is very low, I'm wondering if we really need 3 queues. We could probably just get a new identity instead of adding more logic to this class.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
vinagesh: Scratch that. Looks like you only added it to tools but there is no exposed API for customers to call yet.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -137,99 +154,440 @@ public static RegistryManager getRegistyManager(String iotHubConnectionString) t
...
+                        testX509ModuleQueue.add((TestModuleIdentity) testIdentity);
+                    }
+                }
+            }
+        }
+    }
+
+    private static void disposeTestDevices(Iterable<Device> devices, String iotHubConnectionString)
+    {
+        List<Device> subList = new ArrayList<>();
```
vinagesh, line 559: It would be good to print out the total devices being disposed just to know how many we created for the tests.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -137,99 +154,440 @@ public static RegistryManager getRegistyManager(String iotHubConnectionString) t
...
-            throw new IllegalArgumentException("Test code has not been written for this path yet");
+            List<Device> devicesToDelete = new ArrayList<>();
+            for (TestIdentity testIdentity : testIdentities)
+            {
+                devicesToDelete.add(testIdentity.getDevice());
+            }
+
+            // No need to dispose of modules in particular.
+            // Deleting a device that has a module will delete both the device and module.
+            disposeTestDevices(devicesToDelete, iotHubConnectionString);
```
vinagesh, line 467: I think I'm mis-reading something here. If RECYCLE_TEST_IDENTITIES is set to true, why are we still removing all the test devices?

vinagesh (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
timtay-microsoft: I'm not sure I follow how you picture this change. This if statement is within a foreach loop, so I always want to add each device to the sublist. It's just when the sublist gets big enough that something different happens

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
timtay-microsoft: Unfortunately the number of tests that need a fresh twin is rather high. I'd estimate that we have ~400 test runs that assume no initial twin is present. We have 

number of twin operations (desired property tests, query, reported property, getTwin, fault injection for each as well)

times

number of twin protocols (4)

times

device client vs module client (2)

timtay-microsoft: The real benefit here is to allow an identity used in a twin test to be re-used in a telemetry/methods/c2d message test

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
timtay-microsoft: Oops, I had this backwards. Line 450 should have been "if recycleIdentities then dispose each identity" where the method call to dispose test identity just sticks each identity into the appropriate recycling queue.

PR #1175, opened 2021-04-15
Fix the remainder of actionable warnings.

azabbasi:


.gitignore
```diff
@@ -52,7 +52,6 @@ x64/
 *.vssscc
 .builds
 *.pidb
-*.log
```
azabbasi, line 55: redundant

vinagesh (approved):

barustum (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/generated/implementation/DigitalTwinsImpl.java
```diff
@@ -210,8 +200,7 @@ public Void call(ServiceResponseWithHeaders<Void, DigitalTwinUpdateDigitalTwinHe
         }
         Validator.validate(digitalTwinPatch);
         final String apiVersion = "2020-09-30";
-        final String ifMatch = null;
```
timtay-microsoft, line 213: This one worries me a bit. The change is fine, but please follow up with the PnP swagger folks as to why the if match header can't be set here. This seems like that issue we had in ADT land where the if-none-match header was always set to "*" because it only had one value defined in the swagger

timtay-microsoft (approved):

PR #1184, opened 2021-04-22
Merge AAD auth support from feature branch into master

timtay-microsoft:
This won't be merged until it is closer to time to release this feature, as the service itself doesn't support this in every region just yet


vsts/sdl.yaml
```diff
@@ -20,6 +20,7 @@ phases:
     displayName: CredScan
     inputs:
       debugMode: false
+      suppressionsFile: vsts/CredScanSuppressions.json
```
timtay-microsoft, line 23: This PR brought new samples with it, and one of the samples has a false postive credscan case that this suppressions file accounts for

barustum (approved):

Issue #1182, opened 2021-04-22
[feature request] Device Client File upload APIs don't include IoT Hub status codes in thrown exceptions


sanjay-90210:
**Is your feature request related to a problem? Please describe.**

While using DeviceClient.getFileUploadSasUri, the real cause of the issue is hidden and a generic IOException with the message "Sas URI response message had no payload" is thrown. This hides the real issue that caused the empty payload and so programmatically addressing the exception is a lot of guess work or near impossible.

The above exception can be thrown in the following scenarios that I am running into:
1. A storage container is not associated at the iot hub end for file uploads. ([link](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-file-upload#associate-an-azure-storage-account-with-iot-hub))

     - This gets mapped to HTTP 400 - BAD_FORMAT

2. The number of outstanding SAS URIs may be more than that is allowed by the quota and limits ([link](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#iot-hub-limits))

**File upload operations** | 83.33 file upload initiations/sec/unit (5,000/min/unit) (for S3),  1.67 file upload initiations/sec/unit (100/min/unit) (for S1 and S2).  10,000 SAS URIs can be out for an Azure Storage account at one time. **10 SAS URIs/device can be out at one time.**

       - This gets mapped to HTTP 403 - TOO_MANY_DEVICES
 
3. The device may no longer exist just before the SAS URI is requested. The device may be deleted or disabled.

      - This gets mapped to HTTP 404 - HUB_OR_DEVICE_ID_NOT_FOUND

In all the above cases, the only exception that I get is a vanilla IOException that says "Sas URI response message had no payload" and so cannot take any situation specific action.

Even though the IotHubStatusCode that reports these specific cases reach all the way till the ResponseMessage in the code below, it is ignored and a plain IOException is thrown from the function.

```
    public FileUploadSasUriResponse getFileUploadSasUri(FileUploadSasUriRequest request) throws IOException
    {
        IotHubTransportMessage message = new IotHubTransportMessage(request.toJson());
        message.setIotHubMethod(IotHubMethod.POST);

        ResponseMessage responseMessage;
        httpsTransportManager.open();
        responseMessage = httpsTransportManager.getFileUploadSasUri(message);
        httpsTransportManager.close();

        if (responseMessage.getBytes() == null || responseMessage.getBytes().length == 0)
        {
[...]

PR #1186, opened 2021-04-23
fix(iot-dev): Make file upload APIs throw exceptions with the IoT Hub status code in them

timtay-microsoft:
Previously these methods would only throw if an IOException occurred, and would ignore the IoT Hub service level status codes. This made it impossible to tell if their requests were being throttled, for instance.

#1182 for additional context


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTask.java
```diff
@@ -181,12 +182,29 @@ public FileUploadSasUriResponse getFileUploadSasUri(FileUploadSasUriRequest requ
...
+
+        IotHubServiceException serviceException =
+            IotHubStatusCode.getConnectionStatusException(
+                responseMessage.getStatus(),
+                responseMessagePayload);
+
+        // serviceException is only not null if the provided status code was a non-successful status code like 400, 429, 500, etc.
+        if (serviceException != null)
+        {
+            throw new IOException("Failed to get the file upload SAS URI", serviceException);
```
timtay-microsoft, line 199: This method only declares that is throws IOExceptions, so the closest thing we can do to throwing an IotHubServiceException is to nest it within an IOException.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTask.java
```diff
@@ -199,9 +208,31 @@ public IotHubStatusCode sendNotification(FileUploadCompletionNotification fileUp
...
+
+        IotHubServiceException serviceException =
+            IotHubStatusCode.getConnectionStatusException(
+                responseMessage.getStatus(),
+                responseMessagePayload);
+
+        // serviceException is only not null if the provided status code was a non-successful status code like 400, 429, 500, etc.
+        if (serviceException != null)
+        {
+            throw new IOException(errorMessage, serviceException);
```
timtay-microsoft, line 232: These file upload methods only declare that they throw IOExceptions, so the closest thing we can do to throwing an IotHubServiceException is to nest it within an IOException.

bikamani, line 232: Is it right to throw IOException here? I understand you may not want to change behavior but shouldn't it be wrapped in IoTHub Exception type?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTask.java
```diff
@@ -199,9 +208,31 @@ public IotHubStatusCode sendNotification(FileUploadCompletionNotification fileUp
         ResponseMessage responseMessage = httpsTransportManager.sendFileUploadNotification(message);
         httpsTransportManager.close();
 
+        validateServiceStatusCode(responseMessage, "Failed to complete the file upload notification");
```
bikamani, line 211: Is this to add additional context to why 403 for example?
Does service not return error message/description?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTask.java
```diff
@@ -199,9 +208,31 @@ public IotHubStatusCode sendNotification(FileUploadCompletionNotification fileUp
...
+    private void validateServiceStatusCode(ResponseMessage responseMessage, String errorMessage) throws IOException
+    {
+        String responseMessagePayload = null;
+        if (responseMessage.getBytes() != null && responseMessage.getBytes().length > 0)
+        {
+            responseMessagePayload = new String(responseMessage.getBytes(), DEFAULT_IOTHUB_MESSAGE_CHARSET);
+        }
+
+        IotHubServiceException serviceException =
+            IotHubStatusCode.getConnectionStatusException(
```
bikamani, line 225: What happens when payload is null?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTask.java
```diff
@@ -199,9 +208,31 @@ public IotHubStatusCode sendNotification(FileUploadCompletionNotification fileUp
         ResponseMessage responseMessage = httpsTransportManager.sendFileUploadNotification(message);
         httpsTransportManager.close();
 
+        validateServiceStatusCode(responseMessage, "Failed to complete the file upload notification");
+
         return responseMessage.getStatus();
     }
 
+    private void validateServiceStatusCode(ResponseMessage responseMessage, String errorMessage) throws IOException
```
abhipsaMisra, line 216: why not have this method return the responseMessagePayload so that it doesn't need to be calculated again in L188?

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTaskTest.java
```diff
@@ -764,4 +771,67 @@ public void runSendNotificationThrows() throws IOException, IllegalArgumentExcep
...
+                mockResponseMessage.getStatus();
+                result = IotHubStatusCode.UNAUTHORIZED;
+            }
+        };
+
+        // act
+        try
+        {
+            fileUploadTask.getFileUploadSasUri(mockFileUploadSasUriRequest);
+            fail("Test should have thrown an IOException with a nested BadFormatException");
```
abhipsaMisra, line 797: typo:  BadFormatException -> UnauthorizedException

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTaskTest.java
```diff
@@ -764,4 +771,67 @@ public void runSendNotificationThrows() throws IOException, IllegalArgumentExcep
...
+            VALID_BLOB_NAME, mockInputStream, VALID_STREAM_LENGTH, mockHttpsTransportManager, mockIotHubEventCallback, VALID_CALLBACK_CONTEXT);
+
+        new Expectations()
+        {
+            {
+                mockHttpsTransportManager.getFileUploadSasUri((IotHubTransportMessage) any);
+                result = mockResponseMessage;
+
+                mockResponseMessage.getStatus();
+                result = IotHubStatusCode.UNAUTHORIZED;
```
abhipsaMisra, line 789: this is just arbitrary for the test, right?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTask.java
abhipsaMisra: why not pull the specific error message out of the service exception returned, rather then it being generic text?

abhipsaMisra (approved):
A couple of comments, otherwise lgtm!

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTask.java
timtay-microsoft: You have it backwards. In order to not cause a breaking change, the IoTHubServiceException needs to be the nested exception within an IOException

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTask.java
timtay-microsoft: The nested IotHubServiceException is what will contain the important details. At this level, all the method knows is that completing the file upload notification failed

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTask.java
timtay-microsoft: Then the IotHubServiceException will not have an error message property. It will just be the exception type that contains useful information for the user

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTaskTest.java
timtay-microsoft: Yeah

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTaskTest.java
timtay-microsoft: Good catch!

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTask.java
timtay-microsoft: Sure, that seems reasonable

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUploadTask.java
timtay-microsoft: That specific error message is embedded within the nested IoTHubServiceException, rather than the top level IOException

Issue #1187, opened 2021-04-26
[Bug Report] MultiplexingClient closes connection if its only registered device drops its AMQP session


timtay-microsoft:
This check assumes that the set of configs being greater than 1 signals that the layer is multiplexing, but it is possible to open a multiplexed connection that only has 1 config in that list for a time.

https://github.com/Azure/azure-iot-sdk-java/blob/767389dd24124d825201602f1499da454bf0fddf/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java#L723

There should be a flag within the AMQP layer that signals whether it is multiplexing or not to avoid this type of assumption

timtay-microsoft:
Closing this issue as the fix has been released as of device client version 1.30.0

PR #1192, opened 2021-04-29
fix(iot-dev): Fix bug where amqp layer treated single device multiplexed connections as though they weren't multiplexed

timtay-microsoft:
A multiplexed connection shouldn't drop just because one of its devices dropped its session, even if that is the only session.

#1187

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -99,7 +99,9 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
     private final Set<DeviceClientConfig> multiplexingClientsToRegister;
     private final Set<DeviceClientConfig> multiplexingClientsToUnregister;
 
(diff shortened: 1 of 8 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -688,7 +694,7 @@ public void onMessageReceived(IotHubTransportMessage message)
     @Override
     public void onAuthenticationFailed(String deviceId, TransportException transportException)
     {
-        if (this.deviceClientConfigs.size() > 1)
```
timtay-microsoft, line 691: This line was the source of this bug. The assumption previously was that if there was only one device session on the connection, that the connection was not multiplexing. However, you can open a multiplexed connection with only one device session so this assumption was wrong

timtay-microsoft: For multiplexed connections, we don't drop the full connection just because a single device session drops, so it makes sense to do the same even when there is only one device registered

vinagesh (approved):

PR #1194, opened 2021-04-30
feat(service-client): Add support for user defined msi in import export jobs.

vinagesh:
# Checklist

The JobProperties can now have a new property called 'identity'. If we pass the correct identity in the import/export jobs, the IoThub will use it internally to interact with the storage account to do things like upload devices.txt. From the SDK's perspective, we support this new property and the capability to use it in import and export jobs.
This PR has:
* Adding the new identity property to job properties
* Sample to show the usage of this in import and export jobs

This PR is against a feature branch. There are no gates enabled yet because the feature is only available on a whitelisted subscription.

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
@@ -6,6 +6,7 @@
 package com.microsoft.azure.sdk.iot.service;
 
 import com.microsoft.azure.sdk.iot.deps.serializer.JobPropertiesParser;
+import com.microsoft.azure.sdk.iot.deps.serializer.ManagedIdentity;
 import com.microsoft.azure.sdk.iot.deps.serializer.StorageAuthenticationType;
 
 import java.util.Date;
@@ -206,6 +207,22 @@ public class JobProperties
         this.excludeKeysInExport = excludeKeysInExport;
     }
 
+    /**
+     * @return The managed identity used to access the storage account for import and export jobs.
+     * For more information, see TODO
+     */
+    public ManagedIdentity getIdentity() {
+        return identity;
+    }
+
+    /**
+     * @param identity The managed identity used to access the storage account for import and export jobs.
+     * For more information, see TODOe
+     */
+    public void setIdentity(ManagedIdentity identity) {
+        this.identity = identity;
+    }
+
     public enum JobStatus
     {
         UNKNOWN,
@@ -237,6 +254,7 @@ public class JobProperties
     private boolean excludeKeysInExport;
     private String failureReason;
     private StorageAuthenticationType storageAuthenticationType;
+    private ManagedIdentity identity;
 
     /**
      * Constructs a new JobProperties object using a JobPropertiesParser object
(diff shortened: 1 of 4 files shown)


service/iot-service-samples/device-manager-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceManagerImoprtExportWithIdentitySample.java
```diff
@@ -0,0 +1,108 @@
...
+        // the job will fail.
+        // If StorageAuthenticationType is set to IdentityBased the userAssignedIdentity property is
+        // null, the jobs will use system defined identity. If the IoT hub is not configured with the
+        // user defined managed identity, the job will fail.
+        // If StorageAuthenticationType is set to IdentityBased and neither user defined nor system defined
+        // managed identities are configured on the hub, the job will fail.
+        ManagedIdentity identity = new ManagedIdentity();
+        identity.setUserAssignedIdentity(userDefinedManagedIdentityResourceId);
+
+        JobProperties jobProperties = new JobProperties();
```
vinagesh, line 53: Is there a better way to do this setting of job properties?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
```diff
@@ -303,14 +323,16 @@ JobPropertiesParser toJobPropertiesParser()
      *                              Including a SAS token is dependent on the StorageAuthenticationType
      * @param outputBlobContainerUri URI to a blob container. This is used to output the status of the job and the results.
      *                               Including a SAS token is dependent on the StorageAuthenticationType
+     * @param identity the managed identity used to access the storage account for import jobs.
      * @return An instance of JobProperties
      */
     public static JobProperties createForImportJob(
```
timtay-microsoft, line 329: Do customers call this method? If so, this looks like a breaking change

timtay-microsoft (approved):
Just the one comment, otherwise looks good

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/JobProperties.java
vinagesh: Why would it be a breaking change? It is just adding a new optional property. Customers can either set it or leave it alone.

timtay-microsoft: Java doesn't have optional parameters, so they can't leave it alone. If someone was calling this method with 2 parameters before, it wouldn't compile once you add a third parameter

timtay-microsoft: Typically when you add an optional parameter to a java method, you need to create a new method like

```java
public void myMethod(int arg1, int arg2)
{
    myMethod(arg1, arg2, 0)
}

public void myMethod(int arg1, int arg2, int arg3)
{

}
```

vinagesh: Ahh I think I get it now. Some how your comment did not show up at the right method. How about creating an overload for the createForImportJob and createForExportJob ?

timtay-microsoft (approved):

Issue #1188, opened 2021-04-27
[Bug Report] MultiplexingClient does not notify state properly


sanjay-90210:
# Context

- **OS and version used:** Ubuntu 16.04.7 LTS
- **Java runtime used:** openjdk version "1.8.0_282"
OpenJDK Runtime Environment (build 1.8.0_282-8u282-b08-0ubuntu1~16.04-b08)
OpenJDK 64-Bit Server VM (build 25.282-b08, mixed mode)
- **SDK version used:** iot-device-client 1.29.4

## Description of the issue

**Devices:**

    1. GOOD     - Enabled and working device with proper connect string
    2. DISABLED - Device is present in the iot hub but is disabled
    3. MISSING  - Device is not there in the iot hub, old connect string 
                    is in use

**Test program options:**

    1. Setting add.last.later system property allows the last device specified
        in the command line argument list of connect string to be added 15 secs
        later to simulate continued onboarding of devices while the system is
        running.
    
    2. Setting disable.start.empty system property may be used to create a 
        multiplexed client that is not started before adding any devices.

**Scenarios with pre-opened AMQP muliplexed connection,
all devices are added at the same time:**

    1. 2 GOOD devices
        - OK    - Add devices works fine, both device in CONNECTED state
        - BAD   - MultiplexingClient ConnectionStatusChangeCallback is not
                    called and so muxStatus is null

    2. 1 GOOD, 1 MISSING
        - OK    - GOOD device works fine and continues to be connected and
                    working
        - OK    - MISSING device is rejected with
                    MultiplexingClientDeviceRegistrationAuthenticationException
        - BAD   - MultiplexingClient ConnectionStatusChangeCallback is not
                    called and so muxStatus is null
        - BAD   - Exception reported while closing the mux client
       
```
  [azure-iot-sdk-ReactorRunner] WARN com.microsoft.azure.sdk.iot.device.transport.amqps.AmqpsIotHubConnection - Amqp connection was closed, creating a thread to notify transport layer
com.microsoft.azure.sdk.iot.device.exceptions.MultiplexingDeviceUnauthorizedException: One or more multiplexed devices failed to authenticate
    at com.microsoft.azure.sdk.iot.device.transport.amqps.AmqpsIotHubConnection.onAuthenticationFailed(AmqpsIotHubConnection.java:693)
[...]

PR #1198, opened 2021-05-06
fix(iot-dev): Fix bug where transport layer treated multiplexed connections with one or fewer devices as a non-multiplexed connection

timtay-microsoft:
The expected connection status callbacks were not executed due to this bug. For instance, opening a multiplexed connection with no devices registered wouldn't execute the multiplexing client level connection status callback.

This also fixes a similar bug where the multiplexing client level retry policy isn't used under the same conditions

#1188 for additional context

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -1238,7 +1238,7 @@ public class IotHubTransport implements IotHubListener
             reconnectionAttempts++;
 
             RetryPolicy retryPolicy;
-            if (this.deviceClientConfigs.size() != 1)
+            if (isMultiplexing)
             {
                 retryPolicy = multiplexingRetryPolicy;
             }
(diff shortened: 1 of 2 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1560,7 +1560,7 @@ private void updateStatus(IotHubConnectionStatus newConnectionStatus, IotHubConn
             log.debug("Invoking connection status callbacks with new status details");
             invokeConnectionStateCallback(newConnectionStatus, reason);
 
-            if (deviceClientConfigs.size() < 2 || newConnectionStatus != IotHubConnectionStatus.CONNECTED)
+            if (!isMultiplexing || newConnectionStatus != IotHubConnectionStatus.CONNECTED)
             {
```
abhipsaMisra, line 1564: is "newConnectionStatus" the state of the mux client => tcp connection?

timtay-microsoft: Yes. Within the context of this method, the ```newConnectionStatus``` argument is the TCP level connection

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1238,7 +1238,7 @@ private void reconnect(TransportException transportException)
             reconnectionAttempts++;
 
             RetryPolicy retryPolicy;
-            if (this.deviceClientConfigs.size() != 1)
```
timtay-microsoft, line 1241: Assuming that configs.size != 1 means multiplexing is an incorrect assumption, so that is corrected here. The ```isMultiplexing``` flag is set at constructor time and is determined higher up in the code. For instance, when a IotHubTransport instance is created by a multiplexingClient, this flag is set to true.

abhipsaMisra (approved):

PR #1202, opened 2021-05-07
fix(iot-dev): Fix a few small multiplexing bugs

timtay-microsoft:
- Fixed a bug where AmqpSessionHandler would report that a session was closing as expected when it was actually closing unexpectedly

- Fixed a bug where the IotHubTransport layer would close the entire multiplexed connection if a single multiplexed device session exhausted its retry attempts

- Fixed a bug where IotHubTransport layer would execute the connection status callback for an already disconnected multiplexed device when the user closed the multiplexed connection

- Fixed a log statement in AMQP layer

- Fixed a log statement in IotHubTransport layer so that it didn't report that the SDK was attempting a retry until after the retry decision has been made

Fixes some aspects of #1188


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1161,7 +1161,6 @@ private void reconnectDeviceSession(TransportException transportException, Strin
         while (this.deviceConnectionStates.get(deviceId) == IotHubConnectionStatus.DISCONNECTED_RETRYING
                 && !hasReconnectOperationTimedOut)
         {
-            log.trace("Attempting to reconnect device session: attempt {}", reconnectionAttempts);
```
timtay-microsoft, line 1164: The next lines of code determine if a retry attempt should even be made. It makes more sense to move this line after that decision is made so that it doesn't appear the SDK is retrying when it actually isn't

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1183,30 +1184,20 @@ private void reconnectDeviceSession(TransportException transportException, Strin
...
-            }
+            this.updateStatus(IotHubConnectionStatus.DISCONNECTED, IotHubConnectionStatusChangeReason.RETRY_EXPIRED, transportException, deviceId);
+            log.debug("Reconnection for device {} was abandoned due to the retry policy", deviceId);
         }
-        catch (DeviceClientException ex)
+        else if (this.hasOperationTimedOut(reconnectionStartTimeMillis))
         {
-            log.error("Encountered an exception while closing the client object, client instance should no longer be used as the state is unknown", ex);
-            this.updateStatus(IotHubConnectionStatus.DISCONNECTED, IotHubConnectionStatusChangeReason.COMMUNICATION_ERROR, transportException, deviceId);
+            this.updateStatus(IotHubConnectionStatus.DISCONNECTED, IotHubConnectionStatusChangeReason.RETRY_EXPIRED, transportException, deviceId);
```
timtay-microsoft, line 1194: Rather than call ```this.close(...)``` which closes the entire multiplexed connection, just update the state of the device that failed to reconnect.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -1080,7 +1080,7 @@ private void openAsync() throws TransportException
 
     private void closeAsync()
     {
-        log.trace("OpenAsync called for amqp connection");
+        log.trace("CloseAsync called for amqp connection");
```
timtay-microsoft, line 1083: This method is ```CloseAsync``` not ```OpenAsync```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -153,7 +156,7 @@ public void onSessionLocalOpen(Event e)
     public void onSessionRemoteClose(Event e)
     {
         Session session = e.getSession();
-        if (session.getLocalState() == EndpointState.ACTIVE)
+        if (session.getLocalState() == EndpointState.ACTIVE || !isClosing)
```
timtay-microsoft, line 159: In my testing, there were times where the session was closing unexpectedly when the local state of the session wasn't ACTIVE yet, so I'm adding this additional flag to ensure we know when a session closure is expected or not. Our transport layer handles unexpected session closures differently than expected session closures, so it's important to get this distinction correct

azabbasi (approved):
Do you think there is value in adding regression tests for these bug fixes?

PR #1203, opened 2021-05-07
refactor(e2e): Remove redundant multiplexing test, fix deviceTwinWithVersionTest to recycle identities

timtay-microsoft:
deviceTwinWithVersion tests were still creating and removing test identities on their own rather than going through the centralized function that allows for recycling of test identities

The multiplexing test is arguably redundant because we already have a test that you can open a mux connection with the max number of devices. The only difference is registering the devices before vs after open. Since we already have a test that makes sure that registering after open works just like registering before open, there is no need for this test.

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
@@ -88,7 +88,7 @@ import static junit.framework.TestCase.*;
 @RunWith(Parameterized.class)
 public class MultiplexingClientTests extends IntegrationTest
 {
-    private static final int DEVICE_MULTIPLEX_COUNT = 10;
+    private static final int DEVICE_MULTIPLEX_COUNT = 3;
 
     private static final int MESSAGE_SEND_TIMEOUT_MILLIS = 60 * 1000;
     private static final int FAULT_INJECTION_RECOVERY_TIMEOUT_MILLIS = 2 * 60 * 1000;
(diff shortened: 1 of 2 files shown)


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
```diff
@@ -88,7 +88,7 @@
 @RunWith(Parameterized.class)
 public class MultiplexingClientTests extends IntegrationTest
 {
-    private static final int DEVICE_MULTIPLEX_COUNT = 10;
```
timtay-microsoft, line 91: To help save on test identity creation, we can test muxed connections of 3 devices rather than 10

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
```diff
@@ -474,33 +474,6 @@ public void sendMessagesMaxDevicesAllowedTimes10MultiplexingClientsSerialOpen()
         log.debug("Close time: " + (finishCloseTime - startCloseTime) / 1000.0);
     }
 
-    @ContinuousIntegrationTest
-    @Test
-    public void sendMessagesMaxDevicesAllowedRegisterAfterOpen() throws Exception
```
timtay-microsoft, line 479: See the description of this PR for more details on why this is being removed

timtay-microsoft: By removing this test, we also save 1500 test identities from creation. In total, our e2e tests use about 4500 identities, so this would bring us down to 3000. The identity recycling logic I added earlier helps, but there is no guarantee that all 1500 over these test identities would be recycled identities or that they will be recycled later either.

timtay-microsoft: The other 3000 identities can be broken into 1500 for the ```sendMessagesMaxDevicesAllowed()``` test that makes this deleted test redundant, and the rest of the 1500 are scattered among our fault injection tests, twin tests, method tests, telemetry tests, file upload tests, etc...

timtay-microsoft: One interesting discussion we may want to have in the future is if we want to disable the ```sendMessagesMaxDevicesAllowed()``` test as well. Our .NET SDK doesn't have a comparable test, and removing it would save another 1500 device identities.

PR #1204, opened 2021-05-10
fix(iot-dev): Make device client throw UnsupportedOperationException for twin/methods when using HTTPS

timtay-microsoft:
IoT Hub itself does not support twin or method APIs when connecting over HTTP, so the SDK should notify users accordingly

A customer recently tried to use twin APIs over HTTPS and the SDK did not give descriptive errors to signal to the user that the service doesn't support the scenario, so these UnsupportedOperationExceptions will do that instead.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
@@ -39,6 +39,12 @@ public class InternalClient
     static final String SET_HTTPS_CONNECT_TIMEOUT = "SetHttpsConnectTimeout";
     static final String SET_HTTPS_READ_TIMEOUT = "SetHttpsReadTimeout";
 
+    private static final String TWIN_OVER_HTTP_ERROR_MESSAGE =
+        "Twin operations are only supported over MQTT, MQTT_WS, AMQPS, and AMQPS_WS";
+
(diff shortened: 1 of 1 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -1155,12 +1169,19 @@ private void verifyReportedProperties(Set<Property> reportedProperties) throws I
...
-    private void verifyReportedPropertiesAndVersion(Set<Property> reportedProperties, Integer version) throws IOException {
-        verifyReportedProperties(reportedProperties);
-
-        if (version != null && version < 0) {
-            throw new IllegalArgumentException("Version cannot be negative.");
+    private void verifyTwinOperationsAreSupported()
+    {
+        if (this.config.getProtocol() == HTTPS)
+        {
+            throw new UnsupportedOperationException(TWIN_OVER_HTTP_ERROR_MESSAGE);
```
abhipsaMisra, line 1176: concern about this being a breaking change? 

also, the .NET sdk could also use an inspection of the flow. (it looks like only one method call throws an exception and even the exception message is outdated: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/HttpTransportHandler.cs#L254)

drwill-ms (approved):

PR #1208, opened 2021-05-11
refactor(ado): Re-enable javadocs validation at the gate

timtay-microsoft:
A few javadoc issues have slipped by the gate recently. Since it doesn't take long to add this additional validation, it will save us from running into javadoc issues at release time (when it is required to be correct).


vsts/build_repo.ps1
```diff
@@ -12,4 +12,4 @@ else
     Write-Host "Pull request build detected"
 }
 
-mvn -DRUN_PROVISIONING_TESTS="$Env:runProvisioningTests" -DRUN_DIGITAL_TESTS="$Env:runDigitalTwinTests" -DRUN_IOTHUB_TESTS="$Env:runIotHubTests" -DIS_PULL_REQUEST="$isPullRequestBuild" install -T 2C `-Dmaven.javadoc.skip
```
timtay-microsoft, line 15: In case you are curious, the backtick characters here work like escape characters when passing in arguments to maven

jamdavi (approved):

PR #1209, opened 2021-05-11
fix(iot-dev): Make multiplexed connections handle device registrations during reconnection more gracefully

timtay-microsoft:
Rather than throw an exception internally when a device gets unregistered during its own reconnection, simply end the reconnection process since the user doesn't want the device on the multiplexed connection anymore anyways

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -1163,7 +1163,15 @@ public class IotHubTransport implements IotHubListener
         {
             reconnectionAttempts++;
 
-            RetryPolicy retryPolicy = this.getConfig(deviceId).getRetryPolicy();
+            DeviceClientConfig config = this.getConfig(deviceId);
+
+            if (config == null)
+            {
+                log.debug("Reconnection for device {} was abandoned because it was unregistered while reconnecting", deviceId);
+                return;
+            }
+
+            RetryPolicy retryPolicy = config.getRetryPolicy();
             retryDecision = retryPolicy.getRetryDecision(reconnectionAttempts, transportException);
             if (!retryDecision.shouldRetry())
             {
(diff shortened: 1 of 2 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1296,21 +1304,24 @@ else if (transportException != null && !transportException.isRetryable())
...
     private DeviceClientConfig getConfig(String deviceId)
     {
-        DeviceClientConfig config = this.deviceClientConfigs.get(deviceId);
-
-        if (config != null)
-        {
-            return config;
-        }
-
-        throw new IllegalStateException(String.format("Device client config does not exist for device %s", deviceId));
```
timtay-microsoft, line 1313: This method just returns a null value rather than throw an exception now. Each method that calls this method will handle this case a little differently (abandoning reconnection, for instance)

vinagesh (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1163,7 +1163,15 @@ private void reconnectDeviceSession(TransportException transportException, Strin
         {
             reconnectionAttempts++;
 
-            RetryPolicy retryPolicy = this.getConfig(deviceId).getRetryPolicy();
+            DeviceClientConfig config = this.getConfig(deviceId);
+
+            if (config == null)
+            {
+                log.debug("Reconnection for device {} was abandoned because it was unregistered while reconnecting", deviceId);
```
vinagesh, line 1170: Did you want to add the devicId here in the log of just empty braces?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1296,21 +1304,24 @@ else if (transportException != null && !transportException.isRetryable())
     // Still need to check the device connection status before you can report the device to be re-connected.
     private void singleDeviceReconnectAttemptAsync(String deviceId)
     {
-        DeviceClientConfig config = getConfig(deviceId);
+        DeviceClientConfig config = this.getConfig(deviceId);
+
+        if (config == null)
+        {
+            log.debug("Reconnection for device {} was abandoned because it was unregistered while reconnecting", deviceId);
```
vinagesh, line 1311: same

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
timtay-microsoft: No, this is the correct way for SLF4j log statements to look. For a statement like 

```java
log.debug("device {} has state {} at time {}", deviceId, state, time);
```

it knows to fill in the list of arguments from left to right

```
"device <deviceId> has state <state> at time <time>"
```

vinagesh: Nice :)

drwill-ms (approved):

PR #1210, opened 2021-05-13
refactor(all): Fixes outstanding unused declarations

jamdavi:
In most cases I suppressed the message because the private variable was "unassigned" and therefore unused.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -344,6 +344,9 @@ public void sendReportedProperties(Set<Property> reportedProperties) throws IOEx
      */
     public void sendReportedProperties(Set<Property> reportedProperties, int version) throws IOException, IllegalArgumentException
     {
+        if (version < 0) {
```
jamdavi, line 347: One of my helper methods was never called, so I moved this logic here.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/Message.java
```diff
@@ -110,52 +111,6 @@ public void setUserId(String userId)
...
-    /**
-    * [Required in feedback messages] Used in feedback messages generated by IoT Hub.
-    **/
-    private String feedbackDeviceId;
-
-    /**
-    * [Required in feedback messages] Used in feedback messages generated by IoT Hub.
-    **/
-    private String feedbackDeviceGenerationId;
-
```
jamdavi, line 158: All of these are not popualted anywhere. Things compile, but if anyone knows why we should keep them, let me know.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/digitaltwin/generated/implementation/DigitalTwinsImpl.java
```diff
@@ -370,7 +371,6 @@ public Object call(ServiceResponseWithHeaders<Object, DigitalTwinInvokeRootLevel
             throw new IllegalArgumentException("Parameter commandName is required and cannot be null.");
         }
         final String apiVersion = "2020-09-30";
-        final Object payload = null;
```
jamdavi, line 373: Was used in a method but the method just has "null" instead of this variable.

timtay-microsoft (approved):

PR #1214, opened 2021-05-14
refactor(all): Redundant throws [done by IDE]

jamdavi:
I used the IntelliJ tool to remove all of these redundant throws. There are a number of public methods that got changed and if this is not desirable I can revert this and do it one by one.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
@@ -575,9 +575,8 @@ public final class DeviceClient extends InternalClient implements Closeable
      * @param request The request details for getting the SAS URI, including the destination blob name.
      * @return The file upload details to be used with the Azure Storage SDK in order to upload a file from this device.
      * @throws IOException If this HTTPS request fails to send.
-     * @throws URISyntaxException If the returned sas uri cannot be constructed correctly
      */
-    public FileUploadSasUriResponse getFileUploadSasUri(FileUploadSasUriRequest request) throws IOException, URISyntaxException
+    public FileUploadSasUriResponse getFileUploadSasUri(FileUploadSasUriRequest request) throws IOException
     {
         if (this.fileUploadTask == null)
         {
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
@@ -71,13 +71,12 @@ public class ModuleClient extends InternalClient
      *                         HostName=xxxx;DeviceId=xxxx;SharedAccessKey=
      *                         xxxx;moduleId=xxxx;HostNameGateway=xxxx
      * @param protocol The protocol to use when communicating with the module
-     * @throws ModuleClientException if an exception is encountered when parsing the connection string
      * @throws UnsupportedOperationException if using any protocol besides MQTT, if the connection string is missing
      * the "moduleId" field, or if the connection string uses x509
      * @throws IllegalArgumentException if the provided connection string is null or empty, or if the provided protocol is null
      * @throws URISyntaxException if the connection string cannot be parsed for a valid hostname
      */
-    public ModuleClient(String connectionString, IotHubClientProtocol protocol) throws ModuleClientException, IllegalArgumentException, UnsupportedOperationException, URISyntaxException
+    public ModuleClient(String connectionString, IotHubClientProtocol protocol) throws IllegalArgumentException, UnsupportedOperationException, URISyntaxException
     {
         //Codes_SRS_MODULECLIENT_34_006: [This function shall invoke the super constructor.]
         super(new IotHubConnectionString(connectionString), protocol, SEND_PERIOD_MILLIS, getReceivePeriod(protocol), null);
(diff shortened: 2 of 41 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -575,9 +575,8 @@ public void uploadToBlobAsync(String destinationBlobName, InputStream inputStrea
      * @param request The request details for getting the SAS URI, including the destination blob name.
      * @return The file upload details to be used with the Azure Storage SDK in order to upload a file from this device.
      * @throws IOException If this HTTPS request fails to send.
-     * @throws URISyntaxException If the returned sas uri cannot be constructed correctly
      */
-    public FileUploadSasUriResponse getFileUploadSasUri(FileUploadSasUriRequest request) throws IOException, URISyntaxException
```
timtay-microsoft, line 580: This would be a breaking change since this is a public method. The best we can do here is suppress the warning and note in the javadocs that URISyntaxException is never thrown

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
```diff
@@ -170,7 +167,7 @@ public ModuleClient(String connectionString, IotHubClientProtocol protocol, Stri
      * See <a href="https://github.com/Azure/azure-iot-sdk-java/blob/master/device/iot-device-samples/send-event-x509/pom.xml">this pom.xml</a> for an example of how to do this.
      */
     @Deprecated
-    public ModuleClient(String connectionString, IotHubClientProtocol protocol, SSLContext sslContext) throws ModuleClientException, URISyntaxException
```
timtay-microsoft, line 173: Same comment for these ModuleClientExceptions. These would be breaking changes to remove

provisioning/security/security-provider/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/SecurityProviderX509.java
```diff
@@ -27,7 +27,7 @@
     abstract public Collection<X509Certificate> getIntermediateCertificatesChain();
 
     @Override
-    public String getRegistrationId() throws SecurityProviderException
```
timtay-microsoft, line 30: This would also be a breaking change since users do call this method directly under some circumstances

provisioning/security/tpm-provider/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/hsm/SecurityProviderTPMHsm.java
```diff
@@ -393,10 +392,9 @@ else if (rc != TPM_RC.HANDLE)
     /**
      * Getter for extracting StorageRootKey from TPM
      * @return The StorageRootKey from TPM
-     * @throws SecurityProviderException If StorageRootKey could not be extracted.
      */
     @Override
-    public byte[] getStorageRootKey() throws SecurityProviderException
```
timtay-microsoft, line 399: Same comment for these two methods. They are typically only used internally, but this class is directly used by customers, so it would risk breaking them

provisioning/security/tpm-provider-emulator/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/hsm/SecurityProviderTPMEmulator.java
```diff
@@ -511,10 +511,9 @@ else if (rc != TPM_RC.HANDLE)
     /**
      * Getter for extracting EndorsementKey from TPM
      * @return The Endorsement Key from TPM
-     * @throws SecurityProviderException If endorsement key could not be extracted.
      */
     @Override
-    public byte[] getEndorsementKey() throws SecurityProviderException
```
timtay-microsoft, line 517: This class is arguably only for testing purposes, and customers aren't expected to use it outside of their own tests, but we may want to not break it just to be safe

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/IotHubConnectionStringBuilder.java
```diff
@@ -26,12 +26,10 @@
      *
      * @param connectionString The connection string
      * @return The IotHubConnectionString object
-     * @throws IOException This exception is never thrown. Users are advised to use {@link #createIotHubConnectionString(String, AuthenticationMethod)} instead
-     * since it's method signature does not declare any thrown exceptions.
      * @deprecated Use {@link #createIotHubConnectionString(String)} since it does not declare any thrown exceptions
      */
     @Deprecated
-    public static IotHubConnectionString createConnectionString(String connectionString) throws IOException
```
timtay-microsoft, line 34: This class is about 1 layer removed from being public, so I'd revert this change. We reference it in a sample or two when demonstrating how to build SAS tokens

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/Message.java
```diff
@@ -210,9 +210,8 @@ public Message(byte[] byteArray)
      * Important: If a string is passed, the HttpBatch.SerializeAsString is set to true,
      * and the internal byte representation is serialized as UTF-8,
      * with HttpBatch.Encoding set to UTF-8.
-     * @throws UnsupportedEncodingException This exception is thrown if unsupported encoding used
      */
-    public Message(String string) throws UnsupportedEncodingException
```
timtay-microsoft, line 215: Users do use this API when sending c2d messages, so this would be a breaking change

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -102,12 +102,11 @@ public static RegistryManager createFromConnectionString(String connectionString
      * @deprecated because this method declares a thrown IOException even though it never throws an IOException. Users
      * are recommended to use {@link #RegistryManager(String, RegistryManagerOptions)} instead since it does not declare this exception even
      * though it constructs the same RegistryManager.
-     * @throws IOException This exception is never thrown.
      */
     @Deprecated
     public static RegistryManager createFromConnectionString(
             String connectionString,
-            RegistryManagerOptions options) throws IOException
```
timtay-microsoft, line 110: Very publicly used method, so this would be a breaking change for a lot of users

timtay-microsoft: Same for the public methods in this class

timtay-microsoft (changes requested):
Lots of breaking changes in here. The client level objects in particular should not remove these exceptions, even though we don't throw them.

jamdavi:
Not going to handle this scenario as most of these are public methods and changing the signature would break a lot of things. Will add this inspection to the ignore list

PR #1216, opened 2021-05-17
fix(iot-service): Fix support for AAD authentication in Fairfax hubs

timtay-microsoft:
Fairfax hubs require a slightly different authentication scope. Other private cloud instances do not, though.

We'll hold off on merging this in until the service team is ready with their side of this as well

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
@@ -6,6 +6,7 @@ package com.microsoft.azure.sdk.iot.service.auth;
 import com.azure.core.credential.AccessToken;
 import com.azure.core.credential.TokenCredential;
 import com.azure.core.credential.TokenRequestContext;
+import lombok.extern.slf4j.Slf4j;
 
 import java.time.Duration;
 import java.time.Instant;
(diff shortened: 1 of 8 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
```diff
@@ -6,33 +6,57 @@
...
 {
     private final static int MINUTES_BEFORE_PROACTIVE_RENEWAL = 9;
     private final TokenCredential tokenCredential;
     private AccessToken accessToken;
+    private final String[] authenticationScopes;
 
-    public static final String[] IOTHUB_PUBLIC_SCOPE = new String[]{"https://iothubs.azure.net/.default"};
-    public static final String BEARER_TOKEN_PREFIX = "Bearer ";
+    private static final String[] IOTHUB_PUBLIC_SCOPES = new String[]{"https://iothubs.azure.net/.default"};
+    private static final String[] IOTHUB_FAIRFAX_SCOPES = new String[]{"https://iothubs.azure.us"};
```
drwill-ms, line 28: How might a customer specify a value, if they are using yet another private cloud (black forest, mooncake, etc.)?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
```diff
@@ -68,9 +94,41 @@ public TokenCredential getTokenCredential()
...
+    /**
+     * Get the authentication scopes to be used when generating AAD access tokens.
+     * @param hostname The hostname of the IoT Hub that the tokens will be used to authenticate against.
+     * @return The authentication scopes to be used when generating AAD access tokens.
+     */
+    public static String[] getAuthenticationScopes(String hostname)
+    {
+        if (isFairfaxHub(hostname))
+        {
+            log.debug("Fairfax IoT Hub detected based on hostname, supplying Fairfax specific authentication scopes.");
```
drwill-ms, line 106: 2 separate clauses, so use a period or semicolon, instead of a comma.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/TokenCredentialCache.java
drwill-ms: My bad. Your comments specify that other private clouds don't require their own.

timtay-microsoft: They don't have to. The service team has confirmed for us that Fairfax is the only special case here.

timtay-microsoft:
Closing this PR as this solution doesn't account for vanity hostnames, which may be coming soon from the service side

PR #1217, opened 2021-05-18
fix(sample): Fix file upload sample.

vinagesh:
# Checklist

# Description of the problem

# Description of the solution
Azure blobClient does not work with FileInputStream as it does not support mark/reset. We can either use BufferedFileInputStream or call uploadFromFile. The later is simpler so  I have updated the samples to use that. Now the samples upload files successfully without any exception.


device/iot-device-samples/file-upload-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/FileUploadSample.java
```diff
@@ -129,53 +129,40 @@ private static void uploadFileOrDirectoryRecursive(DeviceClient client, String b
 
     private static void uploadFile(DeviceClient client, String baseDirectory, String relativeFileName) throws IOException, URISyntaxException {
         File file = new File(baseDirectory, relativeFileName);
-        try (InputStream inputStream = new FileInputStream(file))
```
vinagesh, line 132: The diff looks weird for some reason but line 132 and line 150 are the main things. We no longer need FileInputStream and we use the blobClient.uploadFromFile.

timtay-microsoft (approved):

PR #1218, opened 2021-05-19
fix(iot-dev): Fix a few small multiplexing bugs

timtay-microsoft:
-Fixed a bug where multiplexing client retry logic for a particular device session did not abandon reconnection when a non-retryable exception is thrown

-Fixed a bug where registering a device to an active multiplexed connection times out if the registration fails due to an issue like if the device is disabled

-Fixed a bug where opening a multiplexing client times out if one or more of the registered devices is disabled. Now the open call will succeed, but the connection status of the disabled device will go DISCONNECTED shortly afterwards. This is a little confusing, but the service itself does open the device session briefly, so it isn't possible for us to fail the open call

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -1159,7 +1159,8 @@ public class IotHubTransport implements IotHubListener
         RetryDecision retryDecision = null;
 
         while (this.deviceConnectionStates.get(deviceId) == IotHubConnectionStatus.DISCONNECTED_RETRYING
-                && !hasReconnectOperationTimedOut)
+                && !hasReconnectOperationTimedOut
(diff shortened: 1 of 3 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1159,7 +1159,8 @@ private void reconnectDeviceSession(TransportException transportException, Strin
         RetryDecision retryDecision = null;
 
         while (this.deviceConnectionStates.get(deviceId) == IotHubConnectionStatus.DISCONNECTED_RETRYING
-                && !hasReconnectOperationTimedOut)
+                && !hasReconnectOperationTimedOut
+                && transportException.isRetryable())
```
timtay-microsoft, line 1163: We abandon reconnection on unretryable errors for non multiplexed client cases, so I consider it a bug that we don't do the same for multiplexed cases.

jamdavi (approved):

PR #1226, opened 2021-06-01
feat(iot-service): Add support for arbitrary token credential authentication scopes

timtay-microsoft:
```java
// Fairfax options
String[] scopes = IotHubAuthenticationScopes.GOVERNMENT_CLOUD_AUTHENTICATION_SCOPES;
RegistryManagerOptions fairfaxOptions = RegistryManagerOptions
    .builder()
    .tokenCredentialAuthenticationScopes(scopes)
    .build();

// Default options
RegistryManagerOptions defaultOptions = RegistryManagerOptions.builder().build();

RegistryManager registryManager = new RegistryManager(hostName, tokenCredential, fairfaxOptions);
```

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
@@ -7,6 +7,7 @@ package com.microsoft.azure.sdk.iot.service;
 
 import com.azure.core.credential.AzureSasCredential;
 import com.azure.core.credential.TokenCredential;
+import com.microsoft.azure.sdk.iot.service.auth.TokenCredentialCache;
 import com.microsoft.azure.sdk.iot.service.transport.amqps.AmqpReceive;
 import lombok.extern.slf4j.Slf4j;
 
@@ -136,7 +137,7 @@ public class FeedbackReceiver extends Receiver
 
     FeedbackReceiver(
             String hostName,
-            TokenCredential credential,
+            TokenCredentialCache credentialCache,
             IotHubServiceClientProtocol iotHubServiceClientProtocol,
             ProxyOptions proxyOptions,
             SSLContext sslContext)
@@ -146,13 +147,13 @@ public class FeedbackReceiver extends Receiver
             throw new IllegalArgumentException("hostName cannot be null or empty");
         }
 
-        Objects.requireNonNull(credential);
+        Objects.requireNonNull(credentialCache);
         Objects.requireNonNull(iotHubServiceClientProtocol);
 
         this.amqpReceive =
                 new AmqpReceive(
                         hostName,
-                        credential,
+                        credentialCache,
                         iotHubServiceClientProtocol,
                         proxyOptions,
                         sslContext);
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
@@ -41,7 +41,7 @@ public class RegistryManagerOptions
 
     /**
      * The authentication scopes to use when requesting authentication tokens from Azure Active Directory for
-     * authenticating with IoT Hub. This value is only used by the client if the client is configured to use role based
+     * authenticating with IoT Hub. This value is only used by the client if the client is configured to use role-based
      * access credentials rather than symmetric key based credentials. This value defaults to the authentication scopes
      * that are used for all public cloud deployments and all private cloud deployments other than those in the
      * Fairfax cloud. For Fairfax cloud users, this value must be set to
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
@@ -42,7 +42,7 @@ public class RegistryManagerOptions
     /**
      * The authentication scopes to use when requesting authentication tokens from Azure Active Directory for
      * authenticating with IoT Hub. This value is only used by the client if the client is configured to use role-based
-     * access credentials rather than symmetric key based credentials. This value defaults to the authentication scopes
+     * access credentials rather than symmetric key-based credentials. This value defaults to the authentication scopes
      * that are used for all public cloud deployments and all private cloud deployments other than those in the
      * Fairfax cloud. For Fairfax cloud users, this value must be set to
      * {@link IotHubAuthenticationScopes#FAIRFAX_AUTHENTICATION_SCOPES}.
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
@@ -44,8 +44,8 @@ public class RegistryManagerOptions
      * authenticating with IoT Hub. This value is only used by the client if the client is configured to use role-based
      * access credentials rather than symmetric key-based credentials. This value defaults to the authentication scopes
      * that are used for all public cloud deployments and all private cloud deployments other than those in the
-     * Fairfax cloud. For Fairfax cloud users, this value must be set to
-     * {@link IotHubAuthenticationScopes#FAIRFAX_AUTHENTICATION_SCOPES}.
+     * Azure Government cloud. For Azure Government cloud users, this value must be set to
+     * {@link IotHubAuthenticationScopes#GOVERNMENT_CLOUD_AUTHENTICATION_SCOPES}.
      */
     @Getter
     @Builder.Default
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
@@ -1,6 +1,6 @@
 package com.microsoft.azure.sdk.iot.service;
 
-import com.microsoft.azure.sdk.iot.service.auth.AuthenticationScopes;
+import com.microsoft.azure.sdk.iot.service.auth.IotHubAuthenticationScopes;
 import lombok.Builder;
 import lombok.Getter;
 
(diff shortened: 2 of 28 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
```diff
@@ -136,7 +137,7 @@ public FeedbackReceiver(
 
     FeedbackReceiver(
             String hostName,
-            TokenCredential credential,
```
timtay-microsoft, line 139: This change from TokenCredential to TokenCredentialCache is seen a lot in this PR. The reason for this change is that the TokenCredentialCache encapsulates the TokenCredential, but also contains the authentication scope information

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/AuthenticationScopes.java
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License.
+
+package com.microsoft.azure.sdk.iot.service.auth;
+
+/**
+ * This class contains the most commonly used Azure Active Directory token audience scopes. They should be used as
+ * inputs when setting the authentication scopes in service client options classes such as
+ * {@link com.microsoft.azure.sdk.iot.service.RegistryManagerOptions}.
+ */
+public class AuthenticationScopes
```
vinagesh, line 11: If it were C#, this class would be static. Same for java or not?

vinagesh: We could also call this class IotHubAuthenticationScopes and the constants below could be simplified to DEFAULT_AUTHENTICATION_SCOPES and FAIRFAX_AUTHENTICATION_SCOPES for better readability and shorter names.

azabbasi (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/AuthenticationScopes.java
timtay-microsoft: Good idea. Since this class isn't intended to be instantiated, it does make sense to make the class static.

timtay-microsoft: Java only allows inner classes to be static. The best that I can do here is add a private no args constructor so that no user can actually construct this class.

timtay-microsoft: > We could also call this class IotHubAuthenticationScopes and the constants below could be simplified to DEFAULT_AUTHENTICATION_SCOPES and FAIRFAX_AUTHENTICATION_SCOPES for better readability and shorter names.

Sure thing

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
drwill-ms: internal-only impact?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
```diff
@@ -37,4 +38,16 @@
     @Getter
     @Builder.Default
     private final int httpConnectTimeout = DEFAULT_HTTP_CONNECT_TIMEOUT_MS;
+
+    /**
+     * The authentication scopes to use when requesting authentication tokens from Azure Active Directory for
+     * authenticating with IoT Hub. This value is only used by the client if the client is configured to use role based
```
drwill-ms, line 44: role-based

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
```diff
@@ -37,4 +38,16 @@
...
     @Builder.Default
     private final int httpConnectTimeout = DEFAULT_HTTP_CONNECT_TIMEOUT_MS;
+
+    /**
+     * The authentication scopes to use when requesting authentication tokens from Azure Active Directory for
+     * authenticating with IoT Hub. This value is only used by the client if the client is configured to use role-based
+     * access credentials rather than symmetric key based credentials. This value defaults to the authentication scopes
+     * that are used for all public cloud deployments and all private cloud deployments other than those in the
+     * Fairfax cloud. For Fairfax cloud users, this value must be set to
+     * {@link IotHubAuthenticationScopes#FAIRFAX_AUTHENTICATION_SCOPES}.
```
drwill-ms, line 48: I thought Fairfax was an internal code name for US Government Cloud. Should we use this term externally?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClientOptions.java
```diff
@@ -17,11 +18,22 @@
...
      * The SSL context to use when opening the AMQPS/AMQPS_WS connections. If not set, this library will generate the default
      * SSL context that trusts the IoT Hub public certificates.
      */
     @Getter
     private final SSLContext sslContext;
+
+    /**
+     * The authentication scopes to use when requesting authentication tokens from Azure Active Directory for
+     * authenticating with IoT Hub. This value is only used by the client if the client is configured to use role-based
+     * access credentials rather than symmetric key based credentials. This value defaults to the authentication scopes
```
drwill-ms, line 31: key-based

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManagerOptions.java
timtay-microsoft: Good question. I'll run this by some service PMs to get their take on it

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
timtay-microsoft: Yep

service/iot-service-samples/role-based-authorization-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/RoleBasedAuthenticationSample.java
```diff
@@ -74,6 +74,14 @@ public static void main(String[] args)
         runJobClientSample(iotHubHostName, credential);
 
         runDeviceMethodClientSample(iotHubHostName, credential, newDeviceId);
+
+        // For users of IoT Hubs deployed in the Fairfax private cloud, your service client's options must specify these
+        // authentication scopes. The default authentication scopes are valid for all other deployments in public or
+        // private clouds outside of Azure Stack deployments.
+        RegistryManagerOptions fairfaxHubOptions = RegistryManagerOptions
```
drwill-ms, line 81: Prefer not to call this `fairfaxHubOptions` because the object represents all options for registry manager. How about just `registryManagerOptions`?

timtay-microsoft: This is the name of the instance of RegistryManagerOptions, not the class itself. The point is to demonstrate how to build an options object that allows for you to connect to fairfax hubs

drwill-ms (approved):

service/iot-service-samples/role-based-authorization-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/RoleBasedAuthenticationSample.java
drwill-ms: I get that. I still think it is a misleading variable name. I'd rather you use a comment to specify why you are doing it, rather than use a specific variable name, when the options class is available for all kinds of options, not just auth scope.

vinagesh (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/IotHubAuthenticationScopes.java
```diff
@@ -0,0 +1,33 @@
...
+     * deployed IoT Hubs other than those in the Azure Government cloud. For users of IoT Hubs deployed in the Azure Government cloud, the
+     * {@link #GOVERNMENT_CLOUD_AUTHENTICATION_SCOPES} should be used instead of this.
+     */
+    public static final String[] DEFAULT_AUTHENTICATION_SCOPES = new String[]{"https://iothubs.azure.net/.default"};
+
+    /**
+     * The authentication scopes for IoT Hubs deployed in the Fairfax private cloud. Users must provide this value when
+     * constructing the client's options when using role-based access credentials.
+     */
+    public static final String[] GOVERNMENT_CLOUD_AUTHENTICATION_SCOPES = new String[]{"https://iothubs.azure.us/.default"};
```
vinagesh, line 32: I think this should either say US Government Cloud or Fairfax, Just Government Cloud can be confusing to users in other countries.

vinagesh: I had missed your email yesterday. I see the communication on the official naming.
However, it is called AzureGovernment everwhere or Azure US Government cloud. We should probably follow this convention - https://docs.microsoft.com/en-us/dotnet/api/azure.identity.azureauthorityhosts.azuregovernment?view=azure-dotnet#Azure_Identity_AzureAuthorityHosts_AzureGovernment

vinagesh: vAZURE_GOVERNMENT_CLOUD_AUTHENTICATION_SCOPESv and in the comments, explicitly say `Azure US Government cloud`

timtay-microsoft: In our comments, I like spelling out that "AZURE_GOVERNMENT" corresponds to the US government in particular. I'll leave this field as "GOVERNMENT_CLOUD_AUTHENTICATION_SCOPES" but have the header spell out "Azure USA Government cloud"

timtay-microsoft:
Closing this PR since US Government cloud no longer needs a special authentication scope. If necessary, we can add the support for arbitrary token scopes back later.

PR #1229, opened 2021-06-01
refactor(iot-service): Set default values in option objects

timtay-microsoft:
Rather than set the default values for these fields in the constructor that builds these options objects, just set them in the options objects directly

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
@@ -7,7 +7,6 @@ package com.microsoft.azure.sdk.iot.service;
 
 import com.azure.core.credential.AzureSasCredential;
 import com.azure.core.credential.TokenCredential;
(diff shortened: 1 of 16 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -63,10 +62,7 @@ public RegistryManager()
     {
         // This constructor was previously a default constructor that users could use because there was no other constructor declared.
         // However, we still prefer users use the createFromConnectionString method to build their clients.
-        options = RegistryManagerOptions.builder()
-                .httpConnectTimeout(RegistryManagerOptions.DEFAULT_HTTP_CONNECT_TIMEOUT_MS)
-                .httpReadTimeout(RegistryManagerOptions.DEFAULT_HTTP_READ_TIMEOUT_MS)
-                .build();
```
timtay-microsoft, line 69: Having to set these default values in each constructor is unnecessary now that the builder sets the defaults for us

barustum (approved):

PR #1231, opened 2021-06-04
fix(iot-dev): clarify/fix how partial successes work for MultiplexingClient bulk registrations

timtay-microsoft:
When registering X devices to an open connection and 1 of them is disabled, the register call will throw a MultiplexingClientDeviceRegistrationAuthenticationException, but X-1 devices successfully added their session to the active multiplexed connection. This needed to be clarified since the javadocs didn't call this out.

There are two bug fixes here. The first bug fix is simply that the transport layer forgot to remove the device from the set of registered devices if it failed to register in a scenario like the above. It was never successfully registered, so the transport layer should have the same state. 

The second bug is that the MultiplexingClient layer should update it's state accordingly in these partial success scenarios as well. Previously, if it threw an exception in a scenario like the above, its local state would show that all X devices are not registered, even when X-1 of them have open sessions on the actual multiplexed connection. I added a few assert statement to an existing test to validate this change.

One commit on this PR is just to delete some pesky outdated comments

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
@@ -136,7 +136,8 @@ public class MultiplexingClient
      * <p>
      * @throws MultiplexingClientException If any IO or authentication errors occur while opening the multiplexed connection.
      * @throws MultiplexingClientDeviceRegistrationAuthenticationException If one or many of the registered devices failed to authenticate.
-     * Any devices not found in the map of registration exceptions provided by this exception have registered successfully.
+     * Any devices not found in the map of registration exceptions provided by
+     * {@link MultiplexingClientDeviceRegistrationAuthenticationException#getRegistrationExceptions()} have registered successfully.
      * Even when this is thrown, the AMQPS/AMQPS_WS connection is still open, and other clients may be registered to it.
      */
     public void open() throws MultiplexingClientException
@@ -248,8 +249,10 @@ public class MultiplexingClient
      * <p>
      * @throws InterruptedException If the thread gets interrupted while waiting for the registration to succeed. This
      * will never be thrown if the multiplexing client is not open yet.
-     * @throws MultiplexingClientDeviceRegistrationAuthenticationException If one or more devices failed to register. Details for each failure can be found
-     * in this exception.
+     * @throws MultiplexingClientDeviceRegistrationAuthenticationException If the device failed to register. Details for
+     * this failure can be found nested within the map given by
+     * {@link MultiplexingClientDeviceRegistrationAuthenticationException#getRegistrationExceptions()}. If this exception is
+     * thrown, the device was not registered, and therefore it does not need to be unregistered.
      * @throws com.microsoft.azure.sdk.iot.device.exceptions.MultiplexingClientDeviceRegistrationTimeoutException If this operation takes longer than the default timeout allows.
      * @throws MultiplexingClientException If any other Exception is thrown, it will be nested into this exception.
      * @param deviceClient The device client to associate with this multiplexing client.
@@ -298,8 +301,10 @@ public class MultiplexingClient
      * <p>
      * @throws InterruptedException If the thread gets interrupted while waiting for the registration to succeed. This
      * will never be thrown if the multiplexing client is not open yet.
-     * @throws MultiplexingClientDeviceRegistrationAuthenticationException If one or more devices failed to register. Details for each failure can be found
-     * in this exception.
+     * @throws MultiplexingClientDeviceRegistrationAuthenticationException If the device failed to register. Details for
+     * this failure can be found nested within the map given by
+     * {@link MultiplexingClientDeviceRegistrationAuthenticationException#getRegistrationExceptions()}. If this exception is
+     * thrown, the device was not registered, and therefore it does not need to be unregistered.
      * @throws com.microsoft.azure.sdk.iot.device.exceptions.MultiplexingClientDeviceRegistrationTimeoutException If this operation takes longer than the provided timeout allows.
      * @throws MultiplexingClientException If any other Exception is thrown, it will be nested into this exception.
      * @param deviceClient The device client to associate with this multiplexing client.
@@ -349,8 +354,12 @@ public class MultiplexingClient
      * <p>
      * @throws InterruptedException If the thread gets interrupted while waiting for the registration to succeed. This
      * will never be thrown if the multiplexing client is not open yet.
-     * @throws MultiplexingClientDeviceRegistrationAuthenticationException If one or more devices failed to register. Details for each failure can be found
-     * in this exception. Any devices not found in the map of registration exceptions provided by this exception have registered successfully.
+     * @throws MultiplexingClientDeviceRegistrationAuthenticationException If one or more devices failed to register.
+     * Details for each failure can be found in the map provided by
+     * {@link MultiplexingClientDeviceRegistrationAuthenticationException#getRegistrationExceptions()}. Any devices not
+     * found in the map of registration exceptions provided by this exception have registered successfully. Any devices
+     * that are found in the map of registration exceptions provided by this exception were not registered, and therefore
+     * do not need to be unregistered.
      * @throws com.microsoft.azure.sdk.iot.device.exceptions.MultiplexingClientDeviceRegistrationTimeoutException If this operation takes longer than the default timeout allows.
      * @throws MultiplexingClientException If any other Exception is thrown, it will be nested into this exception.
      * @param deviceClients The device clients to associate with this multiplexing client.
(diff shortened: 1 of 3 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -830,6 +776,10 @@ public void registerMultiplexedDeviceClient(List<DeviceClientConfig> configs, lo
                     }
 
                     registrationException.addRegistrationException(deviceId, deviceRegistrationException);
+
+                    // Since the registration failed, need to remove the device from the list of multiplexed devices
+                    DeviceClientConfig configThatFailedToRegister = this.deviceClientConfigs.remove(deviceId);
+                    ((AmqpsIotHubConnection) this.iotHubTransportConnection).unregisterMultiplexedDevice(configThatFailedToRegister);
```
timtay-microsoft, line 782: This is the actual code change that fixes a bug. This code sits within a block that waits while all the devices finish registration either successfully or unsuccessfully. The unsuccessful devices should be removed from this layer and from the AMQP layer so that their state matches that of the MultiplexingClient itself

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -485,13 +500,35 @@ public void registerDeviceClients(Iterable<DeviceClient> deviceClients, long tim
                 log.info("Registering device {} to multiplexing client", configBeingRegistered.getDeviceId());
             }
 
-            this.deviceIO.registerMultiplexedDeviceClient(deviceClientConfigsToRegister, timeoutMilliseconds);
```
timtay-microsoft, line 488: Previously, this call would throw a MultiplexingClientDeviceRegistrationAuthenticationException even in partial success cases, which meant that the block of code below it that updates the local state of the multiplexedDeviceClients list never got updated to match. Now we'll catch the exception, update the local state, and then rethrow the exception

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
```diff
@@ -1517,35 +1517,43 @@ public void disableDeviceAfterOpenAndAfterRegistration() throws Exception
         deviceToDisable.setStatus(DeviceStatus.Disabled);
         registryManager.updateDevice(deviceToDisable);
 
-        // verify that the disabled device loses its device session
```
timtay-microsoft, line 1520: The diff here is a bit hard to follow, but I basically added a try/finally block to always re-enable the device in case it gets recycled by another test

```java
try
{
    // some test code involving a disabled device
}
finally
{
    // re-enable the device no matter what
}
```

timtay-microsoft: And I did that for the three tests in this suite that intentionally disable devices

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -248,8 +251,10 @@ public void close() throws MultiplexingClientException
      * <p>
      * @throws InterruptedException If the thread gets interrupted while waiting for the registration to succeed. This
      * will never be thrown if the multiplexing client is not open yet.
-     * @throws MultiplexingClientDeviceRegistrationAuthenticationException If one or more devices failed to register. Details for each failure can be found
-     * in this exception.
+     * @throws MultiplexingClientDeviceRegistrationAuthenticationException If the device failed to register. Details for
```
vinagesh, line 254: comma after register, not fullstop

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -349,8 +356,12 @@ public void registerDeviceClient(DeviceClient deviceClient, long timeoutMillisec
      * <p>
      * @throws InterruptedException If the thread gets interrupted while waiting for the registration to succeed. This
      * will never be thrown if the multiplexing client is not open yet.
-     * @throws MultiplexingClientDeviceRegistrationAuthenticationException If one or more devices failed to register. Details for each failure can be found
-     * in this exception. Any devices not found in the map of registration exceptions provided by this exception have registered successfully.
+     * @throws MultiplexingClientDeviceRegistrationAuthenticationException If one or more devices failed to register.
```
vinagesh, line 359: fullstop to comma

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -485,13 +500,35 @@ public void registerDeviceClients(Iterable<DeviceClient> deviceClients, long tim
...
+            for (DeviceClient clientsThatAttemptedToRegister : deviceClients)
             {
                 // Only update the local state map once the register call has succeeded
-                String registeredDeviceId = successfullyRegisteredClient.getConfig().getDeviceId();
-                this.multiplexedDeviceClients.put(registeredDeviceId, successfullyRegisteredClient);
+                String deviceIdThatAttemptedToRegister = clientsThatAttemptedToRegister.getConfig().getDeviceId();
+
+                if (!deviceIdsThatFailedToRegister.contains(deviceIdThatAttemptedToRegister))
+                {
+                    this.multiplexedDeviceClients.put(deviceIdThatAttemptedToRegister, clientsThatAttemptedToRegister);
```
vinagesh, line 525: can we not do this on line 508 instead of iterating through the list again?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
```diff
@@ -1517,35 +1517,43 @@ public void disableDeviceAfterOpenAndAfterRegistration() throws Exception
...
-        testSendingMessagesFromMultiplexedClients(testInstance.deviceClientArray.subList(1, DEVICE_MULTIPLEX_COUNT));
+            // Verify that the other devices can still send telemetry
+            testSendingMessagesFromMultiplexedClients(testInstance.deviceClientArray.subList(1, DEVICE_MULTIPLEX_COUNT));
 
-        testInstance.multiplexingClient.close();
+            testInstance.multiplexingClient.close();
+        }
+        finally
+        {
+            deviceToDisable.setStatus(DeviceStatus.Enabled); // re enable the device in case it gets recycled
```
vinagesh, line 1554: Is this to protect against the case where the test throws and exception and the device it is using needs to be recycled?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
barustum: if I am reading this right though, the try block (506 to 508) is not inside a loop, is it?

vinagesh: It should to be inside a bigger loop right?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: I'm confused, are you suggesting it should read "If one or more devices failed to register, details for each failure can be found in the map provided by..."? The period seems appropriate here to me

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
vinagesh: Is it possible to do this by default in disposeTestIdentity in the Tools instead? Unless that won't be called on exception. There is a lot to understand so asking you the question :)

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: Generally, the first line of this exception javadoc explains when it is throw, so "If one or more devices failed to register." seems appropriate as is

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: There was a bigger loop earlier that I could leverage, so I removed this loop in the happy path case. For the case where 1 or more devices fail to register, we still need to compare the list of devices that failed against the list of devices that didn't

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
timtay-microsoft: Typically tests don't disable their own devices at any point, so this try/finally logic is just specific to the tests that do disable devices at this point

PR #1248, opened 2021-06-16
refactor(iot-dev): Add log statements for setOption methods

timtay-microsoft:
These options can drastically alter how the SDK behaves, so for future debugging purposes, they should be logged when set

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
@@ -887,6 +887,7 @@ public class InternalClient
     {
         if (value != null)
         {
+            log.info("Setting path to trusted certificate");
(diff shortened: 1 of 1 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -887,6 +887,7 @@ void setOption_SetCertificatePath(Object value)
     {
         if (value != null)
         {
+            log.info("Setting path to trusted certificate");
```
timtay-microsoft, line 890: Not logging the path being set because it has no value for our debugging purposes. If the path doesn't point at a valid certificate, then the SDK will throw an exception anyways.

npatilsen:
Not a reviewer, but I took a look anyway. Looks alright to me, but take that with a grain of salt 😃

abhipsaMisra (approved):

PR #1250, opened 2021-06-17
Add documentation to repo about which platforms this SDK supports

### LINKED ISSUE #1246: Add specific documentation around what platforms this library supports
It'd be nice to have a formal document where we list which JDK versions and operating systems that we test this SDK against, and lists our stances on other platforms like Kotlin.

Once this document is added, the "new issue" templates should be updated to point users to them so that issues aren't filed for unsupported frameworks/operating systems.

timtay-microsoft:
fixes #1246

Once we settle on a general format, I'll create a similar doc for our .NET repo

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -9,8 +9,8 @@ Due to security considerations, build logs are not publicly available.
 | Preview                  | [![Build Status](https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_apis/build/status/java/pull_request_validation/Java%20Canary?branchName=preview)](https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/latest?definitionId=245&branchName=preview)|
 
 This repository contains the following:
-* **Azure IoT Hub device SDK for Java**: connect client devices to Azure IoT Hub (supports Java 7+)
-* **Azure IoT Hub service SDK for Java**: enables developing back-end applications for Azure IoT (supports Java 8+)
+* **Azure IoT Hub device SDK for Java**: connect client devices to Azure IoT Hub
+* **Azure IoT Hub service SDK for Java**: enables developing back-end applications for Azure IoT
 * **Azure IoT Device Provisioning device SDK for Java**: provision devices to Azure IoT Hub using Azure IoT Device Provisioning
 * **Azure IoT Device Provisioning service SDK for Java**: manage your Provisioning service instance from a back-end Java application
 
@@ -33,8 +33,10 @@ Devices and data sources in an IoT solution can range from a simple network-conn
    * **Working with the SDKs code**: if you are working with the SDKs code to modify it or contribute changes, then you can clone the repository and build the libraries:
       * [Build Device SDK from code][device-code]
       * [Build Service SDK from code][service-code]
-* On Android: our Java device SDK can be used on Android using the API version 17 and higher:
+* On Android: our Java device SDK can be used on Android:
    * [Device SDK][device-android]
+
+For more details on what platforms this SDK supports, see [this document](./supported_platforms.md)
 
 ## API reference
 * [Azure IoT Hub device SDK][java-api-reference-device]
--- a/readme.md
+++ b/readme.md
@@ -36,7 +36,7 @@ Devices and data sources in an IoT solution can range from a simple network-conn
 * On Android: our Java device SDK can be used on Android:
    * [Device SDK][device-android]
 
-For more details on what platforms this SDK supports, see [this document](./supported_platforms.md)
+For more details on what platforms this SDK supports, see [this document](./supported_platforms.md).
 
 ## API reference
 * [Azure IoT Hub device SDK][java-api-reference-device]
--- a/supported_platforms.md
+++ b/supported_platforms.md
@@ -1,0 +1,55 @@
+# Microsoft Azure IoT SDKs for Java
+
+This SDK is tested nightly on Windows 10, Ubuntu 1604, and on an Android emulator. For additional details
+for each tested platform, see the respective sections below. 
+
+While this SDK contains several different packages, each package supports the same platforms and JDKs.
+
+## Supported JDK versions
+
+Currently, this SDK supports JDK versions 8 and 11. This SDK may work with other versions, but there are no guarantees 
+that they will always work for those versions nor are there guarantees that we will fix bugs that are only present on
+those versions.
+
+
+## Windows 10
+
+Nightly test platform details:
+Apache Maven 3.8.1 (05c21c65bdfed0f71a2f2ada8b84da59348c4c5d)
+Java version: 1.8.0_292, vendor: AdoptOpenJDK, runtime: C:\hostedtoolcache\windows\Java_Adopt_jdk\8.0.292-10\x64\jre
+Default locale: en_US, platform encoding: Cp1252
+OS name: "windows server 2016", version: "10.0", arch: "amd64", family: "windows"
+
+Note that, while we only directly test on Windows 10, we do support other Windows versions as well.
+
+## Ubuntu 1604
+
+Nightly test platform details:
+Apache Maven 3.8.1 (05c21c65bdfed0f71a2f2ada8b84da59348c4c5d)
+Java version: 1.8.0_292, vendor: AdoptOpenJDK, runtime: /usr/lib/jvm/adoptopenjdk-8-hotspot-amd64/jre
+Default locale: en_US, platform encoding: UTF-8
+OS name: "linux", version: "4.15.0-1113-azure", arch: "amd64", family: "unix"
+
+Note that, while we only directly test on Ubuntu 1604, we do generally support other popular linux distributions. 
+
+## Android
+
+This SDK supports Android API versions 24, 25, 26, 27, and 28.
+
+The android emulator image used for nightly tests is "system-images;android-28;google_apis;x86_64"
+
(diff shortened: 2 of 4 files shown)


readme.md
```diff
@@ -9,8 +9,8 @@ Due to security considerations, build logs are not publicly available.
 | Preview                  | [![Build Status](https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_apis/build/status/java/pull_request_validation/Java%20Canary?branchName=preview)](https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/latest?definitionId=245&branchName=preview)|
 
 This repository contains the following:
-* **Azure IoT Hub device SDK for Java**: connect client devices to Azure IoT Hub (supports Java 7+)
```
timtay-microsoft, line 12: Don't want duplicate information in this document that may get out of date. The new document will be the source of truth

supported_platforms.md
```diff
@@ -0,0 +1,55 @@
...
+those versions.
+
+
+## Windows 10
+
+Nightly test platform details:
+Apache Maven 3.8.1 (05c21c65bdfed0f71a2f2ada8b84da59348c4c5d)
+Java version: 1.8.0_292, vendor: AdoptOpenJDK, runtime: C:\hostedtoolcache\windows\Java_Adopt_jdk\8.0.292-10\x64\jre
+Default locale: en_US, platform encoding: Cp1252
+OS name: "windows server 2016", version: "10.0", arch: "amd64", family: "windows"
```
timtay-microsoft, line 21: I'd love to have a way for our nightly builds to automatically update these versions as the ADO agents we use update to later maven/JDK/windows versions, but I'll file that one under "Maybe some other time"

supported_platforms.md
```diff
@@ -0,0 +1,55 @@
+# Microsoft Azure IoT SDKs for Java
+
+This SDK is tested nightly on Windows 10, Ubuntu 1604, and on an Android emulator. For additional details
+for each tested platform, see the respective sections below. 
+
+While this SDK contains several different packages, each package supports the same platforms and JDKs.
+
+## Supported JDK versions
+
+Currently, this SDK supports JDK versions 8 and 11. This SDK may work with other versions, but there are no guarantees
```
timtay-microsoft, line 10: @jamdavi I vaguely remember you validating that our SDK works fine with JDK 11, but can you confirm that for me?

supported_platforms.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Ubuntu 1604
+
+Nightly test platform details:
+Apache Maven 3.8.1 (05c21c65bdfed0f71a2f2ada8b84da59348c4c5d)
+Java version: 1.8.0_292, vendor: AdoptOpenJDK, runtime: /usr/lib/jvm/adoptopenjdk-8-hotspot-amd64/jre
+Default locale: en_US, platform encoding: UTF-8
+OS name: "linux", version: "4.15.0-1113-azure", arch: "amd64", family: "unix"
+
+Note that, while we only directly test on Ubuntu 1604, we do generally support other popular linux distributions.
```
timtay-microsoft, line 33: I stopped short of listing out Linux distros in particular because I'm not entirely sure where we'd draw the line between Ubuntu and some random Linux distro created by one developer for personal use. I'm curious if anyone feels I should get more specific here though

supported_platforms.md
```diff
@@ -0,0 +1,55 @@
...
+
+This SDK is tested nightly on Windows 10, Ubuntu 1604, and on an Android emulator. For additional details
+for each tested platform, see the respective sections below. 
+
+While this SDK contains several different packages, each package supports the same platforms and JDKs.
+
+## Supported JDK versions
+
+Currently, this SDK supports JDK versions 8 and 11. This SDK may work with other versions, but there are no guarantees 
+that they will always work for those versions nor are there guarantees that we will fix bugs that are only present on
```
abhipsaMisra, line 11: nit: should we emphasize "may" and "always"? In italics maybe?

supported_platforms.md
abhipsaMisra: this doesn't show up as expected in the readme (with newlines). You probably need to prefix each line with a bullet.

supported_platforms.md
```diff
@@ -0,0 +1,55 @@
...
+those versions.
+
+
+## Windows 10
+
+Nightly test platform details:
+Apache Maven 3.8.1 (05c21c65bdfed0f71a2f2ada8b84da59348c4c5d)
```
abhipsaMisra, line 18: Is this guid of any relevance: 05c21c65bdfed0f71a2f2ada8b84da59348c4c5d? If yes, then a note on that would help.

supported_platforms.md
```diff
@@ -0,0 +1,55 @@
...
+Java version: 1.8.0_292, vendor: AdoptOpenJDK, runtime: C:\hostedtoolcache\windows\Java_Adopt_jdk\8.0.292-10\x64\jre
```
abhipsaMisra, line 19: I don't think the runtime info lends any extra information since we already have the version no specified. Do you think there is any value in including the runtime info?

supported_platforms.md
timtay-microsoft: I can probably get rid of it. It's probably just a download hash or something similar

supported_platforms.md
```diff
@@ -0,0 +1,55 @@
...
+Default locale: en_US, platform encoding: UTF-8
+OS name: "linux", version: "4.15.0-1113-azure", arch: "amd64", family: "unix"
+
+Note that, while we only directly test on Ubuntu 1604, we do generally support other popular linux distributions. 
+
+## Android
+
+This SDK supports Android API versions 24, 25, 26, 27, and 28.
+
+The android emulator image used for nightly tests is "system-images;android-28;google_apis;x86_64"
```
abhipsaMisra, line 39: wouldn't simply specifying that we test against api version 28 be enough, or does this also lend some additional information? It might be easier for customers if we specify the exact info that they require rather than leaving them to parse it. What do you think?

supported_platforms.md
timtay-microsoft: Nope, I agree.

abhipsaMisra (approved):

supported_platforms.md
timtay-microsoft: There are a lot of Android emulators at API level 28, so this does provide a bit more context. The benefit here would be that users could try out the same emulator that we use. Maybe it is a bit too much extra information, though?

supported_platforms.md
timtay-microsoft: Will do

supported_platforms.md
abhipsaMisra: I personally couldn't understand how to interpret this string - should I be looking for this exact string in Android Studio? What if I use a different IDE? How does this string translate to what I need to correctly identify the emulator that our nightly pipeline uses?
I think adding some additional context on what the different parts of this string signify would be helpful here.

readme.md
```diff
@@ -33,9 +33,11 @@ Devices and data sources in an IoT solution can range from a simple network-conn
    * **Working with the SDKs code**: if you are working with the SDKs code to modify it or contribute changes, then you can clone the repository and build the libraries:
       * [Build Device SDK from code][device-code]
       * [Build Service SDK from code][service-code]
-* On Android: our Java device SDK can be used on Android using the API version 17 and higher:
+* On Android: our Java device SDK can be used on Android:
    * [Device SDK][device-android]
 
+For more details on what platforms this SDK supports, see [this document](./supported_platforms.md)
```
drwill-ms, line 39: This sentence could use an ending period.

supported_platforms.md
drwill-ms: Should we capitalize Android?

supported_platforms.md
```diff
@@ -0,0 +1,59 @@
...
+
+The gradle version used in nightly tests is 3.6.3
+
+The build.gradle file that outlines more specifics can be found [here](./iot-e2e-tests/android/app/build.gradle)
```
drwill-ms, line 47: Ending period.

drwill-ms: elsewhere

drwill-ms (approved):

supported_platforms.md
timtay-microsoft: Good point. I can remove this string since adding context to it would probably fall outside of the scope of this document's purpose

supported_platforms.md
ericmitt: It looks like we test on Windows Server, but no mention of Win10 desktop client. We should we specify it somewhere

ericmitt (changes requested):
need to mention Win10

ericmitt (approved):
LGTM

supported_platforms.md
```diff
@@ -0,0 +1,57 @@
...
+Apache Maven 3.8.1
+Java version: 1.8.0_292, vendor: AdoptOpenJDK
+
+Default locale: en_US, platform encoding: Cp1252
+
+OS name: "windows server 2016", version: "10.0", arch: "amd64", family: "windows"
+
+## Ubuntu 1604
+
+Note that, while we only directly test on Ubuntu 1604, we do generally support other popular linux distributions.
```
timtay-microsoft, line 29:
```suggestion
Note that, while we only directly test on Ubuntu 1604, we do generally support other popular Linux distributions. 
```

PR #1251, opened 2021-06-21
refactor(all): Broad fix for warnings, removing unused code

jamdavi:
This is the result of the warning review with @timtay-microsoft today.

This PR will have a number of warnings fixed but the  most common two are listed below.

### Can be final
Suggests the member is only set by the constructor. You will see a number of members with final added to the declaration

### Redundant Suppression
You'll see a lot of removed `@SuppressWarnings()` annotations because likely another class level suppression or a change to the code has fixed this.

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketUpgrade.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketUpgrade.java
@@ -18,6 +18,7 @@ import static org.apache.commons.codec.binary.Base64.encodeBase64String;
 public class WebSocketUpgrade
 {
     public static final String RFC_GUID = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
+    @SuppressWarnings("FieldCanBeLocal")
     private final char _colon = ':';
     private final char _slash = '/';
     private String _host = "";
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
@@ -12,11 +12,9 @@ import javax.net.ssl.SSLContext;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 
 /**
  * A client for creating multiplexed connections to IoT Hub. A multiplexed connection allows for multiple device clients
@@ -519,7 +517,7 @@ public class MultiplexingClient
                     // Only update the local state map once the register call has succeeded
                     String deviceIdThatAttemptedToRegister = clientsThatAttemptedToRegister.getConfig().getDeviceId();
 
-                    if (!e.getRegistrationExceptions().keySet().contains(deviceIdThatAttemptedToRegister))
+                    if (!e.getRegistrationExceptions().containsKey(deviceIdThatAttemptedToRegister))
                     {
                         this.multiplexedDeviceClients.put(deviceIdThatAttemptedToRegister, clientsThatAttemptedToRegister);
                     }
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ReportedPropertiesParameters.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ReportedPropertiesParameters.java
@@ -106,7 +106,7 @@ public class ReportedPropertiesParameters {
     /**
      * The reported properties to send
      */
-    private Set<Property> reportedProperties = null;
+    private final Set<Property> reportedProperties;
     /**
      * The version of the properties
      */
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -289,7 +289,6 @@ public class IotHubTransport implements IotHubListener
                     log.warn("A message was received with a null correlation id.");
                 }
             }
-
         }
         catch (Exception ex)
         {
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -276,12 +276,14 @@ public class IotHubTransport implements IotHubListener
 
         try
         {
-            String messageId = message.getCorrelationId();
-
-            if (messageId != null && correlationCallbacks.containsKey(messageId))
+            if (message != null)
             {
-                Object context = correlationCallbackContexts.get(messageId);
-                correlationCallbacks.get(messageId).onResponseReceived(message, context, e);
+                String messageId = message.getCorrelationId();
+                if (messageId != null && correlationCallbacks.containsKey(messageId))
+                {
+                    Object context = correlationCallbackContexts.get(messageId);
+                    correlationCallbacks.get(messageId).onResponseReceived(message, context, e);
+                }
             }
         }
         catch (Exception ex)
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -284,7 +284,12 @@ public class IotHubTransport implements IotHubListener
                     Object context = correlationCallbackContexts.get(messageId);
                     correlationCallbacks.get(messageId).onResponseReceived(message, context, e);
                 }
+                else
+                {
+                    log.warn("A message was received with a null correlation id.");
+                }
             }
+
         }
         catch (Exception ex)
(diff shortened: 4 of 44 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/CbsSenderLinkHandler.java
```diff
@@ -115,6 +115,7 @@ public int sendAuthenticationMessage(UUID authenticationMessageCorrelationId)
         {
             TokenRequestContext context = new TokenRequestContext().addScopes(IOTHUB_PUBLIC_SCOPE);
             this.currentAccessToken = credential.getToken(context).block();
+            //noinspection ConstantConditions
```
jamdavi, line 118: @timtay-microsoft Here is your reminder about this method.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobClient.java
```diff
@@ -177,7 +177,6 @@ public JobClient(String hostName, AzureSasCredential azureSasCredential, JobClie
      * @throws IOException if the function cannot create a URL for the job
      * @throws IotHubException if the http request failed
      */
-    @SuppressWarnings("deprecation")
```
barustum, line 180: What was the deprecation attribute getting us?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSend.java
```diff
@@ -157,13 +157,15 @@ public AmqpSend(
     /**
      * Create AmqpsSendHandler and store it in a member variable
      */
+    @SuppressWarnings("EmptyMethod")
```
barustum, line 160: are open and close needed?

barustum: ..I guess removing them would be a breaking change.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/jobs/JobClient.java
jamdavi: This was a "redundant suppression"

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpSend.java
jamdavi: Correct

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/CbsSenderLinkHandler.java
timtay-microsoft: Okay, so line 117 may return a null value here, so we would be better off checking if this.currentAccessToken is null and logging an error and returning -1 from this method

timtay-microsoft: ```credential``` is a user provided ```TokenCredential``` interface implementation, so a user could hypothetically pass in an implementation of the interface that just always returns null

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketUpgrade.java
```diff
@@ -18,6 +18,7 @@
 public class WebSocketUpgrade
 {
     public static final String RFC_GUID = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
+    @SuppressWarnings("FieldCanBeLocal")
```
jamdavi, line 21: There are just some constants defined to help build the URI. This field is only used in one method but could possibly be used somewhere else later on.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -519,7 +517,7 @@ public void registerDeviceClients(Iterable<DeviceClient> deviceClients, long tim
                     // Only update the local state map once the register call has succeeded
                     String deviceIdThatAttemptedToRegister = clientsThatAttemptedToRegister.getConfig().getDeviceId();
 
-                    if (!e.getRegistrationExceptions().keySet().contains(deviceIdThatAttemptedToRegister))
+                    if (!e.getRegistrationExceptions().containsKey(deviceIdThatAttemptedToRegister))
```
jamdavi, line 520: This was a suggestion to remove overly complex/inefficient accessors.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ReportedPropertiesParameters.java
```diff
@@ -106,7 +106,7 @@ Object getReportedPropertiesCallbackContext() {
     /**
      * The reported properties to send
      */
-    private Set<Property> reportedProperties = null;
+    private final Set<Property> reportedProperties;
```
jamdavi, line 109: null is implicit

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -276,12 +276,14 @@ else if (message != null)
 
         try
         {
-            String messageId = message.getCorrelationId();
-
-            if (messageId != null && correlationCallbacks.containsKey(messageId))
+            if (message != null)
```
jamdavi, line 279: Detected a null check was required here for messageId

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceOperations.java
```diff
@@ -106,11 +106,7 @@ public static HttpResponse request(
         request.setHeaderField(USER_AGENT, TransportUtils.javaServiceClientIdentifier + TransportUtils.serviceVersion);
         request.setHeaderField(ACCEPT, ACCEPT_VALUE);
         request.setHeaderField(CONTENT_TYPE, ACCEPT_VALUE + "; " + ACCEPT_CHARSET);
-
-        if (payload != null)
-        {
-            request.setHeaderField(CONTENT_LENGTH, String.valueOf(payload.length));
-        }
+        request.setHeaderField(CONTENT_LENGTH, String.valueOf(payload.length));
```
jamdavi, line 109: Null check is unessecary

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/RawTwinQuery.java
```diff
@@ -115,7 +115,6 @@ public RawTwinQuery(String hostName, AzureSasCredential azureSasCredential)
      * @throws IotHubException If IotHub did not respond successfully to the query
      * @throws IOException If any of the input parameters are incorrect
      */
-    @SuppressWarnings("deprecation")
```
jamdavi, line 118: Redundant supression

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/CbsSenderLinkHandler.java
```diff
@@ -169,7 +176,7 @@ private AccessToken getAccessTokenFromSasToken(String sasToken)
 
                 try
                 {
-                    expiryTimeSeconds = Integer.valueOf(expiryTimeValue);
+                    expiryTimeSeconds = Integer.parseInt(expiryTimeValue);
```
jamdavi, line 179: Suggestion to prevent boxing of the int value

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
timtay-microsoft: You may want to add an else case that simply logs a warning that "A message was received with no correlation Id"

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/CbsSenderLinkHandler.java
```diff
@@ -115,6 +115,13 @@ public int sendAuthenticationMessage(UUID authenticationMessageCorrelationId)
         {
             TokenRequestContext context = new TokenRequestContext().addScopes(IOTHUB_PUBLIC_SCOPE);
             this.currentAccessToken = credential.getToken(context).block();
+
+            if (this.currentAccessToken == null)
+            {
+                log.error("The AccessToken supplied by the TokenCrednetial for the CbsSenderLinkHandler was null.");
```
timtay-microsoft, line 121:
```suggestion
                log.error("The AccessToken supplied by the TokenCredential for the CbsSenderLinkHandler was null.");
```

timtay-microsoft: slight typo to fix

timtay-microsoft (approved):

PR #1252, opened 2021-06-22
refactor(all): Remove unused code

jamdavi:
@timtay-microsoft Here is the PR that would remove all unused protected code.

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/RegisterManager.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/RegisterManager.java
@@ -37,14 +37,6 @@ public class RegisterManager
     protected String moduleId = null;
 
     /**
-     * Device generation Id
-     */
-    private static final String GENERATION_ID_TAG = "generationId";
-    @Expose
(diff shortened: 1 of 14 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/RegisterManager.java
```diff
@@ -36,14 +36,6 @@
     @SerializedName(MODULE_ID_TAG)
     protected String moduleId = null;
 
-    /**
```
jamdavi, line 39: Seems like this generated code might need to stay?

I've left the serialized fields alone in the past.

timtay-microsoft: I'd suppress the warning here, yeah. This is a valid field in the json payload

timtay-microsoft: Actually, this entire class seems to be unused outside of test code. You can probably just delete the class and its tests

timtay-microsoft (approved):

Issue #1254, opened 2021-06-24
null serialization value


MikalaiSavenka:
Issue on line: https://github.com/Azure/azure-iot-sdk-java/blob/1ee51d807da156bf035d048cebd87b84ff6d0fe4/device/iot-device-samples/pnp-device-sample/temperature-controller-device-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/device/PnpConvention.java#L67

gson.toJson(telemetryPairs) returns null because telemetryPairs is anonymous class created on line https://github.com/Azure/azure-iot-sdk-java/blob/1ee51d807da156bf035d048cebd87b84ff6d0fe4/device/iot-device-samples/pnp-device-sample/temperature-controller-device-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/device/PnpConvention.java#L54 and Gson cannot serialize it correctly

abhipsaMisra:
Re-opening and closing to work around dashboard issue.

abhipsaMisra:
Closing the issue.

PR #1255, opened 2021-06-24
fix of issue_1254 

MikalaiSavenka:
https://github.com/Azure/azure-iot-sdk-java/issues/1254 fix

Gson cannot serialize  anonymous classes created when use {{}} initialization 

https://stackoverflow.com/questions/39538677/why-does-gson-tojson-always-return-null


timtay-microsoft (approved):
Looks good to me. Thanks for the fix!

PR #1260, opened 2021-07-02
Add sample that shows ecc certificate support

timtay-microsoft:
This feature is supported by IoT Hub as of now, so it is safe for us to add this sample now.


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/DeviceConnectionState.java
```diff
@@ -11,7 +11,11 @@
 public enum DeviceConnectionState
 {
     Disconnected("Disconnected"),
-    Connected("Connected");
+    Connected("Connected"),
+
+    // For some reason, preview hubs send this value with a lower case "disconnected" vs "Disconnected" and "connected" vs "Connected"
```
timtay-microsoft, line 16: We won't keep this when we merge. This is just a workaround for a bug that appears to only be present in preview hubs

PR #1262, opened 2021-07-08
[DRAFT] Adding convention based operations shared libraries (1/X)

jamdavi:
Add the base libraries and implementations for the Convention Based Operations, otherwise known as PnP.

This will be one of many PRs to the `jamdavi/pnpDev` branch which will eventually be merged into the `previews/pnpApi` branch. Which will of course make it to `preview` and then `master`.

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/convention/ConventionConstants.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/convention/ConventionConstants.java
@@ -1,0 +1,39 @@
+package com.microsoft.azure.sdk.iot.deps.convention;
+
+public final class ConventionConstants
+{
+    /// <summary>
+    /// Separator for a component-level command name.
+    /// </summary>
+    public static final char ComponentLevelCommandSeparator = '*';
+
+    /// <summary>
+    /// Marker key to indicate a component-level property.
+    /// </summary>
+    public static final String ComponentIdentifierKey = "__t";
+
+    /// <summary>
+    /// Marker value to indicate a component-level property.
+    /// </summary>
+    public static final String ComponentIdentifierValue = "c";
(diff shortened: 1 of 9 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/convention/ConventionConstants.java
```diff
@@ -0,0 +1,39 @@
+package com.microsoft.azure.sdk.iot.deps.convention;
+
+public final class ConventionConstants
+{
+    /// <summary>
+    /// Separator for a component-level command name.
+    /// </summary>
+    public static final char ComponentLevelCommandSeparator = '*';
```
drwill-ms, line 8: Why not const?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/convention/ConventionConstants.java
```diff
@@ -0,0 +1,39 @@
+package com.microsoft.azure.sdk.iot.deps.convention;
+
+public final class ConventionConstants
```
drwill-ms, line 3: Why not static?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/convention/DefaultPayloadConvention.java
```diff
@@ -0,0 +1,24 @@
+package com.microsoft.azure.sdk.iot.deps.convention;
+
+import lombok.Getter;
+
+/// <summary>
```
drwill-ms, line 5: Does Java support the same /// doc comments as C#??

jamdavi:
Closing draft to create a better set of PRs

PR #1264, opened 2021-07-13
fix(sample): Fix issue in device reconnection sample

timtay-microsoft:
~~The previous implementation did not try to open the connection again if a call to open() threw an exception. This new implementation will retry the open call if that happens.~~

So the only actual issue here is that the doConnect method appeared to not have retry baked in, but it actually did. I'd like to rename the method accordingly so that others don't make the same mistaken assumption that I made


abhipsaMisra:
what about this:https://github.com/Azure/azure-iot-sdk-java/blob/ce71dba21fa98ac992c62594b4314899c2319c0f/device/iot-device-samples/device-reconnection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java#L67?

timtay-microsoft:
> what about this:
> 
> https://github.com/Azure/azure-iot-sdk-java/blob/ce71dba21fa98ac992c62594b4314899c2319c0f/device/iot-device-samples/device-reconnection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java#L67
> 
> ?

Good point, let me rework this a bit

vinagesh (approved):

drwill-ms (approved):

PR #1265, opened 2021-07-13
fix(iot-deps, iot-device): Fix issue where non-alphanumeric ASCII cha…

timtay-microsoft:
…racters could be used for temporary keystore password

Not all bytes in range 0-255 represent alphanumeric characters, and non-alphanumeric characters in keystore passwords may result in an exception depending on the operation system


timtay-microsoft:
My local machine started running into this error after a recent Windows update. While I'm not sure that the Windows update was the cause for this restriction on keystore passwords, this fix resolves it anyways.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
```diff
@@ -280,16 +280,8 @@ private TrustManagerFactory generateTrustManagerFactory(IotHubCertificateManager
 
     private char[] generateTemporaryPassword()
     {
-        byte[] randomBytes = new byte[256];
```
timtay-microsoft, line 283: The [ASCII table](https://www.asciitable.com/) shows that many of the characters on the lower end of the 0-255 range are inappropriate for keystore passwords

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
```diff
@@ -280,16 +280,8 @@ private TrustManagerFactory generateTrustManagerFactory(IotHubCertificateManager
...
-        new SecureRandom().nextBytes(randomBytes);
-
-        for (int i = 0; i < 256; i++)
-        {
-            randomChars[i] = (char) randomBytes[i];
-        }
-
-        return randomChars;
+        // removing all dash characters so that the password is alphanumeric
+        return UUID.randomUUID().toString().replace("-", "").toCharArray();
```
abhipsaMisra, line 284: A sample uuid format (and the corresponding error description it might generate) would help explain while the "-" removal is necessary

abhipsaMisra (approved):

timtay-microsoft:
Actually, scratch this approach. I'd like to still use SecureRandom here rather that UUID. I'll get up a new implementation shortly

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/SSLContextBuilder.java
```diff
@@ -136,13 +136,14 @@ public static RSAPrivateKey parsePrivateKeyString(String privateKeyPEM) throws G
 
     private static char[] generateTemporaryPassword()
```
vinagesh, line 137: Can this method be moved to a shared helper instead? Looks like the same logic is being used in two files.

vinagesh (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/SSLContextBuilder.java
timtay-microsoft: Possibly, but I'd rather not make this a public method.

PR #1272, opened 2021-07-19
break(iot-dev): Remove deprecated APIs from IoT hub device package

timtay-microsoft:
The highlights of this include:
- Removing the old multiplexing client "TransportClient", its tests, and its sample
- Removing some deprecated device client and module client constructors
- Continuing the pattern that was started in #1271 around final getters


device/iot-device-samples/send-receive-x509-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/SSLContextBuilder.java
```diff
@@ -0,0 +1,276 @@
...
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.UUID;
+
+/**
+ * Helper class that demonstrates how to build an SSLContext for x509 authentication from your public and private certificates,
+ * or how to build an SSLContext for SAS authentication from the default IoT Hub public certificates
+ */
+public class SSLContextBuilder
```
timtay-microsoft, line 39: This same class exists in another sample already. Copying it over to this sample as well since it is needed here as well

timtay-microsoft: This sample used to use a device client constructor that read certificates from strings, but that was one of the constructors to go. Now users need to pass in the SSLContext directly if they want to parse certificates

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnectionTest.java
```diff
@@ -478,7 +478,7 @@ public void openThrowsIfProtonReactorThrows() throws TransportException
 
     // Tests_SRS_AMQPSIOTHUBCONNECTION_15_009: [The function shall trigger the Reactor (Proton) to begin running.]
     @Test
-    public void openTriggersProtonReactor(@Mocked final Reactor mockedReactor) throws TransportException, InterruptedException
+    public void openTriggersProtonReactorFinal(@Mocked final Reactor mockedReactor) throws TransportException, InterruptedException
```
drwill-ms, line 481: Final? I thought you removing all copies without Final, and renaming the *Final ones to the base name.

device/iot-device-samples/send-receive-x509-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/SSLContextBuilder.java
drwill-ms: In .NET we'd case this as `SslContextBuilder`. Not the same for Java?

device/iot-device-samples/send-receive-x509-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/SendReceiveX509.java
```diff
@@ -148,7 +150,7 @@ else if (status == IotHubConnectionStatus.CONNECTED)
      * args[1] = number of requests to send
      * args[2] = protocol (optional, one of 'mqtt' or 'amqps' or 'https' or 'amqps_ws')
      */
-    public static void main(String[] args) throws IOException, URISyntaxException
+    public static void main(String[] args) throws IOException, URISyntaxException, GeneralSecurityException
```
drwill-ms, line 153: `GeneralSecurityException`

/salute

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnectionTest.java
timtay-microsoft: Looks like this was just a find + replace error. I'll revert it. This test has nothing to do with the getters that I'm collapsing

drwill-ms (approved):

device/iot-device-samples/send-receive-x509-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/SSLContextBuilder.java
```diff
@@ -0,0 +1,276 @@
...
+    /**
+     * Create an SSLContext instance with the provided public and private keys that also trusts the base iot hub certificates
+     * @param publicKeyCertificateString the public key to use for x509 authentication. Does not need to include the default
+     *                                   Iot Hub trusted certificate as it will be added automatically
+     * @param privateKeyString The private key to use for x509 authentication
+     * @return The created SSLContext that uses the provided public key and private key
+     * @throws GeneralSecurityException If the certificate creation fails, or if the SSLContext creation using those certificates fails
+     * @throws IOException If the certificates cannot be read
+     */
+    public static SSLContext buildSSLContext(String publicKeyCertificateString, String privateKeyString) throws GeneralSecurityException, IOException
```
abhipsaMisra, line 55: With the 2.0.0 packages, all certificates will be read from the device's certificate store, even the certs for an x509 authenticated device/module identity. So I am assuming this sample will be modified once we have the code in place for reading certificates from the trusted certificates store, right?

timtay-microsoft: Kind of. We still need to demonstrate how to build an SSLContext for users who want to trust certificates other than the default ones installed on their machine in case they don't want to install the certs on the machine directly for any reason

timtay-microsoft: Notably, users who are doing x509 auth will still need to construct an SSLContext and pass it in for us to use. There is no way our SDK will be able to read private keys from the device directly without a user providing us the private key store and its password. I'll get more into that in the upcoming PR about reading certs from the device though

abhipsaMisra: Hmm, so this actually touches the point that first sparked this question in my mind. In neither of the sas-based or X509 certificates based flows does iot hub require the private key information, but the private key info seems to be necessary in building the ssl context. Is there no way to implement this in java which would get away from the requirement of the user having to supply the certificate private key to the SDK?

Just for comparison, the .NET SDK does not require the private key info to be sent with the certificate info that is sent to service. Instead, it is read directly from the certificate store by the TCP stack during TLS negotiation.

abhipsaMisra: I've remember this coming up as an issue mostly on linux systems where customers installed their certificates via openSsl and then restricted the access to only the tcp stack.

timtay-microsoft: So they don't even need to pass in a password for the key store that holds the private key? Interesting... I may need to do some more digging to see if we can mimic that in Java. For now though, I'll leave these samples as is

timtay-microsoft: > In neither of the sas-based or X509 certificates based flows does iot hub require the private key information

The client is expected to use the private key locally to encrypt the data that is sent over the wire. While the client doesn't send the private key to IoT Hub, it is still required

timtay-microsoft: > Instead, it (edit: is can be) read directly from the certificate store by the TCP stack during TLS negotiation.

A device can have multiple private keys installed though, right? How does the .NET SDK's TLS stack figure out which private key to use?

abhipsaMisra: Yes, the private key is required, for sure, what I mean is that it is required on a layer lower than the SDK deals with. I am not very confident on how the TLS stack figures out which private key to use, possibly through the common-name and thumbprint combination?

PR #1274, opened 2021-07-20
break(iot-dev): Remove deprecated APIs from IoT hub service package and samples

timtay-microsoft:
There are a few random deprecated methods/constructors that are removed, but the bulk of this is 

```java
@Deprecated
public someServiceClient createFromConnectionString(String connectionString) throws IOException;

public someServiceClient(String connectionString);
```

being simplified to

```java
public someServiceClient(String connectionString);
```

and more cleaning up of the final getters

old:
```java
@Deprecated
public String getFoo() {...}

public final String getFooFinal() {...}
```

new:
```java
public final String getFoo() {...}
```


timtay-microsoft:
There are also a number of client APIs that had an overload with a "forceUpdate" flag argument that did nothing, so those APIs are being removed

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwinDevice.java
```diff
@@ -191,7 +191,7 @@ public Integer getTagsVersion()
         {
             throw new IllegalArgumentException("tag is null");
         }
-        return this.tag.getVersionFinal();
```
drwill-ms, line 194: But what about the [final countdown](https://www.youtube.com/watch?v=9jK-NcRmVcw)?

service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/FeedbackReceiverTest.java
```diff
@@ -34,7 +34,7 @@ public void constructor_input_hostName_null() throws Exception
         String deviceId = "xxx";
         IotHubServiceClientProtocol iotHubServiceClientProtocol = IotHubServiceClientProtocol.AMQPS;
         // Act
-        FeedbackReceiver feedbackReceiver = new FeedbackReceiver(hostName, userName, sasToken, iotHubServiceClientProtocol, deviceId);
+        FeedbackReceiver feedbackReceiver = new FeedbackReceiver(hostName, userName, sasToken, iotHubServiceClientProtocol);
```
drwill-ms, line 37: Can you say more about removing the `deviceId` parameter?

drwill-ms (approved):

service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/FeedbackReceiverTest.java
timtay-microsoft: This constructor was unused, and the test could have just been deleted. The test is fine for testing the same behavior on this other constructor though, so I kept it around

timtay-microsoft:
I need to make sure that the IotHubConnectionStringBuilder matches up with the pattern I'm establishing here. Don't merge this yet, me

service/iot-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/service/FeedbackReceiverTest.java
timtay-microsoft: More generally, the feedback receiver isn't tied to any particular device, so this constructor that took a deviceId never did anything with it to begin with. That was why we deprecated it in the first place. Feedback receivers can get messages for all devices

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/BaseDevice.java
```diff
@@ -188,34 +188,9 @@ public String getSecondaryKey()
...
-    }
-
-    /**
-     * Setter for X509 thumbprint
-     * @param primaryThumbprint the primary thumbprint to set
-     * @param secondaryThumbprint the secondary thumbprint to set
      * @throws IllegalArgumentException if primaryThumbprint or secondaryThumbprint is null or empty
      */
-    public final void setThumbprintFinal(String primaryThumbprint, String secondaryThumbprint)
+    public final void setThumbprint(String primaryThumbprint, String secondaryThumbprint)
```
vinagesh, line 193: We suggested above that only setThumbprint is deprecated, not setThumbprintFinal. Would it not be a breaking change if we rename it?

timtay-microsoft: Yes, this is a breaking change. But this seems like a reasonable enough change since setters don't typically follow the pattern of ```setFooFinal()```

timtay-microsoft:
> I need to make sure that the IotHubConnectionStringBuilder matches up with the pattern I'm establishing here. Don't merge this yet, me

The IotHubConnectionStringBuilder is a bit different from the service clients in that we don't construct an instance of IotHubConnectionStringBuilder. We have static "constructors" that build IotHubConnectionString instances instead. With that in mind, I will keep the IotHubConnectionStringBuilder factory methods as is

PR #1276, opened 2021-07-21
break(dps-service): Remove deprecated APIs from DPS service package and samples

timtay-microsoft:
For the provisioning service client, unlike other service clients, we didn't have any deprecated constructors/factory methods. However, to maintain parity with the service clients in the IoT hub packages that have constructors rather than factory methods, I'm proposing:

current
```java
public static ProvisioningServiceClient createFromConnectionString(String connectionString);

// note that this is currently private
private ProvisioningServiceClient(String connectionString)
```

proposed:
```java
//not deleted since it wasn't deprecated
public static ProvisioningServiceClient createFromConnectionString(String connectionString);

//private -> public
public ProvisioningServiceClient(String connectionString)
```

This change would only be additive, not breaking

### DIFF
--- a/provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/ProvisioningServiceClient.java
+++ b/provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/ProvisioningServiceClient.java
@@ -167,7 +167,7 @@ public final class ProvisioningServiceClient
      * {
      *     // Create a Device Provisioning Service Client.
      *     DeviceProvisioningServiceClient deviceProvisioningServiceClient =
-     *         Devicenew ProvisioningServiceClient(PROVISIONING_CONNECTION_STRING);
+     *         new ProvisioningServiceClient(PROVISIONING_CONNECTION_STRING);
      *
      *     // Create a new individualEnrollment configurations.
      *     Attestation attestation = new TpmAttestation(TPM_ENDORSEMENT_KEY);
@@ -201,7 +201,7 @@ public final class ProvisioningServiceClient
      * {
      *     // Create a Device Provisioning Service Client.
      *     DeviceProvisioningServiceClient deviceProvisioningServiceClient =
-     *         Devicenew ProvisioningServiceClient(PROVISIONING_CONNECTION_STRING);
+     *         new ProvisioningServiceClient(PROVISIONING_CONNECTION_STRING);
      *
      *     // Get the content of the previous individualEnrollment.
      *     IndividualEnrollment individualEnrollment =  deviceProvisioningServiceClient.getIndividualEnrollment(REGISTRATION_ID);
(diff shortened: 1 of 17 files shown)


provisioning/provisioning-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/provisioning/service/configs/EnrollmentGroupTest.java
```diff
@@ -99,24 +99,6 @@ public Attestation getAttestation()
             return mockedAttestation;
         }
 
-        @Mock
```
timtay-microsoft, line 102: Unnecessary to mock these methods

provisioning/provisioning-service-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/provisioning/service/configs/IndividualEnrollmentTest.java
```diff
@@ -79,24 +79,12 @@ public Attestation getAttestation()
             return mockedAttestation;
         }
 
-        @Mock
```
timtay-microsoft, line 82: Unnecessary to mock these methods

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/ProvisioningServiceClient.java
```diff
@@ -160,7 +167,7 @@ private ProvisioningServiceClient(String connectionString)
      * {
      *     // Create a Device Provisioning Service Client.
      *     DeviceProvisioningServiceClient deviceProvisioningServiceClient =
-     *         DeviceProvisioningServiceClient.createFromConnectionString(PROVISIONING_CONNECTION_STRING);
```
abhipsaMisra, line 163: Since this is a major version bump and can have breaking changes, are we planning on deleting `DeviceProvisioningServiceClient.createFromConnectionString` anyway?
If not, do we plan on marking it as deprecated here instead. 

I think having multiple ways to initialize a client with the same parameters can be confusing to customers, and we should provide them with a clear recommended path. What do you think?

abhipsaMisra (approved):

vinagesh:
Change is good but marking the static method as deprecated would be good so that we can remove that in the future. It might be confusing to have two ways to make the client.

vinagesh (approved):

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/ProvisioningServiceClient.java
timtay-microsoft: I'm only hesitant to delete it now because it was never deprecated. As per the point @vinagesh makes below, I'd be happy to mark this static constructor as deprecated to help cut down on the confusion here.

timtay-microsoft:
> Change is good but marking the static method as deprecated would be good so that we can remove that in the future. It might be confusing to have two ways to make the client. Thoughts?

Agreed. I'll go ahead and mark the static constructor as deprecated.

PR #1280, opened 2021-07-23
refactor(iot-dev): Raise the max in flight messages limit for mqtt

timtay-microsoft:
Clients using MQTT were limited to 10 messages in flight at a time, which hurt the client's ability to achieve higher messaging throughputs. Raising this value to a higher number. We can make this value configurable at some point in the future if someone asks for it, but this value should be enough to not hamstring anyone.

"In-flight" messages are messages that have been sent, but haven't been ack'd yet.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
@@ -32,6 +32,8 @@ abstract public class Mqtt implements MqttCallback
     private static final int QOS = 1;
     private static final int MAX_SUBSCRIBE_ACK_WAIT_TIME = 15 * 1000;
 
+    // relatively arbitrary, but only because Paho doesn't have any particular recommendations here. Just a high enough
+    // value that users who are building a gateway type solution don't find this value to be a bottleneck.
     static final int MAX_IN_FLIGHT_COUNT = 65000;
 
     private MqttAsyncClient mqttAsyncClient;
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
@@ -32,8 +32,7 @@ abstract public class Mqtt implements MqttCallback
     private static final int QOS = 1;
     private static final int MAX_SUBSCRIBE_ACK_WAIT_TIME = 15 * 1000;
 
-    // paho mqtt only supports 10 messages in flight at the same time
-    private static final int MAX_IN_FLIGHT_COUNT = 10;
+    static final int MAX_IN_FLIGHT_COUNT = 65000;
 
     private MqttAsyncClient mqttAsyncClient;
     private final MqttConnectOptions connectOptions;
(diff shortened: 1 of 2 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
```diff
@@ -32,8 +32,9 @@
     private static final int QOS = 1;
     private static final int MAX_SUBSCRIBE_ACK_WAIT_TIME = 15 * 1000;
 
-    // paho mqtt only supports 10 messages in flight at the same time
-    private static final int MAX_IN_FLIGHT_COUNT = 10;
+    // relatively arbitrary, but only because Paho doesn't have any particular recommendations here. Just a high enough
+    // value that users who are building a gateway type solution don't find this value to be a bottleneck.
+    static final int MAX_IN_FLIGHT_COUNT = 65000;
```
drwill-ms, line 37: what is the impact of removing `private`? What is the default?

timtay-microsoft: The default is "package private" which means it can now be referenced by other classes in the same package. Since we use this value in a different class in the same package, this modifier change is needed

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -182,6 +183,7 @@ else if (this.config.getAuthenticationType() == DeviceClientConfig.AuthType.X509
         connectOptions.setCleanSession(SET_CLEAN_SESSION);
         connectOptions.setMqttVersion(MQTT_VERSION);
         connectOptions.setUserName(iotHubUserName);
+        connectOptions.setMaxInflight(MAX_IN_FLIGHT_COUNT);
```
drwill-ms, line 186: How does a customer override this default if they want to?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
drwill-ms: In .NET, we prefer to be explicit rather than assume defaults. Not so in Java?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
timtay-microsoft: Currently they can't. Its something that would could add later if a customer asks for it, but no one has asked for it yet.

drwill-ms: Since this is a breaking change, are we worried at all about what the impact might be on customers? Their throughput will increase. That could possibly cause problems, no?

I'm not against the change, just trying to think it through while we have the chance.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
timtay-microsoft: There is no explicit way to declare a field/method "package private"  in Java other than to have no access modifier

```java
private String myPrivateField;
String myPackagePrivateField;
protected String myInheritibleField;
public String myFullyPublicField;
```

drwill-ms (approved):

PR #1282, opened 2021-07-26
feat(pnp): Adding client implementation

jamdavi:
Adding the device client and internal client implementation of the methods.

>NOTE I moved around the Ssl and Sas Methods to group the constructors together.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
```diff
@@ -120,21 +140,6 @@ public DeviceClientConfig(IotHubConnectionString iotHubConnectionString) throws
         configSasAuth(iotHubConnectionString);
     }
 
-    private void configSasAuth(IotHubConnectionString iotHubConnectionString) {
```
jamdavi, line 123: I moved this around to group the constructors together better.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
```diff
@@ -197,46 +201,19 @@ public DeviceClientConfig(IotHubConnectionString iotHubConnectionString, ClientO
         }
     }
 
+    @Builder
     public DeviceClientConfig(IotHubConnectionString iotHubConnectionString, SSLContext sslContext)
     {
         configSsl(iotHubConnectionString, sslContext);
     }
 
-    private void configSsl(IotHubConnectionString iotHubConnectionString, SSLContext sslContext) {
```
jamdavi, line 205: I moved this around to group the constructors together better.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -1216,4 +1221,101 @@ private void verifyMethodsAreSupported()
...
+
+    /**
+     * Set the global command callback handler.
+     * @param callback The callback to be used for commands.
+     * @param callbackContext An optional user context to be sent to the callback.
+     */
+    public void subscribeToCommands(DeviceMethodCallback callback, Object callbackContext)
+    {
+        // Subscribe to methods default handler internally and use the callback received internally to invoke the user supplied command callback.
+        // TODO Implement command handler
```
jamdavi, line 1265: I am actively working on this feature, just needed to get some eyes on this before it gets too long.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceTwin/DeviceTwin.java
```diff
@@ -90,6 +94,14 @@ public IotHubMessageResult execute(Message message, Object callbackContext)
                         if (iotHubStatus == IotHubStatusCode.OK)
                         {
                             TwinState twinState = TwinState.createFromPropertiesJson(new String(dtMessage.getBytes(), Message.DEFAULT_IOTHUB_MESSAGE_CHARSET));
+
+                            if (dtMessage.getClientPropertiesCallback() != null)
+                            {
+                                ClientPropertiesCallback propCallBack = dtMessage.getClientPropertiesCallback();
+                                Object propCallBackContext = dtMessage.getClientPropertiesCallbackContext();
+                                propCallBack.execute(new ClientProperties(ClientPropertyCollection.fromMap(twinState.getReportedProperty()), ClientPropertyCollection.fromMap(twinState.getDesiredProperty())), propCallBackContext);
+                            }
```
jamdavi, line 103: By default we will continue to execute other methods when a customer calls getTwin().

This means that we can have some unexpected doubling of callbacks if a customer implements both convention based and non-convention based operations.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceTwin/DeviceTwin.java
```diff
@@ -90,6 +94,14 @@ public IotHubMessageResult execute(Message message, Object callbackContext)
                         if (iotHubStatus == IotHubStatusCode.OK)
                         {
                             TwinState twinState = TwinState.createFromPropertiesJson(new String(dtMessage.getBytes(), Message.DEFAULT_IOTHUB_MESSAGE_CHARSET));
+
+                            if (dtMessage.getClientPropertiesCallback() != null)
+                            {
+                                ClientPropertiesCallback propCallBack = dtMessage.getClientPropertiesCallback();
+                                Object propCallBackContext = dtMessage.getClientPropertiesCallbackContext();
```
barustum, line 101: You guard against getClientPropertiesCallback. Do we need to guard against getClientPropertiesCallbackContext?

jamdavi: In this case no, since the callback will be implemented by the customer. I'll make sure to add some documentation in the callback class.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientOptions.java
```diff
@@ -3,14 +3,20 @@
...
 import lombok.Getter;
+import lombok.NoArgsConstructor;
 import lombok.Setter;
 
 import javax.net.ssl.SSLContext;
 
 /**
  * Options that allow configuration of the device client instance during initialization.
  */
+@Builder
```
jamdavi, line 18: @timtay-microsoft I don't believe this will break anything, do you foresee any issues with making this a builder class as well?

timtay-microsoft: Since you added the noArgsConstructor annotation as well, you should be fine to do this, yeah

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -13,6 +13,7 @@
 import com.microsoft.azure.sdk.iot.device.transport.amqps.IoTHubConnectionType;
 import com.microsoft.azure.sdk.iot.device.transport.https.HttpsTransportManager;
 import com.microsoft.azure.sdk.iot.provisioning.security.SecurityProvider;
+import lombok.Getter;
```
timtay-microsoft, line 16: Unused?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
```diff
@@ -197,46 +201,19 @@ public DeviceClientConfig(IotHubConnectionString iotHubConnectionString, ClientO
         }
     }
 
+    @Builder
```
timtay-microsoft, line 204: What does this annotation do when added to a constructor rather than to the class?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceTwin/DeviceTwin.java
```diff
@@ -403,6 +417,61 @@ public void subscribeDesiredPropertiesTwinPropertyNotification(Map<Property, Pai
...
+
+        IotHubTransportMessage clientPropertiesRequest = new IotHubTransportMessage(serializedReportedProperties.getBytes(), MessageType.DEVICE_TWIN);
+        clientPropertiesRequest.setCorrelatingMessageCallback(correlatingMessageCallback);
+        clientPropertiesRequest.setCorrelatingMessageCallbackContext(correlatingMessageCallbackContext);
+        clientPropertiesRequest.setConnectionDeviceId(this.config.getDeviceId());
+
+        // MQTT does not have the concept of correlationId for request/response handling but it does have a requestId
+        // To handle this we are setting the correlationId to the requestId to better handle correlation
+        // whether we use MQTT or AMQP.
+        clientPropertiesRequest.setRequestId(UUID.randomUUID().toString());
```
timtay-microsoft, line 455: If we aren't saving this UUID anywhere, do we even need to attach it to the outgoing message?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -1216,4 +1221,101 @@ private void verifyMethodsAreSupported()
             throw new UnsupportedOperationException(METHODS_OVER_HTTP_ERROR_MESSAGE);
         }
     }
+
+    /**
+     * Sends the TelemetryMessage to IoT hub.
+     * @param telemetryMessage The user supplied telemetry message.
+     */
+    public void sendTelemetry(TelemetryMessage telemetryMessage)
```
timtay-microsoft, line 1229: Is this a PnP specific API? The naming is vague enough that I'd expect customers to be confused and use this to send non-pnp telemetry

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -1216,4 +1221,101 @@ private void verifyMethodsAreSupported()
...
+        sendTelemetry(telemetryMessage, null, null);
+    }
+
+    /**
+     * Sends the TelemetryMessage to IoT hub.
+     * @param telemetryMessage The user supplied telemetry message.
+     * @param callback the callback to be invoked when a response is received. Can be {@code null}.
+     * @param callbackContext a context to be passed to the callback. Can be {@code null} if no callback is provided.
+     */
+    public void sendTelemetry(TelemetryMessage telemetryMessage, IotHubEventCallback callback, Object callbackContext)
```
timtay-microsoft, line 1240: What do you think about appending "Async" to the end of these methods that execute callbacks to notify you when the operation has completed?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -1216,4 +1221,101 @@ private void verifyMethodsAreSupported()
...
+    {
+        // In Java we don't return an object for the DeviceTwin, but instead we pass a Map<String, Object> in the form of a Device
+        // this map gets populated with a number of callbacks.
+        try
+        {
+            twin.getClientProperties(callback, callbackContext);
+        }
+        catch (Throwable ex)
+        {
+            throw ex;
```
timtay-microsoft, line 1284: Why catch the exception if you just re-throw it?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -1216,4 +1221,101 @@ private void verifyMethodsAreSupported()
...
+        this.twin.updateClientProperties(clientProperties, null, null,null, callback, callbackContext);
+    }
+
+    /**
+     * Set the global writable properties callback handler.
+     * @param writablePropertyUpdateCallback The callback to be used for writable properties.
+     * @param callbackContext An optional user context to be sent to the callback.
+     * @throws IOException
+     */
+    public void subscribeToWritablePropertiesEvent(WritablePropertiesRequestsCallback writablePropertyUpdateCallback, Object callbackContext) throws IOException
```
timtay-microsoft, line 1317:
```suggestion
    public void subscribeToWritablePropertiesEvents(WritablePropertiesRequestsCallback writablePropertyUpdateCallback, Object callbackContext) throws IOException
```

Since you expect the event to happen more than once?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientOptions.java
```diff
@@ -31,4 +37,12 @@
     @Setter
     @Getter
     public SSLContext sslContext;
+
+    /**
+     * The convention to be used for convention based operations.
+     */
+    @Setter
+    @Getter
+    @Builder.Default
```
abhipsaMisra, line 46: This is interesting, so now any time ClientOptions is created using a builder, the payload convention is set to our default convention (unless explicitly set to something else). Is that something we want to do? Should convention be tied to the existence of a model Id instead?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClientConfig.java
abhipsaMisra: James can correct me if I am wrong but I believe a class with no constructors is given the annotation at class level, while if you have a constructor defined you'll need to add the annotation to the constructor directly

abhipsaMisra: I don't see the other constructors in the class being given this annotation, anything special about this particular constructor?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
abhipsaMisra: Do we have an existing send events API that doesn't take in a callback? Without the callback this API behaves as fire-and-forget. I wonder how useful it would be for users.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -1216,4 +1221,101 @@ private void verifyMethodsAreSupported()
...
+
+        sendEventAsync(telemetryMessage, callback, callbackContext);
+    }
+
+    /**
+     * Set the global command callback handler.
+     * @param callback The callback to be used for commands.
+     * @param callbackContext An optional user context to be sent to the callback.
+     */
+    public void subscribeToCommands(DeviceMethodCallback callback, Object callbackContext)
```
abhipsaMisra, line 1262: We'd be renaming DeviceMethodCallback, right? Replace method -> command

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceTwin/DeviceTwin.java
jamdavi: This UUID can be accessed via the callbacks I implemented.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
jamdavi: We do have some existing methods that are fire and forget.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientOptions.java
jamdavi: That is a good question. In all cases only convention aware methods would use this so having it set to default makes sure if someone use a convention aware method without setting the model id it would always use our convention.

Do we do this in .NET?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
jamdavi: Yes, removed this for now will be adding another PR for this.

PR #1285, opened 2021-07-26
break(deps): Make IotHubSSLContext read certs from device store rather than from source

timtay-microsoft:
Keeping this PR limited to the Deps package changes because this is crucial for us to get right. All the different IoT hub clients and DPS clients pull from this IotHubSSLContext to handle turning certificates into SSLContext instances for them.

IotHubSSLContext.java and its unit test file have been significantly altered, so you may just want to review what it looks like now:

[IotHubSSLContext.java](https://github.com/Azure/azure-iot-sdk-java/blob/timtay/readCertsDeps/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java)

[IotHubSSLContextTest.java](https://github.com/Azure/azure-iot-sdk-java/blob/timtay/readCertsDeps/deps/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContextTest.java)


deps/pom.xml
```diff
@@ -47,16 +47,6 @@
             <artifactId>qpid-proton-j-extensions</artifactId>
             <version>1.2.3</version>
         </dependency>
-        <dependency>
-            <groupId>org.bouncycastle</groupId>
-            <artifactId>bcmail-jdk15on</artifactId>
-            <version>1.69</version>
-        </dependency>
-        <dependency>
-            <groupId>org.bouncycastle</groupId>
```
timtay-microsoft, line 56: We only had this bouncycastle dependency for parsing private keys, and we don't do that anymore

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubCertificateManager.java
```diff
@@ -1,192 +0,0 @@
-// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-package com.microsoft.azure.sdk.iot.deps.auth;
```
timtay-microsoft, line 4: The point of this class was to hold the IoT Hub public certificate in source code. Since we no longer want to do that, this class can be deleted

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
```diff
@@ -5,371 +5,156 @@
...
-        try
-        {
-            // Codes_SRS_IOTHUBSSLCONTEXT_34_031: [This function shall return a Private Key instance created by the provided PEM formatted privateKeyString.]
-            Security.addProvider(new BouncyCastleProvider());
-            PEMParser privateKeyParser = new PEMParser(new StringReader(privateKeyString));
-            Object possiblePrivateKey = privateKeyParser.readObject();
-            return IotHubSSLContext.getPrivateKey(possiblePrivateKey);
-        }
-        catch (Exception e)
+        try (FileInputStream fis = new FileInputStream(trustedCertificatesFilePath))
```
timtay-microsoft, line 125: In Java, the pattern 

```java
try (InputStream is = new InputStream())
{
    // use the stream
}
```

is similar to a "using" block in .NET, so the input stream is closed no matter what when this block ends.

jamdavi (approved):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/auth/IotHubSSLContext.java
```diff
@@ -5,371 +5,156 @@
...
+     * @param trustedCertificatesFilePath the absolute file path of the file that contains the PEM formatted
+     * certificates that this IotHubSSLContext will trust.
+     * @return the created IotHubSSLContext.
+     * @throws CertificateException if the provided trustedCertificates are not PEM formatted and cannot be parsed.
+     * @throws IOException if the provided trustedCertificates cannot be read as a stream.
+     * @throws KeyStoreException if a key store cannot be created.
+     * @throws NoSuchAlgorithmException if your device cannot use x509 certificates or TLS SSLContexts.
+     * @throws KeyManagementException if the created SSLContext cannot be initialized.
+     */
+    public static SSLContext getSSLContextFromFile(String trustedCertificatesFilePath) throws CertificateException, IOException, KeyStoreException, NoSuchAlgorithmException, KeyManagementException
```
timtay-microsoft, line 118: While we aren't parsing private keys anymore, we do have some scenarios where we need to parse public certs specifically around Edgehub, and soon around DPS

PR #1287, opened 2021-07-27
break(iot-dev): Read certs from device rather than source

timtay-microsoft:
samples will be a separate commit in order to keep this commit smaller


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -552,41 +547,6 @@ else if (value == null)
             throw new UnsupportedOperationException("Must re-register this client to a multiplexing client before using it");
         }
 
-        switch (optionName)
```
timtay-microsoft, line 555: Now that we've gotten rid of the transportClient paths, there is less logic needed in the DeviceClient class and we can just rely on the common implementation of this method in ModuleClient.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
```diff
@@ -193,7 +197,41 @@ public static ModuleClient createFromEnvironment(IotHubClientProtocol protocol,
         {
             log.debug("Creating module client with the provided connection string");
 
-            //Codes_SRS_MODULECLIENT_34_020: [If an edgehub or iothub connection string is present, this function shall create a module client instance using that connection string and the provided protocol.]
+            //Check for a different default cert to be used
```
timtay-microsoft, line 200: During the moduleClient's createFromEnvironment flow, there may be a PEM formatted public certificate that the SDK is expected to trust in order to connect to the Edgehub

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
```diff
@@ -193,7 +197,41 @@ public static ModuleClient createFromEnvironment(IotHubClientProtocol protocol,
...
+                    sslContext = IotHubSSLContext.getSSLContextFromFile(alternativeDefaultTrustedCert);
+                }
+                catch (CertificateException | IOException | KeyStoreException | NoSuchAlgorithmException | KeyManagementException e)
+                {
+                    throw new ModuleClientException("Failed to create an SSLContext instance from the provided trusted cert file path", e);
+                }
+            }
+            else
+            {
+                sslContext = new IotHubSSLContext().getSSLContext();
```
timtay-microsoft, line 217: And if that public certificate isn't present, then just create a default SSLContext instead

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
```diff
@@ -460,13 +460,6 @@ public void registerDeviceClients(Iterable<DeviceClient> deviceClients, long tim
                 devicesToRegisterMap.put(deviceClientToRegister.getConfig().getDeviceId(), deviceClientToRegister);
                 DeviceClientConfig configToAdd = deviceClientToRegister.getConfig();
 
-                // Overwrite the sslContext of the new client to match the multiplexing client
-                // settings if it was set. If it wasn't set, defer to the default client to set this ssl context
-                if (this.sslContext != null)
-                {
-                    configToAdd.getAuthenticationProvider().setSSLContext(this.sslContext);
```
timtay-microsoft, line 467: When multiplexing, we never check the SSLContext of each registered device, so there was never any reason for us to assign a device's SSLContext to match the MultiplexingClient's SSLContext here

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubAuthenticationProvider.java
```diff
@@ -25,100 +27,35 @@
     protected String moduleId;
 
     protected IotHubSSLContext iotHubSSLContext;
-    protected boolean sslContextNeedsUpdate;
-    protected String iotHubTrustedCert;
-    protected String pathToIotHubTrustedCert;
```
timtay-microsoft, line 30: SSLContext is now set at client construction time, so there is no case where the SSLContext needs updating, nor is there a means to set a new trusted later.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubSasTokenHardwareAuthenticationProvider.java
```diff
@@ -83,40 +81,6 @@ public boolean canRefreshToken()
         return true;
     }
 
-    /**
-     * Getter for SSLContext
-     * @throws IOException if an error occurs when generating the SSLContext
-     * @return The value of SSLContext
-     */
-    @Override
-    public SSLContext getSSLContext() throws IOException
```
timtay-microsoft, line 92: Can just use the implementation defined in the super class, so no need to override here

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubX509.java
```diff
@@ -1,105 +0,0 @@
-/*
```
timtay-microsoft, line 1: Unused class

device/iot-device-client/src/test/java/tests/unit/com/microsoft/azure/sdk/iot/device/auth/IotHubX509Test.java
```diff
@@ -1,184 +0,0 @@
-/*
```
timtay-microsoft, line 1: The class under test was deleted, so this unit test class is deleted to match

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
```diff
@@ -278,25 +306,25 @@ public static ModuleClient createFromEnvironment(IotHubClientProtocol protocol,
...
                     String trustCertificates = trustBundleProvider.getTrustBundleCerts(edgedUri, DEFAULT_API_VERSION);
-                    moduleClient.setTrustedCertificates(trustCertificates);
+                    sslContext = IotHubSSLContext.getSSLContextFromString(trustCertificates);
                 }
+                else
+                {
+                    sslContext = new IotHubSSLContext().getSSLContext();
+                }
+
+                IotHubAuthenticationProvider iotHubAuthenticationProvider = IotHubSasTokenHsmAuthenticationProvider.create(signatureProvider, deviceId, moduleId, hostname, gatewayHostname, generationId, DEFAULT_SAS_TOKEN_TIME_TO_LIVE_SECONDS, DEFAULT_SAS_TOKEN_BUFFER_PERCENTAGE, sslContext);
```
azabbasi, line 321: this seems it can use some line separation.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
azabbasi: Where is the SSLContext that is used for Multiplexing scenarios?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubAuthenticationProvider.java
```diff
@@ -25,100 +27,35 @@
...
 
     public IotHubAuthenticationProvider(String hostname, String gatewayHostname, String deviceId, String moduleId)
     {
-        if (hostname == null || hostname.isEmpty())
```
azabbasi, line 34: I really like how much this has shrunk!

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubAuthenticationProvider.java
azabbasi: Can you elaborate on why (if you know) we postponed creating the SSL context and had to have all this logic in place?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/hsm/IotHubSasTokenHsmAuthenticationProvider.java
```diff
@@ -98,4 +126,11 @@ private IotHubSasTokenHsmAuthenticationProvider(String hostname, String gatewayH
         this.signatureProvider = signatureProvider;
         this.generationId = generationId;
     }
+
+    private IotHubSasTokenHsmAuthenticationProvider(String hostname, String gatewayHostName, String deviceId, String moduleId, String generationId, String sharedAccessToken, SignatureProvider signatureProvider, int suggestedTimeToLiveSeconds, int timeBufferPercentage, SSLContext sslContext)
```
azabbasi, line 130: separate lines might make this easier to read :)

azabbasi (approved):
LGTM

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MultiplexingClient.java
timtay-microsoft: It is created at constructor time and saved as a local field in this class

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubAuthenticationProvider.java
timtay-microsoft: There are some Edgehub related scenarios where we needed to trust a certificate other than the base IoT hub cert, so we added these features to enable those scenarios. At this point though, it is reasonable to expect a user to know what the SSLContext looks like at client construction time

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubAuthenticationProvider.java
timtay-microsoft: Me too!

PR #1288, opened 2021-07-27
break(iot-dev): Remove Azure Storage SDK dependency, simplify file upload APIs

timtay-microsoft:


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -55,7 +49,6 @@
     private IoTHubConnectionType ioTHubConnectionType = IoTHubConnectionType.UNKNOWN;
 
     private FileUpload fileUpload;
-    private FileUploadTask fileUploadTask;
```
timtay-microsoft, line 58: This task was only used in the once deprecated and now removed file upload APIs

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/FileUpload.java
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) Microsoft. All rights reserved.
```
timtay-microsoft, line 1: Most of this is copied from the existing implementation to a new file that isn't in its own directory. With this move, I made the constructors and methods package private so that users don't see them.

iot-e2e-tests/common/pom.xml
```diff
@@ -97,6 +97,11 @@
             <version>1.18.8</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.azure</groupId>
+            <artifactId>azure-storage-blob</artifactId>
```
timtay-microsoft, line 102: We got this dependency from the IoT Hub device client before, but now that the device client doesn't depend on it anymore, the e2e tests need to declare this dependency

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -322,14 +302,14 @@ public void closeNow() throws IOException
      * @throws IOException If this HTTPS request fails to send.
      * @throws URISyntaxException If the returned sas uri cannot be constructed correctly
```
barustum, line 303: remove the extra throws.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -322,14 +302,14 @@ public void closeNow() throws IOException
...
      */
-    public FileUploadSasUriResponse getFileUploadSasUri(FileUploadSasUriRequest request) throws IOException, URISyntaxException
+    public FileUploadSasUriResponse getFileUploadSasUri(FileUploadSasUriRequest request) throws IOException
```
barustum, line 305: isn't removing the exception being thrown a breaking change?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/FileUpload.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @SuppressWarnings("UnusedReturnValue") // Public method
+    IotHubStatusCode sendNotification(FileUploadCompletionNotification fileUploadStatusParser) throws IOException
+    {
+        IotHubTransportMessage message = new IotHubTransportMessage(fileUploadStatusParser.toJson());
+        message.setIotHubMethod(IotHubMethod.POST);
+
+        ResponseMessage responseMessage = httpsTransportManager.sendFileUploadNotification(message);
+        httpsTransportManager.close();
```
abhipsaMisra, line 58: unrelated to this PR, but any reason why we close the http layer after sending the upload notification? 
what if a customer want to perform > retrieve uri for file 1 -> upload file 1 -> notify completion for file 1 -> retrieve uri for file 2 -> ..

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/FileUpload.java
```diff
@@ -0,0 +1,91 @@
...
+        {
+            throw new IOException(errorMessage, serviceException);
+        }
+
+        return responseMessagePayload;
+    }
+
+    void close()
+    {
+        this.httpsTransportManager.close();
```
abhipsaMisra, line 89: this `this` isn't really required

iot-e2e-tests/common/pom.xml
abhipsaMisra: this is the track 2 version, right?

abhipsaMisra (approved):

iot-e2e-tests/common/pom.xml
timtay-microsoft: Yes, but it may be a couple of minor versions behind the latest

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
timtay-microsoft: Yes, but this is one of the acceptable breaking changes we agreed on making with higher ups

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/FileUpload.java
timtay-microsoft: Both open and close are basically no-ops since the HttpsTransportManager doesn't persist anything between requests. That said, there is no reason to close here. I'll just open the HttpsTransportManager when FileUpload is constructed, and close it when FileUpload is closed.

barustum (approved):

Issue #1286, opened 2021-07-27
[Bug Report] AllocationPolicy field for enrollment group set to "null" when retrieving the group or mapping the creation/update response


dusancvetkov:
# Context

- **OS and version used:** Windows 10 
- **Java runtime used:** openjdk version "11.0.9.1" 2020-11-04
- **SDK version used:** provisioning-service-client 1.7.0 & 1.8.0 (both tested)

## Description of the issue

Java IoT SDK could not map the api response properly when creating/updating/retrieving an enrollment group.
The field name that could not be mapped is "allocationPolicy", if we provide the "CUSTOM" (custom allocation policy) value, it will be saved to the enrollment group in the azure dps service, but when trying to retrieve it the value is just null. 

## Code sample exhibiting the issue

```
EnrollementGroup enrollmentGroup = new EnrollmentGroup(....);
enrollmentGroup.setAllocationPolicy(AllocationPolicy.CUSTOM);

EnrollementGroup resultAfterCreation = provisioningServiceClient.createOrUpdateEnrollmentGroup(enrollmentGroup);
log.info(resultAfterCreation.getAllocationPolicy()); // NULL but allocation policy is properly set in DPS

// if we try to update the group with this "null" value, the default provisioning will be set 
EnrollementGroup resultAfterUpdate = provisioningServiceClient.createOrUpdateEnrollmentGroup(resultAfterCreation);
log.info(resultAfterUpdate.getAllocationPolicy()); // NULL again & now allocation policy in dps for this group is set to default

```

My assumption is that the problem occurs in class: 

**com.microsoft.azure.sdk.iot.provisioning.service.configs.EnrollmentGroup** line **126** when the json string response could not map the enum field **com.microsoft.azure.sdk.iot.provisioning.service.configs.AllocationPolicy** because it returns it with value "custom" and not with value "CUSTOM" or could not map the enum value at all. 

## Console log of the issue

N/A

timtay-microsoft:
Sounds like a bug, yeah. We'll get this fixed ASAP

timtay-microsoft:
#1289 will fix this issue for you.

dusancvetkov:
Thanks @timtay-microsoft for the fast response. When should we expect the release with this bugfix to be pushed?

timtay-microsoft:
I'll try to get a release out with this fix within the next few work days.

timtay-microsoft:
[...]

PR #1289, opened 2021-07-27
fix(prov-serv): Fix issues with deserializing enrollments

timtay-microsoft:
-Fixed issue where the allocation policy was not deserialized correctly because of casing issues (#1286)

-Fixed issue where the device capabilities of an enrollmentGroup were never deserialized

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningServiceClientTests.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningServiceClientTests.java
@@ -1,5 +1,6 @@
 package tests.integration.com.microsoft.azure.sdk.iot.provisioning;
 
+import com.microsoft.azure.sdk.iot.deps.twin.DeviceCapabilities;
 import com.microsoft.azure.sdk.iot.provisioning.service.ProvisioningServiceClient;
 import com.microsoft.azure.sdk.iot.provisioning.service.configs.*;
 import com.microsoft.azure.sdk.iot.provisioning.service.exceptions.ProvisioningServiceClientException;
@@ -13,11 +14,13 @@ import java.util.UUID;
 import static junit.framework.TestCase.assertNotNull;
 import static junit.framework.TestCase.assertTrue;
 import static org.junit.Assert.assertEquals;
+import static tests.integration.com.microsoft.azure.sdk.iot.provisioning.setup.ProvisioningCommon.CUSTOM_ALLOCATION_WEBHOOK_URL_VAR_NAME;
+import static tests.integration.com.microsoft.azure.sdk.iot.provisioning.setup.ProvisioningCommon.DPS_CONNECTION_STRING_ENV_VAR_NAME;
 
 public class ProvisioningServiceClientTests
 {
-    public static final String DPS_CONNECTION_STRING_ENV_VAR_NAME = "IOT_DPS_CONNECTION_STRING";
     public static String provisioningServiceConnectionString = "";
+    public static String customAllocationWebhookUrl = "";
 
     public ProvisioningServiceClient provisioningServiceClient = null;
 
@@ -28,6 +31,7 @@ public class ProvisioningServiceClientTests
     {
         provisioningServiceConnectionString = Tools.retrieveEnvironmentVariableValue(DPS_CONNECTION_STRING_ENV_VAR_NAME);
         provisioningServiceClient = ProvisioningServiceClient.createFromConnectionString(provisioningServiceConnectionString);
+        customAllocationWebhookUrl = Tools.retrieveEnvironmentVariableValue(CUSTOM_ALLOCATION_WEBHOOK_URL_VAR_NAME);
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningServiceClientTests.java
```diff
@@ -13,11 +14,13 @@
...
 import static junit.framework.TestCase.assertTrue;
 import static org.junit.Assert.assertEquals;
+import static tests.integration.com.microsoft.azure.sdk.iot.provisioning.setup.ProvisioningCommon.CUSTOM_ALLOCATION_WEBHOOK_URL_VAR_NAME;
+import static tests.integration.com.microsoft.azure.sdk.iot.provisioning.setup.ProvisioningCommon.DPS_CONNECTION_STRING_ENV_VAR_NAME;
 
 public class ProvisioningServiceClientTests
 {
-    public static final String DPS_CONNECTION_STRING_ENV_VAR_NAME = "IOT_DPS_CONNECTION_STRING";
     public static String provisioningServiceConnectionString = "";
+    public static String customAllocationWebhookUrl = "";
```
abhipsaMisra, line 23: why not simply declare it, without initializing it to empty string?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningServiceClientTests.java
```diff
@@ -86,4 +90,57 @@ public void enrollmentGroupGetAttestationMechanismSymmetricKey() throws Provisio
         assertTrue(retrievedSymmetricKeyAttestation.getPrimaryKey().length() > 0);
         assertTrue(retrievedSymmetricKeyAttestation.getSecondaryKey().length() > 0);
     }
+
+    @Test
+    public void createdEnrollmentGroupMatchesSentEnrollmentGroup() throws ProvisioningServiceClientException
+    {
+        String groupId = testPrefix + UUID.randomUUID();
+        Attestation attestation = new SymmetricKeyAttestation("", "");
```
abhipsaMisra, line 98: create an object will empty-string params?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningServiceClientTests.java
```diff
@@ -86,4 +90,57 @@ public void enrollmentGroupGetAttestationMechanismSymmetricKey() throws Provisio
...
+        EnrollmentGroup enrollment = new EnrollmentGroup(groupId, attestation);
+        enrollment.setAllocationPolicy(AllocationPolicy.CUSTOM);
+        CustomAllocationDefinition customAllocationDefinition = new CustomAllocationDefinition();
+        customAllocationDefinition.setApiVersion("2018-09-01-preview");
```
abhipsaMisra, line 102: is this something users need to be aware of or set?

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/AllocationPolicy.java
```diff
@@ -3,15 +3,24 @@
...
+import com.google.gson.annotations.SerializedName;
+
 /**
  * The Device Provisioning Service enrollment level allocation policies.
  *
  * @see <a href="https://docs.microsoft.com/en-us/azure/iot-dps/tutorial-provision-multiple-hubs#set-the-device-provisioning-service-allocation-policy">
  *     Provision devices across load-balanced IoT hubs</a>
  */
 public enum AllocationPolicy {
+    @SerializedName("hashed")
```
abhipsaMisra, line 15: a unit test for this?

abhipsaMisra (approved):
some questions on tests

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningServiceClientTests.java
timtay-microsoft: Sure thing

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningServiceClientTests.java
timtay-microsoft: When we pass in null or empty symmetric keys, the service will generate a pair for us. Since the key doesn't need to be a particular value for this test, we may as well leave it blank

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningServiceClientTests.java
timtay-microsoft: Yes, albeit I had a hard time finding the expected values here

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/AllocationPolicy.java
timtay-microsoft: Sure thing. I'll add a unit test for each of these possible values in both the individual and group enrollment unit tests

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningServiceClientTests.java
abhipsaMisra: Got it. Wouldn't it be better if we had a no-args constructor though? Requiring customers to pass in null/empty string in order to get service-generated symmetric keys seems unnecessary, right?

timtay-microsoft: We do have a no-args constructor, but it's implementation (for better or worse) generates some keys locally for the user. Since we don't need to generate any keys locally, I avoided that no-args constructor

PR #1290, opened 2021-07-29
refactor(samples): Make device samples demonstrate reading certificates from device

timtay-microsoft:


device/iot-device-samples/send-event-x509/pom.xml
```diff
@@ -35,29 +35,4 @@
             </plugin>
         </plugins>
     </build>
-    <dependencies>
-        <!--suppress MavenDuplicateDependenciesInspection -->
-        <!--
-        This duplicate dependency declaration is only here so that it can demonstrate how to exclude
-        bouncycastle dependency from the device client library.
```
timtay-microsoft, line 42: Since we no longer have the bouncycastle dependency, we no longer need to demonstrate how to remove it

device/iot-device-samples/send-receive-x509-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/SSLContextBuilder.java
```diff
@@ -1,276 +0,0 @@
-/*
```
timtay-microsoft, line 1: Since this file was just a copy from another sample's SSLContextBuilder, I'm removing it and adding a dependency to the other sample

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/SSLContextBuilder.java
```diff
@@ -23,28 +22,31 @@
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.UUID;
 
 public class SSLContextBuilder
```
timtay-microsoft, line 26: This is a copy of the SSLContextBuilder that lives in our samples that we use for our e2e tests. It isn't part of the samples that I'm editing, but it is the same code so it may as well be reviewed at the same time

device/iot-device-samples/send-event-x509/src/main/java/samples/com/microsoft/azure/sdk/iot/SSLContextBuilder.java
```diff
@@ -143,128 +152,4 @@ public static RSAPrivateKey parsePrivateKeyString(String privateKeyPEM) throws I
 
         return randomChars;
     }
-
-    private static TrustManagerFactory generateTrustManagerFactory(KeyStore trustKeyStore)
```
timtay-microsoft, line 147: This logic was to build the TrustManagers out of the Iot hub public certs from source, which we no longer want users to do.

device/iot-device-samples/send-batch-events/src/main/java/samples/com/microsoft/azure/sdk/iot/SendBatchEvents.java
```diff
@@ -152,15 +149,6 @@ else if (protocolStr.equals("mqtt_ws"))
                         protocolStr);
```
barustum, line 149: Remove line 151 from the "before" view

drwill-ms (approved):

PR #1291, opened 2021-07-30
feat(iot-dev): Add ability to configure the maximum number of messages sent per send thread

timtay-microsoft:
Also fix a bug where setting the send and/or receive period after opening the client didn't change the period.

For additional context here, the Java SDK has a periodic thread that checks for queued messages to send, and that thread currently only sends 10 messages at most, even if more messages are queued. This feature allows users to configure this value instead of leaving it hardcoded.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
@@ -796,6 +796,11 @@ public final class DeviceClient extends InternalClient implements Closeable
      *	      in case of MQTT and AMQP protocols, this option specifies the interval in milliseconds
      *	      between spawning a thread that dequeues a message from the SDK's queue of received messages.
      *
+     *	    - <b>SetMaxMessagesSentPerThread</b> - this option is applicable to all protocols.
+     *	      This option specifies how many messages a given send thread should attempt to send before exiting.
+     *	      This option can be used in conjunction with "SetSendInterval" to control the how frequently and in what
+     *	      batch size messages are sent.
+     *
      *	    - <b>SetCertificatePath</b> - this option is applicable only
      *	      when the transport configured with this client is AMQP. This
      *	      option specifies the path to the certificate used to verify peer.
@@ -861,6 +866,7 @@ public final class DeviceClient extends InternalClient implements Closeable
             case SET_HTTPS_READ_TIMEOUT:
             case SET_AMQP_OPEN_AUTHENTICATION_SESSION_TIMEOUT:
             case SET_AMQP_OPEN_DEVICE_SESSIONS_TIMEOUT:
+            case SET_MAX_MESSAGES_SENT_PER_THREAD:
             {
                 break;
             }
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
@@ -799,7 +799,8 @@ public final class DeviceClient extends InternalClient implements Closeable
      *	    - <b>SetMaxMessagesSentPerThread</b> - this option is applicable to all protocols.
      *	      This option specifies how many messages a given send thread should attempt to send before exiting.
      *	      This option can be used in conjunction with "SetSendInterval" to control the how frequently and in what
-     *	      batch size messages are sent.
+     *	      batch size messages are sent. By default, this client sends 10 messages per send thread, and spawns
+     *	      a send thread every 10 milliseconds. This gives a theoretical throughput of 1000 messages per second.
      *
      *	    - <b>SetCertificatePath</b> - this option is applicable only
      *	      when the transport configured with this client is AMQP. This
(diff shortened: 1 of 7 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -208,6 +202,11 @@ void setMultiplexingRetryPolicy(RetryPolicy retryPolicy)
         this.transport.setMultiplexingRetryPolicy(retryPolicy);
     }
 
+    void setMaxNumberOfMessagesSentPerSendThread(int maxNumberOfMessagesSentPerSendThread)
+    {
+        this.transport.setMaxNumberOfMessagesSentPerSendThread(maxNumberOfMessagesSentPerSendThread);
```
timtay-microsoft, line 207: transport is never null, so no null check needed here

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -365,26 +357,28 @@ public long getReceivePeriodInMilliseconds()
...
-            this.receiveTaskScheduler.scheduleAtFixedRate(this.receiveTask, 0,
-                    this.receivePeriodInMilliseconds, TimeUnit.MILLISECONDS);
+            // close the old scheduler and start a new one with the new receive period
+            this.receiveTaskScheduler.shutdown();
+            this.receiveTaskScheduler = Executors.newScheduledThreadPool(1);
+            this.receiveTaskScheduler.scheduleAtFixedRate(
+                this.receiveTask,
+                0,
+                this.receivePeriodInMilliseconds,
+                TimeUnit.MILLISECONDS);
```
timtay-microsoft, line 381: This change was reported to have fixed the issue mentioned in the description of this PR. Close and open a new executor service with the new period

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -1113,6 +1124,22 @@ void setOption_SetAmqpOpenDeviceSessionsTimeout(Object value)
         }
     }
 
+    void setOption_SetMaxMessagesSentPerThread(Object value)
+    {
+        if (value != null)
```
drwill-ms, line 1129: What if `value` is null?

drwill-ms (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
timtay-microsoft: I'll make it throw IllegalArgumentException, sure

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -796,6 +796,11 @@ void setConnectionType(IoTHubConnectionType connectionType)
      *	      in case of MQTT and AMQP protocols, this option specifies the interval in milliseconds
      *	      between spawning a thread that dequeues a message from the SDK's queue of received messages.
      *
+     *	    - <b>SetMaxMessagesSentPerThread</b> - this option is applicable to all protocols.
```
abhipsaMisra, line 799: is there a recommendation for these values that we can provide to the user; eg. the effects of setting a very large value or a very low value?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -365,26 +357,28 @@ public long getReceivePeriodInMilliseconds()
...
-            /* Codes_SRS_DEVICE_IO_21_029: [If the `receiveTask` is null, the setReceivePeriodInMilliseconds shall throw IOException.] */
             if (this.receiveTask == null)
             {
                 throw new IOException("transport receive task not set");
             }
 
-            this.receiveTaskScheduler.scheduleAtFixedRate(this.receiveTask, 0,
-                    this.receivePeriodInMilliseconds, TimeUnit.MILLISECONDS);
+            // close the old scheduler and start a new one with the new receive period
+            this.receiveTaskScheduler.shutdown();
```
abhipsaMisra, line 375: question - when was this old scheduler created? can you set the receive period after client creation or is it set on the transport settings passed in during client creation?

abhipsaMisra (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: These schedulers are created when the connection is first established, and they turn off every time the SDK goes into a reconnection state (since there is no point in trying to send/receive without a connection)

timtay-microsoft: It can be set at any time in the DeviceClient lifespan

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
timtay-microsoft: It'll depend too much on how the customer uses the SDK. I can try to add a bit more context here, though.

PR #1293, opened 2021-08-02
break(prov-dev, security, security-x509): Make SecurityProvider classes read certificates from certificate store instead of from source

timtay-microsoft:
Also make SecurityProviderX509 take the parsed key and certificate objects rather than strings.


iot-e2e-tests/common/pom.xml
```diff
@@ -73,6 +73,16 @@
...
         </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcmail-jdk15on</artifactId>
+            <version>1.69</version>
+        </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcprov-jdk15on</artifactId>
+            <version>1.69</version>
```
timtay-microsoft, line 84: We still use bouncycastle in our e2e tests for generating test certificates

provisioning/security/security-provider/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/SecurityProviderSymmetricKey.java
```diff
@@ -106,43 +105,6 @@ public String getRegistrationId() throws SecurityProviderException
...
 
-    /**
-     * Retrieves the SSL context loaded with trusted certs. In case of X509 SSL context shall be loaded with complete chain
-     * all the way till the leaf along with its private key.
-     *
-     * @return The SSLContext relevant to the flow
-     * @throws SecurityProviderException If ssl context could not be generated for any of the reason
-     */
-    @Override
-    public SSLContext getSSLContext() throws SecurityProviderException
```
timtay-microsoft, line 117: This implementation was shared with the SecurityProviderTpm class, so I moved it up to the more abstract SecurityProvider class while stripping out the reading of certs from source

provisioning/security/security-provider/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/SecurityProviderSymmetricKey.java
```diff
@@ -24,7 +24,6 @@
     private final byte[] primaryKey;
     private byte[] secondaryKey;
     private final String registrationId;
-    private static final String HMAC_SHA256 = "HmacSHA256";
```
timtay-microsoft, line 27: This was already defined on line 23, so I removed it

provisioning/security/x509-provider/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/hsm/SecurityProviderX509Cert.java
```diff
@@ -8,73 +8,44 @@
...
     private final X509Certificate leafCertificatePublic;
     private final Key leafPrivateKey;
     private final Collection<X509Certificate> signerCertificates;
 
-    private final String leafCertificatePublicPem;
-    private final Collection<String> signerCertificatesPem;
     /**
      * Constructor to build the DICE certs from the simulator
      */
-    public SecurityProviderX509Cert(String leafPublicPem, String leafPrivateKey, Collection<String> signerCertificates) throws SecurityProviderException
```
timtay-microsoft, line 46: This API needed to take the security primitives rather than strings because there is no way for us to parse these strings reliably without further details about what kinds of certificates/keys they are (RSA vs ECC, for instance)

provisioning/security/security-provider/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/SecurityProvider.java
```diff
@@ -131,34 +37,24 @@
...
+     * from your device's trusted root certification authorities certificate store. Implementations of {@link SecurityProviderX509}
+     * must override this function so that it returns an SSLContext instance with the required private key and public certificates
+     * loaded into it as well as the default trusted certificates saved in your device's trusted root certification authorities certificate store.
+     * @return The SSLContext instance.
+     * @throws SecurityProviderException If ssl context could not be generated for any of the reason.
      */
-    abstract public SSLContext getSSLContext() throws SecurityProviderException;
-
-    KeyStore getKeyStoreWithTrustedCerts() throws NoSuchAlgorithmException, IOException, CertificateException, KeyStoreException
+    public SSLContext getSSLContext() throws SecurityProviderException
```
drwill-ms, line 47: What are the casing rules for Java for acronyms? Should this be `getSslContext`?

provisioning/security/security-provider/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/SecurityProviderX509.java
```diff
@@ -19,17 +19,16 @@
 
 public abstract class SecurityProviderX509 extends SecurityProvider
 {
-    private static final String ALIAS_CERT_ALIAS = "ALIAS_CERT";
+    private static final String CERT_ALIAS = "ALIAS_CERT";
```
drwill-ms, line 22: cert alias = alias cert? :) Perhaps `ALIAS_CERT`?

drwill-ms (approved):

provisioning/security/security-provider/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/SecurityProvider.java
timtay-microsoft: Java is pretty mixed on this front. Since the type itself is SSLContext, it makes sense to me to keep this as getSSLContext rather than getSslContext

provisioning/security/security-provider/src/main/java/com/microsoft/azure/sdk/iot/provisioning/security/SecurityProviderX509.java
timtay-microsoft:
```suggestion
    private static final String CERT_ALIAS = "IOTHUB_SECURITY_PROVIDER_CERT_ALIAS";
```

timtay-microsoft: I'll make this a bit more descriptive with the above suggestion

PR #1295, opened 2021-08-03
break(all): Removed thrown exception declarations where the exception…

timtay-microsoft:
… isn't thrown

### DIFF
--- a/iot-e2e-tests/edge-e2e/src/main/java/glue/ModuleGlue.java
+++ b/iot-e2e-tests/edge-e2e/src/main/java/glue/ModuleGlue.java
@@ -161,7 +161,8 @@ public class ModuleGlue
             ConnectResponse cr = new ConnectResponse();
             cr.setConnectionId(connectionId);
             handler.handle(Future.succeededFuture(cr));
-        } catch (URISyntaxException | IOException e)
+        }
+        catch (URISyntaxException | IOException e)
         {
             handler.handle(Future.failedFuture(e));
         }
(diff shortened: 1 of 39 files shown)


iot-e2e-tests/edge-e2e/src/main/java/glue/ModuleGlue.java
```diff
@@ -161,7 +161,7 @@ public void connect(String transportType, String connectionString, Certificate c
             ConnectResponse cr = new ConnectResponse();
             cr.setConnectionId(connectionId);
             handler.handle(Future.succeededFuture(cr));
-        } catch (ModuleClientException | URISyntaxException | IOException e)
+        } catch (URISyntaxException | IOException e)
```
drwill-ms, line 164: If you're fixing brace styles, want to fix this one too?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FileUploadNotificationReceiver.java
```diff
@@ -211,14 +210,8 @@ public FileUploadNotification receive(long timeoutMs) throws IOException, Interr
     {
         final CompletableFuture<Void> future = new CompletableFuture<>();
         executor.submit(() -> {
-            try
-            {
-                close();
-                future.complete(null);
-            } catch (IOException e)
-            {
-                future.completeExceptionally(e);
```
drwill-ms, line 220: What is the impact of removing this?

timtay-microsoft: Nothing. Since the IOException was never thrown here, the catch block would never actually execute

iot-e2e-tests/edge-e2e/src/main/java/glue/ModuleGlue.java
timtay-microsoft: Sure, I can do that

drwill-ms (approved):

PR #1298, opened 2021-08-04
fix(iot-dev): Fix bug where multiplexed deviceClients did not maintain twin and/or method subscriptions after reconnection

timtay-microsoft:
If a multiplexed device is subscribed to twin and/or methods, then loses its session due to network issues, it should still be subscribed to twin and/or methods after it finishes reconnection


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -90,14 +90,17 @@
     // Proton-j primitives and wrappers for the device and authentication sessions
     private Connection connection;
     private Reactor reactor;
+    private final Queue<AmqpsSessionHandler> reconnectingDeviceSessionHandlers = new ConcurrentLinkedQueue<>();
```
timtay-microsoft, line 93: The key to this fix is to cache the session handlers for these reconnecting devices since they hold onto all of this twin/methods subscription information

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -172,14 +175,28 @@ public void registerMultiplexedDevice(DeviceClientConfig config)
...
     }
 
-    public void unregisterMultiplexedDevice(DeviceClientConfig config)
+    /**
+     * Asynchronously unregister a multiplexed device from an active multiplexed connection or synchronously unregister.
+     * a multiplexed device from a closed multiplexed connection.
+     * @param config the config of the device that should be unregistered.
+     * @param willReconnect true if the device will be re-registered soon because it is reconnecting.
+     */
+    public void unregisterMultiplexedDevice(DeviceClientConfig config, boolean willReconnect)
```
timtay-microsoft, line 184: Adding a flag to the unregister call that signals to this AMQP layer to cache the unregistered session objects since we will re-register that device session soon.

azabbasi (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -172,14 +175,28 @@ public void registerMultiplexedDevice(DeviceClientConfig config)
         deviceClientConfigs.add(config);
     }
 
-    public void unregisterMultiplexedDevice(DeviceClientConfig config)
+    /**
+     * Asynchronously unregister a multiplexed device from an active multiplexed connection or synchronously unregister.
```
abhipsaMisra, line 179: nit: extra `.` at eol
q - this being async and returning a void, how does a user block on this or be notified when the async operation completes?

abhipsaMisra: user or us if the class is only internally used

abhipsaMisra (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
timtay-microsoft: It's internal, but there is a callback defined elsewhere that allows for the AMQP layer to communicate to the transport agnostic layer when certain events happen (connectionLoss, connectionEstablished, and multiplexedDeviceRegistered/Unregistered)

timtay-microsoft:
```suggestion
     * Asynchronously unregister a multiplexed device from an active multiplexed connection or synchronously unregister
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -940,14 +957,25 @@ private AmqpsSessionHandler createSessionHandler(DeviceClientConfig deviceClient
             }
         }
 
+        // If the device session was temporarily unregistered during reconnection, reuse the cached session handler
+        // since it holds the subscription information needed to fully reconnect all links that were open prior to reconnection.
+        for (AmqpsSessionHandler cachedDeviceSessionHandler : this.reconnectingDeviceSessionHandlers)
+        {
+            if (cachedDeviceSessionHandler.getDeviceId().equals(deviceClientConfig.getDeviceId()))
+            {
+                amqpsSessionHandler = cachedDeviceSessionHandler;
```
jamdavi, line 966: the for loop above has a break statement. Should we add one here?

timtay-microsoft: Good idea

jamdavi (approved):

PR #1300, opened 2021-08-05
Add migration doc to help users upgrade from 1.X.X to 2.X.X releases

timtay-microsoft:
This captures all of the currently approved changes that have been made so far. We can add to it later if we get approval to make further breaking changes


jamdavi (approved):
Should we also call out the Java 7 depreciation?

timtay-microsoft:
> Should we also call out the Java 7 depreciation?

We already kind of deprecated it. See https://github.com/Azure/azure-iot-sdk-java/blob/master/supported_platforms.md#miscellaneous-support-notes

PR #1307, opened 2021-08-11
break(iot-dev): Rename device/module client APIs for consistency

timtay-microsoft:
Move all the common code into InternalClient as well. There are many changed sample/test files that are affected by these renames, but for the sake of simplifying this PR, those changes will be added later.

old:
```java
public void subscribeToDesiredProperties(...);
public void startTwin(...);
public void getTwin(...);
public void subscribeToDeviceMethod(...); //deviceClient
public void subscribeToMethod(...); //moduleClient
public void registerConnectionStatusChangeCallback(...);
```

new
```java
public void subscribeToDesiredPropertiesAsync(...);
public void startTwinAsync(...);
public void getTwinAsync(...);
public void subscribeToMethodsAsync(...); //moduleClient and deviceClient, "Methods" rather than "Method"
public void setConnectionStatusChangeCallback(...); //"set" instead of "register"
```

With no changes in the parameters or return types

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientType.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientType.java
@@ -1,12 +1,11 @@
-package com.microsoft.azure.sdk.iot.device.transport.amqps;
+package com.microsoft.azure.sdk.iot.device;
 
 /**
- * Type of the connection
+ * The type of the device client. Used to differentiate between device clients that own their own connection from device
+ * clients that are multiplexing.
  */
-public enum IoTHubConnectionType
+public enum ClientType
 {
-    UNKNOWN,
-
     /**
      * The connection type is a non-multiplexed, single device identity connection.
      */
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
@@ -6,15 +6,12 @@ package com.microsoft.azure.sdk.iot.device;
 import com.microsoft.azure.sdk.iot.deps.serializer.FileUploadCompletionNotification;
 import com.microsoft.azure.sdk.iot.deps.serializer.FileUploadSasUriRequest;
 import com.microsoft.azure.sdk.iot.deps.serializer.FileUploadSasUriResponse;
-import com.microsoft.azure.sdk.iot.device.DeviceTwin.*;
 import com.microsoft.azure.sdk.iot.device.transport.RetryPolicy;
-import com.microsoft.azure.sdk.iot.device.transport.amqps.IoTHubConnectionType;
 import com.microsoft.azure.sdk.iot.device.transport.https.HttpsTransportManager;
 import com.microsoft.azure.sdk.iot.provisioning.security.SecurityProvider;
 import lombok.extern.slf4j.Slf4j;
 
 import java.io.Closeable;
-import java.io.IOError;
 import java.io.IOException;
 import java.net.URISyntaxException;
 
@@ -36,17 +33,16 @@ import java.net.URISyntaxException;
  * </p>
  * The client supports HTTPS 1.1 and AMQPS 1.0 transports.
  */
-@SuppressWarnings("deprecation") // A lot of references to deprecated APIs, but all of them are still in use
 @Slf4j
 public final class DeviceClient extends InternalClient implements Closeable
 {
-    private static long SEND_PERIOD_MILLIS = 10L;
+    private static final long SEND_PERIOD_MILLIS = 10L;
 
-    private static long RECEIVE_PERIOD_MILLIS_AMQPS = 10L;
-    private static long RECEIVE_PERIOD_MILLIS_MQTT = 10L;
-    private static long RECEIVE_PERIOD_MILLIS_HTTPS = 25*60*1000; /*25 minutes*/
+    private static final long RECEIVE_PERIOD_MILLIS_AMQPS = 10L;
+    private static final long RECEIVE_PERIOD_MILLIS_MQTT = 10L;
+    private static final long RECEIVE_PERIOD_MILLIS_HTTPS = 25*60*1000; /*25 minutes*/
 
-    private IoTHubConnectionType ioTHubConnectionType = IoTHubConnectionType.UNKNOWN;
+    private ClientType clientType = ClientType.SINGLE_CLIENT;
 
     private FileUpload fileUpload;
 
@@ -71,7 +67,7 @@ public final class DeviceClient extends InternalClient implements Closeable
      */
     public DeviceClient(String connString, IotHubClientProtocol protocol) throws URISyntaxException, IllegalArgumentException
     {
-        this(connString, protocol, (ClientOptions) null);
+        this(connString, protocol, null);
     }
 
     /**
@@ -94,10 +90,7 @@ public final class DeviceClient extends InternalClient implements Closeable
     public DeviceClient(String connString, IotHubClientProtocol protocol, ClientOptions clientOptions) throws URISyntaxException, IllegalArgumentException
     {
         super(new IotHubConnectionString(connString), protocol, SEND_PERIOD_MILLIS, getReceivePeriod(protocol), clientOptions);
-
         commonConstructorVerifications();
-
-        commonConstructorSetup();
     }
 
     /**
@@ -128,7 +121,6 @@ public final class DeviceClient extends InternalClient implements Closeable
     {
         super(hostName, deviceId, null, sasTokenProvider, protocol, clientOptions, SEND_PERIOD_MILLIS, getReceivePeriod(protocol));
         commonConstructorVerifications();
-        commonConstructorSetup();
     }
 
     /**
@@ -196,12 +188,6 @@ public final class DeviceClient extends InternalClient implements Closeable
     private DeviceClient(String uri, String deviceId, SecurityProvider securityProvider, IotHubClientProtocol protocol, ClientOptions clientOptions) throws URISyntaxException, IOException
     {
         super(uri, deviceId, securityProvider, protocol, SEND_PERIOD_MILLIS, getReceivePeriod(protocol), clientOptions);
-        commonConstructorSetup();
-    }
-
-    private void commonConstructorSetup()
-    {
-        this.ioTHubConnectionType = IoTHubConnectionType.SINGLE_CLIENT;
     }
 
     private void commonConstructorVerifications() throws UnsupportedOperationException
@@ -235,7 +221,7 @@ public final class DeviceClient extends InternalClient implements Closeable
      */
     public void open(boolean withRetry) throws IOException
     {
-        if (this.ioTHubConnectionType == IoTHubConnectionType.USE_MULTIPLEXING_CLIENT)
+        if (this.clientType == ClientType.USE_MULTIPLEXING_CLIENT)
         {
             throw new UnsupportedOperationException(MULTIPLEXING_OPEN_ERROR_MESSAGE);
         }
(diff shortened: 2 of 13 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientType.java
```diff
@@ -1,12 +1,11 @@
-package com.microsoft.azure.sdk.iot.device.transport.amqps;
```
timtay-microsoft, line 1: Not sure why this ever lived in the amqp package considering it was only used in the device client level package "com.microsoft.azure.sdk.iot.device"

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -36,17 +33,16 @@
...
+    private static final long SEND_PERIOD_MILLIS = 10L;
 
-    private static long RECEIVE_PERIOD_MILLIS_AMQPS = 10L;
-    private static long RECEIVE_PERIOD_MILLIS_MQTT = 10L;
-    private static long RECEIVE_PERIOD_MILLIS_HTTPS = 25*60*1000; /*25 minutes*/
+    private static final long RECEIVE_PERIOD_MILLIS_AMQPS = 10L;
+    private static final long RECEIVE_PERIOD_MILLIS_MQTT = 10L;
+    private static final long RECEIVE_PERIOD_MILLIS_HTTPS = 25*60*1000; /*25 minutes*/
 
-    private IoTHubConnectionType ioTHubConnectionType = IoTHubConnectionType.UNKNOWN;
```
timtay-microsoft, line 49: Calling this a "IoTHubConnectionType" was a bit misleading since it really is only used to differentiate between multiplexed device clients a non-multiplexed device clients. It's an enum that is only used internally though so no user will care.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -196,12 +188,6 @@ public DeviceClient setMessageCallback(MessageCallback callback, Object context)
     private DeviceClient(String uri, String deviceId, SecurityProvider securityProvider, IotHubClientProtocol protocol, ClientOptions clientOptions) throws URISyntaxException, IOException
     {
         super(uri, deviceId, securityProvider, protocol, SEND_PERIOD_MILLIS, getReceivePeriod(protocol), clientOptions);
-        commonConstructorSetup();
-    }
-
-    private void commonConstructorSetup()
-    {
-        this.ioTHubConnectionType = IoTHubConnectionType.SINGLE_CLIENT;
```
timtay-microsoft, line 204: Making this the default for the field at class instantiation time saves us from setting it in this method that was called by every constructor

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -331,270 +315,10 @@ public void completeFileUpload(FileUploadCompletionNotification notification) th
         this.fileUpload.sendNotification(notification);
     }
 
-    /**
-     * Retrieves the twin's latest desired properties
-     * @throws IOException if the iothub cannot be reached
-     */
-    public void getDeviceTwin() throws IOException
```
timtay-microsoft, line 338: All of these twin and method APIs that are shared between DeviceClient and ModuleClient have been moved into their parent class InternalClient for simplicity. In this move, though, many methods were renamed by appending an "Async" suffix to them.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -331,270 +315,10 @@ public void completeFileUpload(FileUploadCompletionNotification notification) th
...
-    public void subscribeToDeviceMethod(DeviceMethodCallback deviceMethodCallback, Object deviceMethodCallbackContext,
-                                        IotHubEventCallback deviceMethodStatusCallback, Object deviceMethodStatusCallbackContext)
-            throws IOException, IllegalArgumentException
-    {
-        this.subscribeToMethodsInternal(deviceMethodCallback, deviceMethodCallbackContext, deviceMethodStatusCallback, deviceMethodStatusCallbackContext);
-    }
-
     // Used by multiplexing clients to signal to this client what kind of multiplexing client is using this device client
-    @SuppressWarnings("SameParameterValue") // The connection type is currently only set to "multiplexing client", but it can be set to the deprecated transport client as well.
-    void setConnectionType(IoTHubConnectionType connectionType)
```
timtay-microsoft, line 458: This method is replaced by ```markAsMultiplexed()``` since it was only ever called with IoTHubConnectionType.Multiplexed and never set back to SINGLE_CLIENT.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -331,270 +315,10 @@ public void completeFileUpload(FileUploadCompletionNotification notification) th
...
-     *      - <b>SetAmqpOpenDeviceSessionsTimeout</b> - this option is applicable for AMQP.
-     *         This option specifies the timeout in seconds to open the device sessions.
-     *         By default, this value is 60 seconds.
-     *         The value is expected to be of type {@code int}.
-     *
-     * @param optionName the option name to modify
-     * @param value an object of the appropriate type for the option's value
-     * @throws IllegalArgumentException if the provided optionName is null
-     */
-    public void setOption(String optionName, Object value) throws IllegalArgumentException
```
timtay-microsoft, line 520: DeviceClient no longer does anything different from ModuleClient in the setOption flow, so no need to override it from InternalClient anymore

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -45,7 +50,7 @@
         "Direct methods are only supported over MQTT, MQTT_WS, AMQPS, and AMQPS_WS";
 
     DeviceClientConfig config;
-    DeviceIO deviceIO;
+    private DeviceIO deviceIO;
```
timtay-microsoft, line 53: DeviceClient and ModuleClient now cannot access this field directly, which will help protect us from making any silly mistakes later when adding device/module APIs

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -76,52 +80,36 @@
         this.deviceIO = new DeviceIO(this.config, sendPeriodMillis, receivePeriodMillis);
     }
 
-    InternalClient(IotHubConnectionString iotHubConnectionString, IotHubClientProtocol protocol, SSLContext sslContext, long sendPeriodMillis, long receivePeriod)
```
timtay-microsoft, line 79: Unused

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -403,7 +381,7 @@ public void sendReportedProperties(Set<Property> reportedProperties, Integer ver
      * @param callbackContext a context to be passed to the callback. Can be {@code null}.
      * @throws IllegalArgumentException if provided callback is null
      */
-    public void registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback callback, Object callbackContext) throws IllegalArgumentException
```
timtay-microsoft, line 406: Possibly a more controversial change, but "set" is a more commonly used verb than "register" for a callback like this

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -534,16 +505,7 @@ else if (value == null)
                 case SET_MINIMUM_POLLING_INTERVAL:
                 case SET_RECEIVE_INTERVAL:
                 {
-                    if (this.deviceIO.isOpen())
-                    {
-                        throw new IllegalStateException("setOption " + optionName +
```
timtay-microsoft, line 539: We don't need to impose this restriction, so I'm removing it

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -862,206 +848,151 @@ void setDeviceIO(DeviceIO deviceIO)
...
+        else
+        {
+            throw new IllegalArgumentException("value is not int = " + value);
         }
     }
 
-    void setOption_SetHttpsReadTimeout(Object value)
+    private void setOption_SetHttpsReadTimeout(Object value)
     {
-        if (value != null)
```
timtay-microsoft, line 888: These values were already null checked in the method that calls this method, so I'm removing them

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
```diff
@@ -343,7 +330,7 @@ public static ModuleClient createFromEnvironment(IotHubClientProtocol protocol,
     }
 
     @SuppressWarnings("SameParameterValue") // The SEND_PERIOD is currently 10ms for all protocols, but can be made configurable in the future.
-    private ModuleClient(IotHubAuthenticationProvider iotHubAuthenticationProvider, IotHubClientProtocol protocol, long sendPeriodMillis, long receivePeriodMillis) throws IOException, TransportException
```
timtay-microsoft, line 346: unthrown exceptions

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -26,17 +31,17 @@
 public class InternalClient
```
jamdavi, line 31: Considering that we're starting to expose more public methods from this class do we want to change the visisbility so there is no confusion about the clients? ie, we only expose a DeviceClient, ServiceClient and ModuleClient?

jamdavi (approved):
Happy to see so much red on this one.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
barustum: What confuses me is that we are moving public APIs that customers call into "InternalClient". But the word Internal throws me off because now it's been accessed by external entities, such as customer code.

barustum: This is more of a comment than a suggestion to change any code :)

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
timtay-microsoft: I'll mark this class and other similar classes package private once this PR is in, and once I get #1308 from master as well

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
timtay-microsoft: While this is a bit confusing for us to read, the good news is that customers won't see this level of detail. They'll just see that both ModuleClient and DeviceClient have APIs like sendEventAsync()

timtay-microsoft: We could certainly rename this client to something like "CommonClient" or something to make it clear that the purpose of this client is not to be internal, but to house the shared APIs

PR #1308, opened 2021-08-12
Rename unit test packages to match the packages of the classes they test

timtay-microsoft:
This is necessary if we want to make a package private class in our shipped code still be unit testable

In the upcoming 2.0.0 version we will mark a lot of public classes as package private in order to hide them from users. This change isn't a breaking change though, so it can go directly into our master branch

https://stackoverflow.com/questions/18507311/private-class-unit-tests


abhipsaMisra:
> This is necessary if we want to make a package private class in our shipped code still be unit testable
> 
> In the upcoming 2.0.0 version we will mark a lot of public classes as package private in order to hide them from users. This change isn't a breaking change though, so it can go directly into our master branch
> 
> https://stackoverflow.com/questions/18507311/private-class-unit-tests

I wasn't aware of this approach, good to know!

abhipsaMisra (approved):

jamdavi (approved):
I'm going to assume all the tests pass :)

PR #1310, opened 2021-08-12
break(iot-serv, iot-dev): Remove unnecessary synchronization from public methods

timtay-microsoft:
The ```synchronized``` keyword on a method simply means that a given instance of the class that contains that method cannot invoke the method or any other synchronized method at the same time. It makes them behave like atomic operations.

A given device method client should be allowed to invoke a method on device "a" while already invoking a method on device "b". The service will protect users from trying to invoke multiple methods on the same device simultaneously.

These APIs do not share any local resources that we needed to synchronize on. They are just simple HTTP requests to the service that are constructed locally within the method call itself.

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
@@ -207,7 +207,7 @@ public class DeviceTwin
      * @throws IOException This exception is thrown if the IO operation failed.
      * @throws IotHubException This exception is thrown if the response verification failed.
      */
-    public synchronized void updateTwin(DeviceTwinDevice device) throws IotHubException, IOException
+    public void updateTwin(DeviceTwinDevice device) throws IotHubException, IOException
     {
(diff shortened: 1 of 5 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/devicetwin/DeviceTwin.java
```diff
@@ -246,34 +246,6 @@ public synchronized void updateTwin(DeviceTwinDevice device) throws IotHubExcept
...
-     * {@link com.microsoft.azure.sdk.iot.service.devicetwin.DeviceTwinDevice#setDesiredProperties(TwinCollection)}
-     * method.
-     *
-     * @param device The device with a valid id for which device twin is to be updated.
-     * @throws UnsupportedOperationException This exception is always thrown.
-     */
-    public void replaceDesiredProperties(DeviceTwinDevice device) throws UnsupportedOperationException
-    {
-        // Currently this is not supported by service
-        throw new UnsupportedOperationException();
```
timtay-microsoft, line 261: I missed this when removing deprecated APIs, but since we never implemented these methods, now is a good time to remove them

abhipsaMisra (approved):

PR #1311, opened 2021-08-12
fix(e2e): fix multiplexing tests not setting test identity's twinUpdated to true

timtay-microsoft:
These identities could have been recycled into other tests where the twin state is important, so need to mark them accordingly


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
```diff
@@ -836,6 +836,7 @@ public void testTwin() throws Exception
 
         for (int i = 0; i < DEVICE_MULTIPLEX_COUNT; i++)
         {
+            testInstance.testDevicesArrayIdentity.get(i).twinUpdated = true;
```
abhipsaMisra, line 839: the next couple of lines don't make it obvious (to me!) why this property needs to be set, a comment would useful here.

abhipsaMisra (approved):

PR #1312, opened 2021-08-13
Mark classes in iot-deps package package private where possible

timtay-microsoft:


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/ConfigurationStatus.java
```diff
@@ -8,20 +8,8 @@
...
     @SerializedName("targeted")
-    TARGETED(1),
+    TARGETED,
 
     @SerializedName("applied")
-    APPLIED(2);
-
-    private final int numVal;
-
-    ConfigurationStatus(int numVal)
```
timtay-microsoft, line 18: unused

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/http/HttpResponse.java
```diff
@@ -16,10 +16,10 @@
  */
 public class HttpResponse
 {
-    protected final int status;
-    protected final byte[] body;
-    protected final byte[] errorReason;
-    protected final Map<String, String> headerFields;
+    private final int status;
```
vinagesh, line 19: Curious if we never return this back to the user.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/ConfigurationStatus.java
```diff
@@ -8,20 +8,8 @@
 public enum ConfigurationStatus
 {
     @SerializedName("targeted")
-    TARGETED(1),
+    TARGETED,
```
vinagesh, line 11: Won't this become 0 & 1 now?

vinagesh (approved):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/transport/http/HttpResponse.java
timtay-microsoft: We do in some cases, but the field itself can be private since we have a public getter

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ParserUtility.java
```diff
@@ -456,7 +456,7 @@ else if (jsonElement.isJsonArray())
         }
     }
 
-    public static Object getJsonPrimitiveValue(JsonPrimitive jsonPrimitive)
+    private static Object getJsonPrimitiveValue(JsonPrimitive jsonPrimitive)
```
jamdavi, line 459: I'm so-so on changing these two methods. Since it is a utility it might make sense to keep these as public. We're not really hiding any need-to-be-changed implementations here.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
```diff
@@ -66,7 +66,7 @@ public FileUploadCompletionNotification(String correlationId, Boolean isSuccess,
      * Construct this notification with json
      * @param json the json to parse.
      */
-    public FileUploadCompletionNotification(String json)
+    private FileUploadCompletionNotification(String json)
```
jamdavi, line 69: This seems like the right thing to do. 

Would we ever want a customer to construct their own notification this way? Seems like we wouldn't.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketUpgrade.java
```diff
@@ -43,7 +42,7 @@ public WebSocketUpgrade(String hostName, String webSocketPath, int webSocketPort
      *
      * @param host The host header field value.
      */
-    public final void setHost(String host)
+    private void setHost(String host)
```
jamdavi, line 45: If we're going through the trouble of making them private. I'm more in favor to make all of these simple mutators Lombok'd with a private access level.

jamdavi (approved):
Looks fine over all. My biggest suggestion would be when we can remove simple accessors and mutators we should.

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/FileUploadCompletionNotification.java
timtay-microsoft: No, this is a read-only field that we return. Users never create one to provide to our SDK

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/ws/impl/WebSocketUpgrade.java
timtay-microsoft: Good point. Now is as good a time as any

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ParserUtility.java
timtay-microsoft: This utility class itself is buried far enough away from our public APIs that I'd be surprised if anyone was using it. We can always make it public again later if someone asks for it

PR #1313, opened 2021-08-17
feat(dps-dev): Add missing registration result fields

timtay-microsoft:
The service has sent these fields for a while, but we never exposed them to users for some reason. This brings our registration result class up to par with what the service sends to the SDK

### DIFF
--- a/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/ProvisioningDeviceClientRegistrationResult.java
+++ b/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/ProvisioningDeviceClientRegistrationResult.java
@@ -7,12 +7,33 @@
 
 package com.microsoft.azure.sdk.iot.provisioning.device;
 
+import lombok.Getter;
+
 public class ProvisioningDeviceClientRegistrationResult
 {
     protected String iothubUri;
     protected String deviceId;
     protected ProvisioningDeviceClientStatus provisioningDeviceClientStatus;
     protected String payload;
+
+    @Getter
+    protected String registrationId;
+
+    @Getter
+    protected String createdDateTimeUtc;
+
+    @Getter
+    protected String status;
+
+    @Getter
+    protected ProvisioningDeviceClientSubstatus substatus;
+
+    @Getter
+    protected String eTag;
+
+    @Getter
+    protected String lastUpdatesDateTimeUtc;
+
     /**
      * Empty constructor to let users gather the data.
      */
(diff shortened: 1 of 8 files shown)


provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/ProvisioningDeviceClientRegistrationResult.java
```diff
@@ -7,12 +7,33 @@
 
 package com.microsoft.azure.sdk.iot.provisioning.device;
 
+import lombok.Getter;
+
 public class ProvisioningDeviceClientRegistrationResult
```
timtay-microsoft, line 12: This is the public facing class that our provisioning device client returns to a user after a successful device registration. Previously, it only had the essential fields exposed, but now it exposes all the fields that the service gives us.

provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/ProvisioningDeviceClientSubstatus.java
```diff
@@ -0,0 +1,57 @@
...
+
+package com.microsoft.azure.sdk.iot.provisioning.device;
+
+/**
+ * The possible substatus values of a successful device provisioning.
+ */
+public enum ProvisioningDeviceClientSubstatus
+{
+    /**
+     * Device has been assigned to an IoT hub for the first time
```
timtay-microsoft, line 12: These javadocs are copied directly from the service swagger

vinagesh (approved):

provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/ProvisioningDeviceClientSubstatus.java
jamdavi, line 12:
```suggestion
     * Device has been assigned to an IoT hub for the first time.
```

provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/RegistrationResult.java
```diff
@@ -27,4 +28,34 @@
...
+
+    void setETag(String eTag)
+    {
+        this.eTag = eTag;
+    }
+
+    void setLastUpdatesDateTimeUtc(String lastUpdatesDateTimeUtc)
+    {
+        this.lastUpdatesDateTimeUtc = lastUpdatesDateTimeUtc;
+    }
```
jamdavi, line 60: Any reason not to use Lombok? Might save us some code.

jamdavi (approved):

provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/RegistrationResult.java
timtay-microsoft: The reason I'm not using lombok here is because the fields themselves are defined in the super class. I'd add some protected or package private lombok setters in the super class instead, but the setters are only used within this class's package, and it wouldn't make sense to publicly expose setters for this read-only class

PR #1316, opened 2021-08-19
Reduce access level in classes and methods in IoT hub service client

timtay-microsoft:
And replace all the basic handwritten getters/setters with lombok annotations for simplicity


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ExportImportDevice.java
```diff
@@ -258,84 +147,6 @@ public int hashCode()
...
     }
 
-    /**
-     * Retrieves information from the provided parser and returns it in a new ExportImportDevice instance. All information on this shall be overwritten.
-     * @param parser the parser to read from
-     * @throws IllegalArgumentException if the provided parser is missing the authentication or id fields. It also shall
-     * be thrown if the authentication object in the parser uses SAS authentication and is missing one of the symmetric key fields, or if it uses SelfSigned authentication
-     * and is missing one of the thumbprint fields.
-     */
-    ExportImportDevice(ExportImportDeviceParser parser) throws IllegalArgumentException
```
timtay-microsoft, line 268: unused constructor

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceStatistics.java
```diff
@@ -1,25 +0,0 @@
...
- * Licensed under the MIT license. See LICENSE file in the project root for full license information.
- */
-
-package com.microsoft.azure.sdk.iot.service;
-
-/**
- * Data structure for service statistics received in Json array
- * Provide getters and setters for statistics properties
- */
-public class ServiceStatistics
```
timtay-microsoft, line 12: Unused class

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/BaseDevice.java
```diff
@@ -60,7 +61,7 @@ protected BaseDevice(String deviceId, SymmetricKey symmetricKey)
      * @param deviceId Name of the device (used as device id)
      * @param authenticationType - The type of authentication used by this device.
      */
-    protected BaseDevice(String deviceId, AuthenticationType authenticationType)
+    BaseDevice(String deviceId, AuthenticationType authenticationType)
```
vinagesh, line 64: Does this need the private keyword like other places?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
```diff
@@ -39,9 +39,9 @@
      * @param iotHubServiceClientProtocol protocol to be used
      *
      */
-    public FeedbackReceiver(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol)
+    FeedbackReceiver(String hostName, String userName, String sasToken, IotHubServiceClientProtocol iotHubServiceClientProtocol)
```
vinagesh, line 42: Add the private keyword?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceAuthenticationWithSharedAccessPolicyKey.java
```diff
@@ -11,15 +11,15 @@
...
+class ServiceAuthenticationWithSharedAccessPolicyKey extends AuthenticationMethod
 {
     /**
      * Constructor to create instance from policy name and policy key
      *
      * @param policyName The policy name string
      * @param key The policy key string
      */
-    public ServiceAuthenticationWithSharedAccessPolicyKey(String policyName, String key)
+    ServiceAuthenticationWithSharedAccessPolicyKey(String policyName, String key)
```
vinagesh, line 22: private?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceAuthenticationWithSharedAccessPolicyToken.java
```diff
@@ -47,46 +47,9 @@ public IotHubConnectionString populate(IotHubConnectionString iotHubConnectionSt
      * @param policyName The policy name string
      * @param token The token string
      */
-    public ServiceAuthenticationWithSharedAccessPolicyToken(String policyName, String token)
+    ServiceAuthenticationWithSharedAccessPolicyToken(String policyName, String token)
```
vinagesh, line 50: private?

vinagesh (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceAuthenticationWithSharedAccessPolicyToken.java
timtay-microsoft: This constructor is called from another class within this package, so I can't mark it as private. This lack of modifier makes the constructor "Package private" which means only classes within this package can access it. Classes outside of the package, even if they extend this class, can't access it

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceAuthenticationWithSharedAccessPolicyKey.java
timtay-microsoft: This constructor is called from another class within this package, so I can't mark it as private. This lack of modifier makes the constructor "Package private" which means only classes within this package can access it. Classes outside of the package, even if they extend this class, can't access it

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/FeedbackReceiver.java
timtay-microsoft: This constructor is called from another class within this package, so I can't mark it as private. This lack of modifier makes the constructor "Package private" which means only classes within this package can access it. Classes outside of the package, even if they extend this class, can't access it

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/BaseDevice.java
timtay-microsoft: This constructor is called from another class within this package, so I can't mark it as private. This lack of modifier makes the constructor "Package private" which means only classes within this package can access it. Classes outside of the package, even if they extend this class, can't access it

PR #1317, opened 2021-08-20
break(iot-serv): Remove unused service error codes, add descriptions

timtay-microsoft:
This matches the work we did on the .NET SDK, but actually removing the unused error codes

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
@@ -6,52 +6,153 @@ package com.microsoft.azure.sdk.iot.deps.serializer;
 public enum ErrorCodeDescription
 {
     // These descriptions belong to fully qualified status codes, such as 404001
+    /**
+     * This library could not classify the received error code into a known description.
+     */
     UnclassifiedErrorCode,
+
+    /**
+     * The API version used by the SDK is not supported by the IoT hub endpoint used in this connection.
+     * Usually this would mean that the region of the hub doesn't yet support the API version. One should
+     * consider downgrading to a previous version of the SDK that uses an older API version, or use a hub
+     * in a region that supports it.
+     */
     InvalidProtocolVersion,
-    DeviceInvalidResultCount,
+
+    /**
+     * The client has requested an operation that the hub recognizes as invalid. Check the error message
+     * for more information about what is invalid.
+     */
     InvalidOperation,
+
+    /**
+     * Something in the request payload is invalid. Check the error message for more information about what is invalid.
+     */
     ArgumentInvalid,
+
+    /**
+     * Something in the payload is unexpectedly null. Check the error message for more information about what is invalid.
+     */
     ArgumentNull,
+
+    /**
+     * Returned by the service if a JSON object provided by this library cannot be parsed, for instance, if the
+     * JSON provided for updating a twin is malformed.
+     */
     IotHubFormatError,
-    DeviceStorageEntitySerializationError,
-    BlobContainerValidationError,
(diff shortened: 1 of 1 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
```diff
@@ -6,73 +6,181 @@
...
+     */
     NotFound,
+
+    /**
+     * The general error for service requests that failed because the resource already exists.
+     */
     Conflict,
 
     /**
      * Represents status code for 429 and for status code 429001
```
abhipsaMisra, line 159: we can also update this to call out that this happens during etag mismatch

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
```diff
@@ -6,73 +6,181 @@
...
 
+    /**
+     * The general error for service requests that failed because a service request contained a resource that was larger
+     * than the service allows.
+     */
     RequestEntityTooLarge,
 
     /**
      * Represents status code for 429 and 429001
      */
```
abhipsaMisra, line 171:
```suggestion
     /**
     * IoT hub throttling limits have been exceeded for the requested operation.
     * @see <a href="https://aka.ms/iothubthrottling">More details on throttling limits.</a>
     */
```

abhipsaMisra (approved):

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
```diff
@@ -94,79 +200,42 @@ public static ErrorCodeDescription Parse(int errorCode)
...
-            case(400012):
-                return DeserializationError;
             case(400013):
                 return BulkRegistryOperationFailure;
-            case (400027):
-                return ConnectionForcefullyClosedOnNewConnection;
-            case(400301):
-                return CannotRegisterModuleToModule;
-            case(401001):
-                return IotHubNotFound;
```
abhipsaMisra, line 116: unused?

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
```diff
@@ -94,79 +200,42 @@ public static ErrorCodeDescription Parse(int errorCode)
...
             case(401002):
                 return IotHubUnauthorizedAccess;
-            case(401003):
-                return IotHubUnauthorized;
```
abhipsaMisra, line 120: does service return 401002 for all unauth cases?

timtay-microsoft: Yes, that was our finding from the .NET side of this work

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
timtay-microsoft: Yep, no need to categorize these error codes to descriptions since they aren't returned by the service

deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ErrorCodeDescription.java
timtay-microsoft: I'll also fix that this is supposed to be for 412, not 429. I'm not sure why this comment says 429 here

PR #1318, opened 2021-08-20
Reduce access levels on classes and fields in DPS device client

timtay-microsoft:
Also convert handwritten getters and setters to lombok


provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/RequestData.java
```diff
@@ -7,20 +7,43 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
 import javax.net.ssl.SSLContext;
 
 @SuppressWarnings("SameParameterValue") // Some constructors currently are passed null params, but they are designed to be generic.
 public class RequestData
 {
+    @Getter
+    @Setter(AccessLevel.MODULE)
```
abhipsaMisra, line 20: same question that I had in the other PR, module vs package access level

abhipsaMisra (approved):

provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/RequestData.java
timtay-microsoft: I'll mark these all as Package, sure

PR #1319, opened 2021-08-20
Reduce access levels on classes and fields in DPS service client

timtay-microsoft:
Also convert handwritten getters and setters to lombok


provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/EnrollmentGroupManager.java
```diff
@@ -25,7 +25,7 @@
  * @see <a href="https://docs.microsoft.com/en-us/azure/iot-dps/">Azure IoT Hub Device Provisioning Service</a>
  * @see <a href="https://docs.microsoft.com/en-us/rest/api/iot-dps/deviceenrollmentgroup">Device Enrollment Group</a>
  */
-public class EnrollmentGroupManager
+class EnrollmentGroupManager
```
timtay-microsoft, line 28: This class and other "Manager" class are just 1 level below our public client, but they shouldn't directly be used

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/Tools.java
```diff
@@ -248,23 +248,4 @@ public static void appendJsonAttribute(StringBuilder strBuilder, String name, St
...
     }
-
-    /**
-     * Equality check for objects that accounts for null value comparisons. If both objects are null, this will return false.
-     * Both objects must have .equals(...) implemented correctly for this method to work properly.
-     * @param a the first object
-     * @param b the seconds object
-     * @return if the two are equal
-     */
-    public static boolean areEqual(Object a, Object b)
```
timtay-microsoft, line 259: unused

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/DeviceRegistrationState.java
```diff
@@ -41,59 +42,68 @@
...
+    @Getter
+    private transient Date createdDateTimeUtc;
 
     // last update date and time
     private static final String LAST_UPDATED_DATETIME_UTC_TAG = "lastUpdatedDateTimeUtc";
     @Expose
     @SerializedName(LAST_UPDATED_DATETIME_UTC_TAG)
     @SuppressWarnings("unused") // used by reflection during json serialization/deserialization
-    private String lastUpdatedDateTimeUtc;
-    private transient Date lastUpdatedDateTimeUtcDate;
```
timtay-microsoft, line 66: Swapping the names of these fields a bit to maintain that deviceRegistrationState.getLastUpdatedDateTimeUtc() still returns a date rather than a string

timtay-microsoft: You'll see this same pattern in a few date time fields across a few serialization types

provisioning/provisioning-service-client/src/test/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/TpmAttestationTest.java
```diff
@@ -18,28 +18,6 @@
     private static final String VALID_ENDORSEMENT_KEY = "validEndorsementKey";
     private static final String VALID_STORAGE_ROOT_KEY = "validStorageRootKey";
 
-    /* SRS_TPM_ATTESTATION_21_001: [The constructor shall throw IllegalArgumentException if the provided endorsementKey is null or empty.] */
```
timtay-microsoft, line 21: I removed some unnecessary null/empty checks, so a lot of unit tests needed to be deleted

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/Query.java
```diff
@@ -101,8 +100,7 @@
      * @param pageSize the {@code int} with the maximum number of items per iteration. It cannot be negative.
      * @throws IllegalArgumentException if one of the parameters is invalid.
      */
-    @SuppressWarnings("deprecation")
-    protected Query(ContractApiHttp contractApiHttp, String targetPath, QuerySpecification querySpecification, int pageSize)
+    Query(ContractApiHttp contractApiHttp, String targetPath, QuerySpecification querySpecification, int pageSize)
```
abhipsaMisra, line 103: what about the "deprecation"?

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/auth/ProvisioningConnectionString.java
```diff
@@ -3,32 +3,53 @@
...
 
     // Included in the device connection string
-    protected String hostName;
-    protected String deviceProvisioningServiceName;
-    protected AuthenticationMethod authenticationMethod;
-    protected String sharedAccessKeyName;
-    protected String sharedAccessKey;
-    protected String sharedAccessSignature;
+    @Getter
+    @Setter(AccessLevel.MODULE)
```
abhipsaMisra, line 31: q - is there a reason you chose module instead of package?
looks like they mean the same thing: https://github.com/projectlombok/lombok/blob/e050dafc9016519e79184e383eab9ffbd579bebd/src/core/lombok/javac/handlers/JavacHandlerUtil.java#L845-L847

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/DeviceRegistrationState.java
abhipsaMisra: this sure was confusing to get the flow -> getLastUpdatedDateTimeUtc()  returns Date lastUpdatedDateTimeUtcDate which it constructed out of a string lastUpdatedDateTimeUtc

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/EnrollmentGroup.java
```diff
@@ -94,10 +96,22 @@
...
     private static final String ENROLLMENT_GROUP_ID_TAG = "enrollmentGroupId";
+
+    /**
+     * The unique identifier for this enrollment group.
+     *
+     * <p>
+     * A valid enrollment group Id shall follow this criteria.
+     * A case-sensitive string (up to 128 char long)
+     * of ASCII 7-bit alphanumeric chars
+     * + {'-', ':', '.', '+', '%', '_', '#', '*', '?', '!', '(', ')', ',', '=', '@', ';', '$', '''}.
```
abhipsaMisra, line 108: how about we link to the ms doc instead, so that we don't have to update this if service changes their criteria? we don't do any client side validations, right?

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/X509CertificateInfo.java
```diff
@@ -42,50 +44,66 @@
...
+    @Getter
+    @Setter
     private String sha1Thumbprint;
 
     // the sha256 thumbprint of the X509 certificate
     private static final String SHA256_THUMBPRINT_TAG = "sha256Thumbprint";
     @Expose
     @SerializedName(SHA256_THUMBPRINT_TAG)
+    @Getter
+    @Setter
```
abhipsaMisra, line 64: a couple of these need the @NotNull tag

provisioning/provisioning-service-client/src/test/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/IndividualEnrollmentTest.java
```diff
@@ -228,12 +228,12 @@ public void constructorWithJsonSetsOptionalParametersUsesSetters()
 
         // assert
         assertEquals(VALID_DEVICE_ID, Deencapsulation.getField(enrollment, "deviceId"));
-        assertNotNull(Deencapsulation.getField(enrollment, "registrationState"));
+        assertNotNull(enrollment.getDeviceRegistrationState());
```
abhipsaMisra, line 231: why was this change required?

provisioning/provisioning-service-client/src/test/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/TpmAttestationTest.java
abhipsaMisra: could you add some context to why these null/empty checks were removed? is it because those values would only be set by service and hence would never be null/empty?

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/DeviceRegistrationState.java
timtay-microsoft: Yeah, the real unfortunate part of Lombok is that the getter's name is just get<FieldName> with no room for customization

provisioning/provisioning-service-client/src/test/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/TpmAttestationTest.java
timtay-microsoft: It was a mix of fields where the service would validate them for us and readonly fields that the service wouldn't return null values for

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/Query.java
timtay-microsoft: This was a warning that I missed removing back when I was deleting deprecated methods. This method called into a deprecated method at one point, but no longer does, so we can safely remove this suppression

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/auth/ProvisioningConnectionString.java
timtay-microsoft: No reason in particular, but I can unify them to "PACKAGE" if that would make more sense

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/EnrollmentGroup.java
timtay-microsoft: I like that idea, but this PR is really just moving existing documentation around, so I'll save editing it for later

provisioning/provisioning-service-client/src/test/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/EnrollmentGroupTest.java
```diff
@@ -1170,19 +1170,6 @@ public void setIotHubHostNameSucceed()
         assertEquals(newHostName, Deencapsulation.getField(enrollmentGroup, "iotHubHostName"));
     }
 
-    /* Tests_SRS_ENROLLMENT_GROUP_21_024: [The setInitialTwin shall throw IllegalArgumentException if the initialTwin is null.] */
-    @Test (expected = IllegalArgumentException.class)
-    public void setInitialTwinThrowsOnNull()
```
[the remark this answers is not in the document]
timtay-microsoft: The service does this validation for us, yeah

provisioning/provisioning-service-client/src/test/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/IndividualEnrollmentTest.java
timtay-microsoft: It wasn't required, but the method could be accessed directly rather than through reflection since the field is package private, not private. If we ever refactor the getDeviceRegistrationState() method, this unit test would get updated for free

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/X509CertificateInfo.java
timtay-microsoft: The service will validate this info for us

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/auth/ProvisioningConnectionString.java
abhipsaMisra: I think it'll be good to use one throughout, esp since they fall into the same case. 
+1 to unifying them to "package", since it's closer to java's package package-private access level

(On a related note, I'd have thought Lombok AccessLevel.None would have been java-package-private (no modifier). Interesting that it is "private" instead).

abhipsaMisra (approved):

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/BulkEnrollmentOperationError.java
```diff
@@ -37,53 +38,23 @@
     private static final String REGISTRATION_ID_TAG = "registrationId";
     @Expose
     @SerializedName(REGISTRATION_ID_TAG)
+    @Getter
```
jamdavi, line 41:
```suggestion
    /**
     * Getter for the error registrationId.
     *
     * @return The {@code String} with the registrationId content. It cannot be {@code null}.
     */
    @Getter
```

Just to note that with the @Getter methods you can also include full doc comments that will get generated when the lombok plugin executes for Maven. See

https://projectlombok.org/features/GetterSetter

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/BulkEnrollmentOperationError.java
```diff
@@ -37,53 +38,23 @@
...
     private String registrationId;
 
     // the error code
     private static final String ERROR_CODE_TAG = "errorCode";
     @Expose
     @SerializedName(ERROR_CODE_TAG)
+    @Getter
```
jamdavi, line 48:
```suggestion
    /**
     * Getter for the error code.
     *
     * @return The {@code Integer} with the errorCode content. It cannot be {@code null}.
     */
     @Getter
```

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/configs/BulkEnrollmentOperationError.java
```diff
@@ -37,53 +38,23 @@
...
     private Integer errorCode;
 
     // the error status
     private static final String ERROR_STATUS_TAG = "errorStatus";
     @Expose
     @SerializedName(ERROR_STATUS_TAG)
+    @Getter
```
jamdavi, line 55:
```suggestion
    /**
     * Getter for the error status.
     *
     * @return The {@code String} with the errorStatus content. It can be {@code null}.
     */
     @Getter
```

jamdavi (approved):
Just a note about the @Getter/Setter code comments. 

You don't need to add them for ALL of the instances, but if we have additional context that would be lost it'd be good to get it in a code comment above the annotation. 

Lombok will generate correct code docs for us.

provisioning/provisioning-service-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/service/auth/ProvisioningConnectionString.java
timtay-microsoft: Sure thing, I'll mark all of these as PACKAGE for consistency

PR #1320, opened 2021-08-23
break(iot-dev): Rename namespace to remove capital letters

timtay-microsoft:
Java namespaces should never have capital letters. Also, the "device twin" feature is is used by modules, so reducing the package from "DeviceTwin" to just "twin"


vinagesh (approved):

abhipsaMisra (approved):
Sindhu beat me to it!

PR #1321, opened 2021-08-23
break(all): Fix spelling of "Callback" in various classes and methods

timtay-microsoft:
Callback is one word, so no need to uppercase the "B"


abhipsaMisra (approved):
:open_mouth:

abhipsaMisra:
Do we have a note in our official communication to customers reassuring them that seeing 100s of red squiggly lines for such renaming changes shouldn't (a) scare them, (b) put them off from updating to next major version

timtay-microsoft:
> Do we have a note in our official communication to customers reassuring them that seeing 100s of red squiggly lines for such renaming changes shouldn't (a) scare them, (b) put them off from updating to next major version

Yep, right [here](https://github.com/Azure/azure-iot-sdk-java/blob/feature/majorVersionRev/major%20version%20upgrade%20migration%20guide.md).

I'll try to have an entry in here for each rename so that users have a clear direction on how to resolve each red squiggly

PR #1335, opened 2021-09-02
fix(iot-dev): Fix inefficient way of building strings in https batch messages

timtay-microsoft:
StringBuilder is much more efficient than just concatenating strings directly

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsBatchMessage.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsBatchMessage.java
@@ -53,18 +53,17 @@ public final class HttpsBatchMessage implements HttpsMessage
         StringBuilder batchBodyBuilder = new StringBuilder();
         batchBodyBuilder.append('[');
 
-        boolean atLeastOneMessage = false;
+        boolean isSubsequentMessage = false;
         for (HttpsSingleMessage message : messageList)
         {
-            addJsonToStringBuilder(message, batchBodyBuilder);
-            batchBodyBuilder.append(','); // comma to separate each object in the json array
-            this.numMsgs++;
-            atLeastOneMessage = true;
-        }
+            if (isSubsequentMessage)
+            {
+                batchBodyBuilder.append(','); // comma to separate each object in the json array
+            }
 
-        if (atLeastOneMessage)
-        {
-            batchBodyBuilder.deleteCharAt(batchBodyBuilder.length() - 1); // remove the final comma from the list
+            addJsonToStringBuilder(message, batchBodyBuilder);
+            this.numMsgs++;
+            isSubsequentMessage = true; // the next message, and all subsequent messages, will need a comma before them
         }
 
         batchBodyBuilder.append(']');
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsBatchMessage.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsBatchMessage.java
@@ -3,7 +3,6 @@
 
 package com.microsoft.azure.sdk.iot.device.transport.https;
 
-import com.microsoft.azure.sdk.iot.device.Message;
 import com.microsoft.azure.sdk.iot.device.MessageProperty;
 import com.microsoft.azure.sdk.iot.device.exceptions.IotHubSizeExceededException;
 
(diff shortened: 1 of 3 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsBatchMessage.java
```diff
@@ -41,42 +43,37 @@
     /** The current number of messages in the batch. */
     private int numMsgs;
 
-    /** Constructor. Initializes the batch body as an empty JSON array. */
-    public HttpsBatchMessage()
+    public HttpsBatchMessage(List<HttpsSingleMessage> messageList) throws IotHubSizeExceededException
```
drwill-ms, line 46: Presumably this is not part of the public API and therefore not a breaking change?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsBatchMessage.java
```diff
@@ -41,42 +43,37 @@
...
+        {
+            addJsonToStringBuilder(message, batchBodyBuilder);
+            batchBodyBuilder.append(','); // comma to separate each object in the json array
+            this.numMsgs++;
+            atLeastOneMessage = true;
+        }
+
+        if (atLeastOneMessage)
+        {
+            batchBodyBuilder.deleteCharAt(batchBodyBuilder.length() - 1); // remove the final comma from the list
```
drwill-ms, line 62: Another approach to consider, that doesn't requiring removing a character, but still has the overhead of one boolean checked on each iteration in the loop above (but only requires setting once), is to conditionally add the comma only after the first time.

So something like:
```Java
boolean isSubsequentMessage = false;
        for (HttpsSingleMessage message : messageList)
        {
            if (isSubsequentMessage) {
                batchBodyBuilder.append(','); // comma to separate each object in the json array
            } else {
                isSubsequentMessage = true;
            }
            addJsonToStringBuilder(message, batchBodyBuilder);
            this.numMsgs++;
        }
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsBatchMessage.java
timtay-microsoft: Correct. Users don't create this object.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsBatchMessage.java
```diff
@@ -170,7 +164,7 @@ private static String msgToJson(HttpsSingleMessage msg)
             for (String key : allProperties.keySet())
             {
                 jsonMsg.append("\"").append(key).append("\":");
-                jsonMsg.append("\"").append(allProperties.get(key)).append("\",");
+                jsonMsg.append("\"").append(allProperties.get(key)).append("\","); //TODO
```
drwill-ms: todo what?

drwill-ms (approved):

jamdavi (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsBatchMessage.java
jamdavi: Another thought here too is to consider [String.join](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.Iterable-).

In this instance you'd move the JSON string creation to a method in the `HttpsSingleMessage` class, create the List<String> in `sendMessage` and create a `List<String>` constructor to be used. 

A pro of this is we are using something built into the SDK and it makes it more readable; plus less lines of code. As well you're creating a method to be reused anytime you might want to spit out the JSON of a HttpsSingleMessage.

Some cons are you're moving the string creation up a level into `sendMessage` which may not want and you'd be adding a new method to the  `HttpsSingleMessage` class and it may make the batch message creation too disjointed.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/https/HttpsBatchMessage.java
timtay-microsoft: I'll send a PR out for this later. I've removed this todo for now

PR #1338, opened 2021-09-03
fix(iot-dev): Fix how the client handles failed disconnections

timtay-microsoft:
Previously, the SDK would throw an exception if the service didn't respond to the mqtt/amqp disconnect messages and it would leave the SDK in a weird semi-closed state.

This change makes it so that the close logic in our transport layers always cleans up the local resources even if the service doesn't respond to the disconnection in a timely manner


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -478,7 +478,7 @@ public void close() throws IOException
             if (this.transportClient.getTransportClientState() == TransportClient.TransportClientState.OPENED)
             {
                 // Codes_SRS_DEVICECLIENT_12_006: [If the client has been initialized to use TransportClient and the TransportClient is already opened the function shall throw an IOException.]
-                throw new IOException("Calling closeNow() when using TransportClient is not supported. Use TransportClient.closeNow() instead.");
+                throw new UnsupportedOperationException("Calling closeNow() when using TransportClient is not supported. Use TransportClient.closeNow() instead.");
```
drwill-ms, line 481: Should this be documented up by 466?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -498,26 +498,31 @@ public void close(IotHubConnectionStatusChangeReason reason, Throwable cause) th
...
         {
-            this.iotHubTransportConnection.close();
+            if (this.iotHubTransportConnection != null)
+            {
+                this.iotHubTransportConnection.close();
+            }
         }
+        finally
+        {
+            this.updateStatus(IotHubConnectionStatus.DISCONNECTED, reason, cause);
```
timtay-microsoft, line 510: On the off chance that an unchecked exception is thrown from the transport layer, we still need to update the local state accordingly

timtay-microsoft: And wake up the worker threads so they can finish

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1246,30 +1251,22 @@ private void reconnect(TransportException transportException)
 
         // reconnection may have failed, so check last retry decision, check for timeout, and check if last exception
         // was terminal
-        try
```
timtay-microsoft, line 1249: This section is a bit hard to see the diff on, but basically the try/catch was removed since the caught exception is no longer thrown

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
```diff
@@ -154,13 +152,27 @@ void disconnect() throws TransportException
...
+        {
+            try
+            {
+                this.mqttAsyncClient.close();
+            }
+            catch (MqttException ex)
+            {
+                // As per the documentation for mqttAsyncClient.close() this exception is only thrown if the client is already
+                // closed. No need to re-attempt anything here.
+                log.debug("Mqtt client was already closed, so ignoring the thrown exception", ex);
```
timtay-microsoft, line 174: I need to double check that the exception is a part of this log statement

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
jamdavi: wouldn't we want to add the exception to the throws keyword?

timtay-microsoft: ```UnsupportedOperationException``` is a runtime exception, and we typically don't declare those in the method declaration since Java doesn't require it. This is basically the same as how we have APIs that throw IllegalArgumentException but that don't list it in the declared thrown exceptions list

DBY047:
Anybody to review this PR for @timtay-microsoft? I need this fix ASAP. I can cheery-pick the commits unless you plan to change the code. 

Thanks,

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -241,10 +241,8 @@ private void startWorkerThreads()
      * Must be called to terminate the background thread that is sending data to
      * IoT Hub. After {@code close()} is called, the IoT Hub client is no longer
      *  usable. If the client is already closed, the function shall do nothing.
-     *
-     * @throws IOException if the connection to an IoT Hub cannot be closed.
      */
-    public void close() throws IOException
+    public void close()
```
abhipsaMisra, line 245: public API?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -263,43 +261,11 @@ public void close() throws IOException
...
-                throw new IOException(e);
-            }
-
+            this.transport.close(IotHubConnectionStatusChangeReason.CLIENT_CLOSE, null);
             this.state = IotHubConnectionStatus.DISCONNECTED;
         }
     }
 
-    // Functionally the same as "close()", but without wrapping any thrown TransportException into an IOException
-    public void closeWithoutWrappingException() throws TransportException
```
abhipsaMisra, line 281: is this removed because the close will now silently consume any thrown exceptions?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
```diff
@@ -379,7 +379,7 @@ private DeviceClient(String uri, String deviceId, SecurityProvider securityProvi
         commonConstructorSetup();
     }
 
-    void closeFileUpload() throws IOException
```
drwill-ms, line 382: Is this a breaking change?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -241,10 +241,8 @@ private void startWorkerThreads()
      * Must be called to terminate the background thread that is sending data to
      * IoT Hub. After {@code close()} is called, the IoT Hub client is no longer
      *  usable. If the client is already closed, the function shall do nothing.
-     *
-     * @throws IOException if the connection to an IoT Hub cannot be closed.
      */
-    public void close() throws IOException
```
drwill-ms, line 247: Same; breaking change?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUpload.java
```diff
@@ -116,10 +116,8 @@ public synchronized void uploadToBlobAsync(
 
     /**
      * Close the file upload cancelling all existing uploads and shutting down the thread pool.
-     *
-     * @throws IOException if an I/O error occurs in the inputStream.
      */
-    public void closeNow() throws IOException
+    public void closeNow()
```
abhipsaMisra, line 120: public API?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceClient.java
timtay-microsoft: No, this method isn't accessible to users

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/fileupload/FileUpload.java
timtay-microsoft: Nope. We construct this internally, but we never ask users to create an instance of this class

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: Nope. We construct this internally, but we never ask users to create an instance of this class

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: It's removed because close no longer throws the exception that we used to wrap in an IOException

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
timtay-microsoft: Nope. We construct this internally, but we never ask users to create an instance of this class

drwill-ms (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
abhipsaMisra: what kind of unchecked exceptions might be expected to be thrown here?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -333,10 +333,10 @@ public void close() throws TransportException
             }
             catch (InterruptedException e)
             {
-                throw new TransportException("Interrupted while closing proton reactor", e);
+                // If the SDK's disconnect message doesn't make it to the service, just clean up the local resources.
+                // The service will eventually figure out that the connection was lost since the keep-alive pings stop.
```
abhipsaMisra, line 337: I recently learned that http also supports keep-alive values and uses it for tls renegotiation. Do you know if we set http keep-alive in our sdk?

abhipsaMisra (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
timtay-microsoft: Possibly "out of memory" type exceptions

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
timtay-microsoft: I don't believe we do in the Java SDK, no

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
timtay-microsoft: Yep, this does work the way I thought. Even though there is no "{}" space for it, log.debug knows to print the exception with the base message

Issue #1336, opened 2021-09-03
[Bug Report] Multiplexed AMQP connection starts superfluous AmqpsSasTokenRenewalHandler


niklas-sparfeld-gcx:
# Context

- **OS and version used:** macOS & Ubuntu Linux
- **Java runtime used:** Adopt JDK 11
- **SDK version used:** com.microsoft.azure.sdk.iot:iot-device-client:1.32.1

## Description of the issue

We open a multiplexed AMQPS session - I do not yet know whether the bug also works for normal connections.

We can reproduce the bug with one or more device clients on that connection.

When some Exception is thrown by the underlying AMQP layer (e.g. `TransportException: Timed out waiting for authentication`) the session is re-established and *multiple* `AmqpsSasTokenRenewalHandler` are created for the existing client.

After a while this leads to a lot authentication messages being sent: `[k-ReactorRunner] .a.s.i.d.t.a.AmqpsSasTokenRenewalHandler : Sending authentication message for device ...`.

![Screenshot 2021-09-03 at 12 40 33](https://user-images.githubusercontent.com/70274334/131992874-5dfc41bc-8563-448c-85db-7bf0d011bbf1.png)

Ultimately the IoT Hub is overwhelmed and does not respond anymore.

----

I believe the issue is here:

https://github.com/Azure/azure-iot-sdk-java/blob/d948ef4906d408505fed221604942ecc595ae4f5/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java#L947-L981
`addSessionHandler` is called in order to re-establish the session, it finds the *existing* `AmqpsSessionHandler` instance, but it *appends it redundantly* to the `sessionHandlers`.

![Screenshot 2021-09-03 at 10 10 29](https://user-images.githubusercontent.com/70274334/131992055-b46d84f8-185e-448c-bb9d-39acbdf66ba0.png)

Later in `onConnectionLocalOpen()` a new SAS token handler is instantiated per `AmqpsSessionHandler`.

https://github.com/Azure/azure-iot-sdk-java/blob/d948ef4906d408505fed221604942ecc595ae4f5/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java#L462-L466

In the log output below, you also see multiple `AmqpsSessionHandler  : Opening device session for device 89345678901234567892` for the same client. I think this should never happen.

[...]

PR #1341, opened 2021-09-07
fix(iot-dev): Fix bug where multiplexed client had more session handlers than expected

timtay-microsoft:
If a multiplexing client tried to open, failed due to a network issue or something similar, and then later tried to open again, the session handler list would contain 2 session handlers for each multiplexed device instead of the expected 1. The issue was that the catch logic within the open call didn't clear the session handler list in the event of a network issue or interruption.

#1336


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -952,8 +971,8 @@ private AmqpsSessionHandler addSessionHandler(DeviceClientConfig deviceClientCon
         {
             if (existingAmqpsSessionHandler.getDeviceId().equals(deviceClientConfig.getDeviceId()))
             {
-                amqpsSessionHandler = existingAmqpsSessionHandler;
-                break;
```
timtay-microsoft, line 956: This session handler was retrieved from the local state, but later added to the same list it was retrieved from for some reason. Now we just return it once it is found. No need to add a duplicate session handler to the list of session handlers

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -974,6 +993,11 @@ private AmqpsSessionHandler addSessionHandler(DeviceClientConfig deviceClientCon
         {
             amqpsSessionHandler = new AmqpsSessionHandler(deviceClientConfig, this);
         }
+        else
+        {
+            // Since a session is being added for the reconnecting device, it can be removed from the reconnectingDeviceSessionHandlers collection
+            this.reconnectingDeviceSessionHandlers.remove(amqpsSessionHandler);
```
timtay-microsoft, line 999: This list is of multiplexed devices that are temporarily unregistered for reconnection purposes. Since the addSessionHandler call is used to re-add the unregistered device, now is the appropriate time to remove the device from this list.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -298,6 +305,18 @@ public void open() throws TransportException
         log.debug("Amqp connection opened successfully");
     }
 
+    private void clearLocalState()
+    {
+        this.sessionHandlers.clear();
```
jamdavi, line 310: Would there be any need to nullcheck anything in here? Just seeing it being called in the catch so it might give an unexpected NRE.

jamdavi (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
timtay-microsoft: Nope. Both of these queues are final, so they are instantiated once and never set to null

PR #1344, opened 2021-09-08
fix(iot-serv): Fix bug where service client's thread pool wasn't shutdown when the service client is shutdown

timtay-microsoft:
Previously, async operations may have continued from the service client even after closing it

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
@@ -25,7 +25,7 @@ import java.util.concurrent.Executors;
 @Slf4j
 public class ServiceClient
 {
-    private final ExecutorService executor = Executors.newFixedThreadPool(10);
+    private ExecutorService executor;
 
     private final AmqpSend amqpMessageSender;
     private final String hostName;
@@ -298,6 +298,8 @@ public class ServiceClient
             throw new IOException("AMQP sender is not initialized");
         }
 
+        this.executor = Executors.newFixedThreadPool(10);
+
         log.info("Opening service client...");
 
         this.amqpMessageSender.open();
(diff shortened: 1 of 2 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -298,6 +298,8 @@ public void open() throws IOException
             throw new IOException("AMQP sender is not initialized");
         }
 
+        this.executor = Executors.newFixedThreadPool(10);
```
timtay-microsoft, line 301: We can't simply re-open a previously used executor service since they are single use and can't be re-used after they are closed.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -315,6 +317,8 @@ public void close() throws IOException
             throw new IOException("AMQP sender is not initialized");
         }
 
+        this.executor.shutdownNow();
```
timtay-microsoft, line 320: One other option here is to do ```this.executor.shutdown()``` instead. That approach allows threads to finish "gracefully" but our AMQP handlers aren't really setup to handle an interrupt in the first place so that seemed unnecessary to me.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
abhipsaMisra: When would we need to reopen/reinitialize this executor service? Do service clients follow open-close-open flow?

timtay-microsoft: We do allow for users to re-open a closed service client, yes. As long as they call open, the executor service is re-initialized for them. Users can't do open -> close -> sendAsync, for example. But they can do open -> close -> open -> sendAsync

abhipsaMisra (approved):

PR #1346, opened 2021-09-08
fix(iot-serv): Fix how serviceClient runs openAsync and closeAsync

timtay-microsoft:
As an addendum to #1344, make it so that openAsync and closeAsync don't run on the same thread pool that they are supposed to create/shutdown

Note that #1344 was not released before this addendum was added, so there is no need to include this "fix" in the next release's release notes

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
@@ -382,7 +382,7 @@ public class ServiceClient
             {
                 future.completeExceptionally(e);
             }
-        });
+        }).run();
 
         return future;
     }
@@ -410,7 +410,7 @@ public class ServiceClient
             {
                 future.completeExceptionally(e);
             }
-        });
+        }).run();
 
         return future;
     }
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
@@ -369,7 +369,10 @@ public class ServiceClient
     public CompletableFuture<Void> openAsync()
     {
         final CompletableFuture<Void> future = new CompletableFuture<>();
-        executor.submit(() -> {
+
+        // executor service hasn't been initialized yet, so don't use it to run this thread
+        new Thread(() ->
+        {
             try
             {
                 open();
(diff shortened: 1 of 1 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -369,7 +369,10 @@ public void send(String deviceId, String moduleId, Message message) throws IOExc
     public CompletableFuture<Void> openAsync()
     {
         final CompletableFuture<Void> future = new CompletableFuture<>();
-        executor.submit(() -> {
+
+        // executor service hasn't been initialized yet, so don't use it to run this thread
```
abhipsaMisra, line 373: when is the executor service initialized? 
does this thread now run unmonitored? what will happen if this gets interrupted?

timtay-microsoft: The executor service is initialized in the ```open()``` (synchronous) call. It can't be interrupted because the open call doesn't throw an interrupted exception, and we don't expose this thread to the user to allow them to interrupt it either. The open() call that is run asynchronously actually has no IO operations in it. It basically just initializes the executor service.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/ServiceClient.java
```diff
@@ -379,7 +382,8 @@ public void send(String deviceId, String moduleId, Message message) throws IOExc
             {
                 future.completeExceptionally(e);
             }
-        });
+        }).run();
```
abhipsaMisra, line 385: I'm a bit confused as to what we want done here - do we want to call open() on a different thread rather than the current thread? If so, why? And also, if we want to run it on a different thread don't we need to call start()?

timtay-microsoft: >do we want to call open() on a different thread rather than the current thread?

 Yes, because this method is ```openAsync()```

> If so, why?

If we run ```open()``` on the current thread, then it isn't async at all

> And also, if we want to run it on a different thread don't we need to call start()?

You are right, I shouldn't be using ```run()``` here. I should be using ```start()```

timtay-microsoft: This mess of async flavors of APIs is making me consider just removing them from our v2 client. I'll bring this up with our team in standup soon. They don't really add anything that a customer couldn't create a wrapper to handle

timtay-microsoft:
Closing this for now since #1344 is being reverted for now

PR #1353, opened 2021-09-09
refactor(iot-serv): deprecate async service client and registry manager APIs

timtay-microsoft:
There are no usages of these methods in our e2e tests that we need to update. They do still have some unit tests, though


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/RegistryManager.java
```diff
@@ -265,7 +265,9 @@ public Device addDevice(Device device) throws IOException, IotHubException, Json
      * @return The future object for the requested operation
      * @throws IOException This exception is thrown if the IO operation failed
      * @throws IotHubException This exception is thrown if the response verification failed
+     * @deprecated Use the synchronous version of this API {@link #addDevice(Device)}.
```
abhipsaMisra, line 268: A small description on why we deprecated this API might be helpful, i.e. all this API does is spawn a new thread and have it execute the sync API. So you're better off calling the sync API directly instead.

abhipsaMisra (approved):

Issue #1348, opened 2021-09-09
[Bug Report] IoTHubTransport updateStatus callback called in the wrong order


DBY047:
- **OS and version used:** Android 4.4
- **Java runtime used:** 1.6
- **SDK version used:** Java release 2020-11-16 (1.27.0) and Java release 2021-08-09 (1.32.1)

## Description of the issue
When IoTHubTransport->updateStatus is called, deviceIOConnectionStatusChangeCallback is call after invokeConnectionStatusChangeCallback. 
invokeConnectionStatusChangeCallback handle callback registered from deviceClient->subscribeToDeviceMethod: client code. 
deviceIOConnectionStatusChangeCallback handle callback for the deviceIO object: SDK code.  

When our code receive a CONNECTED event and try to call method like subscribeToDeviceMethod or subscribeToTwinDesiredProperties, IOException is thrown because deviceIO have not been notified yet that the connection is establish. I think deviceIOConnectionStatusChangeCallback shall be call first and then call client connectionStatusChangeCallback. Doing that, deviceIO will call startWorkerThreads and set the state to CONNECTED and be ready for whatever the client code need to do when notified of the connection status change, 

I mean by client code the code that create a new DeviceClient object and subscribe to eventCallback: whatever code that use the SDK. 

This make IotHubConnectionStatusChangeCallback unreliable because the status change to CONNECTED is false. 

## Code sample exhibiting the issue
```java
//invoke connection status callbacks
this.log.debug("Invoking connection status callbacks with new status details");
invokeConnectionStateCallback(newConnectionStatus, reason);
invokeConnectionStatusChangeCallback(newConnectionStatus, reason, throwable);
this.deviceIOConnectionStatusChangeCallback.execute(newConnectionStatus, reason, throwable, null);
```

## Console log of the issue
```
Open the client connection before using it.
java.io.IOException: Open the client connection before using it.
	at com.microsoft.azure.sdk.iot.device.InternalClient.subscribeToMethodsInternal(InternalClient.java:704)
	at com.microsoft.azure.sdk.iot.device.DeviceClient.subscribeToDeviceMethod(DeviceClient.java:668)
	at com.distech.android.configuration.iot.IoTClient.setupDeviceMethodsHandleAndCallback(IoTClient.java:187)
	at com.distech.android.configuration.iot.IoTClient.execute(IoTClient.java:293)
[...]

PR #1355, opened 2021-09-10
fix(iot-dev): Fix connection status reporting CONNECTED before all SDK states are CONNECTED

timtay-microsoft:
We were executing user facing callbacks with CONNECTED before we were executing SDK facing callbacks with the same state. As a result, users would try to do something like send telemetry as soon as CONNECTED state callback executes only to see an error saying that the deviceIO layer wasn't CONNECTED yet, too.

The solution is to make sure all SDK facing state callbacks are executed before user facing state callbacks are executed

#1348


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1539,8 +1541,6 @@ private void updateStatus(IotHubConnectionStatus newConnectionStatus, IotHubConn
             {
                 this.multiplexingStateCallback.execute(newConnectionStatus, reason, throwable, this.multiplexingStateCallbackContext);
             }
-
-            this.deviceIOConnectionStatusChangeCallback.execute(newConnectionStatus, reason, throwable, null);
```
timtay-microsoft, line 1543: The user facing state callbacks were above this SDK facing state callback previously. Now this SDK facing state callback executes first

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -636,7 +636,15 @@ public void onDeviceSessionOpened(String deviceId)
...
-            this.listener.onMultiplexedDeviceSessionEstablished(this.connectionId, deviceId);
+
+            if (isMultiplexing)
+            {
+                // For non-multiplexing cases, the above logic for counting down the latch for the single device session
+                // will notify the open() thread that the reactor thread has successfully opened the device session.
+                // That will then execute the onConnectionEstablished callback on this same listener. Because of that,
+                // there is no listener callback to execute here for non-multiplexing cases.
+                this.listener.onMultiplexedDeviceSessionEstablished(this.connectionId, deviceId);
+            }
```
timtay-microsoft, line 647: I found this bug while investigating this issue. We shouldn't have been calling "on**Multiplexed**DeviceSessionEstablished" when not multiplexing. It resulted in the SDK making multiple internal connection state updates when only one should have happened.

timtay-microsoft: This bug shouldn't have any meaningful impact on a user, but it did make debugging the connection state reporting confusing

abhipsaMisra (approved):

Issue #1349, opened 2021-09-09
iot-device-client 1.32.1 with edgeHub 1.1.4 prints repeated warning "A message was received with a null correlation id"


derSchtefan:
# Context

- Ubuntu 18.04
- openjdk version "11.0.12" 2021-07-20 LTS
- iot-device-client 1.32.1

## Description of the issue
We are processing messages from the iot edge edgeHub  1.1.4 like recommended in the examples:
```java
IotHubClientProtocol protocol = IotHubClientProtocol.MQTT;
log.info("Start to create client with MQTT protocol");
ModuleClient client = ModuleClient.createFromEnvironment(protocol);
log.info("Client created");
client.setMessageCallback(IotHubCallbacks.INPUT_NAME, msgCallback, client);
```

The messages are received from the Azure Industrial-IOT  OpcPublisher. 

When upgrading to iot-device-client 1.32.1, we receive the warning:
` c.m.a.s.i.d.transport.IotHubTransport    : A message was received with a null correlation id.`
logged for each received message. It floods the message log.

It did not happen in earlier versions of the SDK. (1.31.0)

Since we have no control over OpcPublisher and what kind of messages it sends, and those correlation ids are not documented, what are the implications. how can we fix this issue, or do we have to silence and ignore the warning ?

DBY047:
I see the same things on my side but the warning message is wrong. First, the correlationId cannot be null. getCorrelationId will return an empty string if correlationId is null on the Message object. Second, the real error is that correlationCallbacks  containsKey return null for that correlationId: correlationCallbacks is empty when this occurred. I see 2 of these warnings. I added more about the message below. 

```
A message was received with a null correlation id:  
Message details: Correlation Id [c8a6f487-d075-48ef-8bcf-ed513b91b145] Message Id [80c9b4f0-56f4-4106-a26c-27ed2bc30756] Request Id [c8a6f487-d075-48ef-8bcf-ed513b91b145] Device Operation Type [DEVICE_OPERATION_TWIN_GET_RESPONSE] 

A message was received with a null correlation id:  
[...]

PR #1357, opened 2021-09-14
fix(client): Correct verbosity of correlation warnings and fix the callback context map

jamdavi:
This addresses issue #1349 and fixes the following problems

* Too many warning messages for the correlation callback not being set, which is optional
* Fix NPE if no callback context is specified
* Fix NPE if correlationId is null

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -226,7 +226,7 @@ public class IotHubTransport implements IotHubListener
             try
             {
                 String correlationId = message.getCorrelationId();
-                if (correlationId != null && correlationCallbacks.containsKey(correlationId))
+                if (!correlationId.isEmpty())
                 {
                     Object context = correlationCallbackContexts.get(correlationId);
                     correlationCallbacks.get(correlationId).onRequestAcknowledged(packet, context, e);
@@ -242,7 +242,7 @@ public class IotHubTransport implements IotHubListener
             try
             {
                 String correlationId = message.getCorrelationId();
-                if (correlationId != null && correlationCallbacks.containsKey(correlationId))
+                if (!correlationId.isEmpty())
                 {
                     Object context = correlationCallbackContexts.get(correlationId);
                     correlationCallbacks.get(correlationId).onUnknownMessageAcknowledged(message, context, e);
@@ -278,7 +278,7 @@ public class IotHubTransport implements IotHubListener
             if (message != null)
             {
                 String correlationId = message.getCorrelationId();
-                if (correlationId != null && correlationCallbacks.containsKey(correlationId))
+                if (!correlationId.isEmpty())
                 {
                     Object context = correlationCallbackContexts.get(correlationId);
                     correlationCallbacks.get(correlationId).onResponseReceived(message, context, e);
@@ -594,7 +594,7 @@ public class IotHubTransport implements IotHubListener
                     {
                         String correlationId = message.getCorrelationId();
 
-                        if (correlationId != null && correlationCallbacks.containsKey(correlationId))
+                        if (!correlationId.isEmpty())
                         {
                             Object context = correlationCallbackContexts.get(correlationId);
                             correlationCallbacks.get(correlationId).onRequestSent(message, packet, context);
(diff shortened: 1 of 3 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -242,7 +242,7 @@ public void onMessageSent(Message message, String deviceId, Throwable e)
             try
             {
                 String messageId = message.getCorrelationId();
-                if (messageId != null && correlationCallbacks.containsKey(messageId))
```
drwill-ms, line 245: Why remove `messageId != null`? If null an exception would be thrown.

drwill-ms: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.containskey?view=net-5.0#exceptions

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -242,7 +242,7 @@ public void onMessageSent(Message message, String deviceId, Throwable e)
             try
             {
                 String messageId = message.getCorrelationId();
```
drwill-ms, line 244: Our variable name is `messageId`, but the method is to get a correlation Id. Should we rename the variable to match?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
jamdavi: Actually yes, you're right. I'm using a concurrent hashmap so supplying a null will throw an NPE.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
jamdavi: I'm good with this.

timtay-microsoft (approved):

drwill-ms (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1704,12 +1700,15 @@ private void addToWaitingQueue(IotHubTransportPacket packet)
...
-                String messageId = message.getCorrelationId();
-                if (!correlationCallbacks.containsKey(messageId))
+                String correlationId = message.getCorrelationId();
+                if (correlationId != null && !correlationCallbacks.containsKey(correlationId))
                 {
-                    correlationCallbacks.put(messageId, message.getCorrelatingMessageCallback());
-                    correlationCallbackContexts.put(messageId, message.getCorrelatingMessageCallbackContext());
-                    correlationCallbacks.get(messageId).onRequestQueued(message, packet, correlationCallbackContexts.get(messageId));
+                    correlationCallbacks.put(correlationId, message.getCorrelatingMessageCallback());
+                    if (message.getCorrelatingMessageCallbackContext() != null)
```
drwill-ms, line 1707: Consider calling and caching the result, so you don't have to call it twice.

DBY047: I don't see the point of adding this check. message.getCorrelationId(); can't return null. 
```java
    /**
     * Getter for the correlationId property
     * @return The property value
     */
    public String getCorrelationId()    
    {
        // Codes_SRS_MESSAGE_34_045: [The function shall return the message's correlation ID.]
        if (correlationId == null)
        {
            return "";
        }

        return correlationId;
    }
```

jamdavi: Can you link to that? I am looking in our current master and that's not there.

jamdavi: Ah I see, yes, that method is changed on my end (as part of a larger refactor) to a Lombok Getter/Setter but it isn't guarded with @NonNull. Thanks for catching that as I wasn't in the master branch when I was looking at the code.

I've updated the branch I'm working on to keep this behavior.

PR #1362, opened 2021-09-16
feat(pnp): Adding device commands

jamdavi:
Adds the device commands to the Java SDK


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/convention/DeviceCommand.java
```diff
@@ -0,0 +1,182 @@
...
+                            if (commandName.contains("*"))
+                            {
+                                String[] componentAndMethod = methodMessage.getMethodName().split("\\*");
+                                componentName = componentAndMethod[0];
+                                commandName = componentAndMethod[1];
+
+                            }
+
+                            log.trace("Executing command invocation callback for component {} with command name {} for message {}",  componentName == null ? "default" : componentName , commandName, methodMessage);
+                            DeviceCommandRequest commandRequest = new DeviceCommandRequest(componentName, commandName, methodMessage.getBytes(), payloadConvention);
```
jamdavi, line 83: The device command request allows us to be specific with the payload. Therefore a customer can use the convention without being aware of it.

deps/pom.xml
```diff
@@ -39,6 +39,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <junit-version>4.12</junit-version>
         <mockito-version>1.10.19</mockito-version>
+        <maven.javadoc.skip>true</maven.javadoc.skip>
```
jamdavi, line 42: Will remove this before merge.

PR #1365, opened 2021-09-17
Make SDK explicitly use UTF-8 for converting strings to bytes

timtay-microsoft:
IoT device client, IoT service client, deps, DPS device client, DPS service client, tpm security provider, and symmetric key security provider

This shouldn't impact most users since this is the default encoding anyways, but this is flagged as an issue in Spotbugs

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/TwinParser.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/TwinParser.java
@@ -21,7 +21,7 @@ public class TwinParser
 
     private TwinChangedCallback onDesiredCallback = null;
     private TwinChangedCallback onReportedCallback = null;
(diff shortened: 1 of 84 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/TwinParser.java
```diff
@@ -107,7 +107,7 @@ public void setReportedCallback(TwinChangedCallback onReportedCallback)
      */
     public void setTagsCallback(TwinChangedCallback onTagsCallback)
     {
-        TwinParser.onTagsCallback = onTagsCallback;
```
timtay-microsoft, line 110: This was another warning caught by Spotbugs. This was unusual of us to have a static field with an instance method setter. This class is deleted in v2 though, and is unused currently, so this change won't impact anyone

abhipsaMisra (approved):

PR #1366, opened 2021-09-20
refactor(all): fix more spotbugs warnings across all the packages

timtay-microsoft:
Mostly enforcing UTF-8 encoding rather than allowing for the default encoding


deps/src/test/java/com/microsoft/azure/sdk/iot/deps/serializer/ConfigurationParserTest.java
```diff
@@ -179,7 +179,7 @@ public void gettersAndSetters()
         //assert
         assertEquals(id, parser.getId());
         assertEquals(schemaVersion, parser.getSchemaVersion());
-        assertEquals("123", parser.getLabels().get(new String("xyz")));
```
timtay-microsoft, line 182: No need to ```new String()``` something that is already a string

abhipsaMisra (approved):

azabbasi (approved):

PR #1370, opened 2021-09-27
doc(iot-dev): Add additional documentation to connection status callback

timtay-microsoft:
A few customers have been confused about this point, so it is worth adding some clarification around it.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
@@ -420,6 +420,11 @@ public class InternalClient
      * with a status and a reason why the device's status changed. When the callback is fired, the provided context will
      * be provided alongside the status and reason.
      *
+     * This connection status callback is not triggered by any upstream connection change events. For example, if
+     * if the connection status callback is set for a module on an IoT Edge device and that IoT Edge device
+     * loses connection to the cloud, this connection status callback won't execute since the connection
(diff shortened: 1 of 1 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -420,6 +420,11 @@ public void sendReportedProperties(Set<Property> reportedProperties, Integer ver
      * with a status and a reason why the device's status changed. When the callback is fired, the provided context will
      * be provided alongside the status and reason.
      *
+     * This connection status callback is not triggered by any upstream connection change events. For example, if
```
drwill-ms, line 423: Are you also planning customer-facing docs?

drwill-ms (approved):

azabbasi (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
timtay-microsoft: What do you mean? These are the customer facing docs. This class is "InternalClient" but the method and its javadocs are exposed via inheritance in both ModuleClient and DeviceClient

drwill-ms: I assumed the fact it is named internal client meant it wasn't customer facing. My mistake. :)

PR #1371, opened 2021-09-27
feat(iot-dev): Add support for Plug and Play feature over AMQP

timtay-microsoft:
Includes some multiplexing scenario PnP tests since customers will be trying PnP out in their protocol gateways

One interesting discussion we could have here is whether we need to release this in preview first or not. You'll notice that this PR doesn't make any API changes to enable this new feature, so I'm not worried about needing to make a breaking change in our API design later. With that in mind, I feel comfortable forgoing preview as long as we are committed to supporting PnP over AMQP. I'm open to hearing otherwise, though.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsTelemetrySenderLinkHandler.java
```diff
@@ -28,7 +28,7 @@
 
     AmqpsTelemetrySenderLinkHandler(Sender sender, AmqpsLinkStateCallback amqpsLinkStateCallback, DeviceClientConfig deviceClientConfig, String linkCorrelationId)
     {
-        super(sender, amqpsLinkStateCallback, linkCorrelationId);
+        super(sender, amqpsLinkStateCallback, linkCorrelationId, deviceClientConfig.getModelId());
```
azabbasi, line 31: Does that mean deviceClientConfig will never be null? do we need a nullcheck?

timtay-microsoft: It won't be null here, no

jamdavi:
As part of PnP support we would need to add the componentName to the Message I would think. This would allow a customer to construct an Event (Telemetry) with the correct properties.

Or would this be out of scope for this particular implementation?

timtay-microsoft:
> As part of PnP support we would need to add the componentName to the Message I would think. This would allow a customer to construct an Event (Telemetry) with the correct properties.
> 
> Or would this be out of scope for this particular implementation?

Isn't that transport agnostic? From the samples for PnP, I see that that twin message is constructed with the componentName, and then the message is sent just like any other.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
```diff
@@ -183,6 +190,84 @@ public void getDigitalTwin() {
...
+        String thermostatDeviceId = multiplexedDeviceClients.get(0).getConfig().getDeviceId();
+        BasicDigitalTwin response = digitalTwinClient.getDigitalTwin(thermostatDeviceId, BasicDigitalTwin.class);
+        ServiceResponseWithHeaders<BasicDigitalTwin, DigitalTwinGetHeaders> responseWithHeaders =
+            digitalTwinClient.getDigitalTwinWithResponse(thermostatDeviceId, BasicDigitalTwin.class);
+
+        // assert
+        assertEquals(response.getMetadata().getModelId(), E2ETestConstants.THERMOSTAT_MODEL_ID);
+        assertEquals(responseWithHeaders.body().getMetadata().getModelId(), E2ETestConstants.THERMOSTAT_MODEL_ID);
+
+        // act
```
drwill-ms, line 221: I don't love unit tests that get overly complicated beyond a single pass at arrange/act/assert.

Should this be 2 unit tests? Can the acts and asserts be grouped together?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsMethodsSenderLinkHandler.java
```diff
@@ -33,7 +33,7 @@
 
     AmqpsMethodsSenderLinkHandler(Sender sender, AmqpsLinkStateCallback amqpsLinkStateCallback, DeviceClientConfig deviceClientConfig, String linkCorrelationId)
     {
-        super(sender, amqpsLinkStateCallback, linkCorrelationId);
+        super(sender, amqpsLinkStateCallback, linkCorrelationId, deviceClientConfig.getModelId());
```
drwill-ms, line 36: Do we need to worry about URL escaping the model Id for AMQP? I forget. I know we had to for MQTT.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSenderLinkHandler.java
```diff
@@ -36,10 +37,16 @@
...
 
-    AmqpsSenderLinkHandler(Sender sender, AmqpsLinkStateCallback amqpsLinkStateCallback, String linkCorrelationId)
+    AmqpsSenderLinkHandler(Sender sender, AmqpsLinkStateCallback amqpsLinkStateCallback, String linkCorrelationId, String modelId)
     {
         this.amqpProperties = new HashMap<>();
         this.amqpProperties.put(Symbol.getSymbol(API_VERSION_KEY), TransportUtils.IOTHUB_API_VERSION);
+
+        if (modelId != null && !modelId.isEmpty())
+        {
+            this.amqpProperties.put(Symbol.getSymbol(PNP_MODEL_ID_KEY), modelId);
```
drwill-ms, line 47: Do we need to worry about URL escaping the model Id for AMQP? I forget. I know we had to for MQTT.

drwill-ms (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/digitaltwin/DigitalTwinClientTests.java
timtay-microsoft: This is an integration test. I can remove one of the "act" blocks here, though

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSenderLinkHandler.java
timtay-microsoft: Looks like we don't need to. We URL escaped the model Id in MQTT because it was embedded into a URL. Here in AMQP though, it is just the value in a key-value pair. I checked, and we are not URL escaping the model Id for AMQP in the .NET SDK either.

jamdavi:
> > As part of PnP support we would need to add the componentName to the Message I would think. This would allow a customer to construct an Event (Telemetry) with the correct properties.
> > Or would this be out of scope for this particular implementation?
> 
> Isn't that transport agnostic? From the samples for PnP, I see that that twin message is constructed with the componentName, and then the message is sent just like any other.

It's only agnostic for Twin and Methods. For events we need to have the componentName and the correct system property. Abhipsa has the AMQP implementation in the below PR.

https://github.com/Azure/azure-iot-sdk-csharp/pull/1467/files

timtay-microsoft:
> It's only agnostic for Twin and Methods. For events we need to have the componentName and the correct system property. Abhipsa has the AMQP implementation in the below PR.

I see. I pushed some changes to capture that then. I noticed that we weren't doing this in MQTT either, so I added the component name to the telemetry for both AMQP and MQTT

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
```diff
@@ -328,29 +329,20 @@ public IotHubTransportMessage receive() throws TransportException
...
     {
-        TransportException ex = null;
-
         log.warn("Mqtt connection lost", throwable);
 
         this.disconnect();
 
         if (this.listener != null)
         {
-            if (ex == null)
```
timtay-microsoft, line 339: ```ex``` was always null, so I removed some unnecessary blocks of code here

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSenderLinkHandler.java
```diff
@@ -273,13 +280,22 @@ MessageImpl iotHubMessageToProtonMessage(Message message)
...
+        Map<Symbol, Object> messageAnnotationsMap = new HashMap<>();
         if (message.isSecurityMessage())
         {
-            userProperties.put(MessageProperty.IOTHUB_SECURITY_INTERFACE_ID, MessageProperty.IOTHUB_SECURITY_INTERFACE_ID_VALUE);
+            messageAnnotationsMap.put(Symbol.valueOf(MessageProperty.IOTHUB_SECURITY_INTERFACE_ID), MessageProperty.IOTHUB_SECURITY_INTERFACE_ID_VALUE);
         }
 
-        ApplicationProperties applicationProperties = new ApplicationProperties(userProperties);
-        outgoingMessage.setApplicationProperties(applicationProperties);
+        if (message.getComponentName() != null)
```
jamdavi, line 292: Let's also skip if the componentName isEmpty. From what I understand of the DTDL the the component name can't be empty, it has to be 1 or more Alphanumeric characters and underscores are allowed after the first character which must be a-z.

```suggestion
        if (message.getComponentName() != null && !message.getComponentName().isEmpty())
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttMessaging.java
```diff
@@ -108,6 +108,11 @@ public void send(Message message) throws TransportException
             separatorNeeded = appendPropertyIfPresent(stringBuilder, separatorNeeded, MQTT_SECURITY_INTERFACE_ID, MessageProperty.IOTHUB_SECURITY_INTERFACE_ID_VALUE, false);
         }
 
+        if (message.getComponentName() != null)
```
jamdavi, line 111: Same here.

```suggestion
        if (message.getComponentName() != null && !message.getComponentName().isEmpty())
```

PR #1376, opened 2021-10-04
fix(iot-dev): Fix endless loop when messages get requeued after a AMQP session has been closed

niklas-sparfeld-gcx:
Signed-off-by: Niklas Sparfeld <niklas.sparfeld@grandcentrix.net>

# Checklist

# Description of the problem

This draft PR fixes an issue in IoT Device Client`AmqpsIotHubConnection`. The *symptom* we see on our system is an endless loop of messages being re-queued again and again: 

Screenshot 2021-10-04 at 09 53 56

My investigation shows that this happens when a session is unexpectedly closed *while* a message is being handled.
The message is not sent successfully, its session is no longer in `this.sessionHandlers` and when it is requeued it will be requeued forever.

By now we can confirm that we see this behavior in multiplexed as well as simplex clients.

---

I think the *underlying* problem is a bit different, and I do not want to pollute the discussion here too much: The `AmqpsIotHubConnection` relies a lot on the fact that `onReactorFinal` is fired. But when the underlying Proton library runs into an exception (which is *does*), it can happen that the `CONNECTION_FINAL` event is never placed in the reactor (`ReactorImpl.process()`).

I'll prepare an issue for that discussion some time this week I think.

# Description of the solution

Instead of opening an issue, I wanted to give it a try and fix it myself. Which is to say: The solution might be complete bonkers ;)

You will see that I simply to not re-queue messages when no session is responsible for it anymore.

I did not yet have the time to get your unit testing suite running and write a test case.

Also I have to say that I'm not able to *demonstrate* the issue, because it is so hard to reproduce.

Thanks for anything that helps :) Feel free to add changes to this PR as you see fit, I do not have much time myself.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -871,8 +871,22 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
         while (message != null && messagesAttemptedToBeProcessed < MAX_MESSAGES_TO_SEND_PER_CALLBACK)
         {
             messagesAttemptedToBeProcessed++;
-            boolean lastSendSucceeded = sendQueuedMessage(message);
 
+            AmqpsSessionHandler existingSessionHandler = null;
+            for (AmqpsSessionHandler sessionHandler : this.sessionHandlers)
+            {
+                if (sessionHandler.getDeviceClientConfig().getDeviceId().equals(message.getConnectionDeviceId())) {
+                    existingSessionHandler = sessionHandler;
+                    break;
+                }
+            }
+
+            if(existingSessionHandler == null) {
+                log.trace("Amqp message failed to send, because no session handler exists for the client. Not adding it back to the queue ({})", message);
+                continue;
+            }
+
+            boolean lastSendSucceeded = existingSessionHandler.sendMessage(message);
             if (!lastSendSucceeded)
             {
                 //message failed to send, likely due to lack of link credit available. Re-queue and try again later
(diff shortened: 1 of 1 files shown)


drwill-ms (approved):

timtay-microsoft:
I'll take a deeper look at this for you @niklas-sparfeld-gcx. I'm a little worried about just discarding a message like this. We do have some callbacks we could invoke around the place you made the change that would notify you that the message failed to send, but we may be able to handle this more gracefully. I'd expect most unexpected session drops to be quickly followed by some retry logic to re-establish the session, and that's likely why we were re-queueing these messages in the first place.

In your repro's of this issue, are you causing these session drops manually or on purpose? Or is this just typical networking issues? And what kind of retry policies do you have attached to your device's whose sessions are being lost here?

timtay-microsoft (changes requested):
Working with @niklas-sparfeld-gcx to understand the scenario a little better before we merge anything

timtay-microsoft:
My initial thoughts after investigating this issue are that, when a session is lost during a send, one of two things should happen depending on if a retry policy is set or not:

- If the SDK will attempt to reconnect the session (default retry policy), the message should be requeued. 
  - Once the SDK stops retrying to establish the session (if retry expires), then the message should stop being requeued, and the message callback should execute with a failed status.
- If the SDK will not attempt to reconnect the session (no retry policy), the message should not be requeued, and the message callback should execute with a failed status

Do these expectations seem reasonable to you, @niklas-sparfeld-gcx ?

niklas-sparfeld-gcx:
Thanks for reaching out :) 

>I'd expect most unexpected session drops to be quickly followed by some retry logic to re-establish the session, and that's likely why we were re-queueing these messages in the first place.

Agreed. If the session is about to be opened again, I would also re-queue the messages. 

Note that *currently* this re-queuing is happening independently from the "high-level" RetryPolicy.

> these session drops manually or on purpose? Or is this just typical networking issues?

Both ;) We run into the endless loop in two scenarios:

a) We periodically re-create all clients from time to time, because we noticed that after e.g. 24h the connection silently stops *receiving responses* from IoT Hub. Might be a deadlock in the receive-task thread 🤷 
b) We run into NullPointerExceptions in proton-j regularly rather soon after `deviceClient.open()` & `startDeviceTwin()`

> And what kind of retry policies do you have attached to your device's whose sessions are being lost here?

We had a standard ExponentialRetryPolicy attached to it, but with reduced retryCount=3 - everything else default.

---

and last but not least:

>Do these expectations seem reasonable to you, @niklas-sparfeld-gcx ?

Yes, that is what I - and I hope everyone else as well - would have expected. As you said initially "I'm a little worried about just discarding a message like this." which is exactly what I meant by saying "The solution might be complete bonkers".

timtay-microsoft:
Okay, I've opened #1380 with my proposed fix for the infinitely requeueing message issue that you raised in this thread @niklas-sparfeld-gcx 

Feel free to let me know if you think I missed anything with my fix here. I'm also working on investigating the other issues you/your company reported. If you are able to provide any logs around these issues, I'd appreciate it. Feel free to open issues on this github repo for each of them that you've found so that other users are aware of them as well.

timtay-microsoft:
Sounds good. I do have a thread going with Teresa P, and we can just track the other issues through the github issues that you've opened up then. Do you mind if I close this PR then, since my fix will take care of this issue in its stead?

timtay-microsoft:
Closing as per the above

PR #1380, opened 2021-10-06
fix(iot-dev): Fix issue where amqp layer would requeue messages that …

timtay-microsoft:
…failed to send regardless of if the device session is still around

If a message fails to send because it's device session is no longer active, we need to either requeue the message because its session is reconnecting, or execute the callback to notify the user that the message failed to send because it's device session no longer exists.

Previously, we would always requeue the message even if the user had unregistered the device session on purpose. As a result, the message would stay in our send queue forever unless they re-registered the device session


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -338,110 +338,112 @@ void openLinks()
...
-                    }
+                if (this.explicitInProgressMethodsSubscriptionMessage == null)
+                {
+                    createMethodLinks();
+                    this.explicitInProgressMethodsSubscriptionMessage = transportMessage;
+                    return SendResult.SUCCESS; //connection layer doesn't care about this delivery tag
                 }
-                else //noinspection StatementWithEmptyBody
-                    if (subscriptionType == DEVICE_OPERATION_TWIN_UNSUBSCRIBE_DESIRED_PROPERTIES_REQUEST)
+                else
```
drwill-ms, line 376: 374 returns, so this else isn't really needed, yeah?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -338,110 +338,112 @@ void openLinks()
...
                 {
-                    //TODO: can add logic here to tear down twin links if the user wants to unsubscribe to desired properties
+                    log.debug("Rejecting methods subscription message because that subscription is already in progress");
+                    return SendResult.DUPLICATE_SUBSCRIPTION_MESSAGE;
                 }
-                else if (subscriptionType == DEVICE_OPERATION_TWIN_SUBSCRIBE_DESIRED_PROPERTIES_REQUEST)
+            }
+            else //noinspection StatementWithEmptyBody
```
drwill-ms, line 382: I don't know what to make of this comment

timtay-microsoft: This is a warnings suppression, but I can remove it if I comment out this else if block

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -338,110 +338,121 @@ void openLinks()
...
-                    {
-                        log.debug("Rejecting twin subscription message because that subscription is already in progress");
-                        return false;
-                    }
-                }
+            if (subscriptionType == DEVICE_OPERATION_METHOD_SUBSCRIBE_REQUEST)
+            {
+                return handleMethodSubscriptionRequest(transportMessage);
+            }
+            //else if (subscriptionType == DEVICE_OPERATION_TWIN_UNSUBSCRIBE_DESIRED_PROPERTIES_REQUEST)
```
drwill-ms, line 365: Why is this remaining if commented out?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -338,110 +338,121 @@ void openLinks()
...
+            //{
+            //    //TODO: can add logic here to tear down twin links if the user wants to unsubscribe to desired properties
+            //}
+            else if (subscriptionType == DEVICE_OPERATION_TWIN_SUBSCRIBE_DESIRED_PROPERTIES_REQUEST)
```
drwill-ms, line 369: As 363 returns, doesn't really need `else`

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -338,110 +338,121 @@ void openLinks()
...
-    boolean sendMessage(Message message)
+    SendResult sendMessage(Message message)
     {
         if (this.deviceClientConfig.getDeviceId().equals(message.getConnectionDeviceId()))
         {
-            if (message.getMessageType() == null)
-            {
-                message.setMessageType(DEVICE_TELEMETRY);
-            }
+            return SendResult.WRONG_DEVICE;
```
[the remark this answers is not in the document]
timtay-microsoft: This is a frequently executed line of code that it would be a nuisance to log here. The AMQP connection is iterating over its sessions to find the one corresponding to this device. For instance, if a multiplexed connection had 500 devices, each telemetry message sent out could hit this line of code up to 499 times. I can certainly add some comments around what this means, though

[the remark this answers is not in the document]
drwill-ms: What does that mean about logging then? Still spurious and unhelpful?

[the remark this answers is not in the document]
timtay-microsoft: It turns this line of code into a "Should never happen" type scenario. I'll add a log for it now

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
timtay-microsoft: Got rid of the ```else```, not the ```else if``` as we discussed

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -338,110 +338,120 @@ void openLinks()
...
+        for (AmqpsSenderLinkHandler senderLinkHandler : this.senderLinkHandlers)
+        {
+            if (senderLinkHandler instanceof AmqpsTelemetrySenderLinkHandler && messageType == DEVICE_TELEMETRY
+                    || senderLinkHandler instanceof AmqpsTwinSenderLinkHandler && messageType == DEVICE_TWIN
+                    || senderLinkHandler instanceof AmqpsMethodsSenderLinkHandler && messageType == DEVICE_METHODS)
+            {
+                if (messageType == DEVICE_TWIN)
                 {
-                    if (this.twinSenderLinkOpened && this.twinReceiverLinkOpened)
+                    if (explicitInProgressTwinSubscriptionMessage != null)
```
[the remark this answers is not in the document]
timtay-microsoft: I'll add a more thorough explanation in comments here then

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -338,110 +338,120 @@ void openLinks()
...
-
-                        for (SubscriptionType subscriptionType : this.implicitInProgressSubscriptionMessages.values())
-                        {
-                            if (subscriptionType == SubscriptionType.DESIRED_PROPERTIES_SUBSCRIPTION)
-                            {
-                                // Don't send any twin messages while a twin subscription is in progress. Wait until the subscription
-                                // has been acknowledged by the service before sending it.
-                                return false;
-                            }
+                            return SendResult.SUBSCRIPTION_IN_PROGRESS;
```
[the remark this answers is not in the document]
timtay-microsoft: Knowing that there is at least one subscription of this category in progress is equally as valuable as knowing exactly how many are in progress, if that makes sense. In both cases, we need to hold off on sending twin messages until the subscriptions finish

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -90,8 +90,13 @@
     // Proton-j primitives and wrappers for the device and authentication sessions
     private Connection connection;
     private Reactor reactor;
-    private final Queue<AmqpsSessionHandler> reconnectingDeviceSessionHandlers = new ConcurrentLinkedQueue<>();
-    private final Queue<AmqpsSessionHandler> sessionHandlers = new ConcurrentLinkedQueue<>();
+
+    // keys are device Ids, values are the session handlers associated with that device id
+    private final Map<String, AmqpsSessionHandler> reconnectingDeviceSessionHandlers = new ConcurrentHashMap<>();
```
niklas-sparfeld-gcx, line 95: Should you clear this map in `clearLocalState()`?

I fear that when unexpected errors happen while closing a connection that we might run into the original issue again: The timer-task still fires (because it was not cleaned up for some reason) and a session is left over in `reconnectingDeviceSessionHandlers`.

Then messages would be re-queued endlessly just like before.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -338,120 +338,145 @@ void openLinks()
...
+        if (messageType == null)
+        {
+            // Twin and method messages have a message type assigned to them when they are constructed by this SDK
+            // (users can't construct twin/method messages directly), but telemetry messages don't necessarily have this
+            // type assigned since users may create telemetry messages. By default, assume any messages with an
+            // unassigned type are telemetry messages.
+            message.setMessageType(DEVICE_TELEMETRY);
+        }
+
+        //Check if the message being sent is a subscription change message. If so, open the corresponding links.
```
drwill-ms, line 361: nit: space after //

```suggestion
        // Check if the message being sent is a subscription change message. If so, open the corresponding links.
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -338,120 +338,145 @@ void openLinks()
...
+            return SendResult.SUCCESS;
+        }
+
+        if (this.explicitInProgressMethodsSubscriptionMessage == null)
+        {
+            createMethodLinks();
+            this.explicitInProgressMethodsSubscriptionMessage = transportMessage;
+            return SendResult.SUCCESS;
+        }
+        else
```
drwill-ms, line 465: Same here, else could be left out.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/SendResult.java
```diff
@@ -0,0 +1,45 @@
...
+     */
+    SUCCESS,
+
+    /**
+     * The message belonged to a different device session.
+     */
+    WRONG_DEVICE,
+
+    /**
+     * The message was a twin/method subscription message and the device session has already sent that subscription message.
```
drwill-ms, line 24: What does it mean that the device already sent that subscription message? Is this just an echo back?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/SendResult.java
```diff
@@ -0,0 +1,45 @@
...
+     * message is sent when the twin sender link isn't open yet. This should never happen though, because telemetry links
+     * are always opened along with the session, and our API design makes it so twin/method links are always opened
+     * before any twin/method messages can be sent.
+     */
+    LINKS_NOT_OPEN,
+
+    /**
+     * Default case.
+     */
+    UNKNOWN_FAILURE
```
drwill-ms, line 44: Interesting that this isn't the first enum option given (at least in .NET) an uninitialized enum would have the default value of 0. Too late to change this?

timtay-microsoft: Not to late, since we're just adding this enum now. I can make this the first enum

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/SendResult.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+enum SendResult
+{
+    /**
+     * The message was sent successfully.
+     */
+    SUCCESS,
+
+    /**
+     * The message belonged to a different device session.
```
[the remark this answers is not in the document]
timtay-microsoft: A multiplexed device sends messages only on it's session's links. Not on any other device's session's links. This enum value is the signal that the latter case was attempted.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -338,120 +341,143 @@ void openLinks()
...
+
+        log.debug("Rejecting twin subscription message because that subscription is already in progress");
+        return SendResult.DUPLICATE_SUBSCRIPTION_MESSAGE;
+    }
+
+    private SendResult handleMethodSubscriptionRequest(IotHubTransportMessage transportMessage)
+    {
+        if (this.methodsSenderLinkOpened && this.methodsReceiverLinkOpened)
+        {
+            // No need to do anything. Method links are already opened
```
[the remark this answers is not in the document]
timtay-microsoft: We do ack the message just beneath this comment, but you are right that this comment is a bit misleading. I'll clarify this comment to mention that we still ack the message and to explain why it isn't actually sent.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -338,120 +341,143 @@ void openLinks()
...
+            this.amqpsSessionStateCallback.onMessageAcknowledged(transportMessage, Accepted.getInstance(), this.getDeviceId());
+            return SendResult.SUCCESS;
+        }
+
+        if (this.explicitInProgressMethodsSubscriptionMessage == null)
+        {
+            createMethodLinks();
+            this.explicitInProgressMethodsSubscriptionMessage = transportMessage;
```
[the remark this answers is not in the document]
timtay-microsoft: This subscription message isn't ack'd until the methods links have successfully opened. The ```createMethodLinks()``` call behaves asynchronously here. It locally opens the method sender and receiver links, but the service needs to acknowledge these new links before we are "done" with this subscription message and can ack it. I can rename these ```createXYZLinks()``` methods to make this a bit more clear

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
timtay-microsoft: Actually, after taking a second look at this, I can't don that. The ```reconnectingDeviceSessionHandlers``` are used to persist subscription states between connection attempts. For instance, if a multiplexed device is subscribed to twin, runs into a disconnect, and tries to reconnect, this map's saved state of the device's previous session allows for us to implicitly re-subscribe. Clearing this on open/close would make it so that you'd have to re-subscribe to twin and/or methods for all of your clients manually

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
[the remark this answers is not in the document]
drwill-ms: I see, so we're processing one message which happens to match one we already are waiting on a response for, and we know in this case to do nothing because we have a match. Yes?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/SendResult.java
drwill-ms: My point is, the comment could be more conceptual in nature, rather than talk about the mechanism by which a subscription on a topic is requested.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -871,14 +873,49 @@ private void sendQueuedMessages()
         while (message != null && messagesAttemptedToBeProcessed < MAX_MESSAGES_TO_SEND_PER_CALLBACK)
         {
             messagesAttemptedToBeProcessed++;
-            boolean lastSendSucceeded = sendQueuedMessage(message);
+            SendResult sendResult = sendQueuedMessage(message);
 
-            if (!lastSendSucceeded)
+            // If no active device sessions are responsible for sending messages for the device that the message belongs to,
```
drwill-ms, line 878:
```suggestion
            // If no active device sessions are responsible for sending messages for the device that the AMQP message belongs to,
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -871,14 +873,49 @@ private void sendQueuedMessages()
...
-                log.trace("Amqp message failed to send, adding it back to messages to send queue ({})", message);
+                // Proton-j doesn't handle the scenario of sending twin/method messages on links that haven't been opened remotely yet, so hold
+                // off on sending them until the subscription has finished.
+                log.trace("Attempted to send twin/method message while the twin/method subscription was in progress. Adding it back to messages to send queue to try again after the subscription has finished ({})", message);
+                messagesToSend.add(message);
+            }
+            else if (sendResult == SendResult.LINKS_NOT_OPEN)
+            {
+                // Shouldn't happen
+                log.trace("Failed to send a message because its AMQP links were not open yet. Adding it back to messages to send queue ({})", message);
```
drwill-ms, line 911: warn? error?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -307,183 +310,218 @@ public void onLinkClosedUnexpectedly(ErrorCondition errorCondition)
...
+    private SendResult handleMethodSubscriptionRequest(IotHubTransportMessage transportMessage)
+    {
+        if (this.methodsSenderLinkOpened && this.methodsReceiverLinkOpened)
+        {
+            // No need to do anything besides ack the message. Method links are already opened so there is no need to send
+            // this message over the wire.
+            this.amqpsSessionStateCallback.onMessageAcknowledged(transportMessage, Accepted.getInstance(), this.getDeviceId());
+            return SendResult.SUCCESS;
+        }
+
```
drwill-ms, line 465: // if no subscription is in progress

drwill-ms (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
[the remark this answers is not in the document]
timtay-microsoft: Yes

Issue #1375, opened 2021-09-30
[Bug Report] 


daloiso:
I have this device twin
{
    "etag": "\"AAAAAAAAAAI=\"",
    "deviceId": "dispositivo2",
    "version": 2,
    "tags": {},
    "properties": {
        "desired": {
            "$metadata": {
                "$lastUpdated": "2021-09-30T10:52:00.8193027Z",
                "$lastUpdatedVersion": 2,
                "nested": {
                    "metadata": {
                        "$lastUpdated": "2021-09-30T10:52:00.8193027Z",
                        "$lastUpdatedVersion": 2,
                        "nested": {
                            "lastUpdatedUtc": {
                                "$lastUpdated": "2021-09-30T10:52:00.8193027Z",
                                "$lastUpdatedVersion": 2,
                                "nested": {}
                            },
                            "nested": {
                                "$lastUpdated": "2021-09-30T10:52:00.8193027Z",
                                "$lastUpdatedVersion": 2,
                                "nested": {}
                            }
                        }
                    },
                    "prova": {
                        "$lastUpdated": "2021-09-30T10:52:00.8193027Z",
                        "$lastUpdatedVersion": 2,
                        "nested": {}
                    },
                    "version": {
                        "$lastUpdated": "2021-09-30T10:52:00.8193027Z",
                        "$lastUpdatedVersion": 2,
                        "nested": {}
                    }
                }
            },
            "$version": 2,
            "metadata": {
                "lastUpdatedUtc": "2021-09-29T10:15:30.4060682Z",
                "nested": {}
            },
            "prova": 1,
            "version": 1
        },
        "reported": {
            "$metadata": {
                "$lastUpdated": "2021-09-29T10:15:30.4060682Z",
                "nested": {}
            },
            "$version": 1
        }
    },
    "capabilities": {
        "iotEdge": false
    }
}

I read the twin from java sdk and I have this error:
Exception in thread "main" java.lang.IllegalArgumentException: Twin metadata is inconsistent for property: nested
	at com.microsoft.azure.sdk.iot.deps.twin.TwinCollection.addMetadata(TwinCollection.java:366)

Debugging I have found that nested is the third metadata founded with $lastUpdated and $lastUpdatedVersion

I have the last sdk and last week the twining works.

jamdavi:
@daloiso Is this the twin we see in IoT Hub in Azure? Or, is this coming from another source?

[...]

PR #1381, opened 2021-10-06
fix(client, service): Remove metadata validation for null metadata

jamdavi:
As seen in issue #1375 if the metadata contains an unknown field and that field is null the SDK will throw an exception. There is no harm in removing this check as it will safely ignore any non expected metadata and will not affect the payload.


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/twin/TwinCollection.java
```diff
@@ -360,17 +360,12 @@ private static void addMetadata(TwinCollection twinCollection, Map<? extends Str
...
                 lastUpdatedByDigest = (String) entry.getValue();
             } else {
-                Object valueInCollection = twinCollection.get(key);
-                if (valueInCollection == null) {
-                    // If the property (key) exists in metadata but not in twinCollection metadata is inconsistent.
-                    throw new IllegalArgumentException("Twin metadata is inconsistent for property: " + key);
-                }
-
                 TwinMetadata twinMetadata = TwinMetadata.tryExtractFromMap(entry.getValue());
                 if (twinMetadata != null) {
```
drwill-ms, line 364: Maybe I don't understand this scenario very well, but if it was null, wouldn't that mean we want to delete the key from the twin? Should we have code that handles when it is null? Or, is it null in this case because we couldn't parse it. And if that is true, should we log something that says we couldn't parse something so we're doing nothing with it.

drwill-ms (approved):

Issue #1379, opened 2021-10-06
[Bug Report] AmqpsIotHubConnection.closeReactorLatch is not released in case of proton-j exceptions


niklas-sparfeld-gcx:
# Context

- **OS and version used:** macOS and Ubuntu
- **Java runtime used:** Adopt 11
- **SDK version used:** 1.33.0

## Description of the issue

Sometimes proton-j throws exceptions. For example there might be a `NullPointerException` in `SelectorImpl.update()` when `SelectionKey key` is `null`. I cannot provide a repro for it, it happens rarely after a device client has been online for many hours.

This issue shall be about an issue in the *handling* of these kinds of exceptions :)

Those exceptions are thrown while in `ReactorImpl.process()`. It breaks the processing loop, which means that `Type.REACTOR_FINAL` is never placed on the reactor queue.

This in turn leads to the fact that `AmqpsIotHubConnection.onReactorFinal()` is never called.

And finally we are at the problem I want to point out: `closeReactorLatch` is never released and in `AmqpsIotHubConnection.close()` we await the full *hard coded* 20 seconds until it continues.

This in turn leads to the fact that the application code is notified about the disconnection event only after 20 seconds and all the while the client is not usable, but the application code does not know about it.

## Code sample exhibiting the issue

I use this simple piece of code that works as expected:

```
val log = LoggerFactory.getLogger("main")
    val connectionString = "..."

    val deviceClient = DeviceClient(connectionString, IotHubClientProtocol.AMQPS)
    deviceClient.setRetryPolicy { _, _ -> RetryDecision(false, 0) }
    deviceClient.setOption("SetSASTokenExpiryTime", 10L)

    val latch = CountDownLatch(1)
    deviceClient.registerConnectionStatusChangeCallback({ status,
                                                          statusChangeReason,
                                                          throwable,
                                                          callbackContext ->
        log.debug("status: $status, statusChangeReason: $statusChangeReason, throwable: $throwable")
        if (throwable != null) {
            latch.countDown()
        }
    }, deviceClient)

    deviceClient.open(false)

    latch.await()

    deviceClient.closeNow()
```

I then set a breakpoint in `SelectorImpl.update()`  and let it throw a `NullPointerException` manually via the debugger:

![Screenshot 2021-10-06 at 15 28 44](https://user-images.githubusercontent.com/70274334/136211813-a34e66dc-f690-49ad-994a-879f19d8a9c3.png)

## Console log of the issue

[...]

PR #1383, opened 2021-10-07
fix(iot-dev): Fix bug where AMQP layer would block a while on closing if the proton reactor thread crashed

timtay-microsoft:
#1379 

We had a basic assumption that onReactorFinal would always be called when proton-j's thread was finished, but if proton-j's thread encountered a runtime exception or some handler exception, the thread would exit prematurely and would never execute onReactorFinal. Because of that, we need to release some state latches when this happens so that the user's close thread doesn't block for 20 seconds as discussed in the issue above.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -1238,44 +1243,15 @@ private void executorServicesCleanup()
         }
     }
 
-    private static class ReactorRunner implements Callable<Object>
```
timtay-microsoft, line 1241: This change is a bit hard to follow, but basically I moved this class out to its own file so that this file didn't have other classes defined in it. I then moved all the logic from the IotHubReactor class into this class since IotHubReactor was an unnecessary layer to have here.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/IotHubReactor.java
```diff
@@ -1,39 +0,0 @@
...
-        this.reactor.start();
-
-        //noinspection StatementWithEmptyBody
-        while (this.reactor.process())
-        {
-            // The empty while loop is to ensure that reactor thread runs as long as it has messages to process
-        }
-        this.reactor.stop();
-        this.reactor.process();
-        this.reactor.free();
```
timtay-microsoft, line 32: This one function was the sole reason we had this class abstracting the Reactor. I moved this logic into the ReactorRunner class to cut down on the number of classes we have.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/IotHubReactor.java
```diff
@@ -1,39 +0,0 @@
...
-    }
-
-    public void free()
-    {
-        this.reactor.free();
```
timtay-microsoft, line 37: A perfect illustration of why this layer of abstraction was pretty pointless

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/ReactorRunner.java
```diff
@@ -0,0 +1,67 @@
...
+        {
+            TransportException transportException = new TransportException(e);
+
+            // unclassified exceptions are treated as retryable in ProtonJExceptionParser, so they should be treated
+            // the same way here. Exceptions caught here tend to be transient issues.
+            transportException.setRetryable(true);
+
+            // Notify the AMQP connection layer so it can tear down the reactor's resources that would usually
+            // get cleaned up in a graceful close of the reactor
+            this.reactorRunnerStateCallback.onReactorClosedUnexpectedly();
```
timtay-microsoft, line 56: This is the new stuff that actually fixes the bug. Just another state callback so that we can release the latches like the comment mentions.

jamdavi (approved):

PR #1384, opened 2021-10-08
fix(iot-serv): Fix two bugs with file upload notification receiving

timtay-microsoft:
- Fixed bug where subsequent calls to ```fileUploadNotificationReceiver.receive()``` would return the same notification if no notification was actually received
- Fixed bug where calls to ```fileUploadNotificationReceiver.receive()``` would result in the SDK completing all notifications received instead of just the notification that is returned to the user

Our existing receiver code would get a delivery from the service with 1 to many file upload notifications, but our SDK only returned one of them to the user after acknowledging all of them. To fix this, we will now limit the link credit on these receivers to 1 message. That way the service doesn't send more messages than we are prepared to return to the user.

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpFileUploadNotificationReceive.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpFileUploadNotificationReceive.java
@@ -189,7 +189,12 @@ public class AmqpFileUploadNotificationReceive implements AmqpFeedbackReceivedEv
      */
     public synchronized FileUploadNotification receive(long timeoutMs) throws IOException, InterruptedException
     {
-        if  (isOpen)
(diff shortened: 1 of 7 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/transport/amqps/AmqpFileUploadNotificationReceive.java
```diff
@@ -189,7 +189,12 @@ public synchronized void close()
      */
     public synchronized FileUploadNotification receive(long timeoutMs) throws IOException, InterruptedException
     {
-        if  (isOpen)
+        // This value is set in the reactor thread once we receive the file upload notification from the service over AMQP.
+        // It's important to set this to null since receive may be called multiple times in a row. Otherwise, a single
+        // notification could be returned to the user multiple times if nothing overwrote a previous value.
+        fileUploadNotification = null;
```
timtay-microsoft, line 195: This value not being reset per call was the cause of the first bug.

PR #1391, opened 2021-10-13
feat(iot-dev): Add support for x509 authentication over AMQPS_WS

timtay-microsoft:
Adds tests for that scenario, and for x509 auth over MQTT_WS since that is something we already claim support for

Also update AMQP dependencies to the latest since it has been a while

This feature doesn't work 100% of the time currently, and I am working with service engineers to figure out what is going wrong here. Leaving this PR as a draft until this is fixed.


timtay-microsoft:
Closing this as the support suffers randomly from a service side bug. We'll revisit this after service team has fixed this bug

PR #1394, opened 2021-10-14
feat(iot): Add missing moduleContent field to ConfigurationContent class

timtay-microsoft:
See https://github.com/Azure/azure-iot-sdk-csharp/issues/2185. Java was missing this field as well.


azabbasi (approved):
LGTM

PR #1408, opened 2021-11-04
feat(iot-dev): Add support for configuring keep-alive interval for amqps and mqtt

timtay-microsoft:


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientOptions.java
```diff
@@ -31,4 +33,47 @@
...
     public SSLContext sslContext;
+
+    /**
+     * Gets the keep alive interval in seconds. This value defines the
+     * maximum time interval between messages sent or received. It enables the
+     * client to detect if the server is no longer available, without having to wait
+     * for the TCP/IP timeout. The client will ensure that at least one message
+     * travels across the network within each keep alive period. In the absence of a
+     * data-related message during the time period, the client sends a very small
+     * "ping" message, which the server will acknowledge. The default value is 230 seconds.
```
azabbasi, line 44: Interesting, where does 230 come from?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -113,7 +113,9 @@
 
     private final boolean isMultiplexing;
 
-    public AmqpsIotHubConnection(DeviceClientConfig config, String transportUniqueIdentifier, boolean isMultiplexing)
```
azabbasi, line 116: did you remove the parameter because it can be inferred?

timtay-microsoft: Yes. ```isMultiplexing``` was only ever one value

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientOptions.java
timtay-microsoft: It's mostly a magic number the SDK team agreed on a few years back. I think the inspiration was to have it be just short of 4 minutes for some reason. This PR is not changing the default behavior here, though.

azabbasi (approved):

PR #1409, opened 2021-11-04
refactor(ado): Migrate to latest windows agents to avoid deprecation issues

timtay-microsoft:


vsts/sdl.yaml
```diff
@@ -2,15 +2,15 @@ name: $(BuildID)_$(BuildDefinitionName)_$(SourceBranchName)_$(Date:yyyyMMdd)$(Re
 resources:
 - repo: self
   clean: true
-phases:
```
timtay-microsoft, line 5: "phases" has been deprecated by ado and replaced by "jobs"

vsts/sdl.yaml
```diff
@@ -69,7 +69,7 @@ phases:
       FileDirPath: '$(Build.StagingDirectory)'
       TreatSignatureUpdateFailureAs: 'Warning'
       SignatureFreshness: 'UpToDate'
-      TreatStaleSignatureAs: 'Error'
```
timtay-microsoft, line 72: In the event that the antimalware executable fails to update its signature, we shouldn't fail the SDL run

vsts/sdl.yaml
```diff
@@ -21,6 +21,7 @@ phases:
     inputs:
       debugMode: false
       suppressionsFile: vsts/CredScanSuppressions.json
+      toolMajorVersion: V2
```
timtay-microsoft, line 24: We've been getting warning messages about this as well. v1 is the default and is being deprecated.

timtay-microsoft: We are already using v2 in the .NET analyzer scripts, FYI

jamdavi (approved):

vsts/sdl.yaml
azabbasi: What does it mean for the antimalware executable to fail to update its signature? 
Under what circumstances does that happen?

timtay-microsoft: > What does it mean for the antimalware executable to fail to update its signature? 
It means that whatever server this task requests the latest signature from fails to deliver it for whatever reason. Possibly due to server errors

> Under what circumstances does that happen?
We've only started seeing this recently, so I'd chalk it up as a server error for whatever service manages these signature update requests

azabbasi: The main reason I ask is to ensure that ignoring them is safe

azabbasi (approved):

vsts/sdl.yaml
timtay-microsoft: My understanding is that this is safe since the signature will get updated eventually even if it fails once or twice in a row. Would you prefer I leave the antimalware task alone, though?

azabbasi: Honestly, unless we fully know the impact of this signature getting updated, I would rather keep it as an error until we know more.

timtay-microsoft: Okay, I'll revert this particular change

azabbasi (approved):

PR #1410, opened 2021-11-08
fix AmqpsIotHubConnection changing order of sent messages

niklas-sparfeld-gcx:
Signed-off-by: Niklas Sparfeld <niklas.sparfeld@grandcentrix.net>

# Checklist
- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-java/blob/main/.github/CONTRIBUTING.md).

# Description of the problem

When a message to IoT Hub could not be sent *yet*, e.g. due to SendResult.SUBSCRIPTION_IN_PROGRESS, the message would have been appended to the `messagesToSend` for later.

This potentially leads to messages being sent in a different order than what the user intended:

1. start twin
2. update reported property to 1 -> queued
3. update reported property to 2 -> queued
4. retry update 1 -> queued
5. twin started
6. retry update 2 -> done
7. retry update 1 -> done

This results in the twin reported properties being updated not to the latest value.

# Description of the solution

Instead of re-queueing at the end of the queue, this commit implements a deque and requeues all messages that have not been sent at the front.

This approach is rather naive, but in early tests it fixes the problem for us.

As always I am very open for you to open a separate PR with *real* fixes ;) I am mostly using this PR in order to verify my findings and to communicate the issue more clearly to you guys.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -66,7 +66,7 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
     private static final int MAX_MESSAGES_TO_SEND_PER_CALLBACK = 1000; //Max number of queued messages to send per periodic sending task
 
     // States of outgoing messages, incoming messages, and outgoing subscriptions
-    private final Queue<Message> messagesToSend = new ConcurrentLinkedQueue<>();
+    private final Deque<Message> messagesToSend = new ConcurrentLinkedDeque<>();
     private String connectionId;
     private IotHubConnectionStatus state;
     private final String hostName;
@@ -595,7 +595,7 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
         // all message sending must be done from the proton-j thread that is exposed to this SDK through callbacks
         // such as onLinkFlow(), or onTimerTask()
         log.trace("Adding message to amqp message queue to be sent later ({})", message);
-        messagesToSend.add(message);
+        messagesToSend.addLast(message);
         return IotHubStatusCode.OK;
     }
 
@@ -876,7 +876,9 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
     private void sendQueuedMessages()
     {
         int messagesAttemptedToBeProcessed = 0;
-        Message message = messagesToSend.poll();
+        Message message = messagesToSend.pollFirst();
+        LinkedList<Message> unsendMessages = new LinkedList<>();
+
         while (message != null && messagesAttemptedToBeProcessed < MAX_MESSAGES_TO_SEND_PER_CALLBACK)
         {
             messagesAttemptedToBeProcessed++;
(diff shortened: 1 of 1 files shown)


timtay-microsoft:
It was a mistake for the AMQP layer to do this requeueing in the first place without consulting the retry policy that is housed in the IotHubTransport layer, so I've created #1411 to address that issue first. With this fix, if two queued messages both fail to send, they will both be requeued in the same order which should address the issue you've brought up here.

That said, the SDK really can't promise in-order delivery of each message that is queued in it. The service may reject some messages randomly due to throttling limits and accept others. As a result, you'll still have to worry about the out-of-order twin scenario you've brought up. If in-order delivery is important in your scenario (which it sounds like it is), then I'd consider not queueing the second twin message until you have confirmation that the first twin message has sent successfully.

With all that said, do you have any other concerns or thoughts you'd like to share @niklas-sparfeld-gcx ?

niklas-sparfeld-gcx:
You're right of course: If we really rely on ordered updates, we need to make sure about it ourselves. I would like to differentiate between two scenarios though:

1. "Normal" operation: Opening a twin and putting messages in, they should not be re-ordered just because internally the link has not yet been established - the user cannot see that.
2. "Erroneous" operation: When something goes wrong, users have to put in extra effort.

For our scenario it would be super helpful when scenario 1) works well. We can improve scenario 2) later.

---

That being said I would like to point out issues with your proposed solution in contrast to my code changes:

* The "re-queuing" performed by the retry policy has a jitter, right? One message might be re-scheduled after 205ms the next in 198ms, again leading to unnecessary out-of-order messages.
* *New* messages that arrive *after* the connection is established, *while* older messages are waiting to be re-tried will be send immediately, leading to unnecessary out-of-order messages:

1. Start twin
2. send message 1 -> retry in 200ms
3. twin started
4. send message 2 -> sent
5. retry message 1 after 200ms

For these reasons - from the *outside* perspective - I actually *liked* that the `SUBSCRIPTION_IN_PROGRESS`-case queues differently than error-cases.

sconvent:
Hi @timtay-microsoft,

I'm working with @niklas-sparfeld-gcx  on this issue.
Do you have an update on Niklas' comments?
In a nutshell: Can we rely on in-order processing when the device-twin-connection ist still being established? (Neglecting other potential reasons for out-of-order processing)
Or do we have to take care of that ourselves?

We tested the changes in PR #1411 , but still have some out-of-order messages. Probably due to the resons Niklas mentioned.

Either way it would be really helpful to have a clear decision where this has to be implemented (SDK or our code).

Thanks!

timtay-microsoft:
It took me a while, but I believe I got the v2 SDK into a better state with regards to this issue. The v2 SDK that we are releasing this week will no longer have any retry logic within the AMQP layer. Instead, any message that fails to send over AMQP will be sent back up to the client's retry policy to see if it should be resent. 

There was a comment at one point (from Niklas, I think?) about cases where a message is "expected to fail" to send over AMQP. The case applied when starting twin since two messages were queued (one to open twin links, one to get the initial twin) and the latter would always fail once or twice while waiting for the twin links to open. With the v2 redesign we did, we split up the twin operations into more discrete steps. In v2, you first subscribe to twin (opens twin links), then you can subscribe to desired properties and call getTwin whenever you want.

Would your team be willing to try out the v2 client when it is published and let me know if it satisfies what you needed here?

PR #1411, opened 2021-11-10
fix(iot-dev): Fix bug where amqp layer retried sending messages without consulting retry policy

timtay-microsoft:
Instead of requeueing messages locally, send them back up to the IotHubTransport layer so that it can consult the retry policy and be re-sent/cancelled accordingly.

Also clear the local queue of messages when the amqp connection closes so that the unsent messages can be handled by the IotHubTransport layer instead of just carrying them over between retry attempts.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -267,6 +267,7 @@ public class IotHubTransport implements IotHubListener
             {
                 log.warn("Exception thrown while calling the onUnknownMessageAcknowledged callback in onMessageSent", ex);
             }
+
             log.warn("A message was acknowledged by IoT Hub, but this client has no record of sending it ({})", message);
         }
     }
@@ -1418,7 +1419,7 @@ public class IotHubTransport implements IotHubListener
                 RetryDecision retryDecision = config.getRetryPolicy().getRetryDecision(packet.getCurrentRetryAttempt(), transportException);
                 if (retryDecision.shouldRetry())
                 {
-                    this.taskScheduler.schedule(new MessageRetryRunnable(this.waitingPacketsQueue, packet, this), retryDecision.getDuration(), MILLISECONDS);
+                    this.taskScheduler.schedule(new MessageRetryRunnable(this.waitingPacketsQueue, packet, this.sendThreadLock), retryDecision.getDuration(), MILLISECONDS);
                     return;
                 }
                 else
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -335,6 +335,19 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
         this.executorServicesCleanup();
     }
 
+    private void reportUnsentMessages()
+    {
+        for (Message unsentMessage : this.messagesToSend)
+        {
+            TransportException transportException = new TransportException("AMQP message failed to send because the connection was closed before it could be sent.");
+            transportException.setRetryable(true);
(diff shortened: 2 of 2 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -899,13 +913,16 @@ private void sendQueuedMessages()
                 AmqpsSessionHandler reconnectingDeviceSessionHandler = this.reconnectingDeviceSessionHandlers.get(message.getConnectionDeviceId());
                 if (reconnectingDeviceSessionHandler != null)
                 {
-                    log.trace("Amqp message failed to send because its AMQP session is currently reconnecting. Adding it back to messages to send queue ({})", message);
-                    messagesToSend.add(message);
```
timtay-microsoft, line 903: re-queueing the message that failed to send here skips consulting the retry policy. So instead, we need to kick the message back up a layer to IotHubTransport via the ```listener.onMessageSent()``` callback so that the retry policy gets to validate that a message should be resent.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -918,29 +935,27 @@ else if (sendResult == SendResult.SUBSCRIPTION_IN_PROGRESS)
...
+                TransportException transportException = new TransportException("Amqp message failed to send for unknown reasons. " + message);
+                transportException.setRetryable(true);
+                log.debug("Notifying transport layer that amqp message failed to send.", transportException);
+                this.listener.onMessageSent(message, message.getConnectionDeviceId(), transportException);
             }
-
-            message = messagesToSend.poll();
-        }
-
-        if (message != null)
```
timtay-microsoft, line 940: instead of doing
```
message = queue.poll();
while (message != null)
{
  // send message
  message = queue.poll();
}
queue.enqueue(message);
```

we'll do

```
while (queue.peek() != null)
{
  message = queue.poll();
  // send message
}
```

for simplicity and so that the ordering of messages isn't unnecessarily jumbled

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1418,7 +1419,7 @@ private void handleMessageException(IotHubTransportPacket packet, TransportExcep
                 RetryDecision retryDecision = config.getRetryPolicy().getRetryDecision(packet.getCurrentRetryAttempt(), transportException);
                 if (retryDecision.shouldRetry())
                 {
-                    this.taskScheduler.schedule(new MessageRetryRunnable(this.waitingPacketsQueue, packet, this), retryDecision.getDuration(), MILLISECONDS);
+                    this.taskScheduler.schedule(new MessageRetryRunnable(this.waitingPacketsQueue, packet, this.sendThreadLock), retryDecision.getDuration(), MILLISECONDS);
```
timtay-microsoft, line 1422: This fixes a bug where requeued messages didn't wake up the send thread. They'd have to wait for the next queued message to come along from the user first. We were always supposed to pass in the sendThreadLock here, but somehow ended up just passing in ```this```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -885,9 +899,9 @@ private void addProxyLayer(Transport transport, String hostName)
     private void sendQueuedMessages()
     {
         int messagesAttemptedToBeProcessed = 0;
-        Message message = messagesToSend.poll();
-        while (message != null && messagesAttemptedToBeProcessed < MAX_MESSAGES_TO_SEND_PER_CALLBACK)
+        while (messagesToSend.peek() != null && messagesAttemptedToBeProcessed < MAX_MESSAGES_TO_SEND_PER_CALLBACK)
         {
+            Message message = messagesToSend.poll();
```
jamdavi, line 904: Are we threaded/synchronized in anyway here? I like this order of execution but should we check to see if message != null before executing anything below?

Issue #1412, opened 2021-11-11
[Bug report] DeviceClient threads leak during reconnection events


sorinpenteleiciuc:
When device clients that connect over MQTT lose connection, a 2 news threads are created :
azure-iot-sdk-IotHubReceiveTas, azure-iot-sdk-IotHubSendTask

 
We're currently struggling with an issue that the MQTT connection gets lost after every 3 to 5 minutes.
This leads additionally to the issue that two new threads are started every time the connection gets lost which also leads to the issue that the JVM ran out of memory after a certain time.
 
These are the exceptions for the MQTT connection loss:
 
2021-11-08 23:58:09.666  59562328 [MQTT Rec: 999] WARN  c.m.a.s.i.d.transport.mqtt.Mqtt - Mqtt connection lost
org.eclipse.paho.client.mqttv3.MqttException: Connection lost
            at org.eclipse.paho.client.mqttv3.internal.CommsReceiver.run(CommsReceiver.java:190)
            at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: javax.net.ssl.SSLException: Connection reset
            at java.base/sun.security.ssl.Alert.createSSLException(Alert.java:127)
            at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:326)
            at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:269)
            at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:264)
            at java.base/sun.security.ssl.SSLSocketImpl.handleException(SSLSocketImpl.java:1576)
            at java.base/sun.security.ssl.SSLSocketImpl$AppInputStream.read(SSLSocketImpl.java:974)
            at java.base/sun.security.ssl.SSLSocketImpl$AppInputStream.read(SSLSocketImpl.java:865)
            at java.base/java.io.DataInputStream.readByte(DataInputStream.java:270)
            at org.eclipse.paho.client.mqttv3.internal.wire.MqttInputStream.readMqttWireMessage(MqttInputStream.java:92)
            at org.eclipse.paho.client.mqttv3.internal.CommsReceiver.run(CommsReceiver.java:137)
            ... 1 common frames omitted
            Suppressed: java.net.SocketException: Broken pipe (Write failed)
                        at java.base/java.net.SocketOutputStream.socketWrite0(Native Method)
                        at java.base/java.net.SocketOutputStream.socketWrite(SocketOutputStream.java:110)
[...]

PR #1413, opened 2021-11-11
fix(iot-dev): fix bug where reconnection events wouldn't fully close the send/receive thread pools before creating new thread pools

timtay-microsoft:
As a result, there would be lingering threads that would build up over time.

#1412

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
@@ -219,6 +219,11 @@ public final class DeviceIO implements IotHubConnectionStatusChangeCallback
      */
     private void startWorkerThreads()
     {
+        // while startWorkerThreads should never be called when threads are already active, it doesn't hurt to double
+        // check that any previous thread pools have been shut down.
+        stopWorkerThreads();
+
+        log.info("Starting worker threads");
         this.sendTask = new IotHubSendTask(this.transport);
         this.receiveTask = new IotHubReceiveTask(this.transport);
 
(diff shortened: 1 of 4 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -247,12 +252,16 @@ private void stopWorkerThreads()
     {
         if (this.sendTaskScheduler != null)
         {
-            this.sendTaskScheduler.shutdown();
```
timtay-microsoft, line 250: This was the main culprit. Even though we were calling ```shutdown()``` on the thread pool, the thread pool sometimes stuck around. Switching to ```shutdownNow()``` consistently gives us the termination of all threads in the thread pool like we want.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -219,6 +219,11 @@ void setMaxNumberOfMessagesSentPerSendThread(int maxNumberOfMessagesSentPerSendT
      */
     private void startWorkerThreads()
     {
+        // while startWorkerThreads should never be called when threads are already active, it doesn't hurt to double
+        // check that any previous thread pools have been shut down.
+        stopWorkerThreads();
```
timtay-microsoft, line 224: While not necessary, I figured it didn't hurt to add this check here just in case. When protecting against thread leaking issues, you can't be too careful

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -462,6 +471,14 @@ void registerMultiplexingConnectionStateCallback(IotHubConnectionStatusChangeCal
     @Override
     public void execute(IotHubConnectionStatus status, IotHubConnectionStatusChangeReason statusChangeReason, Throwable throwable, Object callbackContext)
     {
+        log.trace("DeviceIO notified of status {} with reason {}", status, statusChangeReason);
+
+        if (status == this.state)
+        {
+            // no change in status, so no need to start/stop worker threads.
+            return;
```
timtay-microsoft, line 479: Again, while not completely necessary, it helps to keep unusual code paths from causing issues. Typically the status updates only flip from DISCONNECTED/DISCONNECTED_RETRYING to CONNECTED. If a bug in the transport layer causes multiple CONNECTECT status updates to fire, then this layer won't needlessly close and re-open the worker thread pools.

azabbasi (approved):

Issue #1414, opened 2021-11-17
[Bug Report] BaseDevice / DeviceParser silently change connectionStateUpdatedTime from UTC to a Local Time string


derSchtefan:
# Context

- **Java runtime used:** 11
- **SDK version used:** iot-service-client:1.33.2

## Description of the issue
The constructor of BaseDevice in conjunction with DeviceParser causes the connectionStateUpdatedTime field to end up as a local time based string, without reason or indication. This is a problem because:
* All other fields in the SDK offer UTC 
* The way back to a UTC timestamp is awkward and fails if executed during DST change
* This field is used to order events and conditionally update fields in a database, local/UTC timestamp mismatches caused a bug in our solution.

Proposed solution: let the faulty String stay as it is (too much code might rely on that quirky behaviour), but expose the Date object from DeviceParser on BaseDevice.

## Code sample exhibiting the issue

At first `DeviceParser.java` behaves correctly
```java
191:  if (deviceParser.connectionStateUpdatedTimeString != null) {
192:      this.connectionStateUpdatedTimeString = deviceParser.connectionStateUpdatedTimeString;
193:      this.connectionStateUpdatedTime = ParserUtility.getDateTimeUtc(deviceParser.connectionStateUpdatedTimeString);
194:  }
```

With an example of the result being:
```
String connectionStateUpdatedTimeString = 2021-11-17T05:23:04.3511935Z
Date   connectionStateUpdatedTime       = Wed Nov 17 06:23:04 CET 2021  
```

The issue happens then in `BaseDevice.java`:
```java
459:  if (parser.getConnectionStateUpdatedTime() != null)
460:  {
461:      this.connectionStateUpdatedTime = ParserUtility.getDateStringFromDate(parser.getConnectionStateUpdatedTime());
462:  }
```

Resulting in 
```
String connectionStateUpdatedTime       = 2021-11-17T06:23:04
```

Notice how the date suddenly transformed from 5am to 6am, with no reason why it also suddenly has the fractional part truncated.

timtay-microsoft:
Yep, this looks like a bug. I'll try to get a fix checked in soon.

timtay-microsoft:
After some discussion internally, we're worried about this fix being a breaking change for customers. We're planning on bringing it to our upcoming 2.X.X versions of the clients though since we agree the SDK should not be doing this.

derSchtefan:
[...]

PR #1415, opened 2021-11-17
fix(iot-deps, iot-serv): Fix date time parser utility converting dates to local time when converting to strings

timtay-microsoft:
#1414

### DIFF
--- a/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ParserUtility.java
+++ b/deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ParserUtility.java
@@ -13,6 +13,9 @@ import com.google.gson.JsonPrimitive;
 import java.nio.charset.StandardCharsets;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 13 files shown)


deps/src/main/java/com/microsoft/azure/sdk/iot/deps/serializer/ParserUtility.java
```diff
@@ -30,7 +33,10 @@
 {
     private static final String DATEFORMAT = "yyyy-MM-dd'T'HH:mm:ss";
     private static final String OFFSETFORMAT = "yyyy-MM-dd'T'HH:mm:ssXXX";
-    private static final String TIMEZONE = "UTC";
+
+    final static DateTimeFormatter UTC_DATETIME_FORMATTER = DateTimeFormatter.ofPattern(DATEFORMAT);
```
timtay-microsoft, line 37: We will still be returning date time strings in the same format as before, just without the change in hours that happened due to using local time zones.

deps/src/test/java/com/microsoft/azure/sdk/iot/deps/serializer/ParserUtilityTest.java
```diff
@@ -679,23 +680,20 @@ public void mapToJsonElementEmptyMapSucceed() throws ClassNotFoundException
...
     {
         //act
-        ParserUtility.getDateStringFromDate(null);
+        ParserUtility.getUTCDateStringFromDate(null);
     }
 
-    //Tests_SRS_PARSER_UTILITY_34_043: [The provided date shall be converted into this format: "yyyy-MM-dd'T'HH:mm:ss".]
     @Test
-    public void getSimpleDateStringFromDateSuccess()
+    public void getSimpleDateStringFromDateWithTimezone()
```
timtay-microsoft, line 687: Now this unit test does string -> date -> string in order to verify that the string we get from the service is preserved

timtay-microsoft:
Closing this PR as we will only be bringing this fix to our upcoming v2 SDK since it could be taken as a breaking change by users

PR #1431, opened 2021-12-06
Fix the sync from GH issues to ADO

azabbasi:


azabbasi:
Will not fix the issue.

Issue #1429, opened 2021-12-06
[Bug Report] Azure IOT( iot-device-client ) sdk java (1.33.1) Device client gets stuck not receiving messages anymore.


prashantg077:
# Context

- **OS and version used:** (Windows 10, Ubuntu 18.04, etc. )
- **Java runtime used:** jdk 11
- **SDK version used:** 1.33.1
## Description of the issue
Scenario ---> Reli run of application (edge module ) with C2D messages . 
                      Standard S1 Tier
                      75 messages/min sent from an application deployed on Azure cloud.
                      runs for 3-4 hrs without any issues , all messages are received and processed. 
Issue -----> After a while say 4 hrs on run  , Device client gets stuck and stops processing any messages from IOT Hub. Although connection is still established with IoT hub but messages are not handled by device client. 

## Code sample exhibiting the issue
Device Client created using X509 cert based authentication
Device client is opened
Device client has a message callback where it receives and acknowledges the messages.

NOTE --- Upon restart also the messages are not received for an hour untill the messages in the IOT Hub queue are expired with default 1 hr TTL.

## Console log of the issue

Does not show any thing in the logs only the messages last processed is shown. No exceptions are shown.

jamdavi:
We have added more improvements in the latest release that should help with this issue and others.

https://github.com/Azure/azure-iot-sdk-java/releases/tag/2021-12-15

PR #1433, opened 2021-12-07
fix(client): Add logging around unhandled user exceptions and update messaging around the message callback

jamdavi:
While investigating for issue #1429 I could see there is some undocumented behavior around AMQP messaging when receiving C2D messages from IoT hub.

If there is an exception thrown from the callback it will cause the message to be stuck in the queue until it expires. In high volume scenarios this can cause the queue limit to be reached quite quickly and will cause the client to appear as if it's stuck.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubMessageResult.java
```diff
@@ -4,10 +4,36 @@
...
+ *     See <a href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-c2d#the-cloud-to-device-message-life-cycle"/>The cloud-to-device message life cycle</a>
+ *     for more details on the states and what it means for message delivery and retry.
+ * </p>
  */
 public enum IotHubMessageResult
 {
-    COMPLETE, ABANDON, REJECT
+    /**
+     * Instructs IoT hub to complete the message.
+     * <p>
```
azabbasi, line 18: missing end of sentence period.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -966,7 +966,7 @@ private void cancelPendingPackets()
 
     /**
      * If the provided received message has a saved callback, this function shall execute that callback and send the ack
-     * to the service
+     * provided by the callback to the service
```
timtay-microsoft, line 969:
```suggestion
     * chosen by the callback to the service
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -979,8 +979,23 @@ private void acknowledgeReceivedMessage(IotHubTransportMessage receivedMessage)
...
         if (messageCallback != null)
         {
-            log.debug("Executing callback for received message ({})", receivedMessage);
-            IotHubMessageResult result = messageCallback.execute(receivedMessage, messageCallbackContext);
+            // If a message callback throws an exception here the acknowledge will never be sent and this message will
+            // live in Iot hub until it expires.
+            IotHubMessageResult result = IotHubMessageResult.COMPLETE;
+            try
+            {
+                log.debug("Executing callback for received message ({})", receivedMessage);
```
timtay-microsoft, line 987: This could probably live as a trace level log. What do you think, though?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -979,8 +979,23 @@ private void acknowledgeReceivedMessage(IotHubTransportMessage receivedMessage)
...
+                result = messageCallback.execute(receivedMessage, messageCallbackContext);
+            }
+            catch (Throwable ex)
+            {
+                // We want to log this exception and bubble up to the transport
+                log.warn("Exception thrown while calling the message callback for received message {} in acknowledgeReceivedMessage. " +
+                        "This exception is preventing the completion of message delivery and can result in messages being" +
+                        "stuck in IoT hub until they expire. This can prevent the client from receiving futher messages.", receivedMessage);
+                log.warn("Exception: {} ", ex);
```
timtay-microsoft, line 996: You can do 

```
log.warn("This is some error for message {}.", receivedMessage, ex);
```

and it will log the stacktrace alongside the message here. That way we don't have two separate log statements here

timtay-microsoft (approved):
Looks good. Brownie points if you'd like to make this same change for the callback we extend for direct methods and twin updates, too.

PR #1435, opened 2021-12-08
refactor(iot-dev): Add more verbose logging around link credit in amqp links

timtay-microsoft:
We didn't have any logs that would keep us aware of the current link credit on either our sender or receiver links, and this state has come into question in a recent customer issue


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/LoggingFlowController.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.qpid.proton.engine.Receiver;
+
+/**
+ * Copy of {@link org.apache.qpid.proton.reactor.FlowController} but with some small edits for logging purposes.
+ *
+ * Use this FlowController to automatically flow link credit back to the service from this client's receiver links after
+ * each delivery on the receiver link.
+ */
+@Slf4j
+public class LoggingFlowController extends BaseHandler
```
jamdavi, line 19: Can we inherit from org.apache.qpid.proton.reactor.FlowController instead?

timtay-microsoft: The relevant method that is being overidden from FlowController is a private method, so no.

jamdavi (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsReceiverLinkHandler.java
```diff
@@ -118,13 +135,13 @@ public void onLinkRemoteClose(Event event)
         Link link = event.getLink();
         if (link.getLocalState() == EndpointState.ACTIVE)
         {
-            log.debug("{} receiver link with link correlation id {} was closed remotely unexpectedly", getLinkInstanceType(), this.linkCorrelationId);
```
andyk-ms, line 121: should we use trace here instead... line 160 has similar info but that is a trace instead of debug.

andyk-ms (approved):

azabbasi (approved):

PR #1437, opened 2021-12-09
fix(e2e): Add more sanity checks in Android build/test

timtay-microsoft:
Check more return status codes in the build phase since gradle builds appear to fail silently sometimes, and check for the build artifacts at the beginning of the test runs to cut down on confusion if the build failed silently

Also, reverted a few recent dependency upgrades since they were causing compatibility issues with Android API versions 24 and 25


pom.xml
```diff
@@ -47,6 +47,9 @@
 
         <!-- Properties to consolidate slf4j versions -->
         <slf4j-version>1.7.32</slf4j-version>
+        <azure-core-version>1.12.0</azure-core-version>
+        <lombok-version>1.18.8</lombok-version>
+        <azure-storage-blob-version>12.8.0</azure-storage-blob-version>
```
timtay-microsoft, line 52: We'll have to re-evaluate the recent upgrade we did on azure-core and azure-storage-blob since they break our existing Android support. This PR reverts the upgrade to them made by https://github.com/Azure/azure-iot-sdk-java/commit/5a590f59422529e09d56c0f9a44f2518c5acffd5

vsts/gradle_build.cmd
```diff
@@ -9,4 +9,15 @@ if "%TARGET_BRANCH%" == "$(System.PullRequest.TargetBranch)" (SET isPullRequestB
 cd %build-root%\iot-e2e-tests\android
 call gradle wrapper --gradle-version 6.8.3
 call gradlew :clean :app:clean :app:assembleDebug
+
+if %ERRORLEVEL% NEQ 0 (
+  echo "Gradle build failed with error code %ERRORLEVEL%"
+  exit 1
```
timtay-microsoft, line 15: Previously, the gradle build could fail, but the ADO task would be green because we weren't manually returning a failed status code here

jamdavi (approved):

andyk-ms (approved):

azabbasi (approved):

PR #1439, opened 2021-12-13
fix(iot-dev): Fix bug where a retry scheduled message wouldn't wake up worker thread

timtay-microsoft:
The worker thread would still be woken up by a newly queued message. It just wouldn't be woken up if the only newly queued message was a message being retried.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -1418,7 +1418,7 @@ public class IotHubTransport implements IotHubListener
                 RetryDecision retryDecision = config.getRetryPolicy().getRetryDecision(packet.getCurrentRetryAttempt(), transportException);
                 if (retryDecision.shouldRetry())
                 {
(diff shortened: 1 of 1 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1418,7 +1418,7 @@ private void handleMessageException(IotHubTransportPacket packet, TransportExcep
                 RetryDecision retryDecision = config.getRetryPolicy().getRetryDecision(packet.getCurrentRetryAttempt(), transportException);
                 if (retryDecision.shouldRetry())
                 {
-                    this.taskScheduler.schedule(new MessageRetryRunnable(this.waitingPacketsQueue, packet, this), retryDecision.getDuration(), MILLISECONDS);
```
timtay-microsoft, line 1421: Using ```this``` as the lock parameter led to the IotHubTransport instance being notified when the message to retry had be requeued rather than notifying the ```sendThreadLock``` that we use for keeping the worker thread active/asleep

timtay-microsoft: This previous parameter was just a mistake, not by design

jamdavi (approved):

andyk-ms (approved):

PR #1442, opened 2021-12-14
fix(iot-dev): Fix amqp message acknowledgements being sent from outside the reactor thread

timtay-microsoft:
proton-j reactor is not thread safe, so all operations (including sending acks) must be done from the reactor thread itself

See #1429


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -584,6 +586,7 @@ public void onTransportError(Event event)
     public void onTimerTask(Event event)
     {
         sendQueuedMessages();
+        sendQueuedAcknowledgements();
```
abhipsaMisra, line 589: does amqp have an ordering requirement that makes this change necessary?

timtay-microsoft: No. We can acknowledge messages out of order. This change is just to keep amqp operations in the reactor thread.

jamdavi: Thats good, the only other way to handle this is with a uber work queue that just takes things in order. But that might require a bit more refactoring than is necessary.

jamdavi (approved):
This looks good. After testing on my machine for 24 hours it's been stable. The customer who ran into this issue is doing the same.

abhipsaMisra (approved):

Issue #1465, opened 2022-02-08
[Technical Question] Sending a response for the method that was never invoked


DBY047:
Hi,

Regularly, on devices we deployed and on our QA lab, we see episodes of TransportException. By episode, I means the IotHubTransport prints the error below for 37 seconds (from 02-08 02:46:44.072 to 02:47:21.379). There is 369 exceptions like that one in that period of time. Does someone have and explanation for this and if there is something we can do to fix this? QA and clients ask lots of questions about these exceptions. They think messages might be missed... 

```
Encountered exception while sending message with correlation id 9ccdcde3-6055-4942-8be6-f900d142795f
com.microsoft.azure.sdk.iot.device.exceptions.TransportException: Sending a response for the method that was never invoked
	at com.microsoft.azure.sdk.iot.device.transport.mqtt.MqttDeviceMethod.throwMethodsTransportException(MqttDeviceMethod.java:200)
	at com.microsoft.azure.sdk.iot.device.transport.mqtt.MqttDeviceMethod.send(MqttDeviceMethod.java:117)
	at com.microsoft.azure.sdk.iot.device.transport.mqtt.MqttIotHubConnection.sendMessage(MqttIotHubConnection.java:362)
	at com.microsoft.azure.sdk.iot.device.transport.IotHubTransport.sendPacket(IotHubTransport.java:1448)
	at com.microsoft.azure.sdk.iot.device.transport.IotHubTransport.sendMessages(IotHubTransport.java:607)
	at com.microsoft.azure.sdk.iot.device.transport.IotHubSendTask.run(IotHubSendTask.java:53)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:279)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:152)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
	at java.lang.Thread.run(Thread.java:841)
Handling an exception from sending message: Attempt number 0
[...]

PR #1468, opened 2022-02-10
fix(iot-dev): Remove unnecessary tracking of direct method request Ids

timtay-microsoft:
method response messages should always be sent. Tracking them like this only allowed issues like #1465 to happen if the state is lost somehow

This state is already not tracked for methods in AMQP

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttDeviceMethod.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttDeviceMethod.java
@@ -21,7 +21,6 @@ public class MqttDeviceMethod extends Mqtt
 {
     private final String subscribeTopic;
     private final String responseTopic;
-    private final Map<String, DeviceOperations> requestMap = new HashMap<>();
     private boolean isStarted = false;
 
     private static final String POUND = "#";
@@ -60,11 +59,6 @@ public class MqttDeviceMethod extends Mqtt
     public void stop()
     {
         isStarted = false;
-
-        if (!requestMap.isEmpty())
-        {
-            log.trace("Pending {} responses to be sent to IotHub yet unsubscribed", requestMap.size());
-        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttDeviceMethod.java
```diff
@@ -21,7 +21,6 @@
 {
     private final String subscribeTopic;
     private final String responseTopic;
-    private final Map<String, DeviceOperations> requestMap = new HashMap<>();
```
timtay-microsoft, line 24: This map was from requestId (unique identifier of a method invocation) to an enum type of operation. Since this map lives in the context only of device methods, it only ever mapped to one enum value

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttDeviceMethod.java
```diff
@@ -105,18 +99,6 @@ public void send(final IotHubTransportMessage message) throws TransportException
                     throw new IllegalArgumentException("Request id cannot be null or empty");
                 }
 
-                if (requestMap.containsKey(message.getRequestId()))
-                {
-                    if (requestMap.remove(message.getRequestId()) != DeviceOperations.DEVICE_OPERATION_METHOD_RECEIVE_REQUEST)
-                    {
-                        throwMethodsTransportException("Mismatched request and response operation");
```
timtay-microsoft, line 112: Never happened since all the mapped value was always DEVICE_OPERATION_METHOD_RECEIVE_REQUEST

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttDeviceMethod.java
```diff
@@ -105,18 +99,6 @@ public void send(final IotHubTransportMessage message) throws TransportException
...
-                    }
-                }
-                else
-                {
-                    throwMethodsTransportException("Sending a response for the method that was never invoked");
```
timtay-microsoft, line 117: Not a good reason for the SDK to fail to send a method invocation response. We have all the information needed to send the response without this map's contents, so this was unnecessary to worry about

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttDeviceMethod.java
```diff
@@ -105,18 +99,6 @@ public void send(final IotHubTransportMessage message) throws TransportException
...
-                }
-
                 String topic = this.responseTopic + BACKSLASH +
```
timtay-microsoft, line 102: Note that neither this topic nor the message we send on the topic depends on anything saved in the map that is being deleted

drwill-ms (approved):

drwill-ms:
Are you doing the same for v2?

timtay-microsoft:
> Are you doing the same for v2?

Yep. I'll pull this into v2 from main once it is merged.

PR #1476, opened 2022-03-02
refactor(all): update dependencies, move to dependency management in main pom

timtay-microsoft:
The dependency management section in the base pom.xml defines which versions of each library will be used throughout the entire project. This does not mean that each library will take a dependency on each library defined in this dependencyManagement section though. For instance, our device client pom still needs to declare a dependency on our mqtt library. It just doesn't need to specify the version anymore since it is defined in our dependencyManagement in our base pom


pom.xml
```diff
@@ -44,12 +167,6 @@
         <dice-provider-emulator-version>1.1.1</dice-provider-emulator-version>
         <dice-provider-version>1.1.1</dice-provider-version>
         <x509-provider-version>1.1.6</x509-provider-version>
-
-        <!-- Properties to consolidate slf4j versions -->
```
abhipsaMisra, line 48: are these no longer used?

abhipsaMisra (approved):
Nice!

pom.xml
timtay-microsoft: Yep, the dependency management section now holds the "source of truth" on what version we are using.

Issue #1460, opened 2022-01-29
[Bug Report] IoT Device Client - MethodRequest incorrectly implements toJson, makes module method calls fail


derSchtefan:
# Context

- **SDK version used:** iot-device-client:1.34.1

## Description of the issue
Package `com.microsoft.azure.sdk.iot.device.edge`,  `MethodRequest.java`, lines 37, 45, 59, and 80.
`payload` is incorrectly defined as String instead of Object. The toJson method of this class is called during execution and GSON will then encode the String again as JSON string . This makes it impossible to execute any module method calls with the Module Java SDK, unless it uses no payload or is designed to work with this weird quirk. 

Any target module out there just will receive double encoded junk and won't work. A direct call it from IoT hub (or using another SDK for another language) will work correctly.

## Code sample exhibiting the issue
```java
var device = moduleClient.getConfig().getDeviceId();

var body = Map.of("param1", "value1"); // this would usually be new MyRequestModel(.....)
var bodyAsString = (new GsonBuilder()).create().toJson(body);

var request = new MethodRequest("someMethod", bodyAsString); // ctor parameter and payload field are both "String". 
var response = moduleClient.invokeMethod(device, module, request);
```

will send a request of 

```json
{
    "methodName": "someMethod",
    "payload": "{\"param1\":\"value1\"}"
}
```

instead of 
```json
{
    "methodName": "someMethod",
    "payload": {
        "param1": "value1"
    }
}
```

## Suggested change
Make the payload of type Object. This will allow GSON to correctly serialize the MethodRequest. Any existing code out there (if there even is any) will continue to work correctly (since a String input serializes to the same String).

## Suggested workaround
While Microsoft is working on this issue, if anybody else stumbles over this, create this class that extends from MethodRequest and pass it as parameter to `invokeMethod`. I tested this with several 3rd party target modules.

```java
public class CorrectMethodRequest extends MethodRequest {
    private final Map<String,Object> requestMap;

    public CorrectMethodRequest(String methodName, Object payloadObject) {
        // just so this does not look odd in  debugger
        super(methodName, null);

        var map = new HashMap<String, Object>();
        map.put("methodName", methodName);
        map.put("payload", payloadObject);
        this.requestMap = map;
    }


    public CorrectMethodRequest(String methodName, Object payloadObject, Integer responseTimeout, Integer connectionTimeout) {
        // just so this does not look odd in  debugger
        super(methodName, null, responseTimeout, connectionTimeout);

[...]

PR #1477, opened 2022-03-02
Update MethodRequest to take payload in Object type

brycewang-microsoft:
For GH issue #1460 . As changing the type of `payload` from **String** to **Object** could bring some breaking changes to customers who are using V1, we make it happen in V2 instead.


timtay-microsoft:
While this change should work for Edge modules invoking a method, I'd like to see an e2e test here that shows that a method invoked by a service client is received by a device and/or module client with the payload **not** string-ified like it used to. I suspect that this service client -> device/module client method invoke has the same problem that you are fixing for the module client -> module client method invoke.

drwill-ms (approved):

PR #1483, opened 2022-03-09
v2 Migration doc

timtay-microsoft:
Best viewed in [markdown format here](https://github.com/Azure/azure-iot-sdk-java/blob/timtay/migrationDoc/SDK%20v2%20migration%20guide.md)

### DIFF
--- a/SDK v2 migration guide.md
+++ b/SDK v2 migration guide.md
@@ -39,22 +39,22 @@ to migrate to v2 when they have the chance. For more details on LTS releases, se
 
 #### DeviceClient
 
-| Change needed? | V1 class#method  | Equivalent V2 class#method |
+| Changed? | V1 class#method  | Equivalent V2 class#method |
 |:---|:---|:---|
 | yes | DeviceClient#createFromSecurityProvider(String, String, SecurityProvider, IotHubClientProtocol); | new DeviceClient(String, String, SecurityProvider, IotHubClientProtocol);  |
 | yes | DeviceClient#createFromSecurityProvider(String, String, SecurityProvider, IotHubClientProtocol, ClientOptions); | new DeviceClient(String, String, SecurityProvider, IotHubClientProtocol, ClientOptions);  |
 | yes | DeviceClient#setMessageCallback(MessageCallback, Object); | DeviceClient#setMessageCallback(MessageCallback, Object);  |
 | yes | DeviceClient#open(); | DeviceClient#open(boolean);  |
-| yes | DeviceClient#open(boolean); | DeviceClient#open(boolean);  |
+| no | DeviceClient#open(boolean); | DeviceClient#open(boolean);  |
 | yes | DeviceClient#closeNow(); | DeviceClient#close();  |
-| yes | DeviceClient#close(); | DeviceClient#close();  |
-| yes | DeviceClient#getFileUploadSasUri(FileUploadSasUriRequest); | DeviceClient#getFileUploadSasUri(FileUploadSasUriRequest);  |
-| yes | DeviceClient#completeFileUpload(FileUploadCompletionNotification); | DeviceClient#completeFileUpload(FileUploadCompletionNotification);  |
+| no | DeviceClient#close(); | DeviceClient#close();  |
+| no | DeviceClient#getFileUploadSasUri(FileUploadSasUriRequest); | DeviceClient#getFileUploadSasUri(FileUploadSasUriRequest);  |
+| no | DeviceClient#completeFileUpload(FileUploadCompletionNotification); | DeviceClient#completeFileUpload(FileUploadCompletionNotification);  |
 | yes | DeviceClient#uploadToBlobAsync(String, InputStream, long, IotHubEventCallback, Object); | no equivalent method**  |
-| yes | DeviceClient#isMultiplexed(); | DeviceClient#isMultiplexed();  |
-| yes | DeviceClient#sendEventAsync(Message, IotHubEventCallback, Object); | DeviceClient#sendEventAsync(Message, IotHubEventCallback, Object);  |
+| no | DeviceClient#isMultiplexed(); | DeviceClient#isMultiplexed();  |
+| no | DeviceClient#sendEventAsync(Message, IotHubEventCallback, Object); | DeviceClient#sendEventAsync(Message, IotHubEventCallback, Object);  |
 | yes | DeviceClient#registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback, Object); | DeviceClient#setConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback, Object);  |
-| yes | DeviceClient#setRetryPolicy(RetryPolicy); | DeviceClient#setRetryPolicy(RetryPolicy);  |
+| no | DeviceClient#setRetryPolicy(RetryPolicy); | DeviceClient#setRetryPolicy(RetryPolicy);  |
 | yes | DeviceClient#getProductInfo(); | DeviceClient#getProductInfo();  |
 | yes | DeviceClient#subscribeToDeviceMethod(DeviceMethodCallback , Object, IotHubEventCallback, Object); | DeviceClient#subscribeToMethodsAsync(MethodCallback, Object, IotHubEventCallback, Object);  |
 | yes | DeviceClient#startDeviceTwin(IotHubEventCallback, Object, TwinPropertyCallback, Object); | DeviceClient#subscribeToDesiredPropertiesAsync(DesiredPropertiesSubscriptionCallback, Object, DesiredPropertiesCallback, Object);  |
--- a/SDK v2 migration guide.md
+++ b/SDK v2 migration guide.md
@@ -39,36 +39,36 @@ to migrate to v2 when they have the chance. For more details on LTS releases, se
 
 #### DeviceClient
 
-| V1 class#method  | Equivalent V2 class#method |
-|:---|:---|
-| DeviceClient#createFromSecurityProvider(String, String, SecurityProvider, IotHubClientProtocol); | new DeviceClient(String, String, SecurityProvider, IotHubClientProtocol);  |
-| DeviceClient#createFromSecurityProvider(String, String, SecurityProvider, IotHubClientProtocol, ClientOptions); | new DeviceClient(String, String, SecurityProvider, IotHubClientProtocol, ClientOptions);  |
-| DeviceClient#setMessageCallback(MessageCallback, Object); | DeviceClient#setMessageCallback(MessageCallback, Object);  |
-| DeviceClient#open(); | DeviceClient#open(boolean);  |
-| DeviceClient#open(boolean); | DeviceClient#open(boolean);  |
-| DeviceClient#closeNow(); | DeviceClient#close();  |
-| DeviceClient#close(); | DeviceClient#close();  |
-| DeviceClient#getFileUploadSasUri(FileUploadSasUriRequest); | DeviceClient#getFileUploadSasUri(FileUploadSasUriRequest);  |
-| DeviceClient#completeFileUpload(FileUploadCompletionNotification); | DeviceClient#completeFileUpload(FileUploadCompletionNotification);  |
-| DeviceClient#uploadToBlobAsync(String, InputStream, long, IotHubEventCallback, Object); | no equivalent method**  |
-| DeviceClient#isMultiplexed(); | DeviceClient#isMultiplexed();  |
-| DeviceClient#sendEventAsync(Message, IotHubEventCallback, Object); | DeviceClient#sendEventAsync(Message, IotHubEventCallback, Object);  |
-| DeviceClient#registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback, Object); | DeviceClient#setConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback, Object);  |
-| DeviceClient#setRetryPolicy(RetryPolicy); | DeviceClient#setRetryPolicy(RetryPolicy);  |
-| DeviceClient#getProductInfo(); | DeviceClient#getProductInfo();  |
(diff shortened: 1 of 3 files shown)


SDK v2 migration guide.md
```diff
@@ -34,7 +34,7 @@ We have released [one final LTS version](https://github.com/Azure/azure-iot-sdk-
 we will support like any other LTS release (security bug fixes, some non-security bug fixes as needed), but users are still encouraged
 to migrate to v2 when they have the chance.
 
-## Migrating Guide
+## Migration Guide
 
 ### IoT hub Device Client
```
drwill-ms, line 39: To match case elsewhere.

```suggestion
### IoT Hub Device Client
```

drwill-ms (approved):

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+  - ```DeviceClient``` and ```ModuleClient``` had unneccessarily different method names for the same operations (```deviceClient.startDeviceTwin(...)``` vs ```moduleClient.startTwin(...)```) that could be easily unified for consistency.
+  - ```DeviceClient``` and ```ModuleClient``` had many asynchronous methods whose naming did not reflect that they were asynchronous. This led to some users calling these methods as though they were synchronous.
+
+## What will happen to the v1 SDK
+
+We have released [one final LTS version](https://github.com/Azure/azure-iot-sdk-java/releases/tag/2022-03-04) of the v1 SDK that
+we will support like any other LTS release (security bug fixes, some non-security bug fixes as needed), but users are still encouraged
+to migrate to v2 when they have the chance.
+
+## Migration Guide
```
andyk-ms, line 37: Should we say these are "breaking changes" directly?  (after this list of equivalent methods, it said 'other breaking changes")

Do we also say explicitly that V2 do not provide backward compatibility to the V1 methods? (or this is already specified or understood?

andyk-ms (approved):
Minor suggestions provided, non-blocking issues.

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+## Table Of Contents
+
+ - [Why the v1 SDK is being replaced](#Why-the-v1-sdk-is-being-replaced)
+ - [What will happen to the v1 SDK](#What-will-happen-to-the-v1-sdk)
+ - [Migration Guide](#migration-guide)
+   - [IoT Hub Device Client](#iot-hub-device-client)
+   - [IoT Hub Service Client](#iot-hub-service-client)
+   - [DPS Device Client](#dps-device-client)
+   - [DPS Service Client](#dps-service-client)
+   - [IoT Hub Device Client](#iot-hub-device-client)
```
abhipsaMisra, line 16: it looks like we have this link twice (L12 and L16)

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+   - [Deps](#deps)
+ - [Frequently asked questions](#frequently-asked-questions)
+
+## Why the v1 SDK is being replaced
+
+There are a number of reasons why the Azure IoT SDK team chose to do a major version revision. Here are a few of the more important reasons:
+  - [Upcoming certificate changes](./upcoming_certificate_changes_readme.md) dictated that the SDK needed to stop pinning on a specific IoT Hub public certificate and start reading certificates from the device certificate store.
+  - Several 3rd party dependencies (Bouncycastle, JNR Unixsocket, Azure Storage SDK) were becoming harder to carry due to security concerns and they could only be removed by removing or alterring existing APIs.
+  - Many existing client classes (RegistryManager, DeviceTwin, DeviceMethod, ServiceClient, etc.) were confusingly named and contained methods that weren't always consistent with the client's assumed responsibilities.
+  - Many existing client's had a mix of standard constructors (```new DeviceClient(...)```) and static builder constructors (```DeviceClient.createFromSecurityProvider(...)```) that caused some confusion among users.
```
abhipsaMisra, line 27: ~client's~ clients

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+## Migration Guide
+
+### IoT Hub Device Client
+
+#### DeviceClient
+
+| V1 class#method  | Equivalent V2 class#method |
+|:---|:---|
+| DeviceClient#createFromSecurityProvider(String, String, SecurityProvider, IotHubClientProtocol); | nwe DeviceClient(String, String, SecurityProvider, IotHubClientProtocol);  |
```
abhipsaMisra, line 45: Question - I forget, are we including defaulting to mqtt as a part of the v2 changes?

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+| DeviceClient#sendReportedProperties(Set<Property>); | DeviceClient#updateReportedPropertiesAsync(TwinCollection, ReportedPropertiesCallback, Object);  |
+| DeviceClient#sendReportedProperties(Set<Property>, int); | DeviceClient#updateReportedPropertiesAsync(TwinCollection, ReportedPropertiesCallback, Object);  |
+| DeviceClient#sendReportedProperties(ReportedPropertiesParameters); | DeviceClient#updateReportedPropertiesAsync(TwinCollection, ReportedPropertiesUpdateCorrelatingMessageCallback, Object);  |
+| DeviceClient#sendReportedProperties(Set<Property>, Integer, CorrelatingMessageCallback, Object, IotHubEventCallback, Object); | DeviceClient#updateReportedPropertiesAsync(TwinCollection, ReportedPropertiesUpdateCorrelatingMessageCallback, Object);  |
+| DeviceClient#setOption(String, Object); | no equivalent method***  |
+| DeviceClient#setProxySettings(ProxySettings); | no equivalent method****  |
+
+** This method has been split into the three individual steps that this method used to take. See the file upload samples within this repo for an example of how to do file upload using these discrete steps.
```
abhipsaMisra, line 74: nit: we can add a link to the referenced sample

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+*** The options that were previously set in this method are now set at DeviceClient constructor time in the optional ClientOptions parameter.
+
+**** Proxy settings are now set at ModuleClient constructor time in the optional ClientOptions parameter,
```
abhipsaMisra, line 78: nit: ~ModuleClient~ DeviceClient

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+#### MultiplexingClient
+
+No API surface changes have been made to the MultiplexingClient class
+
+#### TransportClient
+
+This client has been removed in v2. It is replaced by the MultiplexingClient. See this repo's multiplexing client sample for more information.
```
abhipsaMisra, line 122: nit: link to the sample here as well?

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+*** The options that were previously set in this method are now set at DeviceClient constructor time in the optional ClientOptions parameter.
+
+**** Proxy settings are now set at ModuleClient constructor time in the optional ClientOptions parameter,
+
+#### ModuleClient
+
+| V1 class#method  | Equivalent V2 class#method |
+|:---|:---|
+| ModuleClient#createFromEnvironment(); | ModuleClient#createFromEnvironment(UnixDomainSocketChannel);  |
```
abhipsaMisra, line 84: Q - do you think it is worth calling out why the default constructor is now taking in the unix socket channel information? (here, and in other APIs as well)?

I think if someone looks at this document as the starting point, then it might be confusing to simply compare the APIs with not much additional information. However, if someone is coming to this document from code (which I think we expect most people to be doing), then it might not be as confusing since our doc comments would have called out the reasoning for the params being required.

Thought?

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+No API surface changes have been made to the MultiplexingClient class
+
+#### TransportClient
+
+This client has been removed in v2. It is replaced by the MultiplexingClient. See this repo's multiplexing client sample for more information.
+
+#### Other notable breaking changes
+
+- Trust certificates are read from the physical device's trusted root certification authorities certificate store rather than from source.
```
abhipsaMisra, line 126: I think seeing the APIs side-by-side with this description might be useful. Otherwise, people will need to look through the client to figure out which API we are talking about here.

What do you thik?

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+  - Users are expected to install the required public certificates into this certificate store if they are not present already.
+  - See [this document](./upcoming_certificate_changes_readme.md) for additional context on which certificates need to be installed.
+- DeviceClient and ModuleClient constructors that take public certificates and private keys as strings have been removed.
+  - Users must provide an instance of SSLContext that has their public certificates and private keys loaded into it instead.
+  - See [this sample](./device/iot-device-samples/send-event-x509) for the recommended way to create this SSLContext and how to construct your DeviceClient and/or ModuleClient with it.
+- deviceClient.uploadToBlobAsync() has been removed.
+  - Users can still use deviceClient.getFileUploadSasUri() to get a SAS URI that can be used with the Azure Storage SDK to upload the file.
+  - See [this sample](./device/iot-device-samples/file-upload-sample) for the recommended way to upload files.
+- The Bouncycastle, JNR Unixsocket, and Azure Storage SDK dependencies have been removed.
```
abhipsaMisra, line 135: Q - do we want to explain what these libraries did and what we are doing now, or atleast call out that we are now implementing <xyz> natively, so essentially there is no loss of functionality?

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+- Trust certificates are read from the physical device's trusted root certification authorities certificate store rather than from source.
+  - Users are expected to install the required public certificates into this certificate store if they are not present already
+  - See [this document](./upcoming_certificate_changes_readme.md) for additional context on which certificates need to be installed.
+- The Bouncycastle dependencies have been removed
+  - The Bouncycastle dependencies were used for some certificate parsing logic that has been removed from the SDK.
+- Reduced access levels to classes and methods that were never intended to be public where possible 
+- Removed service error code descriptions that the service would never return the error code for
+- Reduce default SAS token time to live from 1 year to 1 hour for security purposes
+- Removed unnecessary synchronization on service client APIs to allow for a single client to make multiple service APIs simultaneously
```
abhipsaMisra, line 267: ...make multiple service ~APIs~ _API calls_ simultaneously. ?

abhipsaMisra: Q - if we say "multiple", do you think people will ask to quantify that; i.e. how many calls can we have in parallel?
How about if we say "...to allow for a single client to make service API calls simultaneously."?

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+| FeedbackReceiver#receive(); | no equivalent method**  |
+| FeedbackReceiver#receive(int); | no equivalent method**  |
+
+** Your FileUploadNotificationProcessorClient and MessageFeedbackProcessorClient will start receiving as soon as the client is started
+
+
+#### Other notable breaking changes
+
+- Trust certificates are read from the physical device's trusted root certification authorities certificate store rather than from source.
+  - Users are expected to install the required public certificates into this certificate store if they are not present already
```
abhipsaMisra, line 260: nit: some of these statements are missing (.).

abhipsaMisra (approved):
👍🏽

SDK v2 migration guide.md
abhipsaMisra: Or if we can actually quantify and state it, even better :)

SDK v2 migration guide.md
timtay-microsoft: We don't have a default protocol for v2, no. Users always have to pick one explicitly.

SDK v2 migration guide.md
timtay-microsoft: We allow infinitely many calls in parallel, so I like your suggestion of  "...to allow for a single client to make service API calls simultaneously." here

SDK v2 migration guide.md
timtay-microsoft: Sure, good idea

SDK v2 migration guide.md
timtay-microsoft: Good point. I'll add an asterisk on these cases and explain more beneath the table about this particular change

SDK v2 migration guide.md
```diff
@@ -0,0 +1,399 @@
...
+we will support like any other LTS release (security bug fixes, some non-security bug fixes as needed), but users are still encouraged
+to migrate to v2 when they have the chance.
+
+## Migration Guide
+
+### IoT Hub Device Client
+
+#### DeviceClient
+
+| V1 class#method  | Equivalent V2 class#method |
```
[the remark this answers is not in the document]
timtay-microsoft: Good idea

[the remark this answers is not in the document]
timtay-microsoft: Actually, I think I'll add another column on the left that is just a flag saying if it changed from v1 to v2

SDK v2 migration guide.md
timtay-microsoft: The migration guide section contains mappings from v1 APIs to v2 APIs, but it also covers how some v1 APIs are still present in v2 so we can't call it a "breaking changes" section

I can add a note about there being no backwards compatibility from v2 to v1 since that isn't explicitly stated anywhere yet.

SDK v2 migration guide.md
timtay-microsoft: This is a tricky one since there is no API for loading certificates. We do that for the user when they construct their client and open a connection to Hub/DPS. I can expand a bit on this in the notes here and clarify that, for most users, no action is required for them here. After all, the IoT hub public certs are typically already installed on every device's cert store.

Issue #1478, opened 2022-03-03
[Bug Report] Memory Leak on client unregistration with MultiplexingClient


gmousset:
# Context

- **OS and version used:** Docker container based on alpine 3.15 (eclipse-temurin:17-alpine)
- **Java runtime used:** openjdk version "17.0.1" 2021-10-19
- **SDK version used:** iot-device-client:1.34.2

## Description of the issue
I work on a gateway for iot devices. Devices are connected to my server application and the application is connected to the IoT Hub through a multiplexed AMQP connection (MultiplexingClient).
Each time a device connects to the gateway, a DeviceClient is created and registered to the MultiplexingClient.
Each time a device disconnects, the DeviceClient is unregisters from the MultiplexingClient.
To send command to the devices through the IoT Hub I subscribe to deviceMethod.

Connections, disconnection, device-to-cloud and cloud-to-device messages, everything seems ok but, during stress test with a lot of connections/disconnections the application crash after some hours because of excessive memory consumption (the application container is managed by k8s and we set limits).

We also noticed than the more the application run, the more the response time to cloud-to-device request is bad: about 90 ms after the beginning more than 10s after few hours.

I profiled the application and it seems there is a memory leak. All objects are not released after device disconnection, and specifically objects on DeviceMethod.

I supposed there's a link with my increasingly bad response times.

Below the capture of the profiler after few hours:
- 749 living DeviceClient objects is more or less the actual devices count connected at this moment
- 15347 living objects for DeviceClientConfig and DeviceMethod callbacks seems the origin of my issues.

![image](https://user-images.githubusercontent.com/766129/156567355-a59221f8-00b3-4ced-8638-4f4998b7a41d.png)

Thanks for your support!

## Code sample exhibiting the issue

Here, a extract of my kotlin code where I register/unregister devices

```
    val deviceConnections = mutableMapOf<String, DeviceClient>()
    private val multiplexingClient = MultiplexingClient(hubProperties.hostName, IotHubClientProtocol.AMQPS)

    init {
        this.multiplexingClient.open()
    }
    
    /**
     * Open connection tu IoT Hub and reference it.
     */
    fun openConnection(
        deviceId: String,
        key: String,
        onConnectionClose: () -> Unit
    ) {
        // check if there's already a connection
        if (this.deviceConnections[deviceId] == null) {
            try {
                val deviceConnectionString = this.connectionStringUtils.deviceConnectionString(deviceId, key)
[...]

PR #1487, opened 2022-03-11
fix(iot-dev): Fix issue where multiplexed connections would leak heap memory for unregistered device clients

timtay-microsoft:
The proton-j reactor powering the multiplexed connection still had references to the various amqp handler objects used in our SDK even after the client using each of these handlers had been unregistered. By clearing the handlers from the reactor, the JVM is free to garbage collect these handlers

Also fix a few small state maps that weren't kept up to date whenever a client was unregistered from a muxed connection

#1478

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -878,6 +878,8 @@ public class IotHubTransport implements IotHubListener
                         throw new MultiplexingClientDeviceRegistrationTimeoutException("Timed out waiting for all device unregistrations to finish.");
                     }
                 }
+
+                this.deviceConnectionStates.remove(newlyUnregisteredConfig.getDeviceId());
             }
         }
     }
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -216,6 +216,7 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
         }
 
         clientConfigurations.remove(config);
+        deviceSessionsOpenedLatches.remove(config.getDeviceId());
     }
 
     public void open() throws TransportException
@@ -1197,6 +1198,7 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
         for (ClientConfiguration successfullyUnregisteredConfig : configsUnregisteredSuccessfully)
         {
             this.multiplexingClientsToUnregister.remove(successfullyUnregisteredConfig);
+            this.deviceSessionsOpenedLatches.remove(successfullyUnregisteredConfig.getDeviceId());
         }
 
         this.clientConfigurations.removeAll(configsUnregisteredSuccessfully);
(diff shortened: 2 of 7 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -878,6 +878,8 @@ public void unregisterMultiplexedDeviceClient(List<ClientConfiguration> configs,
                         throw new MultiplexingClientDeviceRegistrationTimeoutException("Timed out waiting for all device unregistrations to finish.");
                     }
                 }
+
+                this.deviceConnectionStates.remove(newlyUnregisteredConfig.getDeviceId());
```
timtay-microsoft, line 882: This was missing because otherwise the last connection state of a newly unregistered device would stick around indefinitely. Not the cause of the broader memory leak, though.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -216,6 +216,7 @@ public void unregisterMultiplexedDevice(ClientConfiguration config, boolean will
         }
 
         clientConfigurations.remove(config);
+        deviceSessionsOpenedLatches.remove(config.getDeviceId());
```
timtay-microsoft, line 219: Like the above, this state map was holding onto state that is no longer relevant once a device is unregistered.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
```diff
@@ -97,9 +97,6 @@
     private static final int FAULT_INJECTION_RECOVERY_TIMEOUT_MILLIS = 2 * 60 * 1000;
     private static final int FAULT_INJECTION_TIMEOUT_MILLIS = 60 * 1000;
     private static final int DEVICE_METHOD_SUBSCRIBE_TIMEOUT_MILLISECONDS = 60 * 1000;
-    private static final int TWIN_SUBSCRIBE_TIMEOUT_MILLIS = 60 * 1000;
-    private static final long MAXIMUM_TIME_TO_WAIT_FOR_DESIRED_PROPERTY_SUBSCRIPTION_ACKNOWLEDGEMENT = 500; // .5 seconds
-    private static final long MAXIMUM_TIME_TO_WAIT_FOR_REPORTED_PROPERTY_ACKNOWLEDGEMENT = 1000; // 1 second
```
timtay-microsoft, line 102: Unused

brycewang-microsoft (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
drwill-ms: Any chance this call could fail?

drwill-ms: Or threading issues with a device leaving multiplexing and then being added right away?

timtay-microsoft: We have some synchronization in place at the multiplexing client level that prevents simultaneous registration and unregistration, so we are safe in that regard.

drwill-ms (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
timtay-microsoft: > Any chance this call could fail?

No. If the key isn't present in the map, the method simply returns null. We don't care about the returned object here regardless.

PR #1490, opened 2022-03-14
fix(iot-dev): Fix bug where connection level issues caused multiplexed device clients to temporarily close

timtay-microsoft:
The device client would still reconnect after this event, but this temporary close would execute a connection status callback to notify the user that their client has reached a terminal state and their twin/method subscriptions would be erased unnecessarily.

Another symptom that came from this issue was that the IotHubTransport layer's retry logic would be executing both for the multiplexing client and for each of the registered multiplexed clients leading to some race condition caused NPEs. Now the IotHubTransport layer is only notified to start the multiplexing client level retry logic in cases like these.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
@@ -317,17 +317,7 @@ public final class AmqpsIotHubConnection extends BaseHandler implements IotHubTr
                         CountDownLatch deviceSessionOpenedCountDownLatch = this.deviceSessionsOpenedLatches.get(config.getDeviceId());
                         int timeout = config.getAmqpOpenDeviceSessionsTimeout();
 
-                        /*try
-                        {*/
-                            deviceSessionsOpenTimedOut = !deviceSessionOpenedCountDownLatch.await(timeout, TimeUnit.SECONDS);
-                        /*}
-                        catch (NullPointerException e)
-                        {
-                            // indicates that the device session was unregistered while opening the connection, likely due to
-                            // retry logic for both the mux level connection and a particular multiplexed device running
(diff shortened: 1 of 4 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -842,7 +856,11 @@ public void onSessionClosedUnexpectedly(ErrorCondition errorCondition, String de
         // If the session closes during an open call, need to decrement this latch so that the open call doesn't wait
         // for this session to open. The above call to onMultiplexedDeviceSessionRegistrationFailed will report
         // the relevant exception.
-        this.deviceSessionsOpenedLatches.get(deviceId).countDown();
+        CountDownLatch deviceSessionOpenedLatch = this.deviceSessionsOpenedLatches.get(deviceId);
+        if (deviceSessionOpenedLatch != null)
+        {
+            deviceSessionOpenedLatch.countDown();
```
timtay-microsoft, line 862: Adding a null check here just in case a client is unregistered from the multiplexing client during open somehow (this latch would be null in that case, and we wouldn't need to wait on it anymore)

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
```diff
@@ -1166,7 +1166,7 @@ public void multiplexedConnectionRecoversFromDeviceSessionDropsParallel() throws
     }
 
     // Open a multiplexed connection, send a fault injection message to drop the TCP connection, and ensure that the multiplexed
-    // connection recovers
+    // connection recovers and that no twin/method subscriptions were lost
```
timtay-microsoft, line 1169: Adding more checks to the existing TCP connection drop recovery test

PR #1491, opened 2022-03-15
refactor(iot-dev): Rework how reconnection is done

timtay-microsoft:
Previously, the amqp/mqtt layer would spawn a thread each time a disconnect occurred. For mux scenarios, this meant 1 thread spawned per device session lost and 1 thread spawned if the mux connection as a whole was lost. This didn't scale well and caused some interesting race conditions if a mux-client level retry happened at the same time as a device session retry.

With this change, now there is a dedicated thread that is woken up each time a disconnect happens that handles all disconnected connections and muxed device sessions.

The biggest benefit here is the scale of handling cases where many muxed device sessions are lost at once. Another big benefit is the abilitiy to abandon device session specific retry if the muxed connection is lost and instead reopen the whole connection. By not have so many retry threads floating around, we also cut down on race conditions that were prevalent when a muxed connection retry start happens shortly after a device session retry starts.

In refactoring this code, I have also fixed a bug where a multiplexed device session would erroneously report DISCONNECTED when it was closed as part of the broader multiplexing client's reconnection attempt.


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
```diff
@@ -1547,69 +1652,6 @@ public void disableDeviceAfterOpenAndAfterRegistration() throws Exception
         }
     }
 
-    // If you register a disabled device to a multiplexed connection that hasn't opened yet, the open call should succeed
-    // but the disabled device's session should drop shortly afterwards and the other devices on the multiplexed connection
-    // should be unaffected.
-    @ContinuousIntegrationTest
-    @Test
-    public void disableDeviceBeforeOpen() throws Exception
```
timtay-microsoft, line 1555: This test depended on inconsistent service behavior, so I'm removing it. Basically, it was random if opening a mux client with a disabled device would cause the client to fail on open, or fail on connection status callback for the disabled device. Since we can't predict how fast the service will respond on these device disabled cases, it doesn't make sense to have a test around it

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -37,11 +33,13 @@
     private final IotHubTransport transport;
     private final IotHubSendTask sendTask;
     private final IotHubReceiveTask receiveTask;
+    private final IotHubReconnectTask reconnectTask;
```
timtay-microsoft, line 36: Alongside the existing threads dedicated to sending and receiving messages, we'll now have a dedicate thread for responding to disconnection events

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubListener.java
```diff
@@ -60,8 +60,10 @@
      * @param connectionId the Id of the connection, used to identify which of possible many reconnection attempts
      *                     this event belongs to.
      * @param deviceId the Id of the device that the session belongs to
+     * @param shouldReconnect false if this session was lost because the user opted to unregister the client from
+     * the multiplexed connection.
      */
-    void onMultiplexedDeviceSessionLost(TransportException e, String connectionId, String deviceId);
+    void onMultiplexedDeviceSessionLost(TransportException e, String connectionId, String deviceId, boolean shouldReconnect);
```
timtay-microsoft, line 66: Part of the bug I'm solving here is tied to ambiguity around when a device session closes. This new flag makes it clear that the session was either lost because it was closed by the user, or because it was closed as a part of a broader reconnection

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubSendTask.java
```diff
@@ -21,7 +23,7 @@
     // as this SDK would otherwise periodically spawn new threads of this type that would do nothing. It is the IotHubTransport
     // layer's responsibility to notify this thread when a message is queued to be sent or when a callback is queued to be executed
     // so that this thread can handle it.
-    private final Object sendThreadLock;
+    private final Semaphore sendThreadSemaphore;
```
timtay-microsoft, line 26: Changing this to a semaphore to avoid any potential out-of-order problems where the lock is notified but no thread is waiting for it since this could cause deadlocks. With semaphores, the release can be called prior to anyone trying to acquire it, so the order of operations no longer matters

timtay-microsoft: This isn't as big of a deal in send/receive threads since typically users send/receive more than one message per device client, but for the new reconnect thread, this is very important since the disconnection event is much more rare.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1711,62 +1763,28 @@ private void addToWaitingQueue(IotHubTransportPacket packet)
...
-     * Sleep for a length of time without interruption
+     * Sleep for a length of time
      *
-     * @param sleepFor length of time to sleep for
-     * @param unit time unit associated with sleepFor
+     * @param sleepFor length of time to sleep for, in milliseconds
      */
-    @SuppressWarnings("SameParameterValue")
-    // The TimeUnit is currently always MilliSeconds, but this method can be used generically as well.
-    private static void sleepUninterruptibly(long sleepFor, TimeUnit unit)
```
timtay-microsoft, line 1742: Not sure why we wanted the retry logic to be uninterruptable, but now we do want it to be interruptable if the client is closing, so we can get rid of this method and replace it with a simple Thread.sleep call

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/ReconnectionNotifier.java
```diff
@@ -1,32 +0,0 @@
-/*
- *  Copyright (c) Microsoft. All rights reserved.
- *  Licensed under the MIT license. See LICENSE file in the project root for full license information.
- */
-
-package com.microsoft.azure.sdk.iot.device.transport;
-
-import com.microsoft.azure.sdk.iot.device.exceptions.TransportException;
-
-public final class ReconnectionNotifier
```
timtay-microsoft, line 10: The purpose of this class was to spawn a thread to notify the transport layer when a connection was lost and to handle the reconnection logic. Now that we have a dedicated thread for this, we can delete this class

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsIotHubConnection.java
```diff
@@ -72,9 +72,7 @@
     private final Set<ClientConfiguration> clientConfigurations;
     private IotHubListener listener;
     private TransportException savedException;
-    private boolean reconnectionScheduled = false;
     private final Object executorServiceLock = new Object();
-    private final Map<String, Boolean> reconnectionsScheduled = new ConcurrentHashMap<>();
```
timtay-microsoft, line 77: We used this map to keep track of which device sessions had already started reconnection so that we could not spawn two threads to reconnect the same device. Now, notifying the transport layer of a disconnection can be done multiple times without any negative impact so it doesn't make sense to track this anymore

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
```diff
@@ -1166,20 +1176,27 @@ public void multiplexedConnectionRecoversFromDeviceSessionDropsParallel() throws
     }
 
     // Open a multiplexed connection, send a fault injection message to drop the TCP connection, and ensure that the multiplexed
-    // connection recovers
+    // connection recovers and that no twin/method subscriptions were lost
```
timtay-microsoft, line 1179: Adding this additional check to the test since we didn't have coverage for this scenario already

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubListener.java
```diff
@@ -60,8 +60,10 @@
      * @param connectionId the Id of the connection, used to identify which of possible many reconnection attempts
      *                     this event belongs to.
      * @param deviceId the Id of the device that the session belongs to
+     * @param shouldReconnect false if this session was lost because the user opted to unregister the client from
```
drwill-ms, line 63: I think another way of saying this is whether the disconnect was desired by the user, or the consequence of an issue where we'd want to retry connection.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubReconnectTask.java
```diff
@@ -0,0 +1,73 @@
...
+            throw new IllegalArgumentException("Parameter 'transport' must not be null");
+        }
+
+        this.transport = transport;
+        this.reconnectThreadLock = this.transport.getReconnectThreadSemaphore();
+    }
+
+    public void run()
+    {
+        String threadName = this.transport.getDeviceClientUniqueIdentifier() + "-" + "Cxn" + transport.getTransportConnectionId() + "-" + THREAD_NAME;
```
drwill-ms, line 37: Line is getting a bit long. Break it up?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubReconnectTask.java
```diff
@@ -0,0 +1,73 @@
...
+@Slf4j
+public final class IotHubReconnectTask implements Runnable
+{
+    private static final String THREAD_NAME = "azure-iot-sdk-IotHubReconnectTask";
+    private final IotHubTransport transport;
+
+    // This lock is used to communicate state between this thread and the IoTHubTransport layer. This thread will
+    // wait until a disconnection event occurs in that layer before continuing. This means that if the transport layer
+    // has no connectivity problems, then this thread will do nothing and cost nothing.
+    private final Semaphore reconnectThreadLock;
```
drwill-ms, line 22: Rename to `reconnectThreadSemaphore`?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -157,38 +161,47 @@ public IotHubTransport(
...
+        for (MultiplexedDeviceState multiplexedDeviceState : this.multiplexedDeviceConnectionStates.values())
         {
-            return this.callbackPacketsQueue.size() > 0;
+            if (multiplexedDeviceState.getConnectionStatus() == IotHubConnectionStatus.DISCONNECTED_RETRYING)
+            {
+                return true;
+            }
         }
+
+        return this.connectionStatus == IotHubConnectionStatus.DISCONNECTED_RETRYING;
```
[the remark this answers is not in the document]
timtay-microsoft: True. Checking the broader connection state first would be faster in most cases. I'll change the order here

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -495,18 +510,150 @@ public void close(IotHubConnectionStatusChangeReason reason, Throwable cause)
...
+
+            RetryPolicy retryPolicy; // retry policy to be used for connection level retry, not device session specific retry
+            if (isMultiplexing)
+            {
+                retryPolicy = multiplexingRetryPolicy;
+            }
+            else
+            {
+                retryPolicy = this.getDefaultConfig().getRetryPolicy();
+            }
```
drwill-ms, line 543: Does Java have the turnery operator?

```suggestion
                retryPolicy = isMultiplexing
                    ? multiplexingRetryPolicy
                    : this.getDefaultConfig().getRetryPolicy();
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -495,18 +510,150 @@ public void close(IotHubConnectionStatusChangeReason reason, Throwable cause)
...
+     * @param deviceSessionToReconnect the deviceId of the device session to check on.
+     * @return true if the reconnection attempt has reached a terminal state (CONNECTED or DISCONNECTED), and false otherwise.
+     */
+    private boolean checkIfPreviousReconnectionAttemptFinished(String deviceSessionToReconnect)
+    {
+        MultiplexedDeviceState lastReconnectAttemptsDeviceSession = this.multiplexedDeviceConnectionStates.get(deviceSessionToReconnect);
+
+        if (lastReconnectAttemptsDeviceSession == null)
+        {
+            return true; // the device was unregistered during its reconnection, so it's reconnection attempts can stop
```
drwill-ms, line 597:
```suggestion
            return true; // the device was unregistered during its reconnection, so its reconnection attempts can stop
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -340,19 +353,21 @@ public void onMultiplexedDeviceSessionEstablished(String connectionId, String de
...
 
     @Override
-    public void onMultiplexedDeviceSessionLost(TransportException e, String connectionId, String deviceId)
+    public void onMultiplexedDeviceSessionLost(TransportException e, String connectionId, String deviceId, boolean shouldReconnect)
     {
         if (connectionId.equals(this.iotHubTransportConnection.getConnectionId()))
         {
             log.debug("The device session in the multiplexed connection to the IoT Hub has been lost for device {}", deviceId);
-            if (e == null)
+            if (e == null && !shouldReconnect)
```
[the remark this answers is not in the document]
timtay-microsoft: ```e``` being null isn't relevant anymore now that we have the ```shouldReconnect``` flag, so I can remove it as part of this if statement. Conceptually, what I want to happen here is to call DISCONNECTED only if ```shouldReconnect``` is false

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
timtay-microsoft: It does, yeah. I'll change this to use it

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1109,206 +1256,100 @@ private void openConnection() throws TransportException
         this.updateStatus(IotHubConnectionStatus.CONNECTED, IotHubConnectionStatusChangeReason.CONNECTION_OK, null);
     }
 
-    /**
-     * Attempts to reconnect. By the end of this call, the state of this object shall be either CONNECTED or DISCONNECTED
-     *
-     * @param transportException the exception that caused the disconnection
-     */
-    private void handleDisconnection(TransportException transportException)
+    //For reconnecting multiplexed devices only. Since this triggers asynchronous functions in the AMQP layer, there
```
drwill-ms, line 1259:
```suggestion
    // For reconnecting multiplexed devices only. Since this triggers asynchronous functions in the AMQP layer, there
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1109,206 +1256,100 @@ private void openConnection() throws TransportException
...
+    // is no guarantee that the reconnect worked just because the unregister/register calls return successfully.
+    // Still need to check the device connection status before you can report the device to be re-connected.
```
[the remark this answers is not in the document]
drwill-ms: Or just `connected`?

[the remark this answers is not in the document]
timtay-microsoft: I like "connected" here better, yeah

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
drwill-ms: Obviously, you've never cared for an infant. Everyone benefits from uninterrupted sleep!

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1382,7 +1423,7 @@ private void handleMessageException(IotHubTransportPacket packet, TransportExcep
                 RetryDecision retryDecision = config.getRetryPolicy().getRetryDecision(packet.getCurrentRetryAttempt(), transportException);
                 if (retryDecision.shouldRetry())
                 {
-                    this.taskScheduler.schedule(new MessageRetryRunnable(this.waitingPacketsQueue, packet, this.sendThreadLock), retryDecision.getDuration(), MILLISECONDS);
+                    this.taskScheduler.schedule(new MessageRetryRunnable(this.waitingPacketsQueue, packet, this.sendThreadSemaphore), retryDecision.getDuration(), MILLISECONDS);
```
drwill-ms, line 1426: line gettin' long

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubReconnectTask.java
```diff
@@ -0,0 +1,73 @@
...
+        Thread.currentThread().setName(threadName);
+
+        try
+        {
+            try
+            {
+                if (!transport.needsReconnect())
+                {
+                    // IotHubTransport layer will notify this thread once a disconnection event occurs. Until then, do nothing.
+                    this.reconnectThreadLock.acquire();
```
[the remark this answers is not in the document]
timtay-microsoft: You're right, yes. I'm using these a bit backwards compared to typical semaphore usage. I'll give a thorough explanation in the comments here

PR #1495, opened 2022-03-16
refactor(e2e): Mark a c2d test as standard tier only

timtay-microsoft:
Basic tier hubs do not support cloud to device telemetry, so we can't test it.

The other c2d tests were already marked this way, but this test was not. This test was failing in CI builds of basic tier hubs. The only reason this hasn't been a problem yet is that this test was skipped in PR builds.

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/serviceclient/MessagingClientTests.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/serviceclient/MessagingClientTests.java
@@ -366,6 +366,7 @@ public class MessagingClientTests extends IntegrationTest
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/serviceclient/MessagingClientTests.java
```diff
@@ -386,8 +387,6 @@ public void cloudToDeviceTelemetryWithTooLargeMessageThrows() throws IOException
 
         Device device = testDeviceIdentity.getDevice();
 
-        Device deviceGetBefore = registryClient.getDevice(device.getDeviceId());
```
timtay-microsoft, line 389: Unused return value, unnecessary call to getDevice

drwill-ms (approved):

PR #1497, opened 2022-03-17
fix(iot-dev): Fix a bug where updating a twin object in place didn't take

timtay-microsoft:
Also add two helper methods for deciphering IoT hub status codes. I'm reworking the device reconnection sample a bit and found these helper methods useful.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/twin/TwinProperties.java
```diff
@@ -215,12 +207,7 @@ JsonElement toJsonElementWithMetadata()
      */
     public TwinCollection getDesired()
     {
-        /* SRS_TWIN_PROPERTIES_21_009: [The getDesired shall return a TwinCollection with the stored desired property.] */
-        if (this.desired == null)
-        {
-            return null;
-        }
-        return new TwinCollection(this.desired);
```
timtay-microsoft, line 223: By returning a clone of the desired properties twin collection instead of the local reference, we could not update them in place. See the modified e2e test to demonstrate what used to not work

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/TwinCommon.java
```diff
@@ -229,11 +229,10 @@ public void testBasicTwinFlow(boolean subscribe) throws InterruptedException, IO
         // create some reported properties
         final String reportedPropertyKey = UUID.randomUUID().toString();
         final String reportedPropertyValue = UUID.randomUUID().toString();
-        TwinCollection reportedProperties = new TwinCollection();
-        reportedProperties.put(reportedPropertyKey, reportedPropertyValue);
+        twin.getReportedProperties().put(reportedPropertyKey, reportedPropertyValue);
```
timtay-microsoft, line 232: Without the bug fix, the ```twin.getReportedProperties().put(...)``` call wouldn't change the twin object in any way.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubStatusCode.java
```diff
@@ -123,4 +123,36 @@ public static IotHubStatusCode getIotHubStatusCode(int httpsStatus)
 
         return iotHubStatus;
     }
+
+    /**
+     * Returns true if this event callback signals that the asynchronous action was unsuccessful, but could be retried.
+     * Returns false if it was successful, or it was unsuccessful but should not be retried.
+     * @param statusCode The status code.
+     * @return true if this event callback signals that the asynchronous action was unsuccessful, but could be retried, and false otherwise
```
drwill-ms, line 131:
```suggestion
     * @return true if this event callback signals that the asynchronous action was unsuccessful, but could be retried, and false otherwise.
```

drwill-ms (approved):

PR #1498, opened 2022-03-17
break(iot-dev): Add twin version as part of reported properties response callback

timtay-microsoft:
Users need to know the next version of the twin that was returned by the service after a successful update to reported properties in order to set the version for the next reported property update request.

Also convert this version from String to int since it is always supposed to be a number


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTwin.java
```diff
@@ -363,20 +360,4 @@ else if (topic.length() > PATCH.length() && topic.startsWith(PATCH))
             return message;
         }
     }
-
-    @SuppressWarnings("SameParameterValue") // This method currently has a single caller (with a single value for "message"),
-    // but can be used to create a new TransportException with any message string.
-    private void throwDeviceTwinTransportException(String message) throws TransportException
```
timtay-microsoft, line 369: Unused functions

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/twin/ReportedPropertiesCallback.java
```diff
@@ -15,9 +15,10 @@
      * The callback that is executed when the reported properties update request has completed.
      * @param statusCode The service's status code for the request. If {@link IotHubStatusCode#OK} then the request was
      * processed succesfully.
+     * @param version The new version of the twin after this request.
```
timtay-microsoft, line 18:
```suggestion
     * @param version The new version of the twin's reported properties after this request.
```

drwill-ms:
Is this a breaking change in v1?

timtay-microsoft:
> Is this a breaking change in v1?

Yes. We'll need to get this in before we do the v2 release

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/twin/ReportedPropertiesUpdateResponse.java
```diff
@@ -0,0 +1,24 @@
...
+     * The status code returned by the service for this request. May indicate failure to send the reported properties.
+     */
+    @Getter
+    private final IotHubStatusCode statusCode;
+
+    /**
+     * The new version of the reported properties after a successful reported properties update.
+     */
+    @Getter
+    private final int version;
```
abhipsaMisra, line 23: We had to add a special note in the .NET sdk that edge module property update requests would not have the updated value in here, since edge performed the actual update offline. Would that be relevant for the java sdk as well?

abhipsaMisra (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/twin/ReportedPropertiesUpdateResponse.java
timtay-microsoft: Sounds like a good idea, yeah. Where is this note in the .NET SDK?

abhipsaMisra: https://github.com/Azure/azure-iot-sdk-csharp/blob/preview/iothub/device/src/ClientPropertiesUpdateResponse.cs#L28

timtay-microsoft:
> > Yes. We'll need to get this in before we do the v2 release
> 
> Oh, so `main` is now stages for the first v2 release, yes?

Yes

PR #1500, opened 2022-03-23
break(iot-dev, iot-serv): Allow versions to be null, fix version not …

timtay-microsoft:
…showing when service client gets twin

Nullable version means that users can opt out of caring about versions and just send property updates.

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/TwinCollection.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/TwinCollection.java
@@ -5,6 +5,7 @@ package com.microsoft.azure.sdk.iot.service.twin;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
 import com.microsoft.azure.sdk.iot.service.ParserUtility;
 
 import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/TwinCollection.java
```diff
@@ -329,8 +330,13 @@ private static void addMetadata(TwinCollection twinCollection, Map<? extends Str
      *
      * @return The {@code JsonElement} with the content of this class.
      */
-    public JsonElement toJsonElement() {
-        return ParserUtility.mapToJsonElement(this);
+    public JsonObject toJsonObject() {
+        JsonObject object = ParserUtility.mapToJsonElement(this).getAsJsonObject();
+        if (this.version != null)
+        {
+            object.add(VERSION_TAG, new JsonPrimitive(this.version));
```
brycewang-microsoft, line 337: How about changing to 
```suggestion
            object.addProperty(VERSION_TAG, this.version);
```
?

brycewang-microsoft (approved):
Leave a small comment but everything else looks good.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/TwinCollection.java
timtay-microsoft: I like that better, yeah. Good idea

drwill-ms (approved):

PR #1502, opened 2022-03-24
break(iot-dev): Rework thrown exceptions, add new device reconnection sample

timtay-microsoft:
All of DeviceClient, ModuleClient and MultiplexingClient will now share an exception class 

```java
public class IotHubClientException extends Exception
{
    @Getter
    boolean isRetryable;

    @Getter
    IotHubStatusCode iotHubStatusCode;
}
```

We'll still expose the IotHubStatusCode so users can catch for each specific error scenario that they are worried about and have a more general catch for scenarios they are less worried about. For instance:

```java
try
{
    deviceClient.sendEvent(new Message("some message"));
}
catch (IotHubClientException e)
{
    if (e.isRetryable())
    {
        // retry logic
    }
    else
    {
        if (e.getStatusCode() == MESSAGE_TOO_LARGE)
        {
            // break message into smaller chunks
        }
    }
}
```

In general, here is the before and after for non-open/close operations:

```java
// current sync pattern
public IotHubStatusCode sendEvent(Message message) throws InterruptedException, TimeoutException, IllegalStateException

// current async pattern
public void sendEventAsync(Message message, IotHubEventCallback callback, Object context) throws IllegalStateException 

interface IotHubEventCallback 
{
    void execute(IotHubStatusCode responseStatus, Object context);
}
```

```java
// proposed sync pattern
public void sendEvent(Message message) throws InterruptedException, IllegalStateException, IotHubClientException

// proposed async pattern
public void sendEventAsync(Message message, IotHubEventCallback callback, Object context) throws IllegalStateException

interface IotHubEventCallback 
{
    void execute(IotHubStatusCode responseStatus, IotHubClientException e, Object context);
}
```

I want to stop returning IotHubStatusCodes to the user and expecting them to write logic to interpret them, so instead of returning the status code in the sync methods, we'll throw the IotHubClientException derived from the status code instead. Similarly, we'll include that exception in the async callback alongside the status code so that users can ignore the status code and just check the exception.

As a part of this PR, I've included my take on the new "device reconnection sample" that demonstrates best practices for keeping the connection alive and how to retry on each particular operation. It leans heavily on the proposed exception structure. It may be easiest to review the exception structure changes by reviewing this sample

I also renamed a few callback interfaces to keep the ```subscribeToXYZ()``` methods consistent with each other

I also added the sent message as a parameter in the callback for when a message was acknowledged by the service

Once this code is signed off, I plan on deleting the current device reconnection sample and replacing it with the new sample I wrote. That way, any user that already had a link to our device reconnection sample will still have a link to our recommended patterns.


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/FileUpload.java
```diff
@@ -27,50 +30,60 @@
...
-        HttpsResponse responseMessage = httpsTransportManager.getFileUploadSasUri(message);
+        try
+        {
+            HttpsResponse responseMessage = httpsTransportManager.getFileUploadSasUri(message);
+            String responseMessagePayload = validateServiceStatusCode(responseMessage);
 
-        String responseMessagePayload = validateServiceStatusCode(responseMessage, "Failed to get the file upload SAS URI");
+            if (responseMessagePayload == null || responseMessagePayload.isEmpty())
+            {
+                throw new IOException("Sas URI response message had no payload");
```
drwill-ms, line 45:
```suggestion
                throw new IOException("SAS URI response message had no payload");
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
```diff
@@ -322,9 +322,15 @@ public static ModuleClient createFromEnvironment(UnixDomainSocketChannel unixDom
 
                 return new ModuleClient(iotHubAuthenticationProvider, protocol);
             }
-            catch (IOException | TransportException | HsmException | URISyntaxException | CertificateException | NoSuchAlgorithmException | KeyStoreException | KeyManagementException e)
+            catch (HsmException | URISyntaxException | CertificateException | NoSuchAlgorithmException | KeyStoreException | KeyManagementException e)
             {
-                throw new ModuleClientException(e);
+                // TODO
```
drwill-ms, line 327: todo?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubStatusCode.java
```diff
@@ -13,7 +14,7 @@
     OK,
     BAD_FORMAT,
     UNAUTHORIZED,
-    TOO_MANY_DEVICES,
```
timtay-microsoft, line 16: Incorrectly labeled when this enum was first created to represent 403 errors

device/iot-device-samples/device-connection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java
```diff
@@ -0,0 +1,415 @@
...
+
+public class DeviceClientManager implements DesiredPropertiesCallback, MethodCallback, IotHubEventCallback, MessageCallback, GetTwinCallback, ReportedPropertiesCallback
+{
+    // The client. Can be replaced with a module client for writing the equivalent code for a module.
+    private final DeviceClient deviceClient;
+
+    // The twin for this client. Stays up to date as reported properties are sent and desired properties are received.
+    private Twin twin;
+
+    // outgoing work queue of the client
```
drwill-ms, line 24:
```suggestion
    // Outgoing work queue of the client
```

drwill-ms: and elsewhere

device/iot-device-samples/device-connection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java
```diff
@@ -0,0 +1,415 @@
...
+                catch (IllegalStateException e)
+                {
+                    System.out.println("Client was closed while doing work. Re-opening client before doing more work");
+                    return;
+                }
+            }
+        }
+        catch (InterruptedException e)
+        {
+            System.out.println("connection thread interrupted, likely because the sample has ended");
```
drwill-ms, line 120:
```suggestion
            System.out.println("Connection thread interrupted, likely because the sample has ended");
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ModuleClient.java
timtay-microsoft: This PR will need some further work after we've agreed on the principles of what we want to do. I have some cases like this and some documentation to iron out still though

device/iot-device-samples/device-connection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java
```diff
@@ -0,0 +1,415 @@
...
+        }
+        else
+        {
+            // if any previous message failed to send, retry sending it before moving on to new messages
+            messageToSend = this.telemetryToResend.remove();
+        }
+
+        try
+        {
+            this.deviceClient.sendEventAsync(messageToSend, this, messageToSend);
```
drwill-ms, line 246: Why is `messageToSend` included twice? I'm guessing for the callback to be added to `telemetryToSend` on failure? Is that worth a code comment about this sample's strategy?

device/iot-device-samples/device-connection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java
```diff
@@ -0,0 +1,415 @@
...
+
+    // callback for when a telemetry message is sent
+    @Override
+    public void execute(IotHubStatusCode responseStatus, IotHubClientException e, Object callbackContext)
+    {
+        Message message = (Message) callbackContext;
+        if (e == null)
+        {
+            System.out.println("Successfully sent message with correlation id " + message.getCorrelationId());
+        }
```
drwill-ms, line 264: Instead of else ifs and elses, how about just return on 264 inside the brace?

device/iot-device-samples/device-connection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java
```diff
@@ -0,0 +1,415 @@
...
+        else if (e.isRetryable())
+        {
+            System.out.println("Failed to send message with correlation id " + message.getCorrelationId() + " due to retryable error with status code " + responseStatus.name() + ". Requeueing message.");
+            telemetryToResend.add(message);
+        }
+        else
+        {
+            System.out.println("Failed to send message with correlation id " + message.getCorrelationId() + " due to an unretryable error with status code " + responseStatus.name() + ". Discarding message as it can never be sent");
```
drwill-ms, line 272: Is this only encountered when the connection state reaches a terminal state?

drwill-ms, line 272:
```suggestion
            System.out.println("Failed to send message with correlation Id " + message.getCorrelationId() + " due to an unretryable error with status code " + responseStatus.name() + ". Discarding message as it can never be sent");
```

device/iot-device-samples/device-connection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java
```diff
@@ -0,0 +1,415 @@
...
+    {
+        System.out.println("Method " + methodName + " invoked on device.");
+        return new DirectMethodResponse(200, null);
+    }
+    // endregion
+
+    // region cloud to device messaging
+    // callback for when a cloud to device message is received by this device
+    @Override
+    public IotHubMessageResult execute(Message message, Object callbackContext)
```
drwill-ms, line 409: Can we find a better method name for this, perhaps something like `onC2dMessageReceived`?

device/iot-device-samples/device-connection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceConnectionSample.java
```diff
@@ -0,0 +1,119 @@
...
+        System.out.println("Received initial twin from twin service client");
+        System.out.println(twin);
+
+        while (deviceThread.isAlive())
+        {
+            Thread.sleep(10000);
+
+            try
+            {
+                directMethodsClient.invoke(d.getDeviceId(), "someMethodName");
```
drwill-ms, line 48: Is this sample making use of the service SDKs too? I think some code comments would be helpful in this file.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/IotHubClientException.java
```diff
@@ -0,0 +1,180 @@
...
+                || e instanceof AmqpConnectionThrottledException
+                || e instanceof AmqpConnectionForcedException
+                || e instanceof MqttServerUnavailableException)
+        {
+            return new IotHubClientThrottledException("Failed to open the client due to a throttling error", e, true);
+        }
+        else if (e instanceof AmqpInternalErrorException
+                || e instanceof InternalServerErrorException)
+        {
+            return new IotHubClientInternalServerErrorException("Failed to open the client due to the service encountering an internal service error", e, true);
```
azabbasi:
```suggestion
            return new IotHubClientInternalServerErrorException("Failed to open the client due to the service encountering an internal server error", e, true);
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/IotHubClientException.java
```diff
@@ -0,0 +1,180 @@
...
+    {
+        return fromStatusCode(statusCode, "");
+    }
+
+    public static IotHubClientException fromStatusCode(IotHubStatusCode statusCode, String errorMessage)
+    {
+        boolean isRetryable = IotHubStatusCode.isRetryable(statusCode);
+        switch (statusCode)
+        {
+            case SERVER_BUSY:
```
azabbasi: is this the equivelant of a 503? Is mapping it to a throttling exception the right thing to do here? seems like it can have its own exception type.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/IotHubClientException.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    public IotHubClientException(Throwable cause, boolean isRetryable)
+    {
+        super(cause);
+        this.isRetryable = isRetryable;
+    }
+
+    @Getter
+    protected boolean isRetryable;
```
azabbasi: Did we have this originally? how do you feel about calling this `isTransient`?

device/iot-device-samples/device-connection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java
timtay-microsoft: No. It is encountered when there is something fundamentally wrong with the message that was sent. For instance, if it was too large.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/IotHubClientException.java
timtay-microsoft: We did not have this originally, no. I like isRetryable a bit better if only because "transient" might require some users to lookup its definition

SDK v2 migration guide.md
```diff
@@ -56,13 +56,13 @@ to migrate to v2 when they have the chance. For more details on LTS releases, se
 | DeviceClient#registerConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback, Object); | yes | DeviceClient#setConnectionStatusChangeCallback(IotHubConnectionStatusChangeCallback, Object);  |
 | DeviceClient#setRetryPolicy(RetryPolicy); | no | DeviceClient#setRetryPolicy(RetryPolicy);  |
 | DeviceClient#getProductInfo(); | yes | DeviceClient#getProductInfo();  |
-| DeviceClient#subscribeToDeviceMethod(DeviceMethodCallback , Object, IotHubEventCallback, Object); | yes | DeviceClient#subscribeToMethodsAsync(MethodCallback, Object, IotHubEventCallback, Object);  |
```
timtay-microsoft, line 59: We had used IotHubEventCallback as the callback to notify users when most any action had completed, but it could be simplified a bit for cases like when a subscription has been acknowledged

SDK v2 migration guide.md
```diff
@@ -56,13 +56,13 @@ to migrate to v2 when they have the chance. For more details on LTS releases, se
...
-| DeviceClient#startDeviceTwin(IotHubEventCallback, Object, TwinPropertyCallback, Object); | yes | DeviceClient#subscribeToDesiredPropertiesAsync(DesiredPropertiesSubscriptionCallback, Object, DesiredPropertiesCallback, Object);  |
-| DeviceClient#startDeviceTwin(IotHubEventCallback, Object, TwinPropertiesCallback, Object); | yes | DeviceClient#subscribeToDesiredPropertiesAsync(DesiredPropertiesSubscriptionCallback, Object, DesiredPropertiesCallback, Object);  |
-| DeviceClient#startDeviceTwin(IotHubEventCallback, Object, PropertyCallBack<Type1, Type2>, Object); | yes | DeviceClient#subscribeToDesiredPropertiesAsync(DesiredPropertiesSubscriptionCallback, Object, DesiredPropertiesCallback, Object);  |
```
timtay-microsoft, line 62: Instead of defining a callback for when a desired properties subscription is acknowledged, I made it more generic so that it could be reused for method subscriptions as well.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -575,7 +580,7 @@ public void sendEventAsync(Message message, IotHubEventCallback callback, Object
      * @throws IllegalArgumentException if the message provided is {@code null}.
      * @throws IllegalStateException if the client has not been opened yet or is already closed.
      */
-    public void sendEventAsync(List<Message> messages, IotHubEventCallback callback, Object callbackContext)
```
timtay-microsoft, line 578: Making this plural since it sends a batch of messages

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubEventCallback.java
```diff
@@ -1,22 +0,0 @@
...
-package com.microsoft.azure.sdk.iot.device;
-
-/**
- * An interface for an IoT Hub event callback.
- *
- * Developers are expected to create an implementation of this interface,
- * and the transport will call {@link IotHubEventCallback#execute(IotHubStatusCode, Object)}
- * upon receiving a response from an IoT Hub.
- */
-public interface IotHubEventCallback
```
timtay-microsoft, line 13: Replaced with MessageSentCallback, MessagesSentCallback, and SubscriptionAcknowledgedCallback

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
```diff
@@ -106,43 +102,19 @@ void open(boolean withRetry) throws IOException
             {
                 this.transport.open(withRetry);
             }
-            catch (DeviceClientException e)
+            catch (TransportException e)
```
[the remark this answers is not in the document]
timtay-microsoft: Yes

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/IotHubClientException.java
```diff
@@ -0,0 +1,54 @@
...
+import com.microsoft.azure.sdk.iot.device.IotHubStatusCode;
+import lombok.Getter;
+
+/**
+ * Exception that can occur when using a
+ * {@link com.microsoft.azure.sdk.iot.device.DeviceClient}, {@link com.microsoft.azure.sdk.iot.device.ModuleClient},
+ * or a {@link com.microsoft.azure.sdk.iot.device.MultiplexingClient}. Each instance contains the status code of the error
+ * as well as the recommendation on whether the request that resulted in this exception is retryable or not.
+ */
+public class IotHubClientException extends Exception
```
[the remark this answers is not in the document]
timtay-microsoft: While their official guidance is for exceptions like these to be unchecked, the Azure SDK folks admitted that it isn't a very firm requirement. If we felt it was worthwhile to make an exception checked, then they wouldn't object. From our previous design meeting, it seemed like most of us were in favor of checked exceptions, so I went with that.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MessagesSentCallback.java
```diff
@@ -0,0 +1,27 @@
...
+ * by the service. Used by {@link InternalClient#sendEventsAsync(List, MessagesSentCallback, Object)}
+ */
+public interface MessagesSentCallback
+{
+    /**
+     * The callback that is executed once the messages associated with this callback has been acknowledged by the service.
+     *
+     * These messages are sent in bulk and are either all succeeded or all failed to send.
+     *
+     * @param sentMessages the messages that either were sent or failed to send.
```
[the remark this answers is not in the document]
timtay-microsoft: It's an all or nothing operation currently, so there is no need to distinguish

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/IotHubClientException.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public IotHubClientException(IotHubStatusCode statusCode, String errorMessage)
+    {
+        super(errorMessage);
+        this.isRetryable = IotHubStatusCode.isRetryable(statusCode);
+        this.statusCode = statusCode;
+    }
+
+    public IotHubClientException(IotHubStatusCode statusCode, Exception nestedException)
```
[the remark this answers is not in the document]
timtay-microsoft: Good catch

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubStatusCode.java
```diff
@@ -3,26 +3,91 @@
...
+     * The request failed because of network level issues.
+     */
+    IO_ERROR,
+
+    /**
+     * The request failed because it took longer than the device operation timeout as defined in {@link DeviceClient#setOperationTimeout(long)}.
+     */
+    DEVICE_OPERATION_TIMED_OUT;
 
     public static IotHubServiceException getConnectionStatusException(IotHubStatusCode statusCode, String statusDescription)
```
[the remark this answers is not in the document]
timtay-microsoft: This exception type already exists within the SDK and they are one of the sub-types of exceptions for TransportException. Generally speaking, they are supposed to encapsulate an http request failure such as sending a message that is too large. This is as opposed to a protocol level failure like "amqp:connection:forced" or something similar.

This method is just used to turn http level status codes into exceptions that can be thrown within the IotHubTransport or lower levels

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubStatusCode.java
```diff
@@ -3,26 +3,91 @@
...
+     */
     PRECONDITION_FAILED,
+
+    /**
+     * The request failed because the request payload exceeded IoT Hub's size limits.
+     */
     REQUEST_ENTITY_TOO_LARGE,
+
+    /**
+     * The request was rejected by the service because it is handling too many requests right now.
```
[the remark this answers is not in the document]
timtay-microsoft: I'll clarify this because my use of "it" was a bit too vague here. This is for when the service is handling too many requests, though. So it isn't a client failure

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/exceptions/IotHubClientException.java
azabbasi: fair enough

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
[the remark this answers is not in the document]
drwill-ms: Would the TransportException become the inner exception though? In that case, it isn't fully hidden from the user, just not the parent-level type they need to catch.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/FileUpload.java
```diff
@@ -27,50 +27,59 @@
...
-        String responseMessagePayload = validateServiceStatusCode(responseMessage, "Failed to get the file upload SAS URI");
+            if (responseMessagePayload == null || responseMessagePayload.isEmpty())
+            {
+                throw new IotHubClientException(IotHubStatusCode.ERROR, "SAS URI response message had no payload");
+            }
 
-        if (responseMessagePayload == null || responseMessagePayload.isEmpty())
+            return new FileUploadSasUriResponse(responseMessagePayload);
+        }
+        catch (IOException e)
```
drwill-ms, line 47: If there is any other kind of exception that could be thrown that we don't anticipate, should we have a general exception type we catch and still translate into an `IotHubClientException`?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -419,22 +420,28 @@ public ReportedPropertiesUpdateResponse updateReportedProperties(TwinCollection
...
      * Get the twin for this client. This client must have subscribed to desired properties before this method can be called.
      *
      * @return The twin for this client
-     * @throws TimeoutException if the service fails to acknowledge the getTwin request within the default timeout.
      * @throws InterruptedException if the operation is interrupted while waiting on the getTwin request to be acknowledged by the service.
      * @throws IllegalStateException if this client is not open or if this client has not subscribed to desired properties yet.
+     * @throws IotHubClientException if the request is rejected by the service for any reason of if the synchronous operation times out.
      */
-    public Twin getTwin() throws InterruptedException, TimeoutException, IllegalStateException
+    public Twin getTwin() throws InterruptedException, IllegalStateException, IotHubClientException
```
drwill-ms, line 444: Where did you land on having multiple types of exceptions we throw, keeping the native types. I'm guessing based on this you've kept them, yes?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
drwill-ms: Does it really send them in a batch?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubStatusCode.java
```diff
@@ -3,26 +3,91 @@
...
+     * The request failed because the provided credentials are out of date or incorrect.
+     */
     UNAUTHORIZED,
-    TOO_MANY_DEVICES,
-    HUB_OR_DEVICE_ID_NOT_FOUND,
+
+    /**
+     * The request failed because the quota for such operations has been exceeded. For file upload operations, this
+     * signifies that the maximum number of concurrent file upload operations are already happening. For telemetry operations,
+     * this signifies that the IoT Hub has reached its daily quota for the number of messages ingested.
```
drwill-ms, line 33:
```suggestion
     * this signifies that the IoT hub has reached its daily quota for the number of messages ingested.
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubStatusCode.java
```diff
@@ -3,26 +3,91 @@
...
+
+    /**
+     * The request failed because the quota for such operations has been exceeded. For file upload operations, this
+     * signifies that the maximum number of concurrent file upload operations are already happening. For telemetry operations,
+     * this signifies that the IoT Hub has reached its daily quota for messages ingested.
+     */
+    QUOTA_EXCEEDED,
+
+    /**
+     * The request failed because the resource the request targeted does not exist.
```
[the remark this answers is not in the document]
drwill-ms: `has` is not grammatically necessary

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubStatusCode.java
```diff
@@ -40,10 +105,10 @@ public static IotHubServiceException getConnectionStatusException(IotHubStatusCo
             case UNAUTHORIZED:
                 transportException = new UnauthorizedException(statusDescription);
                 break;
-            case TOO_MANY_DEVICES:
+            case QUOTA_EXCEEDED:
                 transportException = new TooManyDevicesException(statusDescription);
```
drwill-ms, line 109: `TOO_MANY_DEVICES` was renamed to `QUOTA_EXCEEDED` but the associated method is still `TooManyDevicesException()`.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubStatusCode.java
```diff
@@ -97,10 +162,10 @@ public static IotHubStatusCode getIotHubStatusCode(int httpsStatus)
                 iotHubStatus = UNAUTHORIZED;
                 break;
             case 403:
-                iotHubStatus = TOO_MANY_DEVICES;
+                iotHubStatus = QUOTA_EXCEEDED;
                 break;
```
drwill-ms, line 166: Can this be combined with the case on 108? If you can't double them up like one can in .NET, I'd at least put them next to each other.

drwill-ms: Same with case 412

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MessagesSentCallback.java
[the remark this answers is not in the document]
azabbasi: As we discussed offline, I wonder what it means for partial success scenarios on protocols other than HTTP.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/FileUpload.java
timtay-microsoft: For cases like "JVM out of memory" types of exceptions? That would be pretty atypical as far as I'm aware.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
timtay-microsoft: I smooshed the TimeoutException into the IotHubClientException, but otherwise it is the same as we went over in the design meeting. IllegalStateException and InterruptedException are special enough in Java to warrant their own catch statement.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
timtay-microsoft: For HTTP, yes. Just talked with Azad about this, and I'm considering making this API for HTTP only for now. Our current implementation is such that MQTT/AMQP clients just send each message individually, and this API isn't set up to handle partial successes/failures

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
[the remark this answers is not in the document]
timtay-microsoft: It would be the inner exception, yes. Users probably won't catch or even check the inner exception since an "AmqpConnectionForcedException" means nothing to them, but they will be there for debugging purposes when a user complains about the SDK throwing a particular IotHubClientException

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
drwill-ms: I'm very much against the "false" batch send we have in .NET for MQTT (and anywhere else). It is deceiving to the customer who thinks it is batch but really no better than a foreach they can implement.

And what happens if it fails at 61/100 messages? When an exception is thrown how will someone know which message to start resending and not get dupes on the service? I'd rather we just not include it than have partial/poor support for it.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubStatusCode.java
drwill-ms: Different cases that have the same action taken.

timtay-microsoft: Not unless we are consolidating these enum values then, no. A user's application may treat a SERVER_BUSY and a THROTTLED case the same way, but there's no reason for this particular class to do that, too. The ```isRetryable``` flag on the exception that holds this enum will do the heavy lifting on that.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/MessagesSentCallback.java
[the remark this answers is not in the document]
timtay-microsoft: I went ahead and made it so the batch messaging APIs are reserved for HTTP for now. If we add support for batch messaging over AMQP and/or MQTT, then this API design should suffice.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubStatusCode.java
drwill-ms: Never mind, bad read.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/TransportException.java
```diff
@@ -0,0 +1,135 @@
...
+{
+    protected boolean isRetryable = false;
+
+    public enum IotHubService
+    {
+        TWIN,
+        TELEMETRY,
+        METHODS,
+        FILE_UPLOAD,
+        NOT_APPLICABLE
```
drwill-ms, line 33: Should NA be the default?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/TransportException.java
```diff
@@ -0,0 +1,135 @@
...
+import com.microsoft.azure.sdk.iot.device.transport.mqtt.exceptions.MqttServerUnavailableException;
+import com.microsoft.azure.sdk.iot.device.transport.mqtt.exceptions.MqttUnauthorizedException;
+
+import static com.microsoft.azure.sdk.iot.device.transport.TransportException.IotHubService.NOT_APPLICABLE;
+
+public class TransportException extends Exception
+{
+    protected boolean isRetryable = false;
+
+    public enum IotHubService
```
drwill-ms, line 27: Is C2D not an option?

drwill-ms, line 27: The name seems too generic to me. What about IotHubOperation?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/TransportException.java
timtay-microsoft: This entire enum can just be deleted, actually. It isn't used anywhere

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/TransportException.java
timtay-microsoft: I'll delete this enum as per the above

timtay-microsoft: I'll delete this enum as per the above

drwill-ms (approved):

timtay-microsoft:
Using admin privileges to squash and merge this without horton passing since I need to make changes to horton like we've done a few times now.

azabbasi (commented):
LGTM

PR #1503, opened 2022-03-28
fix(iot-dev): Fix client not cancelling queued or sent packets for a device that was unregistered

timtay-microsoft:


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -270,7 +270,10 @@ public void onMessageSent(Message message, String deviceId, TransportException e
         }
         else
         {
-            log.warn("A message was acknowledged by IoT Hub, but this client has no record of sending it ({})", message);
+            // For instance, a message is sent by a multiplexed device client, the client is unregistered, and then the
+            // client receives the acknowledgement for that sent message. Safe to ignore since the user has opted to stop
+            // tracking it.
+            log.trace("A message was acknowledged by IoT Hub, but this client has already stopped tracking it ({})", message);
```
drwill-ms, line 276:
```suggestion
            log.trace("A message was acknowledged by IoT hub, but this client has already stopped tracking it ({})", message);
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1028,6 +1031,35 @@ public void unregisterMultiplexedDeviceClient(List<ClientConfiguration> configs,
                 this.multiplexedDeviceConnectionStates.remove(newlyUnregisteredConfig.getDeviceId());
             }
         }
+
+        // when a client is unregistered, remove all "waiting" and "in progress" messages that it had queued
```
drwill-ms, line 1035:
```suggestion
        // When a client is unregistered, remove all "waiting" and "in progress" messages that it had queued.
```

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1028,6 +1031,35 @@ public void unregisterMultiplexedDeviceClient(List<ClientConfiguration> configs,
...
+        for (IotHubTransportPacket waitingPacket : this.waitingPacketsQueue)
+        {
+            String deviceIdForMessage = waitingPacket.getDeviceId();
+            for (ClientConfiguration unregisteredConfig : configs)
+            {
+                if (unregisteredConfig.getDeviceId().equals(deviceIdForMessage))
+                {
+                    this.waitingPacketsQueue.remove(waitingPacket);
+                    this.addToCallbackQueue(waitingPacket);
```
drwill-ms, line 1044: Based on the comment, I don't understand why it is being removed from this queue.

timtay-microsoft: The unsent message here is being removed from the waitingPacketsQueue and added to the callbacks queue so that the user is notified via callback that the message was cancelled

timtay-microsoft: This is analogous to how we cancel all queued or in progress messages for a non-multiplexed client if the user closes that client

brycewang-microsoft (approved):

Issue #1507, opened 2022-03-31
RegistryManager getModulesOnDevice should work when module doesn't have auth type


ancaantochi:
# Context

- **OS and version used:** <VERSION> (Windows 10, Ubuntu 18.04, etc. )
- **Java runtime used:** <VERSION> (in a command prompt: java -version )
- **SDK version used:** com.microsoft.azure.sdk.iot:iot-service-client:1.33.1 or latest

## Description of the issue

Modules on edge devices get created without authentication type being set (via deployment or when an edge device is created automatically edgeHub and edgeAgent modules get created). It should work to call getModulesOnDevice and get the list of modules.

## Code sample exhibiting the issue

```
Exception in thread "main" java.lang.IllegalArgumentException: deviceParser must have an
authentication type assigned
        at com.microsoft.azure.sdk.iot.service.BaseDevice.<init>(BaseDevice.java:439)
        at com.microsoft.azure.sdk.iot.service.Module.<init>(Module.java:164)
        at com.microsoft.azure.sdk.iot.service.RegistryManager.getModulesOnDevice(RegistryManager.java:1166)
```

timtay-microsoft:
This bug has been fixed as of 2.0.1, so I'm closing this issue

PR #1508, opened 2022-03-31
fix(iot-service): Fix RegistryIdentity parser throwing if no authentication type present

timtay-microsoft:
#1507

There are some scenarios where the SDK receives some Module json to parse where there is no authentication type, so don't check for it on the SDK side.


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/registry/RegistryIdentity.java
```diff
@@ -275,20 +275,51 @@ else if (this.authentication.getAuthenticationType() == AuthenticationType.SAS)
      */
     RegistryIdentity(RegistryIdentityParser parser) throws IllegalArgumentException
     {
-        if (parser.getAuthenticationParser() == null || parser.getAuthenticationParser().getType() == null)
```
drwill-ms, line 278: I find that handling error cases immediately makes for a better reading experience. Otherwise one must scroll down a ways, try to figure out the indenting level, and find the relevant exception associated with these being null.

drwill-ms (changes requested):
Can you point out the specific change that matters here?

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/registry/RegistryIdentity.java
timtay-microsoft: I can reorder this a bit, but this is no longer an error case. It'll be something like

```java
//do all other parsing here, saving authentication type for last

if (parser.getAuthenticationParser() == null || parser.getAuthenticationParser().getType() == null)
{
    return;
}

// assign the authentication type here
```

drwill-ms (approved):

Issue #1523, opened 2022-04-07
[Bug Report] iot-device-client defines SLF4J binding as dependency


g0rill4z:
# Context

- **OS and version used:** macOS 12.3.1
- **Java runtime used:** OpenJDK 11.0.11
- **SDK version used:** 2.0.0 / 2.0.1

## Description of the issue
`iot-device-client` 2.0.0 and 2.0.1 define a SLF4J binding as dependency in its parent POM (`iot-device-client-parent`). I assume this is not intended as it can mess up the logging of applications using the dependency. Also `iot-sdk-java` includes the `log4j-slf4j-impl` twice in its dependency management.

brycewang-microsoft:
Hi @g0rill4z , thanks for bring this to us! I have a small [fix ](https://github.com/Azure/azure-iot-sdk-java/pull/1524) for it.

timtay-microsoft:
Fixed by 2.0.2 which has just been released, so closing this thread

PR #1524, opened 2022-04-07
refactor: Remove unnecessary SLF4J dependency

brycewang-microsoft:
For GH issue #1523


timtay-microsoft (approved):
Double check that you can still see SDK logs when you run the device samples, but this should be good.

PR #1529, opened 2022-04-12
Guava upgrade

timtay-microsoft:
Adding a proxy implementation to our code so that we can make some modifications so it is compatible with more recent guava versions. For the most part, this proxy implementation is equivalent to the dependency we previously took

All this proxy code now lives in the "tests/integration/com/microsoft/azure/sdk/iot/helpers/proxy" path.


iot-e2e-tests/common/pom.xml
```diff
@@ -84,24 +84,22 @@
             <artifactId>jmockit</artifactId>
             <version>1.24</version>
         </dependency>
-        <!-- https://mvnrepository.com/artifact/org.littleshoot/littleproxy -->
-        <dependency>
-            <groupId>org.littleshoot</groupId>
-            <artifactId>littleproxy</artifactId>
```
timtay-microsoft, line 90: This dependency isn't being maintained anymore, so I'm porting the code into our repo so that we can upgrade it as necessary. In particular, this library took a dependency on an older guava version and we couldn't simply force it to use a newer guava version without code changes

abhipsaMisra (approved):
Other than changing the guave dependency version, was there any other change required in the code that was copied over? Also, do we have the original sourcecode location documented somewhere so that we have a reference location to look at for any issues/updates etc. (if any)?

drwill-ms (approved):
Approved with the caveat that add more documentation about where we got the code from, why we copied it, what the license is, that we only use it for testing, etc.

PR #1531, opened 2022-04-13
refactor(sample): Refactor device twin sample

timtay-microsoft:
Previous implementation failed with "precondition failed" error because it wasn't using the reported properties version correctly

Converting the async twin API uses into sync API uses since we were just blocking on them anyways. Also adding a lot more comments explaining how to keep the versions of your reported properties up to date and what to do when they are out of date and your request fails with "precondition failed"

The module and device samples here are identical, so the same changes have been made to both

### DIFF
--- a/device/iot-device-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java
+++ b/device/iot-device-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java
@@ -154,7 +154,7 @@ public class DeviceTwinSample
 
         try
         {
-            System.out.println("Opening connection to IoT Hub");
+            System.out.println("Opening connection to IoT hub");
             client.open(false);
 
             System.out.println("Subscribing to desired properties");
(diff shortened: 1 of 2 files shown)


device/iot-device-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java
```diff
@@ -135,76 +154,52 @@ else if (protocolStr.equals("mqtt_ws"))
 
         try
         {
-            System.out.println("Open connection to IoT Hub.");
+            System.out.println("Opening connection to IoT Hub");
```
drwill-ms, line 157:
```suggestion
            System.out.println("Opening connection to IoT hub");
```

device/iot-device-samples/module-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ModuleTwinSample.java
```diff
@@ -11,12 +11,9 @@
 import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
-import java.util.concurrent.CountDownLatch;
-
-import static com.microsoft.azure.sdk.iot.device.IotHubStatusCode.OK;
 
 /**
- * Device Twin Sample for sending module twin updates to an IoT Hub. Default protocol is to use
+ * Twin Sample for sending module twin updates to an IoT Hub. Default protocol is to use
```
drwill-ms, line 16:
```suggestion
 * Twin Sample for sending module twin updates to an IoT hub. Default protocol is to use
```

drwill-ms (approved):

PR #1535, opened 2022-04-14
refactor(sample): Rework reconnection sample a bit

timtay-microsoft:
Previously, the sample wouldn't work well for users who either don't send telemetry/reported properties often or who don't do it at all. This is because the client manager would only attempt to reconnect once it tried to send telemetry/reported properties and failed because the client was closed. This meant that a client could be offline (and missing direct method calls and desired property updates) for a while.

With this rework, the sample will now immediately start reconnecting if a terminal disconnection event occurs regardless of if there is any outgoing telemetry and/or reported properties. The only drawback here is that we now have two threads in the sample which may be confusing. I've created a more robust readme to explain those threads though.

Also replacing System.out.print calls with log.debug calls so that this code can be more easily copied into a user's application.

### DIFF
--- a/device/iot-device-samples/device-reconnection-sample/readme.md
+++ b/device/iot-device-samples/device-reconnection-sample/readme.md
@@ -1,56 +1,63 @@
 # Device Reconnection Sample
 
-This sample code demonstrates the various connection status changes and connection status change reasons the Device Client can return, and how to handle them.
+This sample shows the best practices for handling a device (or module) client using a stateful protocol (AMQPS or MQTT) 
+over the course of its lifetime. This sample is intended to be a starting point for a user developing a production application 
+wherein their device client must handle unexpected connection loss events and transient errors such as throttling, internal server errors, and timeouts. 
 
-The device client exhibits the following connection status changes with reason:
+Included in this sample is the recommended patterns for sending reported properties and receiving desired properties, for
+sending telemetry, for receiving cloud to device messages, and for receiving direct methods.
 
-<table>
-  <tr>
-    <th> Connection Status </th>
-    <th> Change Reason </th>
-    <th> Ownership of connectivity </th>
-    <th> Comments </th>
-  </tr>
-  <tr>
-    <td> CONNECTED </td>
-    <td> CONNECTED_OK </td>
-    <td> SDK </td>
-    <td> SDK tries to remain connected to the service and can carry out all operation as normal </td>
-  </tr>
-  <tr>
-    <td rowspan="2"> DISCONNECTED_RETRYING </td>
-    <td> NO_NETWORK </td>
-    <td rowspan="2"> SDK </td>
-    <td rowspan="2"> When disconnection happens because of any reason (network failures, transient loss of connectivity etc.), SDK makes best attempt to connect back to IotHub. The RetryPolicy applied on the DeviceClient will be used to determine the count of reconnection attempts for <em>retriable</em> errors </td>
-  </tr>
-  <tr>
-    <td> COMMUNICATION_ERROR </td>
-  </tr>
-  <tr>
-    <td rowspan="5"> DISCONNECTED </td>
-    <td> CLIENT_CLOSE </td>
-    <td rowspan="5"> Application </td>
-    <td> This is state when SDK was asked to close the connection by application </td>
-  </tr>
-  <tr>
-    <td> BAD_CREDENTIAL </td>
-    <td> Supplied credential isn’t good for device to connect to service - Fix the supplied credentials before attempting to reconnect again </td>
-  </tr>
-  <tr>
-    <td> EXPIRED_SAS_TOKEN </td>
-    <td> Supplied SAS Token has expired - Fix the supplied SAS token before attempting to reconnect again </td>
-  </tr>
-  <tr>
-    <td> COMMUNICATION_ERROR </td>
-    <td> This is the state when SDK landed up in a non-retriable error during communication - Inspect the <em>throwable</em> supplied in the <code>IotHubConnectionStatusChangeCallback</code> to get the stacktrace of the exception </td>
-  </tr>
-  <tr>
-    <td> RETRY_EXPIRED </td>
-    <td> This is the state when SDK attempted maximum retries set by the application </td>
-  </tr>
-</table>
+## Sample Structure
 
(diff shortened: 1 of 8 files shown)


device/iot-device-samples/device-reconnection-sample/readme.md
```diff
@@ -1,56 +1,63 @@
...
-</table>
-
-For status `CLIENT_CLOSE`, `COMMUNICATION_ERROR` and `RETRY_EXPIRED`, the application can attempt to reconnect by closing and opening the client again eg. calling `deviceClient.close();` `deviceClient.open();` in a separate thread.
-
-We would like to highlight here that a separate thread should be used to call `open()`/ `closeNow()`. These calls should never be made from the same thread in the `IotHubConnectionStatusChangeCallback`. 
+This sample shows the best practices for handling a device (or module) client using a stateful protocol (AMQPS or MQTT) 
+over the course of its lifetime. This sample is intended to be a starting point for a user developing a production application 
+wherein their device client must handle unexpected connection loss events and transient errors such as throttling, internal server errors, and timeouts. 
 
+Included in this sample is the recommended patterns for sending reported properties and receiving desired properties, for
```
timtay-microsoft, line 7:
```suggestion
Included in this sample are the recommended patterns for sending reported properties and receiving desired properties, for
```

device/iot-device-samples/device-reconnection-sample/readme.md
```diff
@@ -1,56 +1,63 @@
...
+This sample shows the best practices for handling a device (or module) client using a stateful protocol (AMQPS or MQTT) 
+over the course of its lifetime. This sample is intended to be a starting point for a user developing a production application 
+wherein their device client must handle unexpected connection loss events and transient errors such as throttling, internal server errors, and timeouts. 
 
+Included in this sample are the recommended patterns for sending reported properties and receiving desired properties, for
+sending telemetry, for receiving cloud to device messages, and for receiving direct methods.
+
+## Sample Structure
+
+The sample uses two threads to accomplish these best practices. The threads, and their purpose is as follows:
```
drwill-ms, line 12:
```suggestion
The sample uses two threads to accomplish these best practices. The threads, and their purpose, is as follows:
```

device/iot-device-samples/device-reconnection-sample/readme.md
```diff
@@ -1,56 +1,63 @@
...
+- "Iot-Hub-Connection-Manager-Thread"
+  - This thread is responsible for ensuring that the device client is either connected to the service or is attempting to reconnect.
+  - When the device client is connected or when the device client's internal retry logic is running, this thread is dormant.
+  - Once a terminal disconnection event happens, this thread is woken up to attempt to reconnect indefinitely. 
+- "Iot-Hub-Worker-Thread"
+  - This thread queues outgoing work such as sending device to cloud telemetry and updating reported properties.
```
drwill-ms, line 18:
```suggestion
  - This thread queues outgoing work such as sending device-to-cloud telemetry and updating reported properties.
```

device/iot-device-samples/device-reconnection-sample/readme.md
```diff
@@ -1,56 +1,63 @@
 # Device Reconnection Sample
 
-This sample code demonstrates the various connection status changes and connection status change reasons the Device Client can return, and how to handle them.
-
-The device client exhibits the following connection status changes with reason:
-
-<table>
```
brycewang-microsoft, line 7: I am wondering why this table gets removed. I think listed connection status & explanation is still useful (especially to new users) as a reference even after this reworking.

timtay-microsoft: I feel that showing is better than telling here. The connection status callback that we define in this sample synthesizes this table into the actual code that a user should write. The contents of this table may make for good javadocs on the connection status change reason enums themselves, though

brycewang-microsoft:
Should we use only one indent [here ](https://github.com/Azure/azure-iot-sdk-java/blob/52f3738a5dd6e8a2941997127a6795f4a9c72f47/device/iot-device-samples/device-reconnection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java#L57)from parent line?

device/iot-device-samples/device-reconnection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java
```diff
@@ -247,18 +288,18 @@ public void onMessageSent(Message sentMessage, IotHubClientException e, Object c
...
         {
-            System.out.println("Successfully sent message with correlation Id " + sentMessage.getCorrelationId());
+            log.debug("Successfully sent message with correlation Id {}", sentMessage.getCorrelationId());
             return;
         }
 
         if (e.isRetryable())
         {
-            System.out.println("Failed to send message with correlation Id " + sentMessage.getCorrelationId() + " due to retryable error with status code " + e.getStatusCode().name() + ". Requeueing message.");
+            log.warn("Failed to send message with correlation Id {} due to retryable error with status code {}. Requeueing message.", sentMessage.getCorrelationId(), e.getStatusCode().name());
```
brycewang-microsoft, line 297: A single line seems a bit too long here

device/iot-device-samples/device-reconnection-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceClientManager.java
```diff
@@ -247,18 +288,18 @@ public void onMessageSent(Message sentMessage, IotHubClientException e, Object c
...
             telemetryToResend.add(sentMessage);
         }
         else
         {
-            System.out.println("Failed to send message with correlation Id " + sentMessage.getCorrelationId() + " due to an unretryable error with status code " + e.getStatusCode().name() + ". Discarding message as it can never be sent");
+            log.error("Failed to send message with correlation Id {} due to an unretryable error with status code {}. Discarding message as it can never be sent", sentMessage.getCorrelationId(), e.getStatusCode().name());
```
brycewang-microsoft, line 302: How about breaking this line to shorter ones? Same for L369, L378, L384

brycewang-microsoft (approved):
Approve with comments for some small changes.

drwill-ms (approved):

PR #1536, opened 2022-04-14
fix(iot-dev): Fix bug where twin/method subscription wasn't cleared when client is closed

timtay-microsoft:
As a result, this call would fail:

```java
deviceClient.open();
deviceClient.subscribeToMethods(...);
deviceClient.close();
deviceClient.open();
deviceClient.subscribeToMethods(...); // would always timeout since the callbacks pointed at the previous client's method state
```

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
@@ -168,6 +168,8 @@ public class InternalClient
     public void close()
     {
         this.deviceIO.close();
+        this.method = null;
+        this.twin = null;
     }
 
     /**


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
```diff
@@ -168,6 +168,8 @@ public void open(boolean withRetry) throws IotHubClientException
     public void close()
     {
         this.deviceIO.close();
+        this.method = null;
+        this.twin = null;
```
timtay-microsoft, line 172: These maintain the twin and methods callbacks for this client and are assigned when the client first subscribes to twin/methods. Because they weren't being unassigned, they carried over between open/close attempts carrying some out-of-date callbacks

drwill-ms: Could we get NPEs as a result?

timtay-microsoft: No. There is a null check before these are used

drwill-ms (approved):

azabbasi (approved):

PR #1540, opened 2022-04-22
Device side Plug and Play support

timtay-microsoft:
Still a work in progress.


PnpReview.md
```diff
@@ -0,0 +1,100 @@
+# PnP Java API Review
```
timtay-microsoft, line 1: TODO delete this since it is out of date

PR #1545, opened 2022-05-02
fix(iot-dev): fix bug where subsequent calls to subscribe to twin/methods times out

timtay-microsoft:
Subsequent calls would overwrite some callbacks but not all callbacks. This makes each call overwrite all the callbacks instead.

```java
deviceClient.open(false);
deviceClient.subscribeToMethods(...); // succeeds
deviceClient.subscribeToMethods(...); // times out, sends notifications to callbacks registered in the previous call
```java

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java
@@ -675,11 +675,7 @@ public class InternalClient
             throw new IllegalStateException("Open the client connection before using it");
         }
 
-        if (this.twin == null)
-        {
-            this.twin = new DeviceTwin(this);
-        }
-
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/InternalClient.java [hunk 1]
```diff
@@ -675,11 +675,7 @@ public void subscribeToDesiredPropertiesAsync(
             throw new IllegalStateException("Open the client connection before using it");
         }
 
-        if (this.twin == null)
```
drwill-ms, line 678: Null check not necessary?

drwill-ms: I see, so this is the override?

drwill-ms (approved):

brycewang-microsoft (approved):

PR #1556, opened 2022-05-27
Add Java 11 testing

timtay-microsoft:
[Java releases for reference](https://whichjdk.com/#releases)

The majority of our non-android users use Java 11 currently, so this PR adds PR and nightly testing of Java 11 to our gates. This PR also relegates Java 8 testing to nightly only in order to save time at the gate.

### DIFF
--- a/vsts/build_repo.ps1
+++ b/vsts/build_repo.ps1
@@ -12,11 +12,11 @@ else
     Write-Host "Pull request build detected"
 }
 
-if ($env:JAVA_VERSION == 8)
+if (($env:JAVA_VERSION).equals(("8"))
 {
     mvn -DRUN_PROVISIONING_TESTS="$Env:runProvisioningTests" -DRUN_DIGITAL_TESTS="$Env:runDigitalTwinTests" -DRUN_IOTHUB_TESTS="$Env:runIotHubTests" -DIS_PULL_REQUEST="$isPullRequestBuild" install -T 2C
 }
-elseif ($env:JAVA_VERSION == 11)
+elseif (($env:JAVA_VERSION).equals(("11"))
 {
     mvn install -DskipTests -T 2C
 
--- a/vsts/build_repo.ps1
+++ b/vsts/build_repo.ps1
@@ -20,7 +20,7 @@ elseif (($env:JAVA_VERSION).equals("11"))
 {
     mvn install -DskipTests -T 2C
 
-    cd "./azure-iot-sdk-java/iot-e2e-tests"
+    cd "./iot-e2e-tests"
 
     $env:JAVA_HOME=$env:JAVA_HOME_11_X64
 
--- a/vsts/build_repo.ps1
+++ b/vsts/build_repo.ps1
@@ -33,10 +33,11 @@ elseif (($env:JAVA_VERSION).equals("11"))
 {
     $env:JAVA_HOME=$env:JAVA_HOME_11_X64
 }
-elseif (($env:JAVA_VERSION).equals("17"))
-{
-    $env:JAVA_HOME=$env:JAVA_HOME_17_X64
-}
+# Leaving this commented out to make it easy to add Java 17 support later
+#elseif (($env:JAVA_VERSION).equals("17"))
+#{
+#    $env:JAVA_HOME=$env:JAVA_HOME_17_X64
+#}
 else
 {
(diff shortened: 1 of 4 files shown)


timtay-microsoft:
TODO update supported platforms doc

vsts/windowsLinuxAndAndroidBuildMatrixConfig.yaml
```diff
@@ -16,10 +16,18 @@ jobs:
   ### Windows ###
   - job: Windows
     timeoutInMinutes: 180
+    strategy:
+      maxParallel: 1
```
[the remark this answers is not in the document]
timtay-microsoft: I would like to get the SDL task down from its 30 minute time, but that is pretty orthogonal to this PR, so I'll take that on separately.

vsts/windowsLinuxAndAndroidBuildMatrixConfig.yaml
```diff
@@ -16,26 +16,34 @@ jobs:
...
+          JAVA_VERSION: 11
+        JDK 8: # only run Java 8 in nightly or CI builds.
+          JAVA_VERSION: 8
     pool:
       vmImage: windows-latest
     displayName: Windows
-    condition: succeeded()
     steps:
       - powershell: ./vsts/echo_versions.ps1
+        condition: or(eq(variables['JAVA_VERSION'], 11) , ne(variables['Build.Reason'], 'PullRequest'))
```
timtay-microsoft, line 31: I couldn't find a better way to skip Java 8 if it is a pull request since this is part of a matrix strategy. A bit frustrating to have copy-pasted this condition multiple places, so if anyone knows a better way, please let me know.

timtay-microsoft: I don't like the approach we did in .NET since it involves copying the entire job and just changing the matrix values to the pri 2 targets. I'll see if your second suggestion can be used here though

timtay-microsoft: Defining the check for PR or Java 11 as a variable does work, so I'll go with that approach since there is no copy pasting of jobs or of conditionals

timtay-microsoft: Or maybe that won't work either :/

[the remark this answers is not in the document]
drwill-ms: Oh, I think I get it now. Run it if: 1) it is Java11 and if not in a PR. Makes sense.

drwill-ms (approved):

vsts/windowsLinuxAndAndroidBuildMatrixConfig.yaml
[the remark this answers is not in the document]
timtay-microsoft: No. 11 will always run. 8 will only run for non-pr builds.

to simplify the conditional for readability:
condition: (is Java 11) || (is not pr build)

vsts/build_repo.ps1
```diff
@@ -12,4 +12,38 @@ else
...
+if (!($env:JAVA_VERSION).equals("8"))
+{
+    #Build the repo with Java 8 as configured in the project itself
+    mvn install -DskipTests -T 2C
+
+    # go into the e2e tests folder so the next mvn install only runs the e2e tests instead of running e2e tests + unit tests.
+    cd "./iot-e2e-tests"
+}
+
+# Set the Java home equal to the Java version under test. Currently, Azure Devops hosted agents only support Java 8, 11 and 17 currently
```
abhipsaMisra, line 26: extremely small nit: can be reworded to remove overuse of "currently".

vsts/windowsLinuxAndAndroidBuildMatrixConfig.yaml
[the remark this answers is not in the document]
abhipsaMisra: nit: I think it would be useful to put this as a comment in the yaml

> condition: (is Java 11) || (is not pr build)

vsts/build_repo.ps1
timtay-microsoft:
```suggestion
# Set the Java home equal to the Java version under test. Currently, Azure Devops hosted agents only support Java 8, 11 and 17
```

PR #1557, opened 2022-06-02
refactor(all): update Gson dependency in main pom

brycewang-microsoft:
Not exactly for Gson-related [dependabot alerts](https://github.com/Azure/azure-iot-sdk-java/security/dependabot) here, as we are using patched version 2.8.9 already. The security alerts were triggered as dependabot could not determine the currently installed version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
             <dependency>
                 <groupId>com.google.code.gson</groupId>
                 <artifactId>gson</artifactId>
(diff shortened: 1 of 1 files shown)


abhipsaMisra (approved):
> The security alerts were triggered as dependabot could not determine the currently installed version.

Do we know how we can get dependabot to pick up the correct version?

drwill-ms (approved):

Issue #1563, opened 2022-06-28
[Bug Report] Thread leak of IotHubReconnectTask


janusz-com2m:
# Context

- **OS and version used:** Docker Image eclipse-temurin:11-alpine (Linux f74e9f6057d9 4.15.0-184-generic #194-Ubuntu SMP Thu Jun 2 18:54:48 UTC 2022 x86_64 Linux)
- **Java runtime used:** openjdk version "11.0.15" 2022-04-19
OpenJDK Runtime Environment Temurin-11.0.15+10 (build 11.0.15+10)
OpenJDK 64-Bit Server VM Temurin-11.0.15+10 (build 11.0.15+10, mixed mode)
- **SDK version used:** 
- iot-device-client: 2.0.3
- provisioning-device-client: 2.0.0

## Description of the issue
We are using the sdk to connect multiple devices within an IoT-Edge module to IoT Central. The DPS is used for the device provisioning. We are seeing an increasing threadcount over the time in the Java Process, until the underlying VM is not able to spawn any more threads.
I have retrieved a threaddump from the process after it was running for ~20 hours and I could find 18 Instances of the IotHubReconnectTask for one device in the threaddump, this behaviour occurs for other devices to. So it looks like that the IotHubReconnectTask-Threads aren't cleaned up correctly.

## Code sample exhibiting the issue
@Service
public class AzureIotEdgeService {

	private static final String TELEMETRY_COMPONENT_NAME = "$.sub";
	private static final String ENCODING_UTF_8 = StandardCharsets.UTF_8.name();
	private static final String CONTENT_APPLICATION_JSON = "application/json";

	private final Logger logger;
	private final AzureDeviceProvisioningService azureDeviceProvisioningService;

	private Map<String, DeviceClient> deviceClients = new ConcurrentHashMap<>();

	public AzureIotEdgeService(
			Logger logger,
			AzureDeviceProvisioningService azureDeviceProvisioningService
	) {
		this.logger = logger;
		this.azureDeviceProvisioningService = azureDeviceProvisioningService;
	}

	@Async
	public void sendMessageToIotHubForDevice(String topic, String payload, Date timestamp) {
		String componentId = topic.split("/")[3].replace("@", "_");
		String subComponent = topic.split("/")[topic.split("/").length - 1];
		Message message = createIotHubMessageUtf8(payload, subComponent, timestamp);

		DeviceClient deviceClient;
		if (deviceClients.containsKey(componentId)) {
[...]

PR #1567, opened 2022-07-06
fix(iot-dev): fix thread-leak issue with IotHubReconnectTask

brycewang-microsoft:
For #1563:

Before this fix, the reconnect thread was not cleared appropriately. As a result, an instance of `IotHubReconnectTask` would be created per connection-status change from "DISCONNECTED_RETRYING" to "CONNECTED", while the previous thread was still waiting until the connection turned into "DISCONNECTED" eventually. 

Thread dump **before** this fix:

Thread dump **after** this fix:

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/DeviceIO.java
@@ -136,13 +136,11 @@ final class DeviceIO implements IotHubConnectionStatusChangeCallback
         // while startWorkerThreads should never be called when threads are already active, it doesn't hurt to double
         // check that any previous thread pools have been shut down.
         stopSendAndReceiveThreads();
-        stopReconnectThreads();
 
         log.debug("Starting worker threads");
 
         this.sendTaskScheduler = Executors.newScheduledThreadPool(1);
         this.receiveTaskScheduler = Executors.newScheduledThreadPool(1);
-        this.reconnectTaskScheduler = Executors.newScheduledThreadPool(1);
 
         // Note that even though these threads are scheduled at a fixed interval, the sender/receiver threads will wait
         // if no messages are available to process. These waiting threads will still count against the pool size defined above,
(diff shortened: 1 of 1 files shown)


drwill-ms (approved):

timtay-microsoft:
I double checked that we didn't have any threads leaking here, and it seems your fix works in that regard. That said, when I go through a basic connection loss event, I see that the SDK behaves like this:

 - Client opens initial connection
   - Send/Receive/Reconnect threads start
 - Client loses connection
   - Send/Receive threads stop
 - Client regains connection
   - Send/Receive threads stop (already stopped, so this is a no-op)
   - Reconnect thread stops*
   - Reconnect thread starts**

I'd rather we didn't stop the reconnect thread here* just to create a new thread pool here**. If the reconnect task is already running, then we don't need to do anything to it after we regain connectivity.

Something like:

```java

private void startWorkerThreads()
{
	stopSendAndReceiveThreads();

	this.sendTaskScheduler = Executors.newScheduledThreadPool(1);
	this.receiveTaskScheduler = Executors.newScheduledThreadPool(1);

	this.sendTaskScheduler.scheduleWithFixedDelay(this.sendTask, 0,
			sendPeriodInMilliseconds, TimeUnit.MILLISECONDS);
	this.receiveTaskScheduler.scheduleWithFixedDelay(this.receiveTask, 0,
			receivePeriodInMilliseconds, TimeUnit.MILLISECONDS);


	// This is only set to null if the client as a whole has been closed. This thread pool stays active through disconnected_retrying
	if (this.reconnectTaskScheduler == null)
	{
		this.reconnectTaskScheduler = Executors.newScheduledThreadPool(1);

		this.reconnectTaskScheduler.scheduleWithFixedDelay(this.reconnectTask, 0,
				receivePeriodInMilliseconds, TimeUnit.MILLISECONDS);	
	}
}
```

brycewang-microsoft:
> 

Yes, this seems more reasonable

brycewang-microsoft:
> I'd rather we didn't stop the reconnect thread here* just to create a new thread pool here**. If the reconnect task is already running, then we don't need to do anything to it after we regain connectivity.


Yes, this seems more reasonable!

timtay-microsoft (approved):

Issue #1571, opened 2022-07-11
[Bug Report] DeviceClient doesn't fully reconnect after temporary network disconnect (AMQP)


BenD10:
# Context

- **OS and version used:** Ubuntu 20.04 LTS
- **Java runtime used:** openjdk version "11.0.15" 2022-04-19
- **SDK version used:** 2.0.4

## Description of the issue
Continuation of #1467 

Performed a test where we temporaily disconnect network connectivity (either pull the ethernet cable or disable VM networking). I left the application running for approx 5 minutes and then restored network connectivity.

We expect that Twin, RPC, and events would resume normal function after network connectivity is restored.
Instead, we observed that Twin and RPC do not function. Events work do though (queued messages are being sent).

## Code sample exhibiting the issue
Reproduced the issue using the following sample from main branch of this repo - https://github.com/Azure/azure-iot-sdk-java/tree/main/device/iot-device-samples/direct-method-sample

1. Connected using AMQPS
2. Invoked a direct method and confirmed it worked successfully, response was recieved.
3. Disabled internet in VM for 4 minutes until I saw the AMQPS connection from the SDK failed.
4. Re-enabled internet, wait until I see connection is successfully established again.
5. All attempts to invoke a direct method will fail:
![image](https://user-images.githubusercontent.com/6621243/178279016-d17eeb61-78fd-4b6d-9bf8-59f51c4fae29.png)

## Console log of the issue
```
Starting...
Beginning setup.
Successfully read input parameters.
Using communication protocol AMQPS.
2022-07-11 09:40:31,479 INFO (main) [com.microsoft.azure.sdk.iot.device.transport.ExponentialBackoffWithJitter] - NOTE: A new instance of ExponentialBackoffWithJitter has been created with the following properties. Retry Count: 2147483647, Min Backoff Interval: 100, Max Backoff Interval: 10000, Max Time Between Retries: 100, Fast Retry Enabled: true
2022-07-11 09:40:31,918 DEBUG (main) [com.microsoft.azure.sdk.iot.device.ClientConfiguration] - Device configured to use software based SAS authentication provider
[...]

PR #1577, opened 2022-07-20
fix(iot-device): fix issue with direct method after reconnect from temporary network disconnect - AMQP

tmahmood-microsoft:
DeviceClient doesn't fully reconnect after temporary network disconnect (AMQP), attempts to invoke direct method fails.
Removed unnecessary boolean which were intended to carry over state between reconnection.

With each reconnection, we instantiate new SessionHandler but only set the boolean to true (and create method/twin links) when sending first message during initial connection, thus during reconnection we never create method/twin links.

github issue: https://github.com/Azure/azure-iot-sdk-java/issues/1571

 

# Checklist
- [ ] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-java/blob/main/.github/CONTRIBUTING.md).
  - [ ] This pull-request is submitted against the `main` branch. 

# Description of the problem

# Description of the solution

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
@@ -35,6 +35,9 @@ class AmqpsSessionHandler extends BaseHandler implements AmqpsLinkStateCallback
     private IotHubTransportMessage explicitInProgressTwinSubscriptionMessage;
     private IotHubTransportMessage explicitInProgressMethodsSubscriptionMessage;
 
+    //Carries over state between reconnections
+    private boolean subscribeToMethodsOnReconnection = false;
+    private boolean subscribeToTwinOnReconnection = false;
     private final AmqpsSessionStateCallback amqpsSessionStateCallback;
 
     //Should not carry over state between reconnects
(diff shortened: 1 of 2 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/amqps/AmqpsSessionHandler.java
```diff
@@ -35,9 +35,6 @@ class AmqpsSessionHandler extends BaseHandler implements AmqpsLinkStateCallback
     private IotHubTransportMessage explicitInProgressTwinSubscriptionMessage;
     private IotHubTransportMessage explicitInProgressMethodsSubscriptionMessage;
 
-    //Carries over state between reconnections
-    private boolean subscribeToMethodsOnReconnection = false;
-    private boolean subscribeToTwinOnReconnection = false;
```
timtay-microsoft, line 40: Keep in mind that not every user wants to subscribe to methods/twin. These flags are in place to track if the session was once subscribed to methods/twin

tmahmood-microsoft: reverted changes to AmqpsSessionhandler and making sure after closing amqp connection due to transport error, we save previous AmqpsSessions so that their twin/methods subscriptions can persist.

timtay-microsoft (approved):
Code change looks good. I'm assuming you've run this test locally and seen that the SDK recovers subscriptions even when onTransportError throws the network loss error?

tmahmood-microsoft:
> Code change looks good. I'm assuming you've run this test locally and seen that the SDK recovers subscriptions even when onTransportError throws the network loss error?

Yes, I was able to test locally and confirm that SDK recovers subscriptions as we do not lose SessionHandler object

timtay-microsoft:
> > Code change looks good. I'm assuming you've run this test locally and seen that the SDK recovers subscriptions even when onTransportError throws the network loss error?
> 
> Yes, I was able to test locally and confirm that SDK recovers subscriptions as we do not lose SessionHandler object

Perfect. Then you are all good to merge.

PR #1597, opened 2022-09-08
Migrate sample from the retired Java sample repo

brycewang-microsoft:
This is the only sample worthy to migrate from the sample repo.


Quickstarts/pom.xml
```diff
@@ -0,0 +1,55 @@
+<!-- Copyright (c) Microsoft. All rights reserved. --><!-- Licensed under the MIT license. See LICENSE file in the project root for full license information. -->
+<project>
+  <modelVersion>4.0.0</modelVersion>
+    <parent>
```
timtay-microsoft, line 4: I don't love this sample being defined in the root/Quickstarts path. Ideally, every sample in our repo would exist within the scope of the package that is being demonstrated. Since this is demonstrating a package we don't own, though, it is a special case. I'd consider moving this in with the IoT hub service client samples. This feature fits in the most with those samples

pom.xml
```diff
@@ -18,6 +18,7 @@
         <module>service</module>
         <module>iot-e2e-tests</module>
         <module>provisioning</module>
+        <module>Quickstarts/read-d2c-messages</module>
```
timtay-microsoft: This module's parent should not be the root pom.xml file. It's parent should be the [iot-service-samples pom](https://github.com/Azure/azure-iot-sdk-java/blob/main/service/iot-service-samples/pom.xml) since the sample will live alongside the service client samples

Quickstarts/read-d2c-messages/pom.xml
```diff
@@ -0,0 +1,40 @@
+<!-- Copyright (c) Microsoft. All rights reserved. --><!-- Licensed under the MIT license. See LICENSE file in the project root for full license information. -->
+<project>
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.microsoft.docs.iothub.samples</groupId>
```
timtay-microsoft: Can update this to match the other samples' groupId

service/iot-service-samples/read-d2c-messages/pom.xml
```diff
@@ -0,0 +1,40 @@
+<!-- Copyright (c) Microsoft. All rights reserved. --><!-- Licensed under the MIT license. See LICENSE file in the project root for full license information. -->
+<project>
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.microsoft.azure.sdk.iot.samples</groupId>
+        <artifactId>iot-service-samples</artifactId>
+        <version>1.18.0</version>
+    </parent>
+    <groupId>com.microsoft.docs.iothub.samples</groupId>
```
timtay-microsoft, line 9: Can change this to match the other samples' groupId

timtay-microsoft (approved):

pom.xml
brycewang-microsoft: This is weird... I changed this before your latest reviews, so the module `read-d2c-messages` shouldn't be in the root pom anymore. Can you still see the changes on root pom [here](https://github.com/Azure/azure-iot-sdk-java/pull/1597/files)?

Quickstarts/read-d2c-messages/pom.xml
brycewang-microsoft: this is weird too, it should not be the case any more.

Issue #1599, opened 2022-09-20
[Bug Report] Importing devices from blob to iothub fails if the Authentication.Type is SELF_SIGNED


yilmaznaslan:
# Context

- **OS and version used:**  MacOs Monterey 12.4
- **Java runtime used:** java 17.0.3.1 2022-04-22 LTS
- **SDK version used:**  `com.microsoft.azure.sdk.iot:iot-service-client:2.1.0` and `com.microsoft.azure:azure-storage:8.0.0`

## Description of the issue
Importing devices from a blob to iothub is causing import errors when the *AuthentcationType* is set to `SELF_SIGNED`. As also can be seen from the message in [importErrors.log](https://github.com/yilmaznaslan/azure-iot-dev-guide/blob/bug/docs/importErrors.log) the `Microsoft.Azure.Devices.Cloud.AuthenticationType` doesn't match with the `com.microsoft.azure.sdk.iot.service.auth.AuthenticationType` due to not matching enum names. In the automatically created [blob](https://github.com/yilmaznaslan/azure-iot-dev-guide/blob/bug/docs/devices.txt) `com.microsoft.azure.sdk.iot.service.auth.AuthenticationType#SELF_SIGNED` should be `com.microsoft.azure.sdk.iot.service.auth.AuthenticationType#selfSigned`. 

A more detailed description can be found in the [readme](https://github.com/yilmaznaslan/azure-iot-dev-guide/tree/bug)

## Code sample exhibiting the issue

You can also find the code in [Github repo](https://github.com/yilmaznaslan/azure-iot-dev-guide/tree/bug)

``
package org.example.azure;

import com.google.gson.Gson;
import com.google.gson.GsonBuilder;
import com.microsoft.azure.sdk.iot.service.auth.AuthenticationMechanism;
import com.microsoft.azure.sdk.iot.service.auth.AuthenticationType;
import com.microsoft.azure.sdk.iot.service.registry.*;
import com.microsoft.azure.storage.CloudStorageAccount;
import com.microsoft.azure.storage.StorageException;
import com.microsoft.azure.storage.blob.*;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

import java.io.ByteArrayInputStream;
import java.io.InputStream;
import java.nio.charset.StandardCharsets;
import java.security.InvalidKeyException;
import java.time.Duration;
import java.time.Instant;
import java.util.Date;
import java.util.EnumSet;

public class MainApplication {

[...]

PR #1600, opened 2022-09-20
fix the authentcationType enums values

yilmaznaslan:
# Checklist

[Importing devices from blob to iothub fails if the Authentication.Type is SELF_SIGNED](https://github.com/Azure/azure-iot-sdk-java/issues/1599)

# Description of the problem
AuthenticationType enum values doesn't match the IoTHub Authentication type values. See the issue for more detial

# Description of the solution
I changed the enum values to correct ones that match to `Microsoft.Azure.Devices.Cloud.AuthenticationType`

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/AuthenticationType.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/AuthenticationType.java
@@ -6,12 +6,12 @@ package com.microsoft.azure.sdk.iot.service.auth;
 /**
  * Enum for the type of authentication used.
  * SAS - shared access signature
- * selfSigned - self signed certificate
+ * SELF_SIGNED - self signed certificate
  * CERTIFICATE_AUTHORITY - certificate authority signed certificate
  */
 public enum AuthenticationType
 {
     SAS,
-    selfSigned,
+    SELF_SIGNED,
     CERTIFICATE_AUTHORITY
 }
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/AuthenticationType.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/AuthenticationType.java
@@ -6,12 +6,12 @@ package com.microsoft.azure.sdk.iot.service.auth;
 /**
  * Enum for the type of authentication used.
  * SAS - shared access signature
- * SELF_SIGNED - self signed certificate
+ * selfSigned - self signed certificate
  * CERTIFICATE_AUTHORITY - certificate authority signed certificate
  */
 public enum AuthenticationType
 {
     SAS,
-    SELF_SIGNED,
+    selfSigned,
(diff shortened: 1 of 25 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/AuthenticationType.java
```diff
@@ -6,12 +6,12 @@
...
  * SAS - shared access signature
- * SELF_SIGNED - self signed certificate
+ * selfSigned - self signed certificate
  * CERTIFICATE_AUTHORITY - certificate authority signed certificate
  */
 public enum AuthenticationType
 {
     SAS,
-    SELF_SIGNED,
+    selfSigned,
```
drwill-ms, line 15: Do we have no control over how these enum values are serialized so we get the proper casing for the wire and not have to compromise on casing standards for Java enum values?

@timtay-microsoft

timtay-microsoft: I'd rather we tag the enum value ```SELF_SIGNED``` with 

```@SerializedName("selfSigned")``` so that we serialize/deserialize with "selfSigned" but don't cause a breaking change on the enum value itself.

Comment (drwill-ms) [reply service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/AuthenticationType.java:15]: Good point; we can't introduce a breaking change.

Comment (drwill-ms) [reply service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/AuthenticationType.java:15]: @yilmaznaslan do you want to resubmit with just that one line change? We do very much appreciate that you've taken the time to submit a PR, of course. :)

Comment (yilmaznaslan) [reply service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/AuthenticationType.java]: I have built and tested it with `@SerializedName("selfSigned")`, it worked out :) I have submitted the change

**Review by drwill-ms — APPROVED** [review]

**Comment by yilmaznaslan** [comment]
The same error message was also occurring for the `CERTIFICATE_AUTHORITY` enum. I also added a fix for that

File: service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/auth/AuthenticationType.java [hunk 2]
```diff
@@ -12,6 +14,8 @@
 public enum AuthenticationType
 {
     SAS,
```
timtay-microsoft, line 16: Looks like the SAS enum value is also incorrect based on the service's swagger doc
```suggestion
    @SerializedName("sas")
    SAS,
```

yilmaznaslan: I have tested it with both **with** `@SerializedName("sas")` and **without** `@SerializedName("sas")` cases, it works in both cases

timtay-microsoft:
Thanks for the fix, @yilmaznaslan!

PR #1612, opened 2022-10-17
Simplify provisioning e2e tests

timtay-microsoft:
- Removes custom allocation policy provisioning tests

From the device side, these provisioning requests aren't any different from a provisioning request with other allocation policies (static, geolatency, etc). From the service side, we don't actually need a valid webhook to test creating enrollments with a custom allocation policy. Our existing tests just use "www.microsoft.com" as the webhook url. DPS will only validate these webhooks when the provisioning request is made from the device. As a result, we no longer need a custom allocation policy webhook.

- Remove reprovisioning tests

These tests featured making a registration request after a successful registration request and testing some service features like allocation policies, flags for migrating/erasing twin state between registrations, and flags for blocking reprovisioning to a new hub. These tests were more about testing the service than testing the device, so we can remove these tests. We still have e2e tests for the service client that ensure we can create enrollments with these flags set correctly, though.

All setup script artifacts tied to the far away hub and the custom allocation policy have been removed because we no longer need either of them.


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningServiceClientTests.java
```diff
@@ -15,14 +15,15 @@
...
-import static tests.integration.com.microsoft.azure.sdk.iot.provisioning.setup.ProvisioningCommon.CUSTOM_ALLOCATION_WEBHOOK_URL_VAR_NAME;
 import static tests.integration.com.microsoft.azure.sdk.iot.provisioning.setup.ProvisioningCommon.DPS_CONNECTION_STRING_ENV_VAR_NAME;
 
 @DeviceProvisioningServiceTest
 public class ProvisioningServiceClientTests
 {
     public static final String provisioningServiceConnectionString = Tools.retrieveEnvironmentVariableValue(DPS_CONNECTION_STRING_ENV_VAR_NAME);
-    public static final String customAllocationWebhookUrl = Tools.retrieveEnvironmentVariableValue(CUSTOM_ALLOCATION_WEBHOOK_URL_VAR_NAME);
+
+    // This is not an actual webhook that works for provisioning, but is fine for the purposes of this test.
```
timtay-microsoft, line 25:
```suggestion
    // This is not an actual webhook that works for provisioning, but is fine for the purposes of these tests as
    they don't involve any devices being provisioned.
```

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningTests.java
```diff
@@ -219,112 +123,26 @@ public void individualEnrollmentGetAttestationMechanismTPM() throws Provisioning
...
 
     private void enrollmentWithInvalidRemoteServerCertificateFails(EnrollmentType enrollmentType) throws Exception
     {
-        if (enrollmentType == EnrollmentType.GROUP && testInstance.attestationType != AttestationType.SYMMETRIC_KEY)
-        {
-            return; // test code not written for the x509 group scenario, and group enrollment does not support tpm attestation
-        }
+        // test code not written for the x509 group scenario, and enrollment groups don't support tpm attestation
+        assumeTrue("Skipping this test because only Symmetric Key attestion can be tested for enrollment groups",
+            this.testInstance.attestationType == AttestationType.SYMMETRIC_KEY);
```
timtay-microsoft, line 145: I'm simplifying a couple of cases like this where the Intellij UI will now show the test as "skipped" rather than "passed" when we don't want to run the scenario

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/setup/ProvisioningCommon.java
```diff
@@ -415,54 +407,11 @@ public ProvisioningStatus registerDevice(ProvisioningDeviceClientTransportProtoc
...
-        if (expectedDeviceCapabilities.isIotEdge())
-        {
-            assertTrue(CorrelationDetailsLoggingAssert.buildExceptionMessageDpsIndividualOrGroup("Provisioned device isn't edge device: " + testInstance.provisionedDeviceId, getHostName(provisioningServiceConnectionString), testInstance.groupId, testInstance.registrationId), twin.getCapabilities().isIotEdge());
-        }
-        else
-        {
-            assertTrue(CorrelationDetailsLoggingAssert.buildExceptionMessageDpsIndividualOrGroup("Provisioned device shouldn't be edge device " + testInstance.provisionedDeviceId, getHostName(provisioningServiceConnectionString), testInstance.groupId, testInstance.registrationId), twin.getCapabilities() == null || !twin.getCapabilities().isIotEdge());
-        }
+        DeviceClient deviceClient = new DeviceClient(iothubUri, deviceId, testInstance.securityProvider, IotHubClientProtocol.MQTT);
+        deviceClient.open(true);
```
abhipsaMisra, line 413: open is an I/o operation, right?

timtay-microsoft: Yes. I'm adding this operation because it was missing before. However, I'm limiting it to only one protocol (MQTT) to limit the number of device connections we make when verifying that the provisioning worked.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningTests.java
```diff
@@ -61,147 +63,49 @@ public ProvisioningTests(ProvisioningDeviceClientTransportProtocol protocol, Att
...
     {
-        ReprovisionPolicy reprovisionPolicy = new ReprovisionPolicy();
-        reprovisionPolicy.setMigrateDeviceData(false);
-        reprovisionPolicy.setUpdateHubAssignment(true);
+        // test code not written for the x509 group scenario, and enrollment groups don't support tpm attestation
+        assumeTrue("Skipping this test because only Symmetric Key attestion can be tested for enrollment groups",
+            this.testInstance.attestationType == AttestationType.SYMMETRIC_KEY);
 
-        reprovisioningFlow(EnrollmentType.GROUP, null, reprovisionPolicy, null, getStartingHubs(), getHubsToReprovisionTo());
+        basicRegistrationFlow(EnrollmentType.INDIVIDUAL);
```
abhipsaMisra, line 81: should this be individual?

abhipsaMisra (approved):

drwill-ms (approved):

Issue #1611, opened 2022-10-17
NullPointerException with TwinState when queryTwins is used with partial twin


ranjith-t:
SDK throws NullPointerException when QueryClient() is used to get twin details from Hub.

SDK Version: 
group: 'com.microsoft.azure.sdk.iot', name: 'iot-service-client', version:'2.1.3'

API used:
        var twinQueryResponse: TwinQueryResponse = queryClient.queryTwins(sqlQuery)
SQL Query:
**select deviceId from devices**
        **val filter: String = "deviceId"** 
        var sqlQuery: String  = **SqlQueryBuilder.createSqlQuery(filter, SqlQueryBuilder.FromType.DEVICES, null, null)**

Trace:

java.lang.NullPointerException
        at com.microsoft.azure.sdk.iot.service.twin.TwinState.getConnectionState(TwinState.java:538)
        at com.microsoft.azure.sdk.iot.service.twin.Twin.fromJson(Twin.java:99)
        at com.microsoft.azure.sdk.iot.service.query.TwinQueryResponse.<init>(TwinQueryResponse.java:50)
        at com.microsoft.azure.sdk.iot.service.query.QueryClient.queryTwins(QueryClient.java:202)
        at com.microsoft.azure.sdk.iot.service.query.QueryClient.queryTwins(QueryClient.java:160)

# Context

- **OS and version used:** <VERSION> Ubuntu
- **Java runtime used:** <VERSION> openjdk version "1.8.0_342"
- **SDK version used:** <VERSION> version:'2.1.3'

## Description of the issue
When the query is tried with "*"(i.e without filter) things work and list of twins are returned.

## Code sample exhibiting the issue
        val filter: String = "deviceId"
        var sqlQuery: String  = SqlQueryBuilder.createSqlQuery(filter, SqlQueryBuilder.FromType.DEVICES, null, null)
        var twinQueryResponse: TwinQueryResponse = queryClient.queryTwins(sqlQuery)

timtay-microsoft:
Sorry for the delay on this issue. #1639 will fix this bug, and we are planning to get it released within a day or two.

ngastelum-ms:
The fix for this issue has been released as of IoT hub service client version 2.1.5, marking this issue as closed

PR #1639, opened 2022-12-08
fix(iot-serv): Fix serialization of incomplete twin objects

timtay-microsoft:
There were some missing null checks and some unnecessary argument-not-null assertions in our twin parsing logic that breaks the ability to query twins with fields filtered down to just one or two fields.

#1611

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/Twin.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/Twin.java
@@ -92,14 +92,41 @@ public class Twin
         TwinState twinState = new TwinState(json);
 
         Twin twin = new Twin(twinState.getDeviceId());
-        twin.setVersion(twinState.getVersion());
-        twin.setETag(twinState.getETag());
-        twin.setStatus(twinState.getStatus());
-        twin.setStatusUpdateTime(twinState.getStatusUpdatedTime());
-        twin.setConnectionState(twinState.getConnectionState());
-        twin.setLastActivityTime(twinState.getLastActivityTime());
-        twin.setCloudToDeviceMessageCount(twinState.getCloudToDeviceMessageCount());
 
+        if (twinState.getVersion() != null)
+        {
+            twin.setVersion(twinState.getVersion());
+        }
+
+        if (twinState.getETag() != null)
+        {
+            twin.setETag(twinState.getETag());
+        }
+
+        if (twinState.getStatus() != null)
+        {
+            twin.setStatus(twinState.getStatus());
+        }
+
+        if (twinState.getStatusUpdatedTime() != null)
+        {
+            twin.setStatusUpdateTime(twinState.getStatusUpdatedTime());
+        }
+
+        if (twinState.getConnectionState() != null)
+        {
+            twin.setConnectionState(twinState.getConnectionState());
+        }
+
+        if (twinState.getLastActivityTime() != null)
+        {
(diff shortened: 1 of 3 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/Twin.java
```diff
@@ -109,29 +109,35 @@ public static Twin fromJson(String json)
         }
 
         // Desired properties
-        twin.getDesiredProperties().setVersion(twinState.getDesiredProperties().getVersion());
```
timtay-microsoft, line 112: We were dereferencing twinState.getDesiredProperties() without null checking it first. Now we will null check first.

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/Twin.java
```diff
@@ -109,29 +109,35 @@ public static Twin fromJson(String json)
...
         }
 
-        twin.getDesiredProperties().setTwinMetadata(twinState.getDesiredProperties().getTwinMetadata());
-        if (twinState.getDesiredProperties().getMetadataMap().size() > 0)
+        if (twinState.getReportedProperties() != null)
         {
-            twin.getDesiredProperties().getMetadataMap().putAll(twinState.getDesiredProperties().getMetadataMap());
-        }
-
-        // Reported properties
```
timtay-microsoft, line 124: Same for reported properties

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/Twin.java
```diff
@@ -166,11 +172,6 @@ public Twin(String deviceId) throws IllegalArgumentException
     {
         this();
 
-        if (Tools.isNullOrEmpty(deviceId))
-        {
-            throw new IllegalArgumentException("deviceId cannot be null or empty.");
```
timtay-microsoft, line 171: Device ID is allowed to be null, interestingly. Only when querying twins and filtering out the deviceId field, though

drwill-ms: Deserves a code comment, somewhere. Thoughts?

drwill-ms (approved):

patilsnr (approved):

service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/Twin.java
```diff
@@ -109,29 +108,35 @@ public static Twin fromJson(String json)
...
+        if (twinState.getReportedProperties() != null)
         {
-            twin.getReportedProperties().putAll(twinState.getReportedProperties());
-        }
-
-        twin.getReportedProperties().setTwinMetadata(twinState.getReportedProperties().getTwinMetadata());
-        if (twinState.getReportedProperties().getMetadataMap().size() > 0)
-        {
-            twin.getReportedProperties().getMetadataMap().putAll(twinState.getReportedProperties().getMetadataMap());
+            // Reported properties
```
brycewang-microsoft, line 128: Maybe move this comment above L126, before we do the null check for reported properties?

brycewang-microsoft (approved):

abhipsaMisra (approved):

andyk-ms (approved):

PR #1640, opened 2022-12-09
fix(prov-dev): Fix bug in failed and disabled statuses during device provisioning

sokolowski-daniel:
# Checklist

# Description of the problem
ProvisioningTask.executeStateMachineForStatus() method doesn't handle switch from Assigning to Failed/Disabled status during device registration.
registrationOperationStatusParser.getRegistrationState() object is used to read error code while this object is null and causes NPE, as the result registration stuck in Assigning status.

# Description of the solution
statusRegistrationOperationStatusParser object is updated in Assigning status and contains required information.
After changing registrationOperationStatusParser.getRegistrationState() to statusRegistrationOperationStatusParser.getRegistrationState(), failed and disabled statuses are reported correctly and ProvisioningTask invokes registration callback with expected exception.


timtay-microsoft:
Thanks for the fix, @sokolowski-daniel! I'd like to add a few other small changes (including unit test coverage for this case), so I plan on merging #1641 instead of this one. I'll make a note that you co-authored this change with me, of course.

sokolowski-daniel:
Thanks @timtay-microsoft

PR #1641, opened 2022-12-09
fix(prov-dev): Fix bug in failed and disabled statuses during device provisioning

timtay-microsoft:
If the provisioning ended with status DISABLED or FAILED, the SDK previously attempted to to assign a null Integer value (registrationOperationStatusParser.getRegistrationState().getErrorCode()) to the primitive int type used by dpsHubException.setErrorCode(...)

Assigning a null Integer value to the primitive int type caused the ProvisioningTask thread to throw an NPE. As a result, the provisioning state machine stayed locked in ASSIGNING perpetually even though the SDK received the terminal DISABLED/FAILED state from the service.

Co-authored by @sokolowski-daniel 

Thanks to #1640 for pointing out this bug and for providing the fix

### DIFF
--- a/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/ProvisioningTask.java
+++ b/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/ProvisioningTask.java
@@ -180,7 +180,6 @@ public class ProvisioningTask implements Callable<Object>
 
     {
         boolean isContinue = false;
-        RegistrationOperationStatusParser statusRegistrationOperationStatusParser = registrationOperationStatusParser;
         ProvisioningStatus nextStatus = ProvisioningStatus.fromString(registrationOperationStatusParser.getStatus());
         log.info("Current provisioning status: {}", nextStatus);
         // continue invoking for status until a terminal state is reached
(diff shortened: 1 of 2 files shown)


provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/ProvisioningTask.java
```diff
@@ -180,7 +180,6 @@ private void executeStateMachineForStatus(RegistrationOperationStatusParser regi
 
     {
         boolean isContinue = false;
-        RegistrationOperationStatusParser statusRegistrationOperationStatusParser = registrationOperationStatusParser;
```
timtay-microsoft, line 183: We were copying the parser class here for no reason in particular. I've removed this and we just reference ```registrationOperationStatusParser``` everywhere below now instead

provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/ProvisioningTask.java
```diff
@@ -247,19 +243,27 @@ private void executeStateMachineForStatus(RegistrationOperationStatusParser regi
                     break;
                 case FAILED:
                     this.dpsStatus = PROVISIONING_DEVICE_STATUS_FAILED;
-                    String errorMessage = statusRegistrationOperationStatusParser.getRegistrationState().getErrorMessage();
+                    String errorMessage = registrationOperationStatusParser.getRegistrationState().getErrorMessage();
                     ProvisioningDeviceHubException dpsHubException = new ProvisioningDeviceHubException(errorMessage);
-                    dpsHubException.setErrorCode(registrationOperationStatusParser.getRegistrationState().getErrorCode());
+                    if (registrationOperationStatusParser.getRegistrationState().getErrorCode() != null)
```
timtay-microsoft, line 248: This null check is the relevant change that fixes the bug here. Going from Integer to int we weren't checking for null

provisioning/provisioning-device-client/src/test/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/ProvisioningTaskTest.java
```diff
@@ -254,6 +254,8 @@ public void invokeRegisterReturnsDisabledSucceeds() throws Exception
                 result = mockedDeviceRegistrationResultParser;
                 mockedDeviceRegistrationResultParser.getErrorMessage();
                 result = "Test Error Message";
+                mockedDeviceRegistrationResultParser.getErrorCode();
+                result = null;
```
timtay-microsoft, line 258: This small update to an existing unit test now will cover this case for us now

drwill-ms (approved):

Issue #1629, opened 2022-10-31
[Bug Report] OutOfMemory Error due to growing correlatingCallbacks map in IoTHubTransport.class


tarekmues:
# Context

- **OS and version used**: Ubuntu Server 22.04, MacOS Ventura
- **Java runtime used**: openjdk-17-alpine (Docker container)
- **SDK version used**: 2.1.2

## Description of the issue
While the app is running and sending twin reported properties for a while, the correlating callbacks map in IotHubTransport.class grows endlessly. Managing several virtual devices in one application, it eventually leads to an OutOfMemory error due to a massive amount of correlating Callbacks to messages which response has already been handled.

To reproduce the error in a short amount of time, I've created a simple code example that just updates the reported properties over and over. Choose a small heapsize (I set 22mb for me in a spring boot environment, took about 15min with the example down below) for the vm in order to reproduce the exception.

## Code sample exhibiting the issue
```
    String connectionString = ""; //insert your connection string here
    DeviceClient client = new DeviceClient(connectionString, IotHubClientProtocol.AMQPS_WS);
    client.open(false);
    client.subscribeToDesiredProperties(new DesiredPropertiesCallback() {
        @Override
        public void onDesiredPropertiesUpdated(Twin twin, Object context) {
            //do nothing
        }
    }, null);`
    
    while (true) {
        TwinCollection reportedProperties = client.getTwin().getReportedProperties();
        client.updateReportedProperties(reportedProperties);
        Thread.sleep(400);
    }
```

## Console log of the issue
```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'test': Injection of autowired dependencies failed; nested exception is java.lang.OutOfMemoryError: Java heap space
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor.postProcessProperties(AutowiredAnnotationBeanPostProcessor.java:405) ~[spring-beans-5.3.23.jar:5.3.23]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.populateBean(AbstractAutowireCapableBeanFactory.java:1431) ~[spring-beans-5.3.23.jar:5.3.23]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:619) ~[spring-beans-5.3.23.jar:5.3.23]
[...]

PR #1647, opened 2022-12-14
fix(iot-dev): Fix bug caused by unclearing correlationCallbacks map

brycewang-microsoft:
Bug fix for issue #1629 

Currently we are never clearing the `correlationCallbacks` map in "IotHubTransport" class, even after the acknowledgement for received cloud to device message has been sent back to the hub. As a result, the size of map will grow endlessly which will result in OutOfMemory Error eventually.

### DIFF
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -33,7 +33,7 @@ public class IotHubTransport implements IotHubListener
 {
     private static final int DEFAULT_MAX_MESSAGES_TO_SEND_PER_THREAD = 10;
 
-    private static final int DEFAULT_CORRELATION_ID_LIVE_TIME = 30000;
+    private static final int DEFAULT_CORRELATION_ID_LIVE_TIME = 20000;
 
     // For tracking the state of this layer in particular. If multiplexing, this value may be CONNECTED while a
     // device specific state is DISCONNECTED_RETRYING. If this state is DISCONNECTED_RETRYING, then the multiplexed
--- a/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -117,9 +117,6 @@ public class IotHubTransport implements IotHubListener
     private final Map<String, CorrelatingMessageCallback> correlationCallbacks = new ConcurrentHashMap<>();
     private final Map<String, Object> correlationCallbackContexts = new ConcurrentHashMap<>();
 
-    // Used to store the number of milliseconds since epoch that this packet was created for a correlationId
-    private final Map<String, Long> correlationStartTimeMillis = new ConcurrentHashMap<>();
-
     /**
      * Constructor for an IotHubTransport object with default values
      *
@@ -1178,7 +1175,6 @@ public class IotHubTransport implements IotHubListener
                         // message has been acknowledged. Otherwise, the size of map will grow endlessly which results in OutOfMemory eventually.
                         correlationCallbacks.remove(correlationId);
                         correlationCallbackContexts.remove(correlationId);
-                        correlationStartTimeMillis.remove(correlationId);
                     }
                 }
                 catch (Exception ex)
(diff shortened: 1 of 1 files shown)


device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1166,6 +1166,10 @@ private void acknowledgeReceivedMessage(IotHubTransportMessage receivedMessage)
                     {
                         CorrelatingMessageCallback callback = correlationCallbacks.get(correlationId);
 
+                        // We need to remove the CorrelatingMessageCallback with the current correlation ID from the map after the received C2D
+                        // message has been acknowledged. Otherwise, the size of map will grow endlessly which results in OutOfMemory eventually.
+                        correlationCallbacks.remove(correlationId);
```
abhipsaMisra, line 1171: Should we wait until after the ack callback has been invoked to remove this from the list, i.e. after L1176?

timtay-microsoft:
One interesting thing to remember is that only the happy path ends when we send the ack for the received twin update status message. It's possible that the service fails to send the final message (due to service bug or due to us losing connection). 

We probably need to periodically check this map for very old messages and delete them. Maybe in the [sendMessages thread loop](https://github.com/Azure/azure-iot-sdk-java/blob/7ce92b326a7be3a6192d83335e07dd6d43d81c53/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java#L723)?

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -34,6 +33,8 @@ public class IotHubTransport implements IotHubListener
 {
     private static final int DEFAULT_MAX_MESSAGES_TO_SEND_PER_THREAD = 10;
 
+    private static final int DEFAULT_CORRELATION_ID_LIVE_TIME = 60000;
```
brycewang-microsoft, line 36: It's hard to know when exactly the callback of a particular correlating message will be completed OR if it will certainly happen (in case that the service fails to send the final message due to service bug or due to us losing connection), so setting this period as the default correlation ID live time. Check if a correlation ID has existed over this period every time while sending messages, if so, remove the ID from maps to avoid of out-of-memory.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -830,6 +835,22 @@ private void checkForExpiredMessages()
...
+    private void checkForOldMessages()
+    {
+        // Check if the "correlationCallbacks" map contains any correlation ID which has existed over the
+        // default correlation ID live time. If so, remove the old correlation IDs from the map. Otherwise,
+        // the size of map will grow endlessly which results in OutOfMemory eventually.
+        for (String correlationId : correlationCallbacks.keySet())
+        {
+            if (System.currentTimeMillis() - correlationStartTimeMillis.get(correlationId) >= DEFAULT_CORRELATION_ID_LIVE_TIME)
+            {
+                correlationCallbacks.remove(correlationId);
```
timtay-microsoft, line 847: I'm pretty sure you can't remove a key from a map while iterating over that map's key set. You'll probably need to move this outside the foreach loop.

timtay-microsoft:
In your first attempt, you removed the callback from the map once you knew the flow had finished successfully. Is there any reason you got rid of that in this fix?

brycewang-microsoft:
> In your first attempt, you removed the callback from the map once you knew the flow had finished successfully. Is there any reason you got rid of that in this fix?

Hmm, just thought that we should be able to cover that case with this new approach as well. But yeah, keeping it (remove when the flow is complete) seems more straightforward so I will get it back.

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1170,7 +1198,14 @@ private void acknowledgeReceivedMessage(IotHubTransportMessage receivedMessage)
                         {
                             Object context = correlationCallbackContexts.get(correlationId);
                             callback.onResponseAcknowledged(receivedMessage, context);
+
+                            correlationCallbackContexts.remove(correlationId);
```
timtay-microsoft, line 1202: Could this be moved down to the same block that clears the other correlation callback states on 1207 and 1208?

timtay-microsoft (approved):

device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
brycewang-microsoft: I prefer to leave it here since it's possible that, for a particular correlationId, we only have key-value object in the `correlationCallbacks` map but no object in the `correlationCallbackContexts` map if there's no CorrelatingMessageCallbackContext for that message ([ref](https://github.com/Azure/azure-iot-sdk-java/blob/main/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java#L1816)).

PR #1650, opened 2023-01-04
Upgrade netty-codec version to 4.1.86.Final

schoims:
netty-codec (v.4.1.77.Final) doesn't have ValueValidator in DefaultHeaders class, so the version needs to be upgraded to 4.1.86.Final.

error message from PR (https://github.com/Azure/azure-iot-sdk-java/pull/1645)
image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -124,7 +124,7 @@
             <dependency>
                 <groupId>io.netty</groupId>
                 <artifactId>netty-codec</artifactId>
-                <version>4.1.77.Final</version> <!--e2e tests and samples only-->
+                <version>4.1.86.Final</version> <!--e2e tests and samples only-->
             </dependency>
             <dependency>
                 <groupId>io.netty</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -195,16 +195,16 @@
         <dice-provider-artifact-id>dice-provider</dice-provider-artifact-id>
         <x509-provider-artifact-id>x509-provider</x509-provider-artifact-id>
 
-        <iot-device-client-version>2.1.2</iot-device-client-version>
-        <iot-service-client-version>2.1.4</iot-service-client-version>
-        <provisioning-device-client-version>2.0.2</provisioning-device-client-version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -124,12 +124,12 @@
             <dependency>
                 <groupId>io.netty</groupId>
                 <artifactId>netty-codec</artifactId>
-                <version>4.1.77.Final</version> <!--e2e tests and samples only-->
+                <version>4.1.86.Final</version> <!--e2e tests and samples only-->
             </dependency>
             <dependency>
                 <groupId>io.netty</groupId>
                 <artifactId>netty-codec-http</artifactId>
-                <version>4.1.77.Final</version> <!--e2e tests and samples only-->
```
abhipsaMisra, line 132: Do we need to update "netty-codec-http2" and "reactor-netty-http" versions as well?

schoims: It seems like the gate is passing without upgrading the versions of those packages.

patilsnr (approved):

pom.xml
abhipsaMisra: I'd update "netty-codec-http2", just so it follows the "netty-codec-http" version. (netty-codec-http2 has a dependency on netty-codec-http, so it should get the vulnerability patch automatically).

@timtay-microsoft - is the http2 lib required for using HTTP2.0?

schoims: Ah sure! Thanks for checking!

timtay-microsoft: Just upgrading the vulnerable package is fine here

timtay-microsoft: Upgrading netty packages has caused some issues with our Azure Core dependencies in the past, so upgrading everything may cause some unusual test failures

timtay-microsoft (approved):

andyk-ms (approved):

pom.xml
schoims: I upgraded the two packages and the gate passed. Am I good to merge it in?

timtay-microsoft: Yeah, go ahead

PR #1652, opened 2023-01-05
Enable invalid cert test

ngastelum-ms:


vsts/windowsLinuxAndAndroidBuildMatrixConfig.yaml
```diff
@@ -138,6 +138,94 @@ jobs:
           JAVA_VERSION: $(JAVA_VERSION)
           COMMIT_FROM: $(COMMIT_FROM)
 
+      - task : Docker@2
```
timtay-microsoft, line 141: These tests will run for android, windows, and linux builds, FYI. Do these steps enable the invalid cert server for those builds as well? Or do we need to duplicate these steps in each of the windows/linux/android builds?

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningTests.java
```diff
@@ -81,15 +81,13 @@ public void enrollmentGroupRegistration() throws Exception
...
 
-    @Ignore // The DPS instance we use for this test is currently offline, so this test cannot be run
     @Test
     @ContinuousIntegrationTest
     public void individualEnrollmentWithInvalidRemoteServerCertificateFails() throws Exception
     {
         enrollmentWithInvalidRemoteServerCertificateFails(EnrollmentType.INDIVIDUAL);
     }
 
-    @Ignore // The DPS instance we use for this test is currently offline, so this test cannot be run
```
schoims, line 92: You can remove import org.junit.Ignore;

iot-e2e-tests/docker/readme.md
```diff
@@ -0,0 +1,37 @@
+# Docker for WSL setup for local test
+
+This script will setup docker images and run instances of these images to support E2E testing.
+
+See [e2eTestsSetup.ps1](https://github.com/Azure/azure-iot-sdk-csharp/blob/main/e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1) script for E2E test setup prerequisites.
+
+Before running docker-setup.sh script:
+
+- Install WSL (See installation details here: https://docs.microsoft.com/en-us/windows/wsl/install). Reboot after install.
```
schoims, line 9:
```suggestion
- Install WSL (See installation details here: https://docs.microsoft.com/windows/wsl/install). Reboot after install.
```

vsts/windowsLinuxAndAndroidBuildMatrixConfig.yaml
ngastelum-ms: Good catch, right now it seems this only works on Linux due to some issue with Docker on Windows.

andyk-ms: Linux is the only one this will work for now (and we should test locally how well it works for android - it should work but this will be the first we test it). Docker for Windows have a bug that is reported but have no resolution and therefore windows agent in the pipeline will not work. On our local test, there are instruction to use WSL (linux) to run this on our windows laptop.

Also, there is a build.ps script for .NET SDK that only run invalid cert tests (Traits) when it is linux (non-window agents). There isn't an equivalent for that here in the changes.

iot-e2e-tests/docker/readme.md
```diff
@@ -0,0 +1,37 @@
+# Docker for WSL setup for local test
+
+This script will setup docker images and run instances of these images to support E2E testing.
+
+See [e2eTestsSetup.ps1](https://github.com/Azure/azure-iot-sdk-csharp/blob/main/e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1) script for E2E test setup prerequisites.
```
timtay-microsoft, line 5: Links to the C# SDK repo?

vsts/windowsLinuxAndAndroidBuildMatrixConfig.yaml
```diff
@@ -138,6 +138,94 @@ jobs:
...
+      - task: Bash@3
+        displayName: 'Cert Validation - Setup local hostname'
+        name: CVTEST_NET
+        inputs:
+          targetType: 'inline'
+          script: |
+            echo "==============="
+            echo "Inspect network"
+            echo "==============="
+            ip -4 addr
```
ngastelum-ms, line 180: Script seems to not work on android as is, so this is currently only for the linux builds

iot-e2e-tests/docker/readme.md
```diff
@@ -0,0 +1,35 @@
+# Docker for WSL setup for local test
+
+This script will setup docker images and run instances of these images to support E2E testing.
+
+Before running docker-setup.sh script:
+
+- Install WSL (See installation details here: https://docs.microsoft.com/windows/wsl/install). Reboot after install.
```
patilsnr, line 7: nit: is it worth specifying which version of WSL to use? I think the default these days is 2 but I'm not entirely sure.

iot-e2e-tests/docker/readme.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Install Docker on WSL (For docker on Ubuntu, see installation details here: https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository, see installation details for other Linux distribution here: https://docs.docker.com/engine/install/). Reboot after install.
```
patilsnr, line 9: nit: for consistency, would it be worth using hyperlinks?

iot-e2e-tests/docker/readme.md
```diff
@@ -0,0 +1,35 @@
...
+- Run wsl (See click on the linux distribution installed or open powershell and enter 'wsl')
+  
+- Change directory to the folder in the repo with the script [docker-setup.sh](https://github.com/Azure/azure-iot-sdk-java/iot-e2e-tests/docker/docker-setup.sh)
+  
+  ```Shell
+  cd <repo directory>/e2e/test/docker
+  .\docker-setup.sh
+  ```
+
+- The script output a list of IPs to hosts mapping at completion, copy these entries to the host system /etc/hosts file (default in $Env:windir\System32\drivers\etc\hosts, not in WSL)
```
patilsnr, line 22: `... not  in WSL).`

iot-e2e-tests/docker/readme.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Common script issues
+
+- This script needs to run each time to setup docker instance with the updated network setup for e2e testing.
+
+- This script needs to re-run after reboot or when the IP address of WSL host and container changes (WSL virtual network IP changes each time it starts).
+
+- Update the host file in the host system where the E2E test code and IDE will run on (not WSL).
+
+- If after running the script, the output does not show valid IP addresses for the hostnames, restart the PC and re-run this script
```
patilsnr, line 32: missing periods on this and next point

patilsnr (approved):
Looks good! I don't know enough about `.cfg` files to give a full judge of what's going on so I would wait for an approval from someone who does before merging.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/serviceclient/MessagingClientTests.java
```diff
@@ -336,11 +332,11 @@ public void messagingClientTokenRenewalWithAzureSasCredential() throws Exception
         Tools.disposeTestIdentity(testDeviceIdentity, iotHubConnectionString);
     }
 
-    @Ignore // The IoT Hub instance we use for this test is currently offline, so this test cannot be run
     @Test
     @ContinuousIntegrationTest
     public void messagingClientValidatesRemoteCertificateWhenSendingTelemetry() throws InterruptedException
     {
+        Assume.assumeTrue(System.getProperty("os.name").toLowerCase().contains("linux"));
```
timtay-microsoft, line 339: Consider moving this to Tools.java where we have a method called "isAndroid" that is similar to what you have here.

```
public static boolean isLinux
{
    return System.getProperty("os.name").toLowerCase().contains("linux");
}
```

So that the test reads a bit more easily

```java
Assume.assumeTrue(Tools.isLinux());
```

timtay-microsoft: It also cuts down on the duplicated code you currently have

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/provisioning/ProvisioningTests.java
```diff
@@ -81,19 +60,19 @@ public void enrollmentGroupRegistration() throws Exception
         basicRegistrationFlow(EnrollmentType.GROUP);
     }
 
-    @Ignore // The DPS instance we use for this test is currently offline, so this test cannot be run
     @Test
     @ContinuousIntegrationTest
     public void individualEnrollmentWithInvalidRemoteServerCertificateFails() throws Exception
     {
+        Assume.assumeTrue(System.getProperty("os.name").toLowerCase().contains("linux"));
```
timtay-microsoft, line 67: duplicated logic for how to check if the test is run on Linux. See my other comment about factoring this out somewhere

timtay-microsoft (approved):

iot-e2e-tests/docker/readme.md
timtay-microsoft: Good idea

schoims (approved):

iot-e2e-tests/docker/readme.md
ngastelum-ms: Looks like the default is wsl 2, I'm thinking it's not worth specifying since the learn docs we link to don't either except in the specific version section.

Issue #1657, opened 2023-01-16
[Bug Report] TwinClient replace method does not work with AzureSasCredential


ranjith-t:
# Context

- **OS Used:** Ubuntu 20.04.2 LTS
- **Java runtime used:** openjdk version "1.8.0_342"
- **SDK version used:** 2.1.3

## Description of the issue
When device twin replace was requested through TwinClient class, it does not update the twin when AzureSasCredential is used for connection. At the IotHub, neither update of device twin nor there is any exception from TwinClient.

## Code sample exhibiting the issue
TwinClient method replace() does not work when the TwinClient uses AzureSasCredential. It looks like the reason being that the iotHubConnectionString is not able to get proper URL for http request. All other classes tries to use 'hostname'(which is set), but in this particular case it tries to use deviceId to derive the url while the iotHubConnectionString is not initialized.

`URL url;
        if (twin.getModuleId() == null || twin.getModuleId().length() == 0)
        {
            url = this.iotHubConnectionString.getUrlTwin(twin.getDeviceId());
        }
        else
        {
            url = this.iotHubConnectionString.getUrlModuleTwin(twin.getDeviceId(), twin.getModuleId());
        }`

brycewang-microsoft:
Hi @ranjith-t, thanks for bringing this to our attention! The [fix ](https://github.com/Azure/azure-iot-sdk-java/pull/1658)has been merged in and I will update here again once we release a newer version including it.

brycewang-microsoft:
The hub service client package [2.1.6](https://github.com/Azure/azure-iot-sdk-java/releases/tag/2023-01-23) has been released with the fix included, so I am closing this thread.

ranjith-t:
Just confirming that this issue has been fixed.

PR #1658, opened 2023-01-18
fix(iot-serv): fix replace call on TwinClient

brycewang-microsoft:
For issue https://github.com/Azure/azure-iot-sdk-java/issues/1657

Currently, request URL within a "replace" call cannot be created properly while initiating a `TwinClient` with `AzureSasCredential`.

### DIFF
--- a/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/TwinClient.java
+++ b/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/TwinClient.java
@@ -321,11 +321,11 @@ public final class TwinClient
         URL url;
         if (twin.getModuleId() == null || twin.getModuleId().length() == 0)
         {
-            url = this.iotHubConnectionString.getUrlTwin(twin.getDeviceId());
+            url = IotHubConnectionString.getUrlTwin(this.hostName, twin.getDeviceId());
         }
         else
         {
-            url = this.iotHubConnectionString.getUrlModuleTwin(twin.getDeviceId(), twin.getModuleId());
+            url = IotHubConnectionString.getUrlModuleTwin(this.hostName, twin.getDeviceId(), twin.getModuleId());
         }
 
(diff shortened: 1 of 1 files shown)


service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/twin/TwinClient.java
```diff
@@ -321,11 +321,11 @@ public Twin replace(Twin twin, String ifMatch) throws IotHubException, IOExcepti
         URL url;
         if (twin.getModuleId() == null || twin.getModuleId().length() == 0)
         {
-            url = this.iotHubConnectionString.getUrlTwin(twin.getDeviceId());
+            url = IotHubConnectionString.getUrlTwin(this.hostName, twin.getDeviceId());
```
abhipsaMisra, line 324: Does this method exist both as a static method and instance method on IotHubConnectionString? If yes, we should probably rename them to highlight their difference?

abhipsaMisra (approved):

service/iot-service-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java
```diff
@@ -105,8 +105,6 @@ private static void replaceDesiredProperties(TwinClient twinClient, Twin deviceT
     {
         deviceTwin.getDesiredProperties().put("temp", new Random().nextInt(TEMPERATURE_RANGE));
 
-        // By replacing the twin rather than patching it, any desired properties that existed on the twin prior to this call
-        // that aren't present on the new set of desired properties will be deleted.
```
brycewang-microsoft, line 109: According to test results, this is not true, so remove it.

brycewang-microsoft: Will check the expected behavior regarding replace call with service team.

PR #1659, opened 2023-01-20
Add invalid cert tests to android

ngastelum-ms:
Build and test are running on different pipeline jobs and VMs, build is on windows and test is macOS. So adding the bash script to build failed due to it being windows


timtay-microsoft:
Closing this PR as "won't add this"

PR #1664, opened 2023-01-26
Fix DeviceTwinSample

brycewang-microsoft:
Detail in comments.

### DIFF
--- a/service/iot-service-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java
+++ b/service/iot-service-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java
@@ -49,22 +49,22 @@ public class DeviceTwinSample
         {
             // Manage complete twin
             // ============================== get initial twin properties =============================
-            Twin device = getInitialState(twinClient, deviceId);
+            getInitialState(twinClient, deviceId);
 
             // ================================ patch desired property ===============================
-            patchDesiredProperties(twinClient, device);
+            patchDesiredProperties(twinClient);
 
             // ================================ replace desired property ===============================
-            replaceDesiredProperties(twinClient, device);
+            replaceDesiredProperties(twinClient);
 
             // ============================ schedule update desired property ==========================
-            scheduleUpdateDesiredProperties(twinClient, device, jobClient);
+            scheduleUpdateDesiredProperties(twinClient, jobClient);
 
             // =================================== cancel job scheduled ===============================
-            cancelJob(twinClient, device, jobClient);
+            cancelJob(twinClient, jobClient);
 
             // ================================= remove desired property ==============================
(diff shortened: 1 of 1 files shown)


service/iot-service-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java
```diff
@@ -49,22 +49,22 @@ public static void main(String[] args) throws Exception
         {
             // Manage complete twin
             // ============================== get initial twin properties =============================
-            Twin device = getInitialState(twinClient, deviceId);
```
brycewang-microsoft, line 52: We don't need to cache the device twin and then pass it in each twin operation method below. When we did so, we were using the **existing** `Twin` instance, as a result, a replace call with the cached device twin would also update the unintended properties. We should've been created a new `Twin` instance each time when we do a twin operation.

service/iot-service-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java
```diff
@@ -88,50 +88,57 @@ private static Twin getInitialState(TwinClient twinClient, String deviceId) thro
...
-        System.out.println(deviceTwin);
+        Twin updatedTwin = twinClient.get(deviceId);
+        System.out.println(updatedTwin);
     }
 
-    private static void replaceDesiredProperties(TwinClient twinClient, Twin deviceTwin) throws IOException, IotHubException
+    private static void replaceDesiredProperties(TwinClient twinClient) throws IOException, IotHubException
     {
-        deviceTwin.getDesiredProperties().put("temp", new Random().nextInt(TEMPERATURE_RANGE));
+        Twin newTwins = new Twin(deviceId);
```
brycewang-microsoft, line 107: This is the new `Twin` object I am talking about above.

service/iot-service-samples/device-twin-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/DeviceTwinSample.java
```diff
@@ -88,50 +88,57 @@ private static Twin getInitialState(TwinClient twinClient, String deviceId) thro
...
+        newTwins.getDesiredProperties().put("temp", new Random().nextInt(TEMPERATURE_RANGE));
 
+        // By replacing the twin rather than patching it, any desired properties that existed on the twin prior to this call
```
brycewang-microsoft, line 110: Get the comments back as the sample is working as expected now.

timtay-microsoft (approved):

schoims (approved):

andyk-ms (approved):

patilsnr (approved):

Issue #1662, opened 2023-01-25
[Bug Report] Unable to upload to Azure Blob Storage after closing and re-opening DeviceClient


gabrielandrade2:
# Context

- **OS and version used:** Ubuntu 22.04.1
- **Java runtime used:** 17.0.5
- **SDK version used:** 2.1.2 and 2.1.3

## Description of the issue
To give a bit of context, I have a piece of code in my project that, after receiving a DirectMethod request, uploads a file to Azure Blob Storage and returns the generated URL as the DirectMethod response. This works just fine as long as I don't have any disconnection events or purposefully disable and re-enable the IoT connection.

After a disconnection and reconnection occur, when I try the same process, the blob upload fails when I call `client.getFileUploadSasUri`, which gives me the stack trace in the log section. This only happens if the disconnection/reconnection occurs after at least one blob has been uploaded before it.

I was also able to reproduce the same behavior using the `FileUploadSimpleSample.java` example code, by modifying it to upload the file more than one time. I attached the source for this in this post.

**_Possible cause:_** Upon some investigation in the SDK source code, I noticed that in the `com.microsoft.azure.sdk.iot.device.FileUpload` class, `httpsTransportManager` is received as a parameter and have `open()` called. 

However, when `DeviceClient`  is closed (in a disconnection event), this object is also closed and unless a new `DeviceClient` is instantiated it is never opened again, so the internal `httpsIotHubConnection` object remains null and causes the Exception, even though we are connected to the cloud.

I believe that `open()` should be called again, either when `DeviceClient.open()` is called again or when the first blob upload occurs.

Thank you for your time in supporting this.

## Code sample exhibiting the issue
```
package samples.com.microsoft.azure.sdk.iot;

import com.azure.storage.blob.BlobClient;
import com.azure.storage.blob.BlobClientBuilder;
import com.microsoft.azure.sdk.iot.device.FileUploadCompletionNotification;
import com.microsoft.azure.sdk.iot.device.FileUploadSasUriRequest;
import com.microsoft.azure.sdk.iot.device.FileUploadSasUriResponse;
import com.microsoft.azure.sdk.iot.device.DeviceClient;
import com.microsoft.azure.sdk.iot.device.IotHubClientProtocol;
import com.microsoft.azure.sdk.iot.device.exceptions.IotHubClientException;

import java.io.File;
import java.io.IOException;
import java.net.URISyntaxException;
import java.nio.charset.StandardCharsets;
[...]

PR #1666, opened 2023-01-27
Fix: Bug fix for SasUri generation failure after disconnecting/connec…

tmahmood-microsoft:
The bug was reproducible when once a file is uploaded to blob storage and after client reconnection, the same process fails with null exception.
The issue was missing open() call to HttpsTransportManager object when the device client is opened after disconnection. 

Original GitHub issue: https://github.com/Azure/azure-iot-sdk-java/issues/1662


patilsnr (approved):

andyk-ms (approved):

timtay-microsoft:
I'd rather see us remove the open/close methods on the ```HttpsTransportManager``` since http clients don't have that concept. Similarly, the ```FileUpload``` class has nothing to open/close, so we can remove that as well.

With the above, we'd instantiate the FileUpload class once (when the first file upload request is made) and never have to open/close/re-instantiate it again. Thoughts?

tmahmood-microsoft:
> I'd rather see us remove the open/close methods on the `HttpsTransportManager` since http clients don't have that concept. Similarly, the `FileUpload` class has nothing to open/close, so we can remove that as well.
> 
> With the above, we'd instantiate the FileUpload class once (when the first file upload request is made) and never have to open/close/re-instantiate it again. Thoughts?

This makes more sense, yes. Since the `HttpsTransportManager` object does not get affected by client reconnection, we dont need to close/open it again.

timtay-microsoft (approved):

Issue #1654, opened 2023-01-09
[Bug Report] Constrained device AMQP sending message rate


timtay-microsoft:
The Edgehub team has a [customer issue](https://github.com/Azure/iotedge/issues/6845) that suggests this SDK has the same bug as the C# SDK used to have as detailed [here](https://github.com/Azure/azure-iot-sdk-csharp/issues/2434). 

The problematic behavior is that there is a 50 ms delay between a message being sent to Edgehub and receiving feedback for that message on the service client side. 

As with the [C# SDK fix](https://github.com/Azure/azure-iot-sdk-csharp/pull/2439), this SDK should not batch ack's for feedback messages. The service only sends feedback messages one at a time, and always waits for the previous message to be ack'd before the next message is sent, so there is no scenario where batching makes sense.

timtay-microsoft:
After some initial digging, we don't seem to even be able to choose to batch these acknowledgements with proton-j, so the issue may lay somewhere else.

timtay-microsoft:
Quick update based on a comment on the other thread: The user switching to MQTT lowered the latency down to 10 ms, so there likely isn't an issue here on the feedback receiver. However, the AMQP protocol on the device side may only be polling once every 50 ms. If so, we should investigate making this configurable.

alexbaltman:
Any update on this?

timtay-microsoft:
Our AMQP library does not appear to make this interval configurable, so it is propagating these messages as fast as it can for now. From the other thread, the MQTT library appears to be faster at this, so users looking for speed here should probably use MQTT instead. Since we are limited by the AMQP library here, I'll tag this as "Underlaying Library Issue"

As for the customer's scenario in particular, it sounds like the user may be sending these messages serially (send message, wait for ack, send message, wait for ack, etc.). In general, for higher throughput scenarios, the user is encouraged to send messages in parallel (send X messages, wait for X acks) instead. Doing that should make this delay less impactful.

timtay-microsoft:
#1673 should help the SDK at least match what the AMQP library can do for sending message throughput, though

maciej-kepa:
> Our AMQP library does not appear to make this interval configurable, so it is propagating these messages as fast as it can for now. From the other thread, the MQTT library appears to be faster at this, so users looking for speed here should probably use MQTT instead. Since we are limited by the AMQP library here, I'll tag this as "Underlaying Library Issue"
> 
[...]

PR #1673, opened 2023-02-13
Fix issue where configured send interval isn't reflected in AMQP layer

timtay-microsoft:
It was hardcoded to 50 ms regardless of what the user set.

Also update some .gitIgnore lines to reflect recent repo re-organization

Probably related to #1654


iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
```diff
@@ -41,6 +41,8 @@
     public static final int DEFAULT_AMQP_OPEN_AUTHENTICATION_SESSION_TIMEOUT_IN_SECONDS = 20;
     public static final int DEFAULT_AMQP_OPEN_DEVICE_SESSIONS_TIMEOUT_IN_SECONDS = 60;
 
+    public static final int DEFAULT_SEND_INTERVAL_IN_MILLISECONDS = 10;
```
abhipsaMisra, line 44: Shouldn't the default value here be 50, so that it isn't a breaking change?

abhipsaMisra (approved):

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
timtay-microsoft: The default send interval at the top level ClientOptions was 10 (and remains 10), so I consider this a bug fix to go from 50 to 10 here in the AMQP layer

abhipsaMisra: So, the 50 was more of an artificial delay than advertised feature behavior? 👍🏽

timtay-microsoft: Exactly

PR #1683, opened 2023-03-01
feat(iot-dev): Add support for x509 authentication over AMQPS_WS

timtay-microsoft:


timtay-microsoft:
This fix only works sometimes, bizarrely. Closing this while we investigate further.

PR #1684, opened 2023-03-02
fix(e2e): Fix test bug where x509 authentication certs were given to SAS  authenticated devices

timtay-microsoft:
This didn't cause any issues as far as I can tell, but we shouldn't be including private keys on device identities who don't use x509 authentication

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/HubTierConnectionTests.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/HubTierConnectionTests.java
@@ -94,10 +94,9 @@ public class HubTierConnectionTests extends IntegrationTest
         hostName = IotHubConnectionStringBuilder.createIotHubConnectionString(iotHubConnectionString).getHostName();
         SSLContext sslContext = SSLContextBuilder.buildSSLContext(x509CertificateGenerator.getX509Certificate(), x509CertificateGenerator.getPrivateKey());
 
-        ClientOptions options = ClientOptions.builder().sslContext(sslContext).build();
+        ClientOptions x509ClientOptions = ClientOptions.builder().sslContext(sslContext).build();
 
         Proxy testProxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(testProxyHostname, testProxyPort));
-        ClientOptions optionsWithProxy = ClientOptions.builder().proxySettings((new ProxySettings(testProxy, testProxyUser, testProxyPass))).build();
 
         return new ArrayList(Arrays.asList(
                 new Object[][]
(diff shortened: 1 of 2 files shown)


iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/HubTierConnectionTests.java
```diff
@@ -94,10 +94,9 @@ public static Collection inputs() throws Exception
         hostName = IotHubConnectionStringBuilder.createIotHubConnectionString(iotHubConnectionString).getHostName();
         SSLContext sslContext = SSLContextBuilder.buildSSLContext(x509CertificateGenerator.getX509Certificate(), x509CertificateGenerator.getPrivateKey());
 
-        ClientOptions options = ClientOptions.builder().sslContext(sslContext).build();
+        ClientOptions x509ClientOptions = ClientOptions.builder().sslContext(sslContext).build();
```
timtay-microsoft, line 97: Renamed to make it clear that these options are only used by x509 devices

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/HubTierConnectionTests.java
```diff
@@ -94,10 +94,9 @@ public static Collection inputs() throws Exception
...
 
         Proxy testProxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(testProxyHostname, testProxyPort));
-        ClientOptions optionsWithProxy = ClientOptions.builder().proxySettings((new ProxySettings(testProxy, testProxyUser, testProxyPass))).build();
```
timtay-microsoft, line 100: Unused

patilsnr (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/SendMessagesCommon.java
```diff
@@ -208,7 +208,16 @@ public SendMessagesTestInstance(IotHubClientProtocol protocol, AuthenticationTyp
 
         public void setup() throws Exception
         {
-            SSLContext sslContext = SSLContextBuilder.buildSSLContext(x509CertificateGenerator.getX509Certificate(), x509CertificateGenerator.getPrivateKey());
+            // Builds an SSL context instance that trusts the certs on the device's certificate store
+            // and has no private keys
+            SSLContext sslContext = SSLContextBuilder.buildSSLContext();
+            if (testInstance.authenticationType == SELF_SIGNED)
```
abhipsaMisra, line 214: Since the SSL context build above does not contain private keys, how are CA certificate signed clients authenticated?

timtay-microsoft: The SSL context instance without private keys is only used when constructing a SAS authenticated device

abhipsaMisra: Oh, interesting. Is this a java specific requirement where the SSL context needs to be constructed even for a sas authenticated client? Is this similar to .NET accessing the device's certificate store directly during TLS negotiation?

timtay-microsoft: And the tests we run for Java only ever used self-signed certs, so we don't need the below 'if' block to include CA signed

PR #1685, opened 2023-03-03
Slim down the e2e tests

timtay-microsoft:
- Remove connection-oriented parameter cases for non-connection focused tests
  - For instance, a test about sending messages does not need to be run for both x509 authentication and SAS authentication. Nor does it need to be run for websocket vs TCP
  - Device/Module cases are still worth testing here because they use different service endpoints when publishing telemetry/methods
- Add connection test class for connection-oriented tests
  - Open/Close connection using all supported permutations of authentication, proxy/no proxy, transport protocol, device/module, TCP vs websocket
- Reorganize/cleanup some test code
  - Some "common" functions were only being used by one test method so they were moved there
- Rewrite some tests to use synchronous functions to simplify test code, remove helpers previously used to wait for async calls to finish


vsts/build_repo.ps1
```diff
@@ -21,7 +21,7 @@ if (($env:JAVA_VERSION).equals("8"))
     if ($isPullRequestBuild.equals("true"))
     {
         Write-Host "Skipping e2e tests since only Java 11 runs e2e tests during a pull request build"
-        mvn install -DskipIntegrationTests -DskipTests -T 2C --batch-mode -q
```
timtay-microsoft, line 24: Java 8 should only skip integration tests during a PR build, not all tests

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/Tools.java
```diff
@@ -256,7 +256,7 @@ private static TestDeviceIdentity getSasTestDevice(String iotHubConnectionString
     {
         // Don't want multiple methods calling this simultaneously and each thinking that they need to create
         // 100 devices. Forcing them to enter this block one at a time means that the first caller creates the 100 devices,
-        // and the subsequent callers just used one of those devices
+        // and the subsequent callers just uses one of those devices
```
patilsnr, line 259: super nit: periods at end of comments

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/ReceiveMessagesCommon.java
```diff
@@ -235,59 +209,60 @@ public IotHubMessageResult onCloudToDeviceMessageReceived(com.microsoft.azure.sd
         }
     }
 
-    public static class MessageCallbackMqtt implements com.microsoft.azure.sdk.iot.device.MessageCallback
+    protected static boolean hasExpectedProperties(com.microsoft.azure.sdk.iot.device.Message msg, Map<String, String> messageProperties)
```
patilsnr, line 212: Are we not able to import `com.microsoft.azure.sdk.iot.device.Message` to make this more readable?

patilsnr (approved):
optimistic sign off

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/setup/ReceiveMessagesCommon.java
timtay-microsoft: I can remove a couple cases of this, but this test class needs to reference the ```Message``` class from the device client side and the ```Message``` class from the service client side. Java doesn't support importing classes and renaming them (import com.microsoft.X as Y) to handle cases like this.

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/connection/ConnectionTests.java
```diff
@@ -0,0 +1,237 @@
...
+            {
+                Proxy testProxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(testProxyHostname, testProxyPort));
+                optionsBuilder.proxySettings(new ProxySettings(testProxy, testProxyUser, testProxyPass));
+            }
+
+            if (clientType == ClientType.DEVICE_CLIENT)
+            {
+                this.identity = Tools.getTestDevice(iotHubConnectionString, this.protocol, this.authenticationType, false, optionsBuilder);
+            }
+            else if (clientType == ClientType.MODULE_CLIENT)
```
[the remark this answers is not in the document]
timtay-microsoft: There are only two types of clients currently, so I'm not that concerned about the else case here.

PR #1691, opened 2023-03-09
Update package-maven-artifacts-for-release.ps1

brycewang-microsoft:
We have a new module "iothub" containing the "device" folder.


timtay-microsoft (approved):
Good catch and thanks for this!

PR #1692, opened 2023-03-10
fix(e2e): fix some CI tests

timtay-microsoft:
SendLargestMessage was never supposed to build a batch message of multiple max-sized messages. It was just supposed to be one max sized message

The mux tests here should not retry on open because they expect to throw an exception

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
@@ -1369,6 +1369,8 @@ public class MultiplexingClientTests extends IntegrationTest
         boolean expectedExceptionThrown = false;
         try
         {
+            // Note that we are opting to not retry on open because the first attempt should result in the caught
+            // exception below.
             testInstance.multiplexingClient.open(false);
         }
         catch (MultiplexingClientRegistrationException e)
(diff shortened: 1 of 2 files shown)


patilsnr (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
```diff
@@ -1369,7 +1369,9 @@ public void registerDeviceWithIncorrectCredentialsBeforeOpenThrowsOnOpen() throw
         boolean expectedExceptionThrown = false;
         try
         {
-            testInstance.multiplexingClient.open(true);
+            // Note that we are opting to not retry on open because the first attempt should result in the caught
+            // exception below.
+            testInstance.multiplexingClient.open(false);
```
abhipsaMisra, line 1374: Shouldn't bad credentials cause the retry policy in place to fail immediately anyway?

abhipsaMisra (approved):

iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
timtay-microsoft: Good point. I'll look into this

brycewang-microsoft (approved):

PR #1700, opened 2023-04-11
feat(dps-device): Add synchronous overloads for registerDevice

timtay-microsoft:
A lot of our samples and tests are simplified with this and a lot of customers would probably benefit the same way.


patilsnr (approved):
Looks like this really cleans up some of the samples. I'd wait for at least one more approval before merging, if you can.

PR #1708, opened 2023-05-08
Pass MqttGatewayHostName field into connection string

brycewang-microsoft:
Per new designs added in the guidance of E4K adapter

### DIFF
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
@@ -86,6 +86,10 @@ public final class ClientConfiguration
     @Setter(AccessLevel.PACKAGE)
     private int sendInterval = DEFAULT_SEND_INTERVAL_IN_MILLISECONDS;
 
+    @Getter
+    @Setter(AccessLevel.PACKAGE)
+    private boolean isConnectingToMqttGateway = false;
+
     private IotHubAuthenticationProvider authenticationProvider;
 
     /**
@@ -143,6 +147,12 @@ public final class ClientConfiguration
      */
     ClientConfiguration(IotHubConnectionString iotHubConnectionString, IotHubClientProtocol protocol) throws IllegalArgumentException
     {
+        String mqttGatewayHostName = iotHubConnectionString.getMqttGatewayHostName();
+        if (mqttGatewayHostName != null && !mqttGatewayHostName.isEmpty())
+        {
+            this.isConnectingToMqttGateway = true;
+        }
+
         this.protocol = protocol;
         configSasAuth(iotHubConnectionString);
     }
@@ -170,6 +180,12 @@ public final class ClientConfiguration
             throw new UnsupportedOperationException("This constructor only support sas token authentication currently");
         }
 
+        String mqttGatewayHostName = authenticationProvider.getMqttGatewayHostname();
+        if (mqttGatewayHostName != null && !mqttGatewayHostName.isEmpty())
+        {
+            this.isConnectingToMqttGateway = true;
+        }
+
         this.protocol = protocol;
         this.authenticationProvider = authenticationProvider;
         this.useWebsocket = false;
@@ -192,6 +208,12 @@ public final class ClientConfiguration
 
     ClientConfiguration(IotHubConnectionString iotHubConnectionString, IotHubClientProtocol protocol, ClientOptions clientOptions)
     {
+        String mqttGatewayHostName = iotHubConnectionString.getMqttGatewayHostName();
+        if (mqttGatewayHostName != null && !mqttGatewayHostName.isEmpty())
+        {
+            this.isConnectingToMqttGateway = true;
+        }
+
         this.protocol = protocol;
 
         if (clientOptions != null && clientOptions.getSslContext() != null)
@@ -265,6 +287,12 @@ public final class ClientConfiguration
 
     ClientConfiguration(IotHubConnectionString iotHubConnectionString, IotHubClientProtocol protocol, SSLContext sslContext)
     {
+        String mqttGatewayHostName = iotHubConnectionString.getMqttGatewayHostName();
+        if (mqttGatewayHostName != null && !mqttGatewayHostName.isEmpty())
+        {
+            this.isConnectingToMqttGateway = true;
+        }
+
         this.protocol = protocol;
         configSsl(iotHubConnectionString, sslContext);
     }
@@ -309,6 +337,12 @@ public final class ClientConfiguration
     ClientConfiguration(IotHubConnectionString connectionString, SecurityProvider securityProvider, IotHubClientProtocol protocol) throws IOException
     {
         commonConstructorSetup(connectionString);
+
+        String mqttGatewayHostName = connectionString.getMqttGatewayHostName();
+        if (mqttGatewayHostName != null && !mqttGatewayHostName.isEmpty())
+        {
+            this.isConnectingToMqttGateway = true;
+        }
 
         this.protocol = protocol;
 
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
@@ -154,6 +154,7 @@ public final class ClientConfiguration
         this.authenticationProvider = new IotHubSasTokenSoftwareAuthenticationProvider(
                 iotHubConnectionString.getHostName(),
                 iotHubConnectionString.getGatewayHostName(),
+                iotHubConnectionString.getMqttGatewayHostName(),
                 iotHubConnectionString.getDeviceId(),
                 iotHubConnectionString.getModuleId(),
                 iotHubConnectionString.getSharedAccessKey(),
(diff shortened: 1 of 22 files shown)


brycewang-microsoft:
Do we consider adding `MqttGatewayHostName` into client [option](iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientOptions.java) as well?

timtay-microsoft:
> Do we consider adding `MqttGatewayHostName` into client [option](iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientOptions.java) as well?

Yes, I think we should add it as an option there as well. Not every device/module client is created with a connection string

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubConnectionString.java
```diff
@@ -103,6 +116,10 @@ else if (attr.toLowerCase().startsWith(GATEWAY_HOST_NAME_ATTRIBUTE.toLowerCase()
             {
                 this.gatewayHostName = attr.substring(GATEWAY_HOST_NAME_ATTRIBUTE.length());
             }
+            else if (attr.toLowerCase().startsWith(MQTT_GATEWAY_HOST_NAME_ATTRIBUTE.toLowerCase()))
```
drwill-ms, line 119: It should be an error to specify both GatewayHostName and MqttGatewayHostName.

drwill-ms:
> > Do we consider adding `MqttGatewayHostName` into client [option](iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientOptions.java) as well?
> 
> Yes, I think we should add it as an option there as well. Not every device/module client is created with a connection string

Same as GatewayHostName, as we did in .NET v2.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubConnectionString.java
```diff
@@ -103,6 +116,16 @@ else if (attr.toLowerCase().startsWith(GATEWAY_HOST_NAME_ATTRIBUTE.toLowerCase()
...
+            {
+                this.mqttGatewayHostName = attr.substring(MQTT_GATEWAY_HOST_NAME_ATTRIBUTE.length());
+            }
+        }
+
+        if (gatewayHostName != null && !gatewayHostName.isEmpty()
+                && mqttGatewayHostName != null && !mqttGatewayHostName.isEmpty())
+        {
+            throw new IllegalArgumentException("[GatewayHostName] and [MqttGatewayHostName] should NOT be specified at the same time.");
         }
```
brycewang-microsoft, line 129: Throw an `IllegalArgumentException` when specifying [GatewayHostName] and [MqttGatewayHostName] at the same time.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubAuthenticationProvider.java
```diff
@@ -16,23 +16,31 @@
...
+    public IotHubAuthenticationProvider(String hostname, String gatewayHostname, String mqttGatewayHostname, String deviceId, String moduleId, SSLContext sslContext)
     {
         Objects.requireNonNull(hostname);
         Objects.requireNonNull(deviceId);
 
+        if (gatewayHostname != null && !gatewayHostname.isEmpty()
+            && mqttGatewayHostname != null && !mqttGatewayHostname.isEmpty())
+        {
+            throw new IllegalArgumentException("[GatewayHostname] and [MqttGatewayHostname] should NOT be specified at the same time.");
+        }
```
brycewang-microsoft, line 39: Same here, throw `IllegalArgumentException` while specifying both.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/hsm/IotHubSasTokenHsmAuthenticationProvider.java
```diff
@@ -169,7 +179,23 @@ private static IotHubSasToken createNewSasToken(
             String data = audience + "\n" + expiresOn;
             String signature = signatureProvider.sign(moduleId, data, generationId);
 
-            String host = gatewayHostName != null && !gatewayHostName.isEmpty() ? gatewayHostName : hostname;
+            if (gatewayHostName != null && !gatewayHostName.isEmpty()
```
timtay-microsoft, line 182: This logic looks like it could sit at a higher level since this class inherits from another. Could you look into moving this up a bit? That way all these AuthenticationProvider classes don't need this same logic within them

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
```diff
@@ -470,6 +476,15 @@ public String getGatewayHostname()
         return this.authenticationProvider.getGatewayHostname();
     }
 
+    /**
+     * Getter for MqttGatewayHostName.
+     * @return the value of MqttGatewayHostName
+     */
+    public String getMqttGatewayHostName()
```
timtay-microsoft, line 483: IIRC, this class or the authentication providers will provide the transport layer the "correct" host to connect to. The MQTT layer doesn't need to know the gateway host name + MQTT gateway host name + IoT hub host name. To simplify things, I'd like to see this be a boolean like "isConnectingToMqttGateway". See #1709

timtay-microsoft: That way the MQTT layer can do things like

```java
if (config.isConnectionToMqttGateway)
{
    topic = "..."
}
else 
{
    topic = "..."
}
```

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/hsm/IotHubSasTokenHsmAuthenticationProvider.java
brycewang-microsoft: I do have the same [logic](https://github.com/Azure/azure-iot-sdk-java/blob/brycewang/MqttGatewayHostName/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubAuthenticationProvider.java#L35-L39) in the ctors of base class inherited by other AuthenticationProvider classes. This logic right here is in creating new SAS token, which is being used only by this class. I prefer to leave it as-is.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
brycewang-microsoft: Added.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubConnectionString.java
```diff
@@ -20,17 +20,29 @@
...
     private static final String MODULE_ID_ATTRIBUTE = "ModuleId=";
 
+    /**
+     * IP address or internet name of the host machine working as a device or protocol gateway.
+     * Used when communicating with Azure Edge devices.
+     * */
     private static final String GATEWAY_HOST_NAME_ATTRIBUTE = "GatewayHostName=";
 
+    /** Used for E4K. */
+    private static final String MQTT_GATEWAY_HOST_NAME_ATTRIBUTE = "MqttGatewayHostName=";
```
abhipsaMisra, line 43: The .NET SDK has some client-side regex validations for the values parsed from the connection string. I see that the java SDK doesn't do any such validations so it isn't an issue here, but we might want to get some information on the character constraints for this hostname.

This is our existing regex validation for both IoT hub hostname and gateway hostname (edge): "[a-zA-Z0-9_\-\.]+$"

brycewang-microsoft: Yeah, this sounds fair to add.

timtay-microsoft: I'm not sure I understand why we would want to validate a host name like this. If the host name is incorrect, then the user will see a HostNotFoundException or something similar.

timtay-microsoft (approved):

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/IotHubConnectionString.java
brycewang-microsoft: > This is our existing regex validation for both IoT hub hostname and gateway hostname (edge): "[a-zA-Z0-9_-.]+$"
Do we define these character constraints ourselves or we refer to guidance somewhere else?

abhipsaMisra: > I'm not sure I understand why we would want to validate a host name like this. If the host name is incorrect, then the user will see a HostNotFoundException or something similar.

I agree with that and that's what we have in v2, but unfortunately this is the existing behavior in v1. We can choose to not add the validation for mqtt gateway hostname but that would change the "expected" behavior of how we treat the rest of the components of a connection string.

brycewang-microsoft: Ah, I see your points.

timtay-microsoft: Sounds good then.

timtay-microsoft:
For anyone looking at this PR to make the same change in .NET/python/node, note that we are likely changing how we get this flag from the user. Ping us to get the updated design doc for this.

PR #1710, opened 2023-05-10
Add parameter GatewayType to specify E4K

brycewang-microsoft:
Per new discussion we have, to specify E4K, more of us prefer to the option of adding a new property `GatewayType` (which can be one of two enum values "Edge" and "E4K") so that it can work with the existing property `GatewayHostName` which already supported for Edge. 

Also reverting the changes of adding `MqttGatewayHostName` committed [before](https://github.com/Azure/azure-iot-sdk-java/commit/8d152f70fe2816af5cd881bae5dc0753e334e6ed) in this PR.

### DIFF
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
@@ -86,6 +86,10 @@ public final class ClientConfiguration
     @Setter(AccessLevel.PACKAGE)
     private int sendInterval = DEFAULT_SEND_INTERVAL_IN_MILLISECONDS;
 
+    @Getter
+    @Setter(AccessLevel.PACKAGE)
+    private boolean isConnectingToMqttGateway = false;
+
     private IotHubAuthenticationProvider authenticationProvider;
 
     /**
@@ -191,6 +195,13 @@ public final class ClientConfiguration
 
     ClientConfiguration(IotHubConnectionString iotHubConnectionString, IotHubClientProtocol protocol, ClientOptions clientOptions)
     {
+        String gatewayHostName = iotHubConnectionString.getGatewayHostName();
+        GatewayType gatewayType = clientOptions.getGatewayType();
+        if (gatewayHostName != null && !gatewayHostName.isEmpty() && gatewayType == GatewayType.E4K)
+        {
+            this.isConnectingToMqttGateway = true;
+        }
+
         this.protocol = protocol;
 
         if (clientOptions != null && clientOptions.getSslContext() != null)
@@ -360,6 +371,14 @@ public final class ClientConfiguration
         // When setting the ClientOptions and a SecurityProvider, the SecurityProvider is responsible for setting the sslContext
         // we do not need to set the context in this constructor.
         this(connectionString, securityProvider, protocol);
+
+        String gatewayHostName = connectionString.getGatewayHostName();
+        GatewayType gatewayType = clientOptions.getGatewayType();
+        if (gatewayHostName != null && !gatewayHostName.isEmpty() && gatewayType == GatewayType.E4K)
+        {
+            this.isConnectingToMqttGateway = true;
+        }
+
         setClientOptionValues(clientOptions);
     }
 
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
@@ -667,14 +667,14 @@ public final class ClientConfiguration
         {
             if (gatewayHostName == null || gatewayHostName.isEmpty())
             {
-                throw new IllegalArgumentException("The value of [GatewayHostName] is NOT provided in the E4K mode.");
+                throw new IllegalArgumentException("The value of [GatewayHostName] is NOT provided while connecting to an E4K MQTT broker.");
             }
 
             this.isConnectingToMqttGateway = true;
 
             if (protocol != MQTT && protocol != MQTT_WS)
             {
-                throw new IllegalArgumentException("The transport protocol should be MQTT or MQTT_WS in the E4K mode.");
+                throw new IllegalArgumentException("The transport protocol should be MQTT or MQTT_WS while connecting to an E4K MQTT broker.");
             }
         }
     }
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
@@ -207,6 +207,11 @@ public final class ClientConfiguration
             {
                 throw new IllegalArgumentException("The value of [GatewayHostName] is NOT provided in the E4K mode.");
             }
+
+            if (protocol != MQTT && protocol != MQTT_WS)
+            {
+                throw new IllegalArgumentException("The transport protocol should be MQTT or MQTT_WS in the E4K mode.");
+            }
         }
 
         this.protocol = protocol;
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
@@ -199,14 +199,12 @@ public final class ClientConfiguration
         GatewayType gatewayType = clientOptions.getGatewayType();
         if (gatewayType == GatewayType.E4K)
         {
-            if (gatewayHostName != null && !gatewayHostName.isEmpty())
-            {
-                this.isConnectingToMqttGateway = true;
-            }
-            else
+            if (gatewayHostName == null || gatewayHostName.isEmpty())
             {
                 throw new IllegalArgumentException("The value of [GatewayHostName] is NOT provided in the E4K mode.");
             }
+
+            this.isConnectingToMqttGateway = true;
 
             if (protocol != MQTT && protocol != MQTT_WS)
             {
@@ -388,13 +386,16 @@ public final class ClientConfiguration
         GatewayType gatewayType = clientOptions.getGatewayType();
         if (gatewayType == GatewayType.E4K)
         {
-            if (gatewayHostName != null && !gatewayHostName.isEmpty())
-            {
-                this.isConnectingToMqttGateway = true;
-            }
-            else
+            if (gatewayHostName == null || gatewayHostName.isEmpty())
             {
                 throw new IllegalArgumentException("The value of [GatewayHostName] is NOT provided in the E4K mode.");
+            }
+
+            this.isConnectingToMqttGateway = true;
+
+            if (protocol != MQTT && protocol != MQTT_WS)
+            {
+                throw new IllegalArgumentException("The transport protocol should be MQTT or MQTT_WS in the E4K mode.");
             }
         }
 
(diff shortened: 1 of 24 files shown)


iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
```diff
@@ -208,8 +195,9 @@ private void configSasAuth(IotHubConnectionString iotHubConnectionString) {
 
     ClientConfiguration(IotHubConnectionString iotHubConnectionString, IotHubClientProtocol protocol, ClientOptions clientOptions)
     {
-        String mqttGatewayHostName = iotHubConnectionString.getMqttGatewayHostName();
-        if (mqttGatewayHostName != null && !mqttGatewayHostName.isEmpty())
+        String gatewayHostName = iotHubConnectionString.getGatewayHostName();
+        GatewayType gatewayType = clientOptions.getGatewayType();
+        if (gatewayHostName != null && !gatewayHostName.isEmpty() && gatewayType == GatewayType.E4K)
```
abhipsaMisra, line 200: Should we throw an exception of the gatewayType is set to E4K, but the hostname isn't supplied?

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientOptions.java
```diff
@@ -136,4 +136,12 @@
...
     @Builder.Default
     private final int receiveInterval = RECEIVE_PERIOD_MILLIS;
+
+    /**
+     * This option specifies the type of gateway to which the device/module client is connecting while
+     * specifying the property of GatewayHostName. By default, the value is EDGE.
+     */
+    @Getter
+    @Builder.Default
+    private final GatewayType gatewayType = GatewayType.EDGE;
```
abhipsaMisra, line 146: We'll eventually want to add in some additional comments that call out the operations that are unsupported over E4K gateway type.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/hsm/IotHubSasTokenHsmAuthenticationProvider.java
```diff
@@ -27,9 +27,8 @@ public class IotHubSasTokenHsmAuthenticationProvider extends IotHubSasTokenWithR
      * @param signatureProvider the signature provider to be used when generating sas tokens
      * @param deviceId the id of the device the module belongs to
      * @param moduleId the id of the module to be authenticated for
-     * @param hostname the hostname of the iothub to be authenticated for. May be null if either gatewayHostname or mqttGatewayHostname is not
-     * @param gatewayHostname the gatewayHostname of the edge hub to be authenticated for. May be null if either hostname or mqttGatewayHostname is not
-     * @param mqttGatewayHostname the mqttGatewayHostname of the edge hub to be authenticated for. May be null if either hostname or gatewayHostname is not
+     * @param hostname the hostname of the iothub to be authenticated for. May be null if gatewayHostname is not
```
abhipsaMisra, line 30: I didn't know that hostname could be null. Wouldn't the gateway still need to know which upstream hub to route to?

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
brycewang-microsoft: Good point!

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/hsm/IotHubSasTokenHsmAuthenticationProvider.java
brycewang-microsoft: This is reverted to what it originally was. But with another look, this statement that hostname could be null is strange indeed as we have some null checks regarding hostName in the downstream logics, like [here](https://github.com/Azure/azure-iot-sdk-java/blob/previews/E4Kadapter/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/auth/IotHubAuthenticationProvider.java#L32). 

@timtay-microsoft Do you have any idea why we have this comment or any objection of removing it?

timtay-microsoft: Allowing the hostname to be null seems wrong to me as well. Go ahead and make it mandatory

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
```diff
@@ -208,10 +195,23 @@ private void configSasAuth(IotHubConnectionString iotHubConnectionString) {
...
+            }
+            else
+            {
+                throw new IllegalArgumentException("The value of [GatewayHostName] is NOT provided in the E4K mode.");
+            }
+
+            if (protocol != MQTT && protocol != MQTT_WS)
+            {
+                throw new IllegalArgumentException("The transport protocol should be MQTT or MQTT_WS in the E4K mode.");
+            }
```
brycewang-microsoft, line 214: only support MQTT and MQTT_WS in E4K mode.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
```diff
@@ -208,10 +195,23 @@ private void configSasAuth(IotHubConnectionString iotHubConnectionString) {
...
         {
-            this.isConnectingToMqttGateway = true;
+            if (gatewayHostName != null && !gatewayHostName.isEmpty())
+            {
+                this.isConnectingToMqttGateway = true;
+            }
+            else
+            {
+                throw new IllegalArgumentException("The value of [GatewayHostName] is NOT provided in the E4K mode.");
+            }
```
timtay-microsoft, line 209:
```suggestion
            if (gatewayHostName != null && !gatewayHostName.isEmpty())
            {
                throw new IllegalArgumentException("The value of [GatewayHostName] is NOT provided in the E4K mode.");            
            }
            this.isConnectingToMqttGateway = true;
```

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
```diff
@@ -400,6 +383,21 @@ else if (securityProvider instanceof SecurityProviderX509)
...
+        GatewayType gatewayType = clientOptions.getGatewayType();
+        if (gatewayType == GatewayType.E4K)
+        {
+            if (gatewayHostName != null && !gatewayHostName.isEmpty())
+            {
+                this.isConnectingToMqttGateway = true;
+            }
+            else
+            {
+                throw new IllegalArgumentException("The value of [GatewayHostName] is NOT provided in the E4K mode.");
```
timtay-microsoft, line 397: Same comment on this if/else block

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientOptions.java
timtay-microsoft: Should the default value here be something like ```NONE```? User's may be confused if they have to pick between E4K and EDGE when they don't want to use a gateway at all

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/GatewayType.java
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the MIT License.
+
+package com.microsoft.azure.sdk.iot.device;
+
+/**
+ * The type of gateway to which the device/module client is connecting while specifying the property of GatewayHostName.
+ */
+public enum GatewayType
+{
+    EDGE, E4K
```
timtay-microsoft, line 11: Add some javadocs explaining each of these options

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientOptions.java
brycewang-microsoft: Yeah, you are right.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientOptions.java
```diff
@@ -138,10 +138,10 @@
...
     /**
-     * This option specifies the type of gateway to which the device/module client is connecting while
-     * specifying the property of GatewayHostName. By default, the value is EDGE.
+     * This option specifies the type of gateway to which the device/module client is connecting. By default,
+     * the value is None.
      */
     @Getter
     @Builder.Default
-    private final GatewayType gatewayType = GatewayType.EDGE;
+    private final GatewayType gatewayType = GatewayType.None;
```
timtay-microsoft, line 146: Sorry to waffle on this, but I realize that we probably can't have "None" be the default since existing edge users would be broken since they now need to pass in this new option. To avoid breaking changes, we'll need just the two gateway types here and the default is Edge.

We'll just need to mention in the javadocs that this field is ignored if no GatewayHostname is provided

brycewang-microsoft: Synced up offline. Ideally, the value of `GatewayHostName` should be ignored if the `GatewayType` is set to "None" (by default or specified explicitly), thus the existing users would have to pass in a `ClientOption` with "Edge" `GatewayType` while connecting to Edge, which is a breaking change for them.

Therefore, changing this back to 2 enum values and adding additional comment to explain this.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
```diff
@@ -208,10 +195,21 @@ private void configSasAuth(IotHubConnectionString iotHubConnectionString) {
 
     ClientConfiguration(IotHubConnectionString iotHubConnectionString, IotHubClientProtocol protocol, ClientOptions clientOptions)
     {
-        String mqttGatewayHostName = iotHubConnectionString.getMqttGatewayHostName();
-        if (mqttGatewayHostName != null && !mqttGatewayHostName.isEmpty())
+        String gatewayHostName = iotHubConnectionString.getGatewayHostName();
```
timtay-microsoft, line 198: This block of code appears twice in this class. Can you factor this out to a private method?

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/GatewayType.java
```diff
@@ -0,0 +1,20 @@
...
+
+package com.microsoft.azure.sdk.iot.device;
+
+/**
+ * The type of gateway to which the device/module client is connecting while specifying the property of GatewayHostName.
+ */
+public enum GatewayType
+{
+    /**
+     * The device/module client is connecting to an Edge hub in the default mode.
```
timtay-microsoft, line 12:
```suggestion
     * The device/module client is connecting to Edgehub.
```

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/GatewayType.java
```diff
@@ -0,0 +1,20 @@
...
+     */
+    EDGE,
+
+    /**
+     * The device/module client is connecting to an Edge hub in the E4K mode.
```
timtay-microsoft, line 17:
```suggestion
     * The device/module client is connecting to an E4K MQTT broker.
```

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/GatewayType.java
timtay-microsoft: FYI "E4K" is not just a flavor of "Edgehub". They are two separate products that behave similarly. Users of existing Edgehubs won't have a toggle to switch into "E4K" mode from "default" mode

timtay-microsoft (approved):

PR #1712, opened 2023-05-11
Update the E4K connect/disconnect message payloads + topics

timtay-microsoft:


iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/E4KConnectionMessagePayload.java
```diff
@@ -0,0 +1,48 @@
...
+    @Setter
+    private String mqttVersion;
+
+    /**
+     * The optional PnP model Id for this client.
+     */
+    @Expose(deserialize = false)
+    @Getter
+    @Setter
+    private String modelId;
```
brycewang-microsoft, line 47: Does model Id appear only with the context of PnP? If not, how about we rename this to "pnpModelId"?

brycewang-microsoft (approved):

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/E4KConnectionMessagePayload.java
drwill-ms: We should leave it to match the property name we have in the SDK as well as the value that was sent in the connect packet.

timtay-microsoft: I'll leave this as-is for now then

PR #1713, opened 2023-05-12
Update Mqtt messaging topics for E4K

brycewang-microsoft:


iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttMessaging.java
```diff
@@ -24,6 +24,16 @@ class MqttMessaging extends Mqtt
     private final String publishTopic;
     private final boolean isEdgeHub;
 
+    private static final String BACKSLASH = "/";
+    private static final String POUND = "#";
```
timtay-microsoft, line 28: I never cared for how we did this when this MQTT layer was first implemented. Remove these constants and just add them to the constants below as needed

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttMessaging.java
```diff
@@ -40,18 +50,31 @@ public MqttMessaging(
...
+
+        // Publish to "$iothub/devices/{clientId}/messages/events/"
+        this.publishTopic = DEVICES +
+                BACKSLASH +
+                clientId +
+                BACKSLASH +
+                EVENTS +
+                BACKSLASH;
+
+        // Subscribe to "$iothub/devices/{clientId}/messages/devicebound/#"
```
timtay-microsoft, line 65: Ideally we would have a constant declared for this string (but for the client Id). That way, we can just do 
```String.format(DEVICEBOUND_TOPIC, clientId)```
instead of all this concatenation below

File: iothub/device/iot-device-client/src/test/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttMessagingTest.java [hunk 3]
```diff
@@ -75,11 +75,11 @@ public void constructorBuildsRoutesWithModuleIdIfPresent(@Mocked final Mqtt mock
         MqttMessaging testMqttMessaging = new MqttMessaging(expectedDeviceId, null, expectedModuleId, false, mockConnectOptions, new HashMap<Integer, Message>(), new ConcurrentLinkedQueue<>());
 
         String actualPublishTopic = Deencapsulation.getField(testMqttMessaging, "publishTopic");
-        assertEquals("devices/" + expectedDeviceId + "/modules/" + expectedModuleId +"/messages/events/", actualPublishTopic);
+        assertEquals("$iothub/devices/" + expectedDeviceId + "/modules/" + expectedModuleId +"/messages/events/", actualPublishTopic);
```
timtay-microsoft, line 78: This change of topics should only happen if the E4K flag is set.

brycewang-microsoft:
> You are changing the topics regardless of if the E4K flag is set

Corrected by checking the value of `isConnectingToMqttGateway` in config passed in ctor of MqttMessaging first

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttIotHubConnection.java
```diff
@@ -242,6 +242,7 @@ else if (proxySettings.getProxy().type() == Proxy.Type.HTTP)
             this,
             moduleId,
             this.config.getGatewayHostname() != null && !this.config.getGatewayHostname().isEmpty(),
+            this.config.isConnectingToMqttGateway(),
```
brycewang-microsoft, line 245: pass the flag `isConnectingToMqttGateway` to ctor of "MqttMessaging", so the topics can be decided accordingly.

timtay-microsoft (approved):

brycewang-microsoft:
We are working on patching the v2 Java SDK to use Mqtt 5 for E4K, so closing this.

PR #1716, opened 2023-05-18
Port existing E4K changes from previews/E4Kadapter branch

timtay-microsoft:
- Send connect/disconnect packet after open/before close
- Add flag for E4K mode
- Create Mqtt5 layer when in E4K mode, otherwise create the existing Mqtt3 layer

### DIFF
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt5/Mqtt5IotHubConnection.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt5/Mqtt5IotHubConnection.java
@@ -3,6 +3,7 @@
 
 package com.microsoft.azure.sdk.iot.device.transport.mqtt5;
 
+import com.google.gson.Gson;
 import com.microsoft.azure.sdk.iot.device.*;
 import com.microsoft.azure.sdk.iot.device.transport.*;
 import com.microsoft.azure.sdk.iot.device.transport.mqtt.*;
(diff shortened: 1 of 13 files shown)


iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt5/Mqtt5IotHubConnection.java
```diff
@@ -206,7 +227,45 @@ else if (proxySettings.getProxy().type() == Proxy.Type.HTTP)
...
+            {
+                log.debug("Sending the \"Opened a connection\" message to the MQTT broker");
+                E4KConnectionMessagePayload connectPayload = E4KConnectionMessagePayload.builder()
+                    .connectionState(E4KConnectionState.Connected)
+                    .modelId(this.config.getModelId())
+                    .mqttVersion(MQTT_VERSION_STRING)
+                    .deviceClientType(this.config.getProductInfo().getUserAgentString())
+                    .build();
+                byte[] serializedPayload = GSON.toJson(connectPayload).getBytes(StandardCharsets.UTF_8);
+                org.eclipse.paho.client.mqttv3.IMqttDeliveryToken token = this.mqttAsyncClient.publish(
```
brycewang-microsoft, line 240: Does Mqtt5 have an object type similar to `mqttv3.IMqttDeliveryToken` here?

timtay-microsoft: Good catch. This was a copy paste error. I have pushed the fix and now this is an Mqtt5 object

iothub/device/iot-device-client/src/test/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/E4KConnectionMessagePayloadTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+public class E4KConnectionMessagePayloadTest
+{
+    @Test
+    public void TestSerialization()
+    {
+        String expectedDeviceClientType = "some device client type";
+        String expectedMqttVersion = "3.1.1";
```
brycewang-microsoft, line 16: I am wondering if we should tie this to v5 only now that E4K requires Mqtt 5?

timtay-microsoft: Yes

brycewang-microsoft (approved):

Issue #1718, opened 2023-05-25
[Bug Report] Application runtime exception in IotHubTransport.


semeneganov-tomtom:
# Context

- **OS and version used:** Android 11 (API 30)
- **Java runtime used:** 1.8
- **SDK version used:** 2.1.5

## Description of the issue
While working throws runtime exception:
```java
java.lang.NullPointerException: Attempt to invoke virtual method 'long java.lang.Long.longValue()' on a null object reference
	at com.microsoft.azure.sdk.iot.device.transport.IotHubTransport.checkForOldMessages(IotHubTransport.java:851)
	at com.microsoft.azure.sdk.iot.device.transport.IotHubTransport.lambda$sendMessages$0$com-microsoft-azure-sdk-iot-device-transport-IotHubTransport(IotHubTransport.java:734)
	at com.microsoft.azure.sdk.iot.device.transport.IotHubTransport$$ExternalSyntheticLambda1.run(Unknown Source:2)
	at java.lang.Thread.run(Thread.java:923)
```

Happens frequently.
MQTT connection is used.

## Code sample exhibiting the issue
Just as sample, nothing wrong, I belive, in here.

```java
public void test() throws IotHubClientException, InterruptedException {
    ScheduledExecutorService executor = Executors.newScheduledThreadPool(1);
    DeviceClient device = new DeviceClient("<<ConnectionString>>", IotHubClientProtocol.MQTT);
    device.open(false);
    Runnable task = () -> {
        try {
            device.sendEvent(new Message("poison"));
        } catch (IotHubClientException|InterruptedException e) {
            System.out.println("Send exception: " + e.getMessage());
        }

        try {
            device.open(false);
        } catch (Exception e) {
            System.out.println("Ignore exception");
        }
    };
    executor.scheduleAtFixedRate(task, 0, 10, TimeUnit.MILLISECONDS);

    Thread.sleep(100000);

	// Then turn off/on the connection and wait till it happen
}
```

## Console log of the issue
Remove partly because of legal and privacy.
Obviously they are not going to show you the whole run.

```
05-03 18:30:42.663  4474 17314 I IotSdkDeviceClient: CONNECTION STATUS UPDATE: CONNECTED
05-03 18:30:42.663  4474 17314 I IotSdkDeviceClient: CONNECTION STATUS REASON: CONNECTION_OK
[...]

PR #1719, opened 2023-05-26
Fix some threading issues with correlation callback maps

timtay-microsoft:
- Make correlation callback cleanup job run less frequently
  - Previously ran several times per second. Now runs once per hour
- Make correlation callback operation blocks atomic
  - We don't want cases where one thread is calling callbacks.get("some id") while another just removed that entry.

#1718

### DIFF
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -124,7 +124,7 @@ public class IotHubTransport implements IotHubListener
     private final Map<String, Long> correlationStartTimeMillis = new ConcurrentHashMap<>();
 
     // A job that runs periodically to remove any stale correlation callbacks
-    private Thread correlationCallbackCleanupThread = new Thread(() -> checkForOldMessages());
+    private final Thread correlationCallbackCleanupThread = new Thread(() -> checkForOldMessages());
     private static final int CORRELATION_CALLBACK_CLEANUP_PERIOD_MILLISECONDS = 60 * 60 * 1000;
     private final Object correlationCallbackOperationLock = new Object();
 
@@ -516,8 +516,6 @@ public class IotHubTransport implements IotHubListener
             openConnection();
         }
 
-        correlationCallbackCleanupThread.start();
-
         log.debug("Client connection opened successfully");
     }
 
@@ -563,8 +561,6 @@ public class IotHubTransport implements IotHubListener
             }
             finally
             {
-                correlationCallbackCleanupThread.interrupt();
-
                 this.updateStatus(IotHubConnectionStatus.DISCONNECTED, reason, cause);
 
                 // Notify send thread to finish up so it doesn't survive this close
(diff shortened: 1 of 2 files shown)


iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -506,6 +516,8 @@ public void open(boolean withRetry) throws TransportException, IotHubClientExcep
             openConnection();
         }
 
+        correlationCallbackCleanupThread.start();
```
timtay-microsoft, line 519: Start the cleanup job when the client has successfully been opened.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -551,6 +563,8 @@ public void close(IotHubConnectionStatusChangeReason reason, Throwable cause)
             }
             finally
             {
+                correlationCallbackCleanupThread.interrupt();
```
timtay-microsoft, line 566: Stop the cleanup job when the client was closed.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -731,7 +745,6 @@ public IotHubClientProtocol getProtocol()
     public void sendMessages()
     {
         checkForExpiredMessages();
-        new Thread(() -> checkForOldMessages()).start();
```
timtay-microsoft, line 734: This method was called very frequently, so we were spawning a lot of threads. Now we just maintain one thread at a time and that thread sleeps for a while after cleaning up stale callbacks

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -123,6 +123,11 @@ public class IotHubTransport implements IotHubListener
     // Used to store the number of milliseconds since epoch that this packet was created for a correlationId
     private final Map<String, Long> correlationStartTimeMillis = new ConcurrentHashMap<>();
 
+    // A job that runs periodically to remove any stale correlation callbacks
+    private final Thread correlationCallbackCleanupThread = new Thread(() -> checkForOldMessages());
+    private static final int CORRELATION_CALLBACK_CLEANUP_PERIOD_MILLISECONDS = 60 * 60 * 1000;
```
brycewang-microsoft, line 128: The only concern I have is that, the original OutOfMemory issue could happen frequently. According to the original [report](https://github.com/Azure/azure-iot-sdk-java/issues/1629), it was reproducible within 15 min. Certainly this depends on the size of environment, but maybe we can decrease the sleeping time length here down to 5 mins or 10 mins perhaps?

timtay-microsoft: I'd expect that these callback maps would get cleaned up as each twin operation finished successfully. This cleanup job is moreso here to catch any entries who belonged to twin operations that didn't finish (service never sent the twin response). With that in mind, having this cleanup job run less frequently should be fine since user's shouldn't expect a lot of twin failures.

That said, I'll run some tests locally to make sure that these maps are being cleaned up even without the cleanup job and let you know.

brycewang-microsoft (approved):

PR #1724, opened 2023-06-16
fix(iot-dev): Add more trace logs to module client createFromEnvironment flow

timtay-microsoft:
We didn't have any way to log the contents of the requests sent to the Edgelet over Unix domain socket

Also wrap a closeable stream in a try block so that it closes automatically

### DIFF
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/hsm/HttpsHsmClient.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/hsm/HttpsHsmClient.java
@@ -260,14 +260,14 @@ public class HttpsHsmClient
 
                     byte[] output = outputStream.toByteArray();
                     log.trace("Writing {} bytes to unix domain socket", output.length);
-                    log.trace("{}", new String(output, StandardCharsets.UTF_8));
+                    log.trace("Contents of the request:\r\n{}", new String(output, StandardCharsets.UTF_8));
                     unixDomainSocketChannel.write(output);
                 }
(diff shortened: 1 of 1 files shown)


iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/hsm/HttpsHsmClient.java
```diff
@@ -251,17 +251,23 @@ private HttpsResponse sendHttpRequestUsingUnixSocket(HttpsRequest httpsRequest,
...
+                // closes the output stream when it exits this block
+                try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream())
+                {
+                    //append http request body to the request bytes
+                    outputStream.write(requestBytes);
+                    outputStream.write(httpsRequest.getBody());
+
+                    byte[] output = outputStream.toByteArray();
+                    log.trace("Writing {} bytes to unix domain socket", output.length);
+                    log.trace("{}", output);
```
brycewang-microsoft, line 263: Maybe specify what we are logging here?
```suggestion
                    log.trace("Contents of the request:\r\n{}", output);
```

brycewang-microsoft (approved):

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/hsm/HttpsHsmClient.java
timtay-microsoft: Sure thing

PR #1739, opened 2023-09-12
Trace reconnecting MQTT errors at INFO level

aalmazanarbs:


aalmazanarbs:
@microsoft-github-policy-service agree

timtay-microsoft:
I'd rather make this kind of behavior opt-in as opposed to the default. How about #1740 instead?

timtay-microsoft:
Closing this as #1740 addresses this issue

PR #1740, opened 2023-09-12
feat(iot-dev): Allow users to opt-out of treating routine disconnects as errors

timtay-microsoft:
SAS authorized MQTT connections have routine disconnects that some users don't want treated as errors or warnings in logs. This adds a flag they can set in client options to opt out of the current default.

Also defer creating a multiplexing retry policy until a multiplexing client is created to avoid unnecessary logs

Like #1739 but requiring opting in to logging these routine disconnects opt-in

### DIFF
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
@@ -91,6 +91,8 @@ public final class ClientConfiguration
 
     private boolean useIdentifiableThreadNames = true;
 
+    private boolean logRoutineDisconnectsAsErrors = true;
+
     private IotHubAuthenticationProvider authenticationProvider;
 
     /**
@@ -223,6 +225,7 @@ public final class ClientConfiguration
         this.threadNamePrefix = clientOptions != null ? clientOptions.getThreadNamePrefix() : null;
         this.threadNameSuffix = clientOptions != null ? clientOptions.getThreadNameSuffix() : null;
         this.useIdentifiableThreadNames = clientOptions == null || clientOptions.isUsingIdentifiableThreadNames();
+        this.logRoutineDisconnectsAsErrors = clientOptions == null || clientOptions.isLoggingRoutineDisconnectsAsErrors();
 
         if (proxySettings != null)
         {
@@ -644,6 +647,12 @@ public final class ClientConfiguration
     {
         // Using a manually written method here to override the name that Lombok would have given it
         return this.useIdentifiableThreadNames;
+    }
+
+    public boolean isLoggingRoutineDisconnectsAsErrors()
+    {
+        // Using a manually written method here to override the name that Lombok would have given it
+        return this.logRoutineDisconnectsAsErrors;
     }
 
     /**
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -1644,7 +1644,7 @@ public class IotHubTransport implements IotHubListener
         {
             if (throwable == null)
             {
-                log.debug("Updating transport status to new status {} with reason {}", newConnectionStatus, reason);
+                log.info("Updating transport status to new status {} with reason {}", newConnectionStatus, reason);
             }
             else if (this.getDefaultConfig() != null
                     && !this.getDefaultConfig().isLoggingRoutineDisconnectsAsErrors()
@@ -1655,7 +1655,7 @@ public class IotHubTransport implements IotHubListener
                 // This is a special case where the user has opted out of treating the routine
                 // MQTT/MQTT_WS SAS token disconnects as an error for logging purposes. As such,
                 // log at the debug level instead of warn or error level.
-                log.debug("Updating transport status to new status {} with reason {}", newConnectionStatus, reason);
+                log.info("Updating transport status to new status {} with reason {}", newConnectionStatus, reason);
             }
             else
             {
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -72,7 +72,7 @@ public class IotHubTransport implements IotHubListener
     private IotHubConnectionStatusChangeCallback multiplexingStateCallback;
     private Object multiplexingStateCallbackContext;
 
-    private RetryPolicy multiplexingRetryPolicy = new ExponentialBackoffWithJitter();
+    private RetryPolicy multiplexingRetryPolicy;
 
     // Callback for notifying the DeviceIO layer of connection status change events. The deviceIO layer
     // should stop spawning send/receive threads when this layer is disconnected or disconnected retrying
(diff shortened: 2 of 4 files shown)


iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
```diff
@@ -327,8 +327,6 @@ public IotHubTransportMessage receive()
     @Override
     public void connectionLost(Throwable throwable)
     {
-        log.warn("Mqtt connection lost", throwable);
```
alejandroaca-ext-inditex, line 330: Removing this, you will not get noticed about disconnecting. In addition, log it as debug and in case where listener is not null, maybe suppose an information lost.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1645,6 +1646,17 @@ private void updateStatus(IotHubConnectionStatus newConnectionStatus, IotHubConn
...
+            else if (this.getDefaultConfig() != null
+                    && !this.getDefaultConfig().isLoggingRoutineDisconnectsAsErrors()
+                    && newConnectionStatus == IotHubConnectionStatus.DISCONNECTED_RETRYING
+                    && reason == IotHubConnectionStatusChangeReason.EXPIRED_SAS_TOKEN
+                    && (protocol == IotHubClientProtocol.MQTT || protocol == IotHubClientProtocol.MQTT_WS))
+            {
+                // This is a special case where the user has opted out of treating the routine
+                // MQTT/MQTT_WS SAS token disconnects as an error for logging purposes. As such,
+                // log at the debug level instead of warn or error level.
+                log.debug("Updating transport status to new status {} with reason {}", newConnectionStatus, reason);
```
alejandroaca-ext-inditex, line 1658: Maybe it is better to log as info, client must know that a disconnect happens. `isLoggingRoutineDisconnectsAsErrors` is about not log as error.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
```diff
@@ -91,6 +91,8 @@ public final class ClientConfiguration
 
     private boolean useIdentifiableThreadNames = true;
 
+    private boolean logRoutineDisconnectsAsErrors = true;
```
alejandroaca-ext-inditex, line 94: It is a nice option, but maybe add an option in `ModuleClient.createFromEnvironment` to avoid to manipulate `ClientOptions` is a better approach. Something like ClientOptionsConfigurer interface to only manipulate the option you need to modify.

timtay-microsoft: ```ModuleClient.createFromEnvironment``` already allows you to pass in a ```ClientOptions``` instance. If you only want to manipulate this new option, then it can already be done:

```java
ClientOptions options = ClientOptions.builder()
        .logRoutineDisconnectsAsErrors(false)
        .build();

ModuleClient.createFromEnvironment(unixDomainSocket, protocol, options);
```

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
timtay-microsoft: I can make this an info level log, sure

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
timtay-microsoft: ```Listener``` is never null here, so moving this line lower doesn't risk losing any log statements

alejandroaca-ext-inditex: Hmm, so we can remove `if (this.listener != null)` and leave as `log.info`? Or even better do this logic:
```
if (transportException.isRetryable()) 
{
    log.info("Mqtt connection lost");
}
else
{
    log.warn("Mqtt connection lost", throwable);
}
```
to avoid add exception when log as info.

timtay-microsoft: MQTT level logs generally aren't done at an Info level because the SDK abstracts most of that logic. The reconnection logic layer will do any info level logging related to disconnects.

> Or even better do this logic

I'm not sure I follow why we would care about retryable vs non-retryable when logging that the connection was lost. I can remove the ```if (this.listener != null)``` though, sure.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
timtay-microsoft: I don't follow. What is stopping you from using ```ModuleClient.createFromEnvironment(unixDomainSocket, protocol, options``` ?

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
alejandroaca-ext-inditex: From our point of view, a retryable exception is not the final error, so maybe, when retrying, you can log it as info otherwise log as warn or debug.

If you don't want to add the "if logic" in the `log.debug`, instead of adding the full exception object as a parameter you can append the message of the exception in order to avoid posting an exception in telemetry.

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ClientConfiguration.java
alejandroaca-ext-inditex: For sure, we can use `ModuleClient.createFromEnvironment(unixDomainSocket, protocol, options)` but the use of the default `ClientOptions` should also be acceptable.

In the case you are proposing, you have to start to care about the construction of `ClientOptions` and if you use `ModuleClient.createFromEnvironment(unixDomainSocket, protocol)` you don't have to do it. Maybe it doesn't have sense to configure `ClientOptions` when you only need to configure logs.

timtay-microsoft: I'm not going to make opting out of these logs as warnings/errors the default behavior, so you will need to start constructing a ```ClientOptions``` here. By changing the default behavior, we risk causing unexpected problems for existing users. The best I can do here is allow you to opt-in via ```ClientOptions```

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/Mqtt.java
timtay-microsoft: > instead of adding the full exception object as a parameter you can append the message of the exception in order to avoid posting an exception in telemetry.

I'd rather not. Omitting the full stacktrace here would make it harder to root-cause disconnects on the off-chance that the exception doesn't make it up to the reconnection logic where it would normally get logged.

The fact that this MQTT-layer log with a stacktrace is done at a Debug level means you can still "hide" this log (and its full stacktrace) by only listening for Info level logs.

alejandroaca-ext-inditex: Ok, close this review.

alejandroaca-ext-inditex (approved):

Issue #1736, opened 2023-08-29
[Bug Report] RegisterTask thread not finalized if ProvisioningDeviceClient is closed too fast


gabrielandrade2:
# Context

- **OS and version used:** Ubuntu 22.04.1
- **Java runtime used:** openjdk version "17.0.8" 2023-07-18
- **SDK version used:** 2.1.1 and 2.1.0

## Description of the issue
To contextualize a bit, I have a service running under my device that implements Azure DPS and allows clients to provision their devices using symmetric keys. I have a task that periodically retries the provisioning in case of failure, however, the user can start/cancel the provisioning process and the retry task (before the device is assigned in the cloud) at any time. Provisioning is done using the async `ProvisioningDeviceClient.registerDevice()` method.

However, during tests with some failure cases, after killing the provisioning process at some times I would have a `azure-iot-sdk-RegisterTask` named thread still lingering around in a `WAITING` state and it never finishes.

After some debugging, I believe that this happens when the provisioning process is asked to be stopped (thus `ProvisioningDeviceClient.close()` is called) too quickly after calling `ProvisioningDeviceClient.registerDevice()`.
It appears to me that, if `close()` is called before the task from `RegisterTask` is done, the thread will never finish.

I was able to create a scenario to reproduce this problem using a sample code from the SDK with some modifications to use the async method instead (attached below). 
If you take a thread dump after the `Press any key to exit...` message appears in the console, there should be a `azure-iot-sdk-RegisterTask` thread running.

If you switch the commented lines and wait a longer time, all DPS threads should be cleared. (During some tests I saw a `ProvisioningTask` thread still running, but that one was finalized after some seconds).

## Code sample exhibiting the issue
```java
package samples.com.microsoft.azure.sdk.iot;

import com.microsoft.azure.sdk.iot.device.*;
import com.microsoft.azure.sdk.iot.device.exceptions.IotHubClientException;
import com.microsoft.azure.sdk.iot.provisioning.device.*;
import com.microsoft.azure.sdk.iot.provisioning.device.internal.exceptions.ProvisioningDeviceClientException;
import com.microsoft.azure.sdk.iot.provisioning.security.SecurityProviderSymmetricKey;

import java.io.IOException;
import java.nio.charset.StandardCharsets;
import java.util.Scanner;

/**
 * Symmetric Key authenticated individual enrollment sample
 */
@SuppressWarnings("CommentedOutCode") // Ignored in samples as we use these comments to show other options.
[...]

PR #1741, opened 2023-09-12
fix(prov-dev): Fix bug where closing provisioning device client didn't cleanup all threads

timtay-microsoft:
If the provisioning task was interrupted by the user closing the client, it needs to shutdown its executor service that runs the RegisterTask

This fix addresses #1736

### DIFF
--- a/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/ProvisioningTask.java
+++ b/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/ProvisioningTask.java
@@ -343,7 +343,7 @@ public class ProvisioningTask implements Callable<Object>
             this.executeStateMachineForStatus(registrationOperationStatusParser);
             this.close();
         }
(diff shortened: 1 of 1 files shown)


provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/task/ProvisioningTask.java
```diff
@@ -343,7 +343,7 @@ public Object call() throws Exception
             this.executeStateMachineForStatus(registrationOperationStatusParser);
             this.close();
         }
-        catch (ExecutionException | TimeoutException | ProvisioningDeviceClientException | SecurityProviderException e)
+        catch (ExecutionException | TimeoutException | ProvisioningDeviceClientException | SecurityProviderException | InterruptedException e)
```
timtay-microsoft, line 346: This catch block cleans up all local resources including the RegisterTask thread. The problem here was that this thread would be interrupted and then exit without ever closing the threads it started.

Now we end all threads this thread created before ending this thread

Issue #1746, opened 2023-10-11
[Bug Report] QOS not set for incoming messages from `$iothub/twin/PATCH/properties/desired/` topic


dzakub:
# Context

- **OS and version used:** - 6.4.12-arch1-1 GNU/Linux
- **Java runtime used:** OpenJDK Runtime Environment 18.9 (build 11.0.12+7)
- **SDK version used:** 2.3.0

## Description of the issue
Messages received from topic: `$iothub/twin/PATCH/properties/desired/` do not set QOS properly. This causes ACK message to not to be sent to a broker. 
Issue was discovered while writing integration tests with fake IOT hub.

## Console log of the issue
```
2023-10-10 17:04:16,262 TRACE - com.hivemq.mqtt.handler.publish.PublishFlowHandler - Client 'iot-hub-fake' completed a PUBLISH flow with QoS 1 or 2 for packet identifier '3'
2023-10-10 17:04:17,026 TRACE - com.microsoft.azure.sdk.iot.device.transport.mqtt.Mqtt - Mqtt message arrived on topic $iothub/twin/PATCH/properties/desired/?$version=2 with mqtt message id 151
2023-10-10 17:04:17,026 INFO  - IotHubFake - IotHubFake: Delivery complete, message id: 3, topics: [$iothub/twin/PATCH/properties/desired/?$version=2], complete: true
2023-10-10 17:06:33,465 TRACE - com.hivemq.mqtt.message.pool.SequentialMessageIDPoolImpl - Tried to return message id 51 although it was already returned. This is could mean a DUP was acked
2023-10-10 17:06:33,465 TRACE - com.microsoft.azure.sdk.iot.device.transport.mqtt.MqttIotHubConnection - Received MQTT device twin message ( Message details: Correlation Id [9410b255-3b4d-4212-bc39-a65962028587] Message Id [0d26490a-371a-474d-83ab-918aba12af34] Device Operation Type [DEVICE_OPERATION_TWIN_SUBSCRIBE_DESIRED_PROPERTIES_RESPONSE] )
[...]

PR #1747, opened 2023-10-12
Fixing issue: QOS not set for incoming messages from $iothub/twin/PAT…

dzakub:
…CH/properties/desired/ topic #1746

 

# Checklist

https://github.com/Azure/azure-iot-sdk-java/issues/1746

# Description of the problem
Problem is described in the issue.

# Description of the solution
Added code to copy QOS from original message.

### DIFF
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTwin.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTwin.java
@@ -342,7 +342,7 @@ class MqttTwin extends Mqtt
                                 {
                                     message = new IotHubTransportMessage(data, MessageType.DEVICE_TWIN);
                                     message.setDeviceOperationType(DeviceOperations.DEVICE_OPERATION_TWIN_SUBSCRIBE_DESIRED_PROPERTIES_RESPONSE);
-                                    message.setQualityOfService(messagePair.getValue().getQos());
+                                    message.setQualityOfService(mqttMessage.getQos());
                                 }
 
                                 // Case for $iothub/twin/PATCH/properties/desired/?$version={new version}
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTwin.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTwin.java
@@ -342,6 +342,7 @@ class MqttTwin extends Mqtt
                                 {
                                     message = new IotHubTransportMessage(data, MessageType.DEVICE_TWIN);
                                     message.setDeviceOperationType(DeviceOperations.DEVICE_OPERATION_TWIN_SUBSCRIBE_DESIRED_PROPERTIES_RESPONSE);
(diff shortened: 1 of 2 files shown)


dzakub:
@microsoft-github-policy-service agree

iothub/device/iot-device-client/src/test/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTwinTest.java
```diff
@@ -904,6 +904,33 @@ public void receiveDoesNotSetVersionForDesiredPropNotifRespIfNotFound() throws T
         }
     }
 
+    @Test
+    public void receiveSetQosForDesiredProp() throws TransportException
+    {
+        final byte[] actualPayload = "NotificationResponseDataContainingDesiredPropertiesDocument".getBytes(StandardCharsets.UTF_8);
+        final String expectedTopic = "$iothub/twin/PATCH/properties/desired/";
+        IotHubTransportMessage receivedMessage = null;
+        int expectedQualityOfService = 2;
```
timtay-microsoft, line 913: Small nitpick to make this unit test closer to what an actual received twin message would look like.

```suggestion
        int expectedQualityOfService = 0;
```

timtay-microsoft:
This change seems reasonable, but these messages always have a QoS of 0, so I'm not sure what change in behavior you'll see from this.

timtay-microsoft (approved):

iothub/device/iot-device-client/src/test/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTwinTest.java
dzakub: In other case I would agree - I'm all for making test data as close to actual production data as possible. But doing so, I would made this test passing regarding of production code change - QOS is held as `int` which during object construction is set to default value: 0.

First I created failing test, and the fix production making test green - this way I knew that test is actually checking case I wanted to cover.

dzakub:
> This change seems reasonable, but these messages always have a QoS of 0, so I'm not sure what change in behavior you'll see from this. Are you expecting QoS other than 0 here?

I didn't know that IOT hub sends always massages with QOS 0 here. If there will be no change in behavior.
I discovered this 'issue' whlie stubbing iot hub for integration tests that were written to document behavior of 1.x clients and to ease upgrading to 2.x. 
Tests were green for 1.x, but started failing after upgrade to 2.x. After investigation I discovered mentioned problem (my fake iot hub was sending messages with QOS 1)

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTwin.java
```diff
@@ -342,6 +342,7 @@ else if (topic.length() > PATCH.length() && topic.startsWith(PATCH))
                                 {
                                     message = new IotHubTransportMessage(data, MessageType.DEVICE_TWIN);
                                     message.setDeviceOperationType(DeviceOperations.DEVICE_OPERATION_TWIN_SUBSCRIBE_DESIRED_PROPERTIES_RESPONSE);
+                                    message.setQualityOfService(messagePair.getValue().getQos());
```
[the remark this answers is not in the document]
dzakub: Sure, thank you.

Issue #1734, opened 2023-08-11
[Bug Report] DeviceClient.sendEventAsync fails and doesn't invoke MessageSentCallback


szlisiecki:
# Context

- **OS and version used:** Ubuntu 20.04.4 LTS
- **Java runtime used:** 11.0.15
- **SDK version used:** iot-device-client 2.1.4

## Description of the issue
Everything works just fine for couple of days and than just stops sending messages to IoTHub. Also messageSentCallback is not invoked. 

## Code sample exhibiting the issue
```
MultiplexingClient multiplexingClient = new MultiplexingClient(host, IotHubClientProtocol.AMQPS, null);
DeviceClient deviceClient = new DeviceClient(connString, IotHubClientProtocol.AMQPS);
multiplexingClient.registerDeviceClient(deviceClient);

deviceClient.sendEventAsync(message, messageSentCallback, null);
```

## Console log of the issue
As you can see, during sending message, I get information that IotHub message was acknowledged. 
For messages that were not sent I can not see in logs information that message was acknowledged by IoTHub.
So I suppose that I can not send messages because IoTHub doesn't want to acknowledge these messages. And the internal state of azure sdk is "some kind of hang up". The processing of message is stoped. Azure sdk doesn't even call a handler (messageSentCallback) to inform that something is wrong with the message.

Logs before crash
```
2023-08-08 20:13:11.041 [XXXXX-Multiplexed-bdf2bbaa-Cxn48ab2303-4071-4407-bb9f-fa1c85c29b75-azure-iot-sdk-IotHubSendTask] TRACE c.m.a.s.i.d.t.IotHubTransport - Dequeued a message from waiting queue to be sent ( Message details: Correlation Id [93ebb01d-8d38-47a2-9607-fc780d044578] Message Id [2310a6c1-fa99-411d-8630-f669a4bc6d25] )

2023-08-08 20:13:11.041 [XXXXX-Multiplexed-bdf2bbaa-Cxn48ab2303-4071-4407-bb9f-fa1c85c29b75-azure-iot-sdk-IotHubSendTask] TRACE c.m.a.s.i.d.t.IotHubTransport - Adding transport message to the inProgressPackets to wait for acknowledgement ( Message details: Correlation Id [93ebb01d-8d38-47a2-9607-fc780d044578] Message Id [2310a6c1-fa99-411d-8630-f669a4bc6d25] )

[...]

PR #1753, opened 2023-10-19
fix(iot-dev): Fix 3 multiplexing scenario bugs

timtay-microsoft:
- Fix bug where a closing multiplexed device would report its state as ```Disconnected``` followed by ```Disconnected_retrying``` and then finally ```Disconnected```
- Fix bug where a multiplexed device that has its device session closed didn't cancel all of its outgoing messages
- Fix bug where terminal retry states (retry expired, non-retryable exception encountered, etc.) didn't end reconnection attempts

### DIFF
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -410,7 +410,9 @@ public class IotHubTransport implements IotHubListener
     @Override
     public void onMultiplexedDeviceSessionLost(TransportException e, String connectionId, String deviceId, boolean shouldReconnect)
     {
-        if (connectionId.equals(this.iotHubTransportConnection.getConnectionId()))
+        if (connectionId.equals(this.iotHubTransportConnection.getConnectionId())
+            && multiplexedDeviceConnectionStates.containsKey(deviceId)
+            && multiplexedDeviceConnectionStates.get(deviceId).getConnectionStatus() == IotHubConnectionStatus.CONNECTED)
         {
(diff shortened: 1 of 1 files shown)


iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1388,6 +1390,7 @@ private void singleDeviceReconnectAttemptAsync(String deviceSessionToReconnect)
             {
                 this.updateStatus(IotHubConnectionStatus.DISCONNECTED, IotHubConnectionStatusChangeReason.RETRY_EXPIRED, transportException, deviceSessionToReconnect);
                 log.debug("Reconnection for device {} was abandoned due to the operation timeout", deviceSessionToReconnect);
+                return;
```
timtay-microsoft, line 1393: Not returning here meant that retry logic would continue even if we declare that retry has expired

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1766,6 +1771,55 @@ private void updateStatus(IotHubConnectionStatus newConnectionStatus, IotHubConn
         else if (newConnectionStatus == IotHubConnectionStatus.DISCONNECTED)
         {
             correlationCallbackCleanupThread.interrupt();
+            finalizeMultiplexedDevicesMessages(deviceId);
```
timtay-microsoft, line 1774: Now when a multiplexed device reaches the DISCONNECTED state, we look for any of its messages in the outgoing queues and cancel them. This should address #1734

andyk-ms (approved):

PR #1762, opened 2023-11-03
feat(iot-dev): Add configurable message expiration check job

timtay-microsoft:
Previously, the client would only check for expired messages each time it tried to send a message. Now it will check for expired messages even if the client is idle.

### DIFF
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
@@ -79,6 +79,9 @@ public class IotHubTransport implements IotHubListener
     // should stop spawning send/receive threads when this layer is disconnected or disconnected retrying
     private final IotHubConnectionStatusChangeCallback deviceIOConnectionStatusChangeCallback;
 
+    // Lock on reading and writing on the waitingPackets queue
+    final private Object waitingPacketsLock = new Object();
+
     // Lock on reading and writing on the inProgressPackets map
     final private Object inProgressMessagesLock = new Object();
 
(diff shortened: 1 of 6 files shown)


iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1766,20 +1816,8 @@ private void updateStatus(IotHubConnectionStatus newConnectionStatus, IotHubConn
             invokeConnectionStatusChangeCallback(newConnectionStatus, previousStatus, reason, throwable, deviceId);
         }
 
-        if (newConnectionStatus == IotHubConnectionStatus.CONNECTED)
-        {
-            try
-            {
-                correlationCallbackCleanupThread.start();
```
timtay-microsoft, line 1773: We shouldn't start/stop this thread when one device in a multiplexed connection is lost. Only start/stop it when the whole connection is opened/closed

iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/IotHubTransport.java
```diff
@@ -1913,28 +1951,6 @@ private boolean hasOperationTimedOut(long startTime)
         return (System.currentTimeMillis() - startTime) > this.getDefaultConfig().getOperationTimeout();
     }
 
-    /**
-     * Returns if the provided packet has lasted longer than the device operation timeout
-     *
-     * @return true if the packet has been in the queues for longer than the device operation timeout and false otherwise
-     */
-    private boolean hasOperationTimedOut(long startTime, String deviceId)
```
timtay-microsoft, line 1921: Unused method

patilsnr (approved):

PR #1778, opened 2024-03-05
Upgrade maven compiler plugin, bouncycastle e2e/sample dependency versions

timtay-microsoft:
Jackson upgrades as well

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,6 +151,21 @@
                 <version>1.70</version>  <!--e2e tests and samples only-->
             </dependency>
             <dependency>
+                <groupId>org.bouncycastle</groupId>
+                <artifactId>bcmail-jdk18on</artifactId>
+                <version>1.77</version>  <!--e2e tests and samples only-->
+            </dependency>
+            <dependency>
+                <groupId>org.bouncycastle</groupId>
+                <artifactId>bcprov-jdk18on</artifactId>
+                <version>1.77</version>  <!--e2e tests and samples only-->
+            </dependency>
+            <dependency>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -142,13 +142,18 @@
             </dependency>
             <dependency>
                 <groupId>org.bouncycastle</groupId>
-                <artifactId>bcmail-jdk15on</artifactId>
-                <version>1.70</version>  <!--e2e tests and samples only-->
+                <artifactId>bcmail-jdk18on</artifactId>
```
timtay-microsoft, line 145: jdk15on versions are no longer maintained, so we need to move to jdk18on to stay up to date on security patches

pom.xml
```diff
@@ -142,13 +142,18 @@
...
             <dependency>
                 <groupId>org.bouncycastle</groupId>
-                <artifactId>bcprov-jdk15on</artifactId>
-                <version>1.70</version>  <!--e2e tests and samples only-->
+                <artifactId>bcprov-jdk18on</artifactId>
+                <version>1.77</version>  <!--e2e tests and samples only-->
+            </dependency>
+            <dependency>
+                <groupId>org.bouncycastle</groupId>
+                <artifactId>bcpkix-jdk18on</artifactId>
```
timtay-microsoft, line 155: This new dependency is a new part of the "jdk18on" set of bouncycastle packages

brycewang-microsoft (approved):

timtay-microsoft:
We still get the old bouncycastle dependency from our TSS library, so I'll close this for now while we work out what we want to do about that dependency

PR #1777, opened 2024-02-23
Fixing issue: Version 2.x of iot-device-client, requires version of reported properties to be sent, while 1.x not require that #1776

dzakub:
# Checklist
- [ x] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-java/blob/main/.github/CONTRIBUTING.md).
- [x] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).
- I submitted this PR against the correct branch: 
  - [x] This pull-request is submitted against the `main` branch. 

# Reference/Link to the issue solved with this PR (if any)

# Description of the problem

# Description of the solution

PR #1779, opened 2024-03-08
Allow device/module client to send reported properties without specifying the version

timtay-microsoft:
#1777 but with an integration test added and a little code cleanup

### DIFF
--- a/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTwin.java
+++ b/iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTwin.java
@@ -130,23 +130,6 @@ class MqttTwin extends Mqtt
                 }
                 break;
             }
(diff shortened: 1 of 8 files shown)


iothub/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/transport/mqtt/MqttTwin.java
```diff
@@ -122,28 +122,13 @@ private String buildTopic(final IotHubTransportMessage message)
                     throw new IllegalArgumentException("Request Id is Mandatory");
                 }
 
-                int version = message.getVersion();
-                topic.append(AND);
-                topic.append(VERSION);
-                topic.append(version);
-                break;
-            }
-            case DEVICE_OPERATION_TWIN_SUBSCRIBE_DESIRED_PROPERTIES_REQUEST:
```
timtay-microsoft, line 131: This particular block would return the topic string to subscribe on, but this method is about building publish payloads, so we can remove this

Issue #118, opened 2017-07-13
Enable Routing on Message Body


it-bytes:
- **OS and version used:** Ubuntu 16.04

- **Java runtime used:** Java(TM) SE Runtime Environment (build 1.8.0_131-b11)

- **SDK version used:** `<dependency>
               <groupId>com.microsoft.azure.sdk.iot</groupId>
               <artifactId>iot-device-client</artifactId>
               <version>1.3.30</version>
</dependency>`

# Description of the issue:

I want to set the System Properties „ContentEncoding“ und „ContentType“ to enable routing on message body. As it mentioned in the article  (https://azure.microsoft.com/de-de/blog/iot-hub-message-routing-now-with-routing-on-message-body/).

Actually, I see no possibilities to set these two system properties with this version of the SDK.

Do I got something wrong here?

# Console log of the issue:
If is send a message with Java SDK it looks like:

12.07.2017 09:41:39> Device: [XXXXXXXXXXXXXXX], Data:[…]Properties:
'source': 'XXXXXXXXXXXXXXXXX'
SYSTEM>x-opt-sequence-number=1808
SYSTEM>x-opt-offset=1767872
SYSTEM>x-opt-enqueued-time=12.07.2017 07:41:39
SYSTEM>iothub-connection-device-id=XXXXXXXXXXXXXXXXXXX
SYSTEM>iothub-connection-auth-method={"scope":"device","type":"sas","issuer":"iothub","acceptingIpFilterRule":null}
SYSTEM>iothub-connection-auth-generation-id=XXXXXXXXXXXXXXXXXX
SYSTEM>iothub-enqueuedtime=12.07.2017 07:41:38
SYSTEM>iothub-message-source=Telemetry
SYSTEM>EnqueuedTimeUtc=12.07.2017 07:41:39
SYSTEM>SequenceNumber=1808
SYSTEM>Offset=1767872
SYSTEM>message-id=4646f130-14f5-4717-baf5-75591ef132b9
SYSTEM>correlation-id=921262c0-bc75-4c92-88b0-a06e94c689ea

And if is send a message with C# SDK and set the two propteries it looks like:
12.07.2017 09:43:52> Device: [XXXXXXXXXXXXXXX], Data:[...]Properties:
'Status': 'Active'
SYSTEM>x-opt-sequence-number=1234
SYSTEM>x-opt-offset=773528
SYSTEM>x-opt-enqueued-time=12.07.2017 07:43:52
[...]

PR #1798, opened 2024-10-27
fix(iot-dev): Fix default Content Type and Content Encoding

nabinbhandari:
Set default Content Type of D2C Message to null and default Content Encoding to UTF-8.

Github issue (fix#118)

 

# Checklist

#118 

# Description of the problem

The existing code set the default content type (instead of content encoding) to "UTF-8". This would prevent using message routing by querying on the message body even if the content type was explicitly set to "application/json".

# Description of the solution
 
Used default IoT Hub message charset name to set the content encoding of the message instead of content type. An application using the client SDK should still explicitly set the content type to "application/json" to enable proper message routing.


nabinbhandari:
@timtay-microsoft Can you please review this?

PR #1806, opened 2025-08-06
[DPS][Public Preview] Provisioniong Service SDK Update

jessHuh:
# Checklist
- [x] I submitted this PR against the correct branch: 

# Description of the problem

Preparing a new public preview API release for DPS Service APIs 

# Description of the solution
 
Adding a new field "CredentialPolicyName" in EnrollmentGruop and IndividualEnrollment. This change will be introduced as part of 2025-07-01-preview API.

### DIFF
--- a/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/SDKUtils.java
+++ b/provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/SDKUtils.java
@@ -19,7 +19,7 @@ import java.util.stream.Collectors;
 @Slf4j
 public class SDKUtils
 {
-    private static final String SERVICE_API_VERSION = "2025-07-01-preview";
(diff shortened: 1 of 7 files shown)


provisioning/provisioning-device-client/src/main/java/com/microsoft/azure/sdk/iot/provisioning/device/internal/SDKUtils.java
```diff
@@ -19,7 +19,7 @@
 @Slf4j
 public class SDKUtils
 {
-    private static final String SERVICE_API_VERSION = "2019-03-31";
+    private static final String SERVICE_API_VERSION = "2025-07-01-preview";
```
timtay-microsoft, line 22: You shouldn't have to make this change since your new service API version is only relevant to the provisioning service SDK and this file is referenced by the provisioning device SDK

jessHuh: Oh you're right. Reverted the change.

PR #1856, opened 2026-08-10
Honor the connect timeout when connecting through an HTTP proxy

ewertons:
# Checklist
- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-java/blob/main/.github/CONTRIBUTING.md).

Nightly `Java Linux` (pipeline definition 533) failures, e.g. builds [161572](https://dev.azure.com/azure-iot-sdks/azure-iot-sdks/_build/results?buildId=161572&view=results) and [161517](https://dev.azure.com/azure-iot-sdks/azure-iot-sdks/_build/results?buildId=161517&view=results).

# Description of the problem

## 1. The device client can block forever when connecting through an HTTP proxy

`ProxiedSSLSocket.connect(SocketAddress, int timeout)` **discarded the `timeout` argument**, and ran the HTTP `CONNECT` tunnel handshake against the proxy socket with `SO_TIMEOUT` left at `0`. A proxy that accepts the TCP connection but stalls before answering `CONNECT` therefore blocks the calling thread indefinitely, with no way for the caller to recover.

Both callers do supply a timeout today, and both were being ignored:

* MQTT/MQTT_WS &mdash; `MqttConnectOptions.getConnectionTimeout()` (default 30s) reaches the socket via `WebSocketSecureNetworkModuleFactory` &rarr; `SSLNetworkModule.setSSLhandshakeTimeout` &rarr; `TCPNetworkModule.setConnectTimeout` &rarr; `socket.connect(addr, conTimeout * 1000)`.
* HTTPS &mdash; `ClientOptions.httpsConnectTimeout` reaches the socket via `HttpsURLConnection.setConnectTimeout` &rarr; `sun.net.NetworkClient.doConnect` &rarr; `socket.connect(addr, connectTimeout)`.

`HttpProxySocketFactory.createSocket()` also connected to the proxy eagerly using the blocking `new Socket(host, port)` constructor, which has no connect timeout either. And `ProxiedSSLSocket.close()` threw a `NullPointerException` when it ran after a failed connect attempt (`sslSocket` is still `null` at that point), which hid the original failure.

This is what breaks the nightly runs. `ConnectionTests.CanOpenConnection[MQTT_WS_SAS_DEVICE_CLIENT_true_*]` fails with

```
org.junit.runners.model.TestTimedOutException: test timed out after 60000 milliseconds
    at ...ConnectionTests.CanOpenConnection(ConnectionTests.java:244)
```

where line 244 is `open(true)` and the blocked frame is `Object.wait` (Paho's connect token). Instead of failing fast and letting `IotHubTransport`'s retry policy recover, the client sat in the unbounded read until the test's own timeout fired.

Looking at the last 12 failed nightly runs, **8 of them fail on this same proxy path**: `CanOpenConnection[MQTT_WS_SAS_DEVICE_CLIENT_true_false]` / `[..._true_true]` (6 runs), `getAndCompleteSasUriWithoutUpload[HTTPS_SAS_true]` + `getSasUriWithoutUploadConnectionToggle[HTTPS_SAS_true]`, and `tokenRenewalWorks` (which creates proxied HTTPS/MQTT_WS/AMQPS_WS clients).

## 2. E2E test harness issues that make the above much worse

* `HttpProxyServer.startAsync(port)` returns a `CompletionStage` that completes when the bind completes, and all four test classes that start a local proxy ignored it. With `<parallel>both</parallel>` + `useUnlimitedThreads`, all ~29 parameterized `ConnectionTests` fire within ~150ms of `@BeforeClass`, so tests can try to connect before the port is listening.
* `ConnectionTests.ConnectionTestInstance.dispose()` was dead code &mdash; nothing ever called it. Every test in the class leaked its `DeviceClient` (and its identity). Those leaked clients keep reconnecting in the background, and once `@AfterClass` closes the proxies they retry forever against a dead proxy. A single failing run's log contains **117** stacks of:

```
Caused by: java.net.ConnectException: Connection refused
    at java.base/java.net.Socket.<init>(Socket.java:287)
    at ...HttpProxySocketFactory.createSocket(HttpProxySocketFactory.java:27)
    at org.eclipse.paho.client.mqttv3.internal.TCPNetworkModule.start(TCPNetworkModule.java:73)
```

  all after `stopProxy()` ran, burning threads and CPU for the remaining ~15 minutes of the job.
* `@Test(timeout = 60000)` on the `ConnectionTests` methods is **exactly equal** to `Mqtt.CONNECTION_TIMEOUT` (60s), so a single slow connect attempt guarantees the test times out before the SDK's own retry can even start.

## 3. `ContractAPIMqttTest` corrupts the JVM on JDK 8

Visible in build 161517 (JDK 8 is the only job that runs unit tests), masked by `rerunFailingTestsCount`:

```
Exception in thread "Reference Handler" java.lang.UnsatisfiedLinkError: java.lang.Object.wait(J)V
...
java.lang.IllegalStateException: Not in the recording phase
    at ...ContractAPIMqttTest.createContractClass(ContractAPIMqttTest.java:93)
```

`ContractAPIMqttTest` declares `@Mocked Object mockSendLock` and `@Mocked Integer mockedInteger`, and `ContractAPIAmqpTest` declares `@Mocked Object mockSendLock`. **None of the three are used anywhere.** Mocking `java.lang.Object` makes JMockit retransform it, which unbinds `Object.wait(J)V` and kills the Reference Handler thread, after which JMockit's phase tracking is inconsistent and unrelated tests in the class fail.

# Description of the solution

## `iot-device-client`

* `ProxiedSSLSocket` now connects to the proxy inside `connect(SocketAddress, int)` using the supplied timeout, and applies the remaining part of that same budget as `SO_TIMEOUT` while waiting for the `CONNECT` response (restoring the previous value afterwards, unless the handshake already closed the socket). When the caller supplies no timeout at all, a 60s default is used rather than blocking forever.
* `HttpProxySocketFactory.createSocket()` returns an unconnected socket so the connect can be deferred to where the timeout is known.
* `ProxiedSSLSocket.close()` is null-safe.

The resulting `SocketTimeoutException` / `ConnectException` are already classified as retryable by `PahoExceptionTranslator`, so `IotHubTransport.open(withRetry = true)` retries instead of hanging.

New unit test `HttpProxySocketFactoryTest.connectTimesOutWhenProxyNeverAnswersConnectRequest` covers this against a `ServerSocket` that accepts but never replies. It fails (hangs) without the fix.

## `iot-e2e-tests`

* New `ProxyServerTools.startProxyServer(server, port)` waits on the future returned by `startAsync`; used by `ConnectionTests`, `FileUploadTests`, `MultiplexingClientTests` and `TokenRenewalTests`.
* `ConnectionTests` disposes its test instance in an `@After`. The one-off ECC identities it creates itself are deleted from the registry rather than recycled, so a device carrying a certificate no other test knows about can never be handed to a later x509 test from the shared pool.
* Removed the per-method `@Test(timeout = 60000)` in `ConnectionTests`; the shared 2 minute `IntegrationTest` timeout rule still bounds these tests.

## `provisioning-device-client`

* Removed the unused `@Mocked Object` / `@Mocked Integer` fields from `ContractAPIMqttTest` and `ContractAPIAmqpTest`.

# Validation

Run locally (JDK 8 for unit tests, matching `vsts/build_repo.ps1`, since JMockit 1.24 cannot run on newer JDKs):

* `mvn install -DskipTests` &mdash; `BUILD SUCCESS` for the whole repo, including `iot-e2e-common` test compilation.
* `mvn test` for `iot-device-client` and `provisioning-device-client` on JDK 8 &mdash; `BUILD SUCCESS`.
  * `HttpProxySocketFactoryTest` passes in 2.168s against its 2s budget (unbounded block before the fix).
  * `ContractAPIMqttTest` 34/34 and `ContractAPIAmqpTest` 31/31, with no `Reference Handler` `UnsatisfiedLinkError` and no `Not in the recording phase`. The "36 tests" reported by CI was 34 real tests plus 2 surefire reruns.

The e2e changes need a gated run against real resources to confirm.

### DIFF
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ProxyServerTools.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/helpers/ProxyServerTools.java
@@ -1,0 +1,34 @@
+/*
+ *  Copyright (c) Microsoft. All rights reserved.
+ *  Licensed under the MIT license. See LICENSE file in the project root for full license information.
+ */
+
+package tests.integration.com.microsoft.azure.sdk.iot.helpers;
+
+import com.github.monkeywie.proxyee.server.HttpProxyServer;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Helpers for the local HTTP proxy servers that the proxy related integration tests run their traffic through.
+ */
+public class ProxyServerTools
+{
+    private static final int PROXY_START_TIMEOUT_SECONDS = 30;
+
+    /**
+     * Start the provided proxy server on the provided port and block until it is actually listening on that port.
+     *
+     * {@link HttpProxyServer#startAsync(int)} only initiates the bind, so tests that don't wait on the returned future
+     * can start sending traffic to the proxy before it is listening. When that happens, the client under test gets a
+     * "Connection refused" instead of a working proxy.
+     *
+     * @param proxyServer the proxy server to start.
+     * @param port the port for the proxy server to listen on.
+     * @throws Exception if the proxy server could not be started within {@link #PROXY_START_TIMEOUT_SECONDS} seconds.
+     */
+    public static void startProxyServer(HttpProxyServer proxyServer, int port) throws Exception
+    {
+        proxyServer.startAsync(port).toCompletableFuture().get(PROXY_START_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+    }
+}
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/FileUploadTests.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/FileUploadTests.java
@@ -36,6 +36,7 @@ import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import tests.integration.com.microsoft.azure.sdk.iot.helpers.annotations.FlakeyTest;
 import tests.integration.com.microsoft.azure.sdk.iot.helpers.IntegrationTest;
+import tests.integration.com.microsoft.azure.sdk.iot.helpers.ProxyServerTools;
 import tests.integration.com.microsoft.azure.sdk.iot.helpers.TestConstants;
 import tests.integration.com.microsoft.azure.sdk.iot.helpers.TestDeviceIdentity;
 import tests.integration.com.microsoft.azure.sdk.iot.helpers.Tools;
@@ -147,12 +148,15 @@ public class FileUploadTests extends IntegrationTest
     }
 
     @BeforeClass
-    public static void startProxy()
+    public static void startProxy() throws Exception
     {
         HttpProxyServerConfig config = new HttpProxyServerConfig();
         config.setHandleSsl(false);
         proxyServer = new HttpProxyServer().serverConfig(config);
-        proxyServer.startAsync(testProxyPort);
+
+        // startAsync() only starts the bind, so the returned future must be waited on. Otherwise the tests in this
+        // class can start connecting before the proxy is listening, and they fail with "Connection refused".
+        ProxyServerTools.startProxyServer(proxyServer, testProxyPort);
     }
 
     @AfterClass
--- a/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
+++ b/iot-e2e-tests/common/src/test/java/tests/integration/com/microsoft/azure/sdk/iot/iothub/MultiplexingClientTests.java
@@ -203,12 +203,15 @@ public class MultiplexingClientTests extends IntegrationTest
     }
 
     @BeforeClass
-    public static void startProxy()
+    public static void startProxy() throws Exception
     {
         HttpProxyServerConfig config = new HttpProxyServerConfig();
         config.setHandleSsl(false);
         proxyServer = new HttpProxyServer().serverConfig(config);
-        proxyServer.startAsync(testProxyPort);
+
+        // startAsync() only starts the bind, so the returned future must be waited on. Otherwise the tests in this
+        // class can start connecting before the proxy is listening, and they fail with "Connection refused".
+        ProxyServerTools.startProxyServer(proxyServer, testProxyPort);
     }
 
     @AfterClass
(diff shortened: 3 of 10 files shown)


timtay-microsoft (approved):

ewertons:
Closing this in favour of #1863, which carries forward the parts that are still relevant.

Most of this PR has been overtaken since it was opened on 10 August, and it is now `CONFLICTING` against `main`.

**Already fixed independently, do not re-apply:**

| This PR proposed | Where it landed |
| --- | --- |
| `ProxiedSSLSocket` honors the connect timeout | #1859 (`90d3ff169`) |
| `HttpProxySocketFactory` returns an unconnected socket | #1859 (`d82729408`) |
| `ProxiedSSLSocket.close()` null-safety | on `main` |

The diagnosis here was correct and matched what #1859 arrived at independently, including the observation that the failures span `CanOpenConnection[MQTT_WS_*]`, `getAndCompleteSasUriWithoutUpload[HTTPS_SAS_true]` and `tokenRenewalWorks` — i.e. the proxied variants across several classes, not one protocol.

**Superseded, and deliberately not carried forward:** removing `@Test(timeout = 60000)`.

This PR read the 60s-vs-60s collision as a race to be won by changing one of the numbers. It is not. These tests call `open(true)`, and the default retry policy is `ExponentialBackoffWithJitter` with `retryCount = Integer.MAX_VALUE`, so the client never abandons the attempt and never throws. Nothing ends the test but the JUnit timeout, and falling back to the shared 2 minute rule would only make it fail slower. #1862 addressed the real cause: the e2e suite ran with `parallel=both` and `useUnlimitedThreads` in a single JVM shared with the embedded proxies, so a burst of test threads starved them. It now runs with `threadCount=6`, measured to cost no additional wall clock, and `CanOpenConnection` logs its connection status transitions so a timeout leaves behind a reason.

**Still relevant, moved to #1863:**

1. `ProxyServerTools` — `startAsync` returns an ignored `CompletionStage`, so tests can hit a proxy before it is listening.
2. `ConnectionTests.dispose()` being dead code — a real client and identity leak, and the same bug #1861 fixed in `TokenRenewalTests`. It also works directly against the parallelism cap from #1862.
3. The unused `@Mocked Object` / `@Mocked Integer` fields in the provisioning contract tests.

On that third item, one correction to the record: I could not reproduce the JVM corruption described here. The full provisioning suite passes on JDK 8 with reruns disabled, and no `ContractAPI*` failure appears in the last seven CI builds. #1863 still removes those fields, because mocking `java.lang.Object` is a genuine hazard and all three are unused, but it does so as hygiene rather than as a fix for an active failure.

Reopening this would mean resolving conflicts in three files whose SDK changes are now redundant, which is more work and more risk than re-applying three small test-harness fixes on top of current `main`. Thanks — the analysis in this PR held up well, and it found two harness bugs that nothing else had.

